PR #8, opened 2017-03-28
ZMS-599:TestMailItem own user

fatkudu:
* Create own test user and delete after
* Added some test data to 2 folders
* Junit 4 style

### DIFF
--- a/store/src/java/com/zimbra/qa/unittest/TestMailItem.java
+++ b/store/src/java/com/zimbra/qa/unittest/TestMailItem.java
@@ -30,7 +30,6 @@ import org.junit.rules.TestName;
 import com.zimbra.common.util.ByteUtil;
 import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.cs.account.Account;
-import com.zimbra.cs.account.Provisioning;
 import com.zimbra.cs.db.DbMailItem;
 import com.zimbra.cs.db.DbResults;
 import com.zimbra.cs.db.DbUtil;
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/qa/unittest/TestMailItem.java
```diff
@@ -34,29 +40,51 @@
...
-    private static final String USER_NAME = "user1";
+    @Rule
+    public TestName testInfo = new TestName();
+    private Account account = null;
     private static final int TEST_CONTACT_ID = 9999;
+    private static final Provisioning prov = Provisioning.getInstance();
 
-    @Override
+    @Before
     public void setUp() throws Exception {
```
grishick: not sure what is the point of an empty setUp() method. Shouldn't cleanUp be called here?

store/src/java/com/zimbra/qa/unittest/TestMailItem.java
```diff
@@ -34,29 +39,47 @@
...
+    }
+
+    @Test
     public void testListItemIds()
     throws Exception {
-        Account account = TestUtil.getAccount(USER_NAME);
+        Account account = TestUtil.createAccount(USER_NAME);
+        // Mailbox.ID_FOLDER_INBOX;
         Mailbox mbox = MailboxManager.getInstance().getMailboxByAccount(account);
+        TestUtil.addMessage(mbox, testInfo.getMethodName() + "missive 1");
```
grishick: nit: since user name is already unique for this test method, there is no need to use test method in message body or subj

fatkudu: I find it helps in finding things in mailbox.log

PR #9, opened 2017-03-28
Zms 598

grishick:
TestServiceServlet: added more test cases to cover possible exploits.

ZimletFile: code changes do not change outcomes of attempting to deploy or undeploy a malicious zimlet, but catch the bad name earlier, making error reporting cleaner


grishick:
This change is not intended for 17.13 and does not have a security impact.

PR #10, opened 2017-03-29
ZMS-603: Modify dynamic ephemeral attributes via Provisioning::modifyAttrs

ilyusha:
Updated _LdapProvisioning::modifyEphemeralAttrs_ to support dynamic ephemeral attributes. This is done by leveraging existing AttributeConverter classes to try to parse the provided as LDAP-formatted strings. If the provided value cannot be parsed with the appropriate AttributeConverter, or if the AttributeConverter does not exist for the ephemeral attribute, then a warning is logged and the  operation is skipped.

### DIFF
--- a/store/src/java/com/zimbra/qa/unittest/TestModifyDynamicEphemeralAttrs.java
+++ b/store/src/java/com/zimbra/qa/unittest/TestModifyDynamicEphemeralAttrs.java
@@ -1,0 +1,62 @@
+package com.zimbra.qa.unittest;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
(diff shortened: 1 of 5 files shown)


store/src/java/com/zimbra/qa/unittest/TestModifyDynamicEphemeralAttrs.java
```diff
@@ -0,0 +1,62 @@
...
+
+public class TestModifyDynamicEphemeralAttrs {
+
+    private static final String acctName = TestModifyDynamicEphemeralAttrs.class.getSimpleName();
+    private static Account acct;
+    private static Provisioning prov = Provisioning.getInstance();
+
+    @BeforeClass
+    public static void init() throws Exception {
+        if (!TestUtil.accountExists(acctName)) {
```
grishick: I think it is better to delete the account before the test even if it exists already in case there is some garbage in the account.

gordyt:
Looks good once previous suggestion implemented.  Be sure and "sign" the CLA.

PR #14, opened 2017-03-30
Bugfix/zms 602 test lmtp

fatkudu:


store/src/java/com/zimbra/qa/unittest/TestUtil.java
```diff
@@ -524,19 +525,49 @@ private static ZMimePart getPart(ZMimePart mimeStructure, String name) {
...
+            String line;
+            while ((line = ir.readLine()) != null) {
+                ZimbraLog.test.info("mailq STDOUT:%s", line);
+            }
+        } catch (IOException e) {
+            ZimbraLog.test.error("Problem logging stream with prefix '%s'", prefix, e);
+        }
+    }
+
+    private static void execMailq(boolean attemptToDeliverAllQueuedMail) {
```
grishick: we have a mechanism for executing postfix CLI commands, take a look com.zimbra.cs.rmgmt.RemoteMailQueue and com.zimbra.cs.service.admin.MailQueueAction. I suspect that using that mechanism would be more friendly to testing in multi-node environments.

store/src/java/com/zimbra/qa/unittest/TestUtil.java
```diff
@@ -555,9 +617,52 @@ private static ZMimePart getPart(ZMimePart mimeStructure, String name) {
...
+             * (delivery temporarily suspended: connect to example.com[XXX.XXX.XXX.XXX]:7025: Connection refused)
+             *                           user1@example.com
+             * and:
+             * DD410183B37     2619 Thu Mar 30 15:59:31  user1@example.com
+             * (host example.com[XXX.XXX.XXX.XXX] said:
+             *     451 4.0.0 Temporary message delivery failure try again (in reply to end of DATA command))
+             *                           user1@example.com
+             */
+            if (postfixFlushDeferredMailQueue(AccountTestUtil.getServer(mbox.getName()),
+                    finishBy - System.currentTimeMillis())) {
```
grishick: My guess is that instead of looking up account's home server, we should be looking up MTA's hostname to figure out where postfix is running.

fatkudu: Been round in circles on this.  I've come to the conclusion that this is probably as good as it gets with the current code.  If there is an issue, it is likely to be that the remote queue handling code works on **Server** objects and assumes that Provisioning.A_zimbraServiceHostname attribute for that Server is the right host to execute postfix queue handling commands on.

grishick:
I don't see the approve option any more, but I think this is good to go

PR #15, opened 2017-03-30
ZMS-573: TestFolder unit test fix

ilyusha:
Removed TestCase subclass from TestFolder; added @BeforeClass/@AfterClass-annotated methods to create/destroy a test account.

### DIFF
--- a/store/src/java/com/zimbra/qa/unittest/TestFolders.java
+++ b/store/src/java/com/zimbra/qa/unittest/TestFolders.java
@@ -25,6 +25,7 @@ import org.junit.After;
 import org.junit.AfterClass;
 import org.junit.Before;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import com.google.common.base.Strings;
@@ -92,14 +93,10 @@ public class TestFolders {
         assertNull(mbox.getFolderByPath(childPath));
     }
 
+    @Ignore
     @Test
     public void testCreateManyFolders() throws Exception {
         //normally skip this test since it takes a long time to complete. just keep it around for quick perf checks
-        boolean skip = true;
-        if (skip) {
-            return;
-        }
-
         ZMailbox mbox = TestUtil.getZMailbox(USER_NAME);
         String parentPath = "/parent";
 
--- a/store/src/java/com/zimbra/qa/unittest/TestFolders.java
+++ b/store/src/java/com/zimbra/qa/unittest/TestFolders.java
@@ -35,21 +35,19 @@ import com.zimbra.cs.account.Provisioning;
 
 public class TestFolders {
     private static String USER_NAME = TestFolders.class.getSimpleName();
-    private static String NAME_PREFIX = "TestFolders";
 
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/qa/unittest/TestFolders.java
```diff
@@ -81,7 +93,8 @@ public void testEmptyLargeFolder()
         assertEquals(0, parent.getMessageCount());
         assertNull(mbox.getFolderByPath(childPath));
     }
-    
+
+    @Test
     public void testCreateManyFolders() throws Exception {
         //normally skip this test since it takes a long time to complete. just keep it around for quick perf checks
         boolean skip = true;
```
grishick: I suggest fixing this using "@Ignore" annotation

store/src/java/com/zimbra/qa/unittest/TestFolders.java
```diff
@@ -17,31 +17,43 @@
...
 
     private String mOriginalEmptyFolderBatchSize;
 
-    @Override
-    protected void setUp()
-    throws Exception {
+
+    @BeforeClass
+    public static void init() throws Exception {
+        TestUtil.deleteAccount(USER_NAME);
```
grishick: if account does not exist, this will log an error in mailbox.log. It is better to check if account exists first.

store/src/java/com/zimbra/qa/unittest/TestFolders.java
```diff
@@ -17,31 +17,43 @@
...
+        TestUtil.createAccount(USER_NAME);
+    }
+
+    @Before
+    public void setUp() throws Exception {
         cleanUp();
```
grishick: I think "cleanUp" method needs rewriting. It currently tries to remove data from existing accounts and it is not reliable. It is better to delete/re-create test accounts before each test run and delete test accounts after each test run.

store/src/java/com/zimbra/qa/unittest/TestFolders.java
```diff
@@ -17,31 +17,43 @@
...
 import com.zimbra.client.ZMailbox;
+import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.cs.account.Provisioning;
 
-public class TestFolders extends TestCase
-{
-    private static String USER_NAME = "user1";
+public class TestFolders {
+    private static String USER_NAME = TestFolders.class.getSimpleName();
     private static String NAME_PREFIX = "TestFolders";
```
grishick: if the user name is unique, there is no need for NAME_PREFIX

store/src/java/com/zimbra/qa/unittest/TestFolders.java
ilyusha: Good point, I'll use TestUtil.deleteAccountIfExists

store/src/java/com/zimbra/qa/unittest/TestFolders.java
ilyusha: Yep, good point

PR #16, opened 2017-03-31
ZCS-1001 CalDav shared calendar fix

rupalid:
Added code to URLDecode calendar path string before comparison. This was a regression. 
The regression caused value: /dav/test1@domain.com/Calendar to be compared to  value: /dav/test1%40domain.com/Calendar/39e4bcf0-c3e7-4603-b202-8b521ca29c64.ics. This resulted in long URL being returned in calendar-query

Checked by installing zm-store jar with the fix on 8.7.5 installation. Configured an account with shared calendar in iCal on Mac and verified that appts  in shared and own calendar are shown and do not disappear. Verified with both recurring and non recurring appts.


Victor-D:
Hello
What is the version this regression was introduced ?
Victor

rupalid:
This is a regression caused by BugFix for 103996.

PR #17, opened 2017-03-31
ZCS-860 : Sieve: addheader/replaceheader must not add a header withou…

chinmay15:
Problem : header name used as space or starting with space should not be allowed.

Fix : left trimmed header name before validation.

Testing done : added unit tests for
1) addheader
    a) header name as single space should fail
    b) header name as multiple spaces should fail
    c) header name starting with space(s) should fail
2) replaceheader
    a) new header name as single space should fail
    b) new header name as multiple spaces should fail
    c) new header name starting with space(s) should fail
    d) header name mentioned as key should fail if same as a, b and c
3) deleteheader
    a) header name as single space should fail
    b) header name as multiple spaces should fail
    c) header name starting with space(s) should fail

Testing to be done by QA : verify all mentioned scenarios and do sanity around the same.

### DIFF
--- a/store/src/java/com/zimbra/cs/filter/jsieve/AddHeader.java
+++ b/store/src/java/com/zimbra/cs/filter/jsieve/AddHeader.java
@@ -17,7 +17,6 @@
 package com.zimbra.cs.filter.jsieve;
 
 import java.io.UnsupportedEncodingException;
-
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.Iterator;
@@ -28,6 +27,7 @@ import javax.mail.MessagingException;
 import javax.mail.internet.MimeMessage;
 import javax.mail.internet.MimeUtility;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.jsieve.Argument;
 import org.apache.jsieve.Arguments;
 import org.apache.jsieve.Block;
@@ -154,6 +154,10 @@ public class AddHeader extends AbstractCommand {
         }
 
         if (!StringUtil.isNullOrEmpty(headerName)) {
+            String tempHeaderName = StringUtils.stripStart(headerName, null);
+            if (!tempHeaderName.equals(headerName)) {
+                throw new SyntaxException("Header name must not start with spaces : \"" + headerName + "\"");
+            }
             if (!CharsetUtil.US_ASCII.equals(CharsetUtil.checkCharset(headerName, CharsetUtil.US_ASCII))) {
                 throw new SyntaxException("AddHeader:Header name must be printable ASCII only.");
             }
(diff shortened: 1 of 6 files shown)


store/src/java/com/zimbra/cs/filter/jsieve/AddHeader.java
```diff
@@ -154,6 +155,12 @@ protected void validateArguments(Arguments arguments, SieveContext context)
         }
 
         if (!StringUtil.isNullOrEmpty(headerName)) {
+            String tempHeaderName = headerName;
```
shrikant-prasad: should move this code to static method in FilterUtil class and call that method in AddHeader and EditHeaderExtension.

chinmay15:
@shrikant-prasad done

store/src/java/com/zimbra/cs/filter/FilterUtil.java
```diff
@@ -1029,5 +1029,14 @@ private static void validateVariableIndex(String srcStr) throws SyntaxException
...
        }
     }
-}
 
+    /**
+     * returns String by removing initial spaces
+     * @param text String to remove initial spaces
+     * @return
+     */
+    public static String removeInitialSpaces(String text) {
```
rohan-ambasta: May be you can use StringUtils.stripStart() instead of this method

chinmay15:
@rohan-ambasta done

store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
```diff
@@ -299,11 +300,15 @@ public void setupEditHeaderData(Arguments arguments, AbstractCommand ac) throws
                         arg = itr.next();
                         if (arg instanceof StringListArgument) {
                             StringListArgument sla = (StringListArgument) arg;
-                            String newName = sla.getList().get(0);
-                            if (StringUtil.isNullOrEmpty(newName)) {
+                            String origNewName = sla.getList().get(0);
```
Prashantsurana: are  you sure to get index 0 from the list? Potential array index out of bound case here. 
in the next line there is check for empty or null string, it indicates there is possibility of empty list. 
Can you please confirm ?

chinmay15: Yes, that's correct. We are checking 1st if the arg is instance of  object of StringListArgument. StringListArguement is created when there is at least 1 String object in the list. So there is no scope for index out of bound.

Prashantsurana: Approving based on above comment.

chinmay15:
merged in feature/sieve

gordyt:
@chinmay15 When you get a chance, please do a clean rebase of the feature/sieve branch onto develop.  Please let me know if you have any questions.

chinmay15:
@gordyt sure

PR #20, opened 2017-03-31
ZMS-225: ZMailbox::alterTag implementation

ilyusha:
Implemented _ZMailbox::alterTag_ method, with corresponding unit test.

### DIFF
--- a/client/src/java/com/zimbra/client/ZMailbox.java
+++ b/client/src/java/com/zimbra/client/ZMailbox.java
@@ -6103,8 +6103,8 @@ public class ZMailbox implements ToZJSONObject, MailboxStore {
     @Override
     public void alterTag(OpContext octxt, Collection<ItemIdentifier> ids, String tagName, boolean addTag)
             throws ServiceException {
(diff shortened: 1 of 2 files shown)


fatkudu:
Looks like you've done the same thing as I've done in the past and created a new bugfix from within an existing bugfix branch - so this code is included in the code for https://github.com/Zimbra/zm-mailbox/pull/21/commits - suggest we stick with just that pull request?

ilyusha:
I was planning on merging this one first so that the commits from this bug are no longer in the zms-226 PR.

grishick:
Should be fine to merge this first if you rebase the other bug branch on develop after merging this one before merging it.

grishick:
Looks like Mailbox.alterTag also handles marking read/unread as a special case. Could you add a test case to check that ZMailbox::alterTag correctly handles marking read/unread as well?

ilyusha:
Actually, the read/unread special case is not applicable here because  _MailboxStore::alterTag_ can only modify tags by name, and this case can only reached by modify a flag. The method where this happens is shared by both tag and flag modification code paths.

PR #21, opened 2017-03-31
ZMS-226: ZMailbox::setTags implementation

ilyusha:
Implemented _ZMailbox::setTags_ method, with corresponding unit test. This was a little more tricky than alterTag, since the ItemAction SOAP handler used to require the folderId to be included in the request in order to perform a tag update operation, while _MailboxStore.setTags_ does not include it. I updated the server code as follows:
1. ItemAction::handleCommon allows for iidFolder to remain null.
2. ItemActionHelper::schedule does not set targeted=true if the operation is UPDATE but the folder id is null.

The SOAP docs do not state that the folder ID is required for the "update tags" operation, so this change actually makes the ItemActionRequest API consistent with the documentation:

> For op="update", caller can specify any or all of: l="{folder}", name="{name}", color="{color}",
> tn="{tag-names}", f="{flags}".  When modifying tags or flags, all specified tags and flags are set, and all others are unset.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/ItemAction.java
+++ b/store/src/java/com/zimbra/cs/service/mail/ItemAction.java
@@ -190,11 +190,14 @@ public class ItemAction extends MailDocumentHandler {
                 localResults = ItemActionHelper.RENAME(octxt, mbox, responseProto, local, type, tcon, name, iidFolder).getResult();
             } else if (opStr.equals(MailConstants.OP_UPDATE)) {
                 String folderId = action.getAttribute(MailConstants.A_FOLDER, null);
-                ItemId iidFolder = new ItemId(folderId == null ? "-1" : folderId, zsc);
-                if (!iidFolder.belongsTo(mbox)) {
-                    throw ServiceException.INVALID_REQUEST("cannot move item between mailboxes", null);
-                } else if (folderId != null && iidFolder.getId() <= 0) {
-                    throw MailServiceException.NO_SUCH_FOLDER(iidFolder.getId());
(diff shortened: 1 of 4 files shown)


store/src/java/com/zimbra/cs/service/mail/ItemAction.java
```diff
@@ -190,11 +190,12 @@ protected String handleCommon(Map<String, Object> context, Element request, Stri
                 localResults = ItemActionHelper.RENAME(octxt, mbox, responseProto, local, type, tcon, name, iidFolder).getResult();
             } else if (opStr.equals(MailConstants.OP_UPDATE)) {
                 String folderId = action.getAttribute(MailConstants.A_FOLDER, null);
-                ItemId iidFolder = new ItemId(folderId == null ? "-1" : folderId, zsc);
-                if (!iidFolder.belongsTo(mbox)) {
-                    throw ServiceException.INVALID_REQUEST("cannot move item between mailboxes", null);
-                } else if (folderId != null && iidFolder.getId() <= 0) {
```
fatkudu: This code was introduced for https://bugzilla.zimbra.com/show_bug.cgi?id=9333 (blame used in zm-store repo to find that)  Commit comment was:

    We were silently ignoring negative target folder IDs in op="update".  Now we throw mail.NO_SUCH_FOLDER.

Need to re-instate code to throw NO_SUCH_FOLDER if a negative folder ID has been supplied.

ilyusha: Good find Gren! I'll make the requested change

PR #22, opened 2017-03-31
Feature/deploy all (reopened)

tmclane:
I got tired of having to wait for the webserver to stop and start again as it deployed each of the 5 pieces of the zm-mailbox project.

This update updates the ant target `deploy-foss` to the zm-mailbox / build.xml:
     * stops the webserver
     * deploys all 5 internal jar files
     * starts the webserver

### DIFF
--- a/build-common.xml
+++ b/build-common.xml
@@ -25,10 +25,10 @@
         </not>
         <then>
           <mkdir dir="${ivy.jar.dir}"/>
-          <!-- download Ivy from web site so that it can be used even without any special installation. 
+          <!-- download Ivy from web site so that it can be used even without any special installation.
              Note, that 'get' task will not try downloading if this file is already downloaded. -->
-          <get src="https://repo1.maven.org/maven2/org/apache/ivy/ivy/${ivy.install.version}/ivy-${ivy.install.version}.jar" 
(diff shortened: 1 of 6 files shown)


fatkudu:
This looks a mess now - what's the plan?

fatkudu:
Worth having a look at the zm-zcs project
    grep target ant-global.xml |grep foss
    <target name="publish-foss-local">
    <target name="deploy-foss">

I suspect that you've added some more useful stuff but it would be good for everything to hook into the right places rather than duplicating things.  For discoverability, it would be good for zm-mailbox root to have targets in build.xml that make sense.

grishick:
LGTM.

tmclane:
Maybe I should replace the 'deploy-foss' in the zm-zcs with my version such that it doesn't stop and start the webserver for every jar.  That was the goal of my 'deploy-all' target after all.
What do you think of that?

tmclane:
This has been updated to cause 'deploy-foss' in zm-mailbox/build.xml to do the intended behavior.
Should quicken the 'deploy-foss' command by quite a bit since the webserver isn't started and stopped unnecessarily.
It's only stopped at the beginning and restarted after the zm-store jar file is copied into place completing the system.

PR #24, opened 2017-04-03
ZCS-871 : Sieve: a match variable  which has a backslash does not mat…

chinmay15:
Added unit test for

1) Comparison of ${1} with header using ":matches"
2) Comparison of ${1} with header using ":is"
3) Comparison of ${1} by setting variable with header using ":matches"
4) Comparison of ${1} by setting variable using ":quotewildcard" with string.

Note :
This bug is working fine. So just added unit tests for this.

### DIFF
--- a/store/src/java-test/com/zimbra/cs/filter/HeaderTest.java
+++ b/store/src/java-test/com/zimbra/cs/filter/HeaderTest.java
@@ -272,4 +272,55 @@ public class HeaderTest {
             fail("No exception should be thrown");
         }
     }
+
+    @Test
+    public void testHeaderMatchWithItself() throws Exception {
+        String script =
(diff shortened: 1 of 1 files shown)


chinmay15:
Code merged to feature/sieve branch.

commit 71bdce610144cca952f130f7cf040d55289a2970
Author: Chinmay Pathak <chinmay.pathak@synacor.com>
Date:   Mon Apr 3 21:04:55 2017 +0530

    ZCS-871 : Sieve: a match variable  which has a backslash does not match the the same object

 store/src/java-test/com/zimbra/cs/filter/HeaderTest.java | 51 +++++++++++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 51 insertions(+)

PR #28, opened 2017-04-04
Bugfix/zms 610cookie

fatkudu:


store/src/java/com/zimbra/qa/unittest/TestCookieReuse.java
```diff
@@ -66,33 +68,43 @@
...
-            TestUtil.deleteAccount(UNAUTHORIZED_USER);
-        }
     }
 
-    @Override
+    @After
     public void tearDown()
     throws Exception {
         cleanUp();
+        Provisioning.getInstance().getLocalServer().setLowestSupportedAuthVersion(currentSupportedAuthVersion);
```
tmclane: I am curious what this line does at this location.
Isn't the test completely over by the time this runs?

fatkudu: It sets the configuration back to how it was before the test, so other tests don't get nasty surprises.  The value is read during setup - if one of the tests that changes it were to assert before setting it back, that would be bad without this.

PR #29, opened 2017-04-05
ZCS-842:fix sieveToJava regex as per RFC 5229

shrikant-prasad:
If there are two or more "*" wildcards in the sieve matching pattern, all wildcards should match non-greedy except the last wildcard. If there is only one wildcard in the sieve pattern, it should be greedy match.

Testing Done: Added JUnit tests for various scenarios tested.


shrikant-prasad:
Merged to feature/sieve
To https://github.com/Zimbra/zm-mailbox.git
   71bdce6..9a7a1bd  feature/sieve -> feature/sieve

PR #30, opened 2017-04-05
ZCS-540 User is able to activate filters with fileinto action...

rohan-ambasta:
1. Added validation to check whether the folder exists or not
2. Account info is required for validation, hence added a constructor with Account param
3. Need to check whether the rule is active or not, hence added FilterRule param in methods

Testing done:
1. If the folder specified in fileIntoAction of inactive filter does not exists, the filter is not activated

Note: Will create a JIRA ticket for UI change


rohan-ambasta:
merged into feature/sieve

PR #36, opened 2017-04-07
ZCS-1006:SIEVE:"i;ascii-numeric" doesn't handle negative number

ykomiyam:
[Problem]
The negative number was treated as positive infinity. The "i;ascii-numeric" comparator does not support negative numbers. In case of negative numbers, exception should be thrown, and the entire filter execution should be canceled.

[Fix]
* When evaluating the test, such as "string", "header", "envelope", "address", "deleteheader", and "replaceheader", under the ascii-numeric comparator, the sieve processor find the negative value as a parameter, it throws the FeatureException. The negative number is defined as a string which starts with '-' (0x2d) followed by a digit. If the given string is empty or started with non-digit character or started with '-' but not followed by a digit, such string is continuously treated as positive infinity.  When the given string is started with a digit, it is continuously treated as digit.

* The signature of the Equals interface is defined inside the jsieve library, and the Equals.equals() method does not throw the exception. To make it throw an exception, a new interface Equals2 is introduced, and the equals2() method throws the FeatureException exception.

[Testing done]
* Unit test for verifying that the filter execution is canceled when the negative number is found during the :is, :count and :value match pattern under the i;ascii-numeric for the tests listed above.

* Ran all the unit tests com.zimbra.cs.filter.* and all PASS.

* Ran automation scripts which test the "i;ascii-numeric"; zm-qa/data/soapvalidator/MailClient/Filters/Sieve/ZCS-865.xml, ZCS-863.xml, ZCS-373.xml and all PASS.

[Testing to be done by QA]
* Once the negative number is found, the filter execution should be canceled and the message will be delivered to the Inbox.  Additionally let's execute the cases for the ZCS-863 to cover the normal case for the :is, :count, and :value match type under the "i;ascii-numeric".

### DIFF
--- a/store/src/java-test/com/zimbra/cs/filter/AddressTest.java
+++ b/store/src/java-test/com/zimbra/cs/filter/AddressTest.java
@@ -352,4 +352,28 @@ public class AddressTest {
             fail("No exception should be thrown");
         }
     }
+
+    @Test
+    public void testNumericNegativeValueIs() {
+        try {
+            Account acct = Provisioning.getInstance().getAccount(MockProvisioning.DEFAULT_ACCOUNT_ID);
+            RuleManager.clearCachedRules(acct);
+            Mailbox mbox = MailboxManager.getInstance().getMailboxByAccount(acct);
+
+            String filterScript = "require \"tag\";\n"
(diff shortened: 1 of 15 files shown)


chinmay15:
You need to finish CLA

PR #39, opened 2017-04-10
ZCS-1195: Fix dependency to generate proper BuildInfoGenerated.java

Prashantsurana:
Fix ant dependency to generate proper BuildInfoGenerated.java


Prashantsurana:
Thanks.

PR #42, opened 2017-04-10
Feature/zms624 - request for feedback

tmclane:
I am still working on testing these changes on my development cluster.
It's not quite complete since I still need to wire in the SOAP request modification such that the additional details are requested from the remote side and are sent back to the caller.
That being said I would like feedback at this time so I am opening the PR now.


fatkudu:
Looks OK.  May be worth looking at the soap JAXB classes and experimenting with the effect on the response XML.  WSDL can be a bit quirky and that may influence how you structure the response.

tmclane:
closing this in favor of the newer combined zms-624 and zms-198 PR

PR #49, opened 2017-04-13
ZMS-194: RemoteImapMailboxStore::addressMatchesAccountOrSendAs

ilyusha:
Implemented _RemoteImapMailboxStore::addressMatchesAccountOrSendAs_, as well as a corresponding unit test. The method uses the same underlying call as _LocalImapMailboxStore_, except it gets the account instance using the _RemoteImapMailboxStore.getAccount()_ method.

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/RemoteImapMailboxStore.java
+++ b/store/src/java/com/zimbra/cs/imap/RemoteImapMailboxStore.java
@@ -50,6 +50,7 @@ import com.zimbra.cs.mailbox.OperationContext;
 import com.zimbra.cs.service.UserServlet;
 import com.zimbra.cs.service.util.ItemId;
 import com.zimbra.cs.store.Blob;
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/imap/RemoteImapMailboxStore.java
```diff
@@ -187,7 +188,7 @@ public int getCurrentMODSEQ(int folderId) throws ServiceException {
 
     @Override
     public boolean addressMatchesAccountOrSendAs(String givenAddress) throws ServiceException {
-        throw new UnsupportedOperationException("RemoteImapMailboxStore method not supported yet");
+        return (AccountUtil.addressMatchesAccountOrSendAs(getAccount(), givenAddress));
```
tmclane: Do the outer parenthesis do something useful in this case?

PR #56, opened 2017-04-17
ZCS-1097 - Make RPM/DEBIAN packages from zm-mailbox

shri314:
This change needs other changes in build and deployment, which are being worked on in zm-build repository.


shri314:
This change needs other changes in build and deployment, which are being worked on in zm-build repository.

PR #59, opened 2017-04-19
Feature/zms 641

grishick:
step 1 for implementing ZContact.getSize and ZContact.getContentStream is to figure out which IMAP calls fail w/o these methods being implemented.


grishick:
changed "not" to "now" in this commit message "all tests from TestImap are now in SharedImapTests"

PR #65, opened 2017-04-20
ZCS-1135:Fixing GetFilterRulesTest:testBug71036_SingleRuleSimpleNeste…

sneha-patil-synacor:
Fixing GetFilterRulesTest:testBug71036_SingleRuleSimpleNestedIfwithMultiConditions02
The response was being compared with static String. The order of the attributes was changed which was causing the unit test failure.

### DIFF
--- a/store/src/java-test/com/zimbra/cs/service/mail/GetFilterRulesTest.java
+++ b/store/src/java-test/com/zimbra/cs/service/mail/GetFilterRulesTest.java
@@ -16,7 +16,6 @@
  */
 package com.zimbra.cs.service.mail;
 
-import com.zimbra.common.util.ZimbraLog;
 import com.google.common.collect.Maps;
 import com.zimbra.common.account.Key;
 import com.zimbra.common.soap.Element;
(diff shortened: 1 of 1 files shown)


store/src/java-test/com/zimbra/cs/service/mail/GetFilterRulesTest.java
```diff
@@ -173,30 +172,41 @@ public void testBug71036_SingleRuleSimpleNestedIfwithMultiConditions02() throws
...
+        Assert.assertEquals("confidential",rules.getElement("filterRule").getElement("nestedRule").getElement("filterTests").getElement("headerTest").getAttribute("value"));
+        Assert.assertEquals("0",rules.getElement("filterRule").getElement("nestedRule").getElement("filterTests").getElement("headerTest").getAttribute("index"));
+        Assert.assertEquals("1",rules.getElement("filterRule").getElement("nestedRule").getElement("filterTests").getElement("sizeTest").getAttribute("s"));
+        Assert.assertEquals("over",rules.getElement("filterRule").getElement("nestedRule").getElement("filterTests").getElement("sizeTest").getAttribute("numberComparison"));
+        Assert.assertEquals("1",rules.getElement("filterRule").getElement("nestedRule").getElement("filterTests").getElement("sizeTest").getAttribute("index"));
+        Assert.assertEquals("priority",rules.getElement("filterRule").getElement("nestedRule").getElement("filterActions").getElement("actionFlag").getAttribute("flagName"));
+        Assert.assertEquals("0",rules.getElement("filterRule").getElement("nestedRule").getElement("filterActions").getElement("actionFlag").getAttribute("index"));
     }
```
rohan-ambasta: There any many failures similar to the above test case. I think we should use xmlunit for this. It is under apache license 2.0, and we have to use only for testing purpose. So, I don't think we will have any issues for legal approval. Basically, this will reduce our coding effort and will be reusable 
http://www.xmlunit.org/

sneha-patil-synacor: Raised legal approval request for the same

PR #67, opened 2017-04-20
Bugfix/ZCS-1105 TestMountpoint

fatkudu:


store/src/java/com/zimbra/qa/unittest/TestMountpoint.java
```diff
@@ -22,66 +22,80 @@
...
-    
-    public void setUp()
-    throws Exception {
-        cleanUp();
-    }
-    
-    public void tearDown()
-    throws Exception {
-        cleanUp();
+    private static String USER_NAME = "user1";
```
grishick: Please use test-specific user names

fatkudu: These are duff initializations.  setUp() changes them to test specific values.  I can submit another pull request if you think it is important.  Certainly confusing.

grishick: ok, that's fine then.

PR #69, opened 2017-04-20
ZCS-1108 TestSearchConv

fatkudu:
This still intermittently fails.  I think there may be a genuine bug underlying this
where **SearchConvResponse** contains the messages out of order.  Comments suggest they should be in reverse age, which would imply the message IDs should be descending but when I see
failures, I see the messages ordered e.g.:

        <m rev="7" s="2280" d="1492727425000" id="261" l="2" cid="260">
        <m rev="8" s="2282" d="1492727425000" id="262" l="2" cid="260">
        <m rev="6" s="2283" d="1492727424000" f="u" id="259" l="2" cid="260">
        <m rev="2" s="539" d="1492727415000" f="su" id="257" l="5" cid="260">

Suggest we stick with this for now and if it continues to fail intermittently, either investigate more or lob a bug in and remove this test from the set that gets run.

### DIFF
--- a/store/src/java/com/zimbra/qa/unittest/TestSearchConv.java
+++ b/store/src/java/com/zimbra/qa/unittest/TestSearchConv.java
@@ -89,6 +89,8 @@ public class TestSearchConv {
         Collections.reverse(msgIds);
     }
 
+    // Newer versions of JUnit would allow us to choose more meaningful test names based on the arguments
+    // e.g. @Parameters(name= "{index}: searchConversation({0},{1},{2},{3})")
     @Parameterized.Parameters()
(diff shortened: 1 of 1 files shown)


gordyt:
@fatkudu is looks like possibly the mis-ordering occurs because the "d" (date?) value is the same for two of the messages:

    These two:
    <m rev="7" s="2280" d="1492727425000" id="261" l="2" cid="260">
    <m rev="8" s="2282" d="1492727425000" id="262" l="2" cid="260">

    <m rev="6" s="2283" d="1492727424000" f="u" id="259" l="2" cid="260">
    <m rev="2" s="539" d="1492727415000" f="su" id="257" l="5" cid="260">

PR #72, opened 2017-04-21
ZCS-1142:adding msgs folder to junit classpath to fix unit test failures

sneha-patil-synacor:
adding msgs folder to junit classpath to fix unit test failures
Fixed following unit test failures:
ZCS-1142, ZCS-1143

### DIFF
--- a/build-common.xml
+++ b/build-common.xml
@@ -116,6 +116,8 @@
     <property name="client.classes.dir" location="${client.dir}/build/classes"/>
     <property name="client.jarfile" location="${client.dir}/build/zimbraclient.jar"/>
 
+    <property name="msgs.dir" location="${zm-mailbox.basedir}/store-conf/conf/msgs"/>
+
     <!-- Emma -->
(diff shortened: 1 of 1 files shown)


fatkudu:
Please change the commit message to mention at least on test that fails as a result of this change.

sneha-patil-synacor:
changed the commit message

build-common.xml
```diff
@@ -356,6 +356,9 @@
         <copy file="${test.src.dir}/log4j-test.properties" tofile="${test.classes.dir}/log4j.properties" failonerror="false"/>
         <junit fork="yes" printsummary="on" showoutput="true" failureproperty="junit.failure" tempdir="${single.test.dir}">
             <classpath refid="test.class.path"/>
+            <classpath>
+                <pathelement path="../store-conf/conf/msgs"/>
+            </classpath>
             <assertions><enable/></assertions>
```
rohan-ambasta: May be you can define the folder as a variable, like we are doing for other directory paths.
And add this directory in test.class.path

PR #77, opened 2017-04-22
Feature/zms 288

grishick:
ZContact will stream VCARD from mailstore to IMAP similar to ZMessage. ZMessage uses  ContentServlet, which does not support Contacts, but UserServlet does, so using UserServlet instead of ContentServlet.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/GetContacts.java
+++ b/store/src/java/com/zimbra/cs/service/mail/GetContacts.java
@@ -34,6 +34,10 @@ import com.zimbra.cs.service.util.ItemId;
 import com.zimbra.cs.service.util.ItemIdFormatter;
 import com.zimbra.cs.session.PendingModifications.Change;
 import com.zimbra.soap.ZimbraSoapContext;
+import com.zimbra.soap.mail.message.GetContactsRequest;
+import com.zimbra.soap.mail.message.GetMsgRequest;
+import com.zimbra.soap.type.AttributeName;
+import com.zimbra.soap.type.Id;
 
 /**
  * @since May 26, 2004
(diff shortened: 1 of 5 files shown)


store/src/java/com/zimbra/cs/service/mail/GetContacts.java
```diff
@@ -74,48 +79,50 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
+        if(reqAttrs != null && reqAttrs.size() > 0) {
+            memberAttrs = new ArrayList<String>();
+            for(AttributeName attrName : reqAttrs) {
+                memberAttrs.add(attrName.getName());
+            }
+        }
+
+        //MailConstants.E_CONTACT
+        List<Id> contactIds = req.getContacts();
+        if(contactIds != null && contactIds.size() > 0) {
```
fatkudu: It looks to me as if there is a bug in the JAXB.  The old code appears to allow multiple E_CONTACT elements, each with an id attribute which may be a comma separated list of IDs.  That seems a bit weird but we should probably preserve that behavior?  So, `req.getContacts()` should return `List<IdList>` where `IdList` is a new type whose ID attribute can be a list of Ids?

store/src/java/com/zimbra/cs/service/mail/GetContacts.java
```diff
@@ -203,7 +210,14 @@ static void partitionItems(ZimbraSoapContext lc, ArrayList<ItemId> ids, ArrayLis
     throws ServiceException {
         List<Element> responses = new ArrayList<Element>();
 
-        // note that we removed all <cn> elements from the request during handle(), above
+        //remove all 'contact' elements from original request
+        for (Element e : request.listElements()) {
```
fatkudu: There is a `List<Element> listElements(String name)` method too which may be worth using?

grishick: old code was also detaching elements whilst looping (line 107 in old code), so I assumed it worked :)

store/src/java/com/zimbra/cs/service/mail/GetContacts.java
grishick: you are right. soap.txt:

<GetContactsRequest [sortBy="nameAsc|nameDesc"] [sync="1"] [l="{folder-id}"] [derefGroupMember="1"] 
    [returnHiddenAttrs="1"] [maxMembers="{max-members}"] [returnCertInfo="1"]>
  <a n="..."/>*
  <ma n="..."/>*
  <cn id="{contact-id}[,{contact-id}...]"/>*
</GetContactsRequest>

PR #82, opened 2017-04-24
account for symbolic links when checking paths in zimlet archives

grishick:
Fix for https://bugzilla.zimbra.com/show_bug.cgi?id=107843
Check zimlet files path against symlink-resolved /opt/zimbra/zimlets-deployed.


plobbes:
LGTM

PR #83, opened 2017-04-24
ZCS-1116 JUnit Updates

tmclane:
```

<RunUnitTestsResponse numFailed="0" numSkipped="0" numExecuted="3" xmlns="urn:zimbraAdmin">
  <results>
    <completed name="testSingleTask" class="com.zimbra.qa.unittest.TestScheduledTaskManager" execSeconds="4.88"/>
    <completed name="testRecurringTask" class="com.zimbra.qa.unittest.TestScheduledTaskManager" execSeconds="1.98"/>
    <completed name="testTaskProperties" class="com.zimbra.qa.unittest.TestScheduledTaskManager" execSeconds="0.32"/>
  </results>
</RunUnitTestsResponse>
```

### DIFF
--- a/store/src/java/com/zimbra/qa/unittest/TestScheduledTaskManager.java
+++ b/store/src/java/com/zimbra/qa/unittest/TestScheduledTaskManager.java
@@ -34,12 +34,12 @@ extends TestCase {
 
     static final String TASK_NAME = "TestTask";
     private static final String USER_NAME = "user1";
-    
+
     public void setUp()
     throws Exception {
         cleanUp();
     }
-    
+
     /**
      * Confirms that a single task is persisted to the database,
      * runs, and is then removed from the database automatically.
@@ -47,7 +47,7 @@ extends TestCase {
     public void testSingleTask()
     throws Exception {
         checkNumPersistedTasks(0);
-        
+
         // Schedule a single-execution task
         TestTask task = new TestTask();
         long now = System.currentTimeMillis();
@@ -55,15 +55,15 @@ extends TestCase {
         Mailbox mbox = TestUtil.getMailbox(USER_NAME);
         task.setMailboxId(mbox.getId());
         ScheduledTaskManager.schedule(task);
-        
+
         // Make sure the task is persisted
         checkNumPersistedTasks(1);
         Thread.sleep(1250);
-        
+
         assertEquals("TestTask was not called", 1, task.getNumCalls());
         checkNumPersistedTasks(0);
     }
-    
+
     /**
      * Confirms that a recurring task is persisted to the database,
      * runs multiple times, and is then removed from the database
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/qa/unittest/TestScheduledTaskManager.java
```diff
@@ -29,21 +35,23 @@
...
     static final String TASK_NAME = "TestTask";
     private static final String USER_NAME = "user1";
+    private static Mailbox mbox;
 
-    public void setUp()
-    throws Exception {
-        cleanUp();
+
+    @Before
+    public void setUp() throws Exception {
```
gordyt: @tmclane Should this method call `tearDown()`?  That seems to be the pattern that most of these test updates are following.

tmclane: I actually removed the call on purpose because it was causing issues.  The reason being that it was using a null mbox attribute.

I could be convinced to add it back in I suppose.

tmclane: Please checkout the most recent commit in response to this comment.  If the call is there I have to check for null before calling mbox.getId()

store/src/java/com/zimbra/qa/unittest/TestScheduledTaskManager.java
```diff
@@ -29,115 +35,115 @@
 import com.zimbra.cs.mailbox.ScheduledTask;
 import com.zimbra.cs.mailbox.ScheduledTaskManager;
 
-public class TestScheduledTaskManager
-extends TestCase {
-
+public class TestScheduledTaskManager {
     static final String TASK_NAME = "TestTask";
     private static final String USER_NAME = "user1";
```
grishick: tests should use test-specific user names instead of "user1", "user2", etc

store/src/java/com/zimbra/qa/unittest/TestScheduledTaskManager.java
```diff
@@ -29,115 +35,115 @@
...
         ScheduledTaskManager.cancel(TestTask.class.getName(), task.getName(), mbox.getId(), true);
         checkNumPersistedTasks(0);
     }
 
-    public void tearDown()
-    throws Exception {
-        cleanUp();
+    @After
+    public void tearDown() throws Exception {
+        if (mbox != null)
```
grishick: nit: formatting. Please keep "{" on the same line as "if"

tmclane:
@gordyt @grishick @ilyusha @fatkudu This has been updated.  Please re-review.

store/src/java/com/zimbra/qa/unittest/TestScheduledTaskManager.java
```diff
@@ -29,115 +35,117 @@
...
+    @Rule
+    public static TestName testInfo = new TestName();
 
     static final String TASK_NAME = "TestTask";
-    private static final String USER_NAME = "user1";
-    
-    public void setUp()
-    throws Exception {
-        cleanUp();
+    private static final String USER_NAME = String.format("%s-%s-user1", TestScheduledTaskManager.class.getSimpleName(), testInfo.getMethodName()).toLowerCase();
```
grishick: this does not break the test, but as in the other PR, getMethodName() will probably return an empty string in this context.

tmclane:
@grishick Updated to move the USER_NAME initialization into setUp like the other tests.

PR #87, opened 2017-04-26
ZCS-1291 Deprecate old sieve ldap attributes and add new

rohan-ambasta:
1. Deprecated old sieve attributes with id 2090, 2091, 2091, 2093, 2094, 2095
2. Added new attributes which can be used in place of the above deprecated attributes. All the attributes are defined at account, cos and domain level.
3. Since there is confusion about the release version, right now I have kept as 8.7.8. Please let me know if that needs to be changed

Testing - 
ant test

Will send a separate pull request for automation


grishick:
Looks like we are deprecating a bunch of attributes we just added, because we added them with wrong inheritance and optionalIn values. Isn't there a more graceful way to fix the code so that it allows changing those parameters without having to replace the attributes with new ones? This is just making such a mess in our schema...

rohan-ambasta:
@grishick I agree with you, but I guess we don't have any other option here

PR #91, opened 2017-04-27
feature/zms-38 updates

gordyt:
### Ehcache

Update Ehcache initialization as follows:

1. When IMAP is running inside of _mailboxd_, initialize the
   persistence layer of Ehcache as before; that it, share the
   same file for persisting cache data as _mailboxd_.
2. When IMAP is not running inside of _mailboxd_, use a different
   file for persisting cache data.  This prevents issues when
   running a decoupled IMAP server on the same host as
   mailboxd.
   
### New LDAP Attributes

The following new LDAP Attributes determine with IMAP servers get started by _zimbra-imap_:

* For IMAP: `zimbraRemoteImapServerEnabled1`
* For IMAPS: `zimbraRemoteImapSSLServerEnabled`

They are equivalent to the settings that control IMAP(S) for _mailboxd_:

* `zimbraImapServerEnabled`
* `zimbraImapSSLServerEnabled`

These new LDAP attributes determine the backend ports that the IMAP(S) servers will use:

* For IMAP: `zimbraRemoteImapBindPort`
* For IMAPS: `zimbraRemoteImapSSLBindPort`

### ImapDaemon

The _zimbra-imap_ service uses a bash script defined in the `zm-core-util` repo called `zmimapdctl` to construct the appropriate JVM options and start the _ImapDaemon_.  The _ImapDaemon_ initializes logging and starts up the IMAP(S) server(s).

### imap.log4j.properties

This file is installed into `/opt/zimbra/conf` and is used by the _ImapDaemon_ to initialize logging.

### RemoteImapConfig

The _RemoteImapConfig_ class is used by the _ImapDaemon_ when initializing the IMAP(S) server(s).  It supports the alternate port bindings defined by the following LDAP attributes:

* For IMAP: `zimbraRemoteImapBindPort`
* For IMAPS: `zimbraRemoteImapSSLBindPort`

Please also see related pull requests for:

* `zm-zcs-lib`
* `zm-nginx-lookup-store`
* `zm-core-utils`
* `zm-build`

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/EhcacheImapCache.java
+++ b/store/src/java/com/zimbra/cs/imap/EhcacheImapCache.java
@@ -45,7 +45,12 @@ final class EhcacheImapCache implements ImapSessionManager.Cache<String, ImapFol
 
     @SuppressWarnings("serial")
     EhcacheImapCache(String name, boolean active) {
-        ehcache = EhcacheManager.getInstance().getEhcache(name);
+        // If running inside mailboxd, share mailboxd cache, else use separate imap cache.  This avoids issues when running
+        // decoupled IMAP service on same host as mailbox.
+        EhcacheManager.Service service = System.getProperty(ImapDaemon.IMAP_SERVER_EMBEDDED, "true").equals("false")
+                ? EhcacheManager.Service.IMAP
+                : EhcacheManager.Service.MAILBOX;
+        ehcache = EhcacheManager.getInstance(service).getEhcache(name);
         this.active = active;
         if (active) {
             activeCacheUpdateTimes = new LinkedHashMap<String, Long>(ACTIVE_CACHE_THRESHOLD, 0.75f, true) {
--- a/store/src/java/com/zimbra/cs/imap/ImapDaemon.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapDaemon.java
@@ -1,0 +1,148 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2004, 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ * ***** END LICENSE BLOCK *****
+ */
+
+package com.zimbra.cs.imap;
+
+import java.io.FileInputStream;
+import java.util.Properties;
+
+import org.apache.log4j.PropertyConfigurator;
+
+import com.zimbra.common.localconfig.LC;
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.cs.account.Provisioning;
+import com.zimbra.common.util.ZimbraLog;
+
+
+public class ImapDaemon {
+
+    public static final String IMAPD_LOG4J_CONFIG = "/opt/zimbra/conf/imapd.log4j.properties";
+    /**
+     * When starting IMAP(S) from ImapDaemon, a System property with the following key will
+     * be set with a value of "false".  This is referenced by the EhcacheImapCache to determine
+     * how the cache is initialized.
+     */
+    public static final String IMAP_SERVER_EMBEDDED = "imap.server.embedded";
+
+    private ImapServer imapServer, imapSSLServer;
+
+    private ImapServer startImapServer(boolean ssl) throws ServiceException {
+        RemoteImapConfig config = new RemoteImapConfig(ssl);
+        ZimbraLog.imap.info("Starting IMAP server, port=%d", config.getBindPort());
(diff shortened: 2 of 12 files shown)


store/src/java/com/zimbra/cs/imap/ImapDaemon.java
```diff
@@ -0,0 +1,137 @@
...
+    private ImapServer startImapServer(boolean ssl) throws ServiceException {
+        RemoteImapConfig config = new RemoteImapConfig(ssl);
+        ZimbraLog.imap.info("Starting IMAP server, port=%d", config.getBindPort());
+        ImapServer server = LC.nio_imap_enabled.booleanValue() ?
+            new NioImapServer(config) : new TcpImapServer(config);
+        server.start();
+        return server;
+    }
+
+	private int startServers() throws ServiceException {
```
plobbes: Is it just me, or did you indentation get off here (and continue for a while)?

store/src/java/com/zimbra/cs/util/EhcacheManager.java
```diff
@@ -69,6 +69,13 @@ private EhcacheManager() {
     }
 
     public static EhcacheManager getInstance() {
+        return getInstance(Service.MAILBOX);
+    }
+
+    public static EhcacheManager getInstance(Service service) {
```
plobbes: Does this method need to be 'syncronized' to avoid race conditions?

fatkudu:
I'm not sure on ImapDaemon as a name on its own.  I think we need something to clearly differentiate between Imap running with the Mailbox JVM and outside.
Perhaps StandaloneImapDaemon?

store/src/java/com/zimbra/cs/imap/ImapDaemon.java
```diff
@@ -0,0 +1,137 @@
...
+	    int cnt = 0;
+        if (isEnabled(Provisioning.A_zimbraRemoteImapServerEnabled)) {
+            imapServer = startImapServer(false);
+            cnt += 1;
+        } else {
+            System.err.println(Provisioning.A_zimbraRemoteImapServerEnabled + " is FALSE");
```
fatkudu: Why not use logging?

store-conf/conf/imap.log4j.properties
```diff
@@ -0,0 +1,53 @@
+# An example log4j configuration file that outputs to System.out.  The
```
fatkudu: Remove this part of the comment - I assume it isn't true...

Have you considered whether we need an analog to `log4j.properties.in`?

store-conf/conf/imap.log4j.properties
```diff
@@ -0,0 +1,53 @@
...
+log4j.rootLogger=INFO,LOGFILE
+#log4j.rootLogger=INFO,LOGFILE,SYSLOG
+
+# Appender LOGFILE writes to the file "/opt/zimbra/log/imap-log4j.log".
+# Daily rolling policy with compressing the old log file while rotating!!
+# The archived log files location can be changed using FileNamePattern value
+log4j.appender.LOGFILE=org.apache.log4j.rolling.RollingFileAppender
+log4j.appender.LOGFILE.RollingPolicy=org.apache.log4j.rolling.TimeBasedRollingPolicy
+log4j.appender.LOGFILE.RollingPolicy.FileNamePattern=/opt/zimbra/log/imap-log4j.log.%d{yyyy-MM-dd}
+log4j.appender.LOGFILE.File=/opt/zimbra/log/imap-log4j.log
```
fatkudu: Why not just `imapd.log` or `imapd-standalone.log`?  I think the log4j bit is self-evident.

store/src/java/com/zimbra/cs/imap/RemoteImapConfig.java
```diff
@@ -0,0 +1,38 @@
...
+    public RemoteImapConfig(boolean ssl) {
+        super(ssl);
+    }
+
+    @Override
+    public int getBindPort() {
+        return isSslEnabled() ?
+            getIntAttr(A_zimbraRemoteImapSSLBindPort, D_REMOTE_IMAP_SSL_BIND_PORT) :
+            getIntAttr(A_zimbraRemoteImapBindPort, D_REMOTE_IMAP_BIND_PORT);
+    }
```
fatkudu: Could create toString() methods for this and ImapConfig if it doesn't have one already which give all the relevant details.  Then logging that currently digs inside this (I think I saw something) could just log all the info.

store/src/java/com/zimbra/cs/imap/EhcacheImapCache.java
```diff
@@ -45,7 +45,10 @@
 
     @SuppressWarnings("serial")
     EhcacheImapCache(String name, boolean active) {
-        ehcache = EhcacheManager.getInstance().getEhcache(name);
+        // If running inside mailboxd, share mailboxd cache, else use separate imap cache.  This avoids issues when running
+        // decoupled IMAP service on same host as mailbox.
+        EhcacheManager.Service service = System.getProperty("jetty.home", null) == null ? EhcacheManager.Service.IMAP : EhcacheManager.Service.MAILBOX;
```
grishick: I think it is better to rely on a property that we can explicitly set on the JVM rather then on an absence of a property.

store/src/java/com/zimbra/cs/util/EhcacheManager.java
grishick: I think it does.

tmclane:
FYI: The CentOS build and install worked fine but I am running into an issue with the mailbox talking to the imap server.  Looking into it now.

store/src/java/com/zimbra/cs/imap/ImapDaemon.java
grishick: I am seeing it too

store-conf/conf/imap.log4j.properties
grishick: I am OK with not adding a *properties.in file until we need to use variables in it, which AFAIK is why we have all *.in files.

PR #94, opened 2017-04-28
ZMS-261 getImapUid support

fatkudu:


store/src/java/com/zimbra/qa/unittest/TestZClient.java
```diff
@@ -29,10 +33,10 @@
 import java.util.List;
 import java.util.Set;
 
-import junit.framework.Assert;
-import junit.framework.TestCase;
-
 import org.apache.commons.io.IOUtils;
+import org.junit.After;
```
grishick: This seems like an unnecessary change that causes a lot more non-functional changes throughout the rest of the file. I don't think it achieves anything other then making the diff very big.

client/src/java/com/zimbra/client/ZMessage.java
```diff
@@ -561,13 +566,33 @@ public MailItemType getMailItemType() {
...
     @Override
     public int getImapUid() {
-        throw new UnsupportedOperationException("ZMessage method not supported yet");
+        if ((null != imapUid) && (imapUid >= 0)) {
+            return imapUid;
+        }
+        ZimbraLog.mailbox.debug("ZMessage getImapUid() - regetting UID");
+        ZMessage zm = null;
+        try {
+            zm = getMailbox().getMessageById(mId);
```
fatkudu: The cache lookup misses if different things were requested - I need another look to check that the lookup fails if Uid wasn't asked for before.

client/src/java/com/zimbra/client/ZContact.java
```diff
@@ -437,8 +443,31 @@ public InputStream getContentStream() throws ServiceException {
...
+        if ((null != imapUid) && (imapUid >= 0)) {
+            return imapUid;
+        }
+        if (isGalContact()) {
+            return 0;  /* TODO: is this the best thing to do? */
+        }
+        ZimbraLog.mailbox.debug("ZContact getImapUid() - regetting UID");
+        ZContact zc = null;
+        try {
+            zc = getMailbox().getContact(mId);
```
fatkudu: I assumed this worked the same as for messages but I don't think it does.  Needs more research.

fatkudu: I think this code is correct.  Added tests that exercise it.

client/src/java/com/zimbra/client/ZMessage.java
fatkudu: Added tests that exercise the code.  I think there are instances when it may be needed.

store/src/java/com/zimbra/qa/unittest/TestZClient.java
fatkudu: Changes now much smaller.  Could do with JUnit 4 changes at some point - but probably better on the develop branch.

fatkudu:
Getting a couple of failures in TestRemoteImapShared -  
`testListContactsContents` fails because:
Caused by: java.lang.UnsupportedOperationException: ZContact method not supported yet
    at com.zimbra.client.ZContact.getContentStream(ZContact.java:443)
`testUidCopy` fails because although it claims to succeed, some info is missing.  I think I saw that Greg had taken on an Imap copy fix, so I'm hoping it will be addressed by that.  Otherwise can look at it later.

grishick:
I pushed a small commit to address ZContact::getContentStream(). You can try rebasing or just plug this simple implementation:
`    @Override
    public InputStream getContentStream() throws ServiceException {
        return mMailbox.getRESTResource(String.format("?id=%s", this.getId()));
    }`

PR #108, opened 2017-05-03
Added target for QA users (especially windows) to use publish-local target and renamed project names

jiteshsojitra:
Added target for QA users (especially windows) to use publish-local target without native make dependency

Renamed project names according to modified structure

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -11,6 +11,14 @@
       <ant dir="./store"  target="create-version-sql" inheritAll="true"/>
    </target>
 
+   <target name="publish-local">
+      <ant dir="./native" target="publish-local" inheritAll="true"/>
+      <ant dir="./common" target="publish-local" inheritAll="true"/>
+      <ant dir="./soap"   target="publish-local" inheritAll="true"/>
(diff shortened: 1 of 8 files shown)


build.xml
```diff
@@ -11,6 +11,14 @@
       <ant dir="./store"  target="create-version-sql" inheritAll="true"/>
    </target>
 
+   <target name="publish-local">
+      <ant dir="./native" target="publish-local" inheritAll="true"/>
```
fatkudu: This is quite similar to a definition in `build-common.xml` which is included in the sub-directory build.xml files.  To prevent duplication, could just do:

        <ant dir="./store" target="publish-foss-local" inheritAll="true"/>

Or alternatively, make that task use your new task?

Could also do something similar with the build-common.xml task "deploy-foss"

fatkudu:
On the importance of keeping project names the same - worth adding a comment in the files?  I suspect this may not be the only attempt to rename the projects.

jiteshsojitra:
Sending

PR #109, opened 2017-05-03
Adding publish-local-all target for QA users (windows) without make dependency.

jiteshsojitra:
Adding publish-local-all target for QA users (windows) without make dependency.


jiteshsojitra:
Ok, so many things for such a small change due to multiple xml files currently. I'm sending fresh review request, thanks all.

PR #111, opened 2017-05-03
Feature/zms 346

grishick:
In addition to new SOAP tests, I've also tested remote IMAP server with apple mail. When mailbox server gets disconnected (tried ifdown/ifup on mailbox server), remote IMAP is able to continue communication after mailbox server connects back. When mailboxd on mailbox server restarts, remote IMAP recreates waitsets. When mailbox server stops responding, remote IMAP drops IMAP sessions for users on the non-responding mailbox server. When mailbox server comes back up, apple mail sends an AUTHENTICATE request and communication between IMAP client and IMAP server gets restored.

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/ImapPath.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapPath.java
@@ -560,6 +560,10 @@ public class ImapPath implements Comparable<ImapPath> {
 
     boolean isVisible() throws ServiceException {
         boolean isMailFolders = Provisioning.getInstance().getLocalServer().isImapDisplayMailFoldersOnly();
+        if(folder == null) {
+            ZimbraLog.imap.error("ImapFolderStore is null. This session was likely terminated.");
+            throw ServiceException.TEMPORARILY_UNAVAILABLE();
+        }
         // check the folder type before hitting a remote server if it's a mountpoint
         if (!(folder.isVisibleInImap(isMailFolders))) {
             return false;
--- a/store/src/java/com/zimbra/cs/imap/ImapServerListener.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapServerListener.java
@@ -20,6 +20,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Future;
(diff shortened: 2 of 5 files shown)


store/src/java/com/zimbra/cs/imap/ImapPath.java
```diff
@@ -560,6 +560,10 @@ boolean isSelectable() throws ServiceException {
 
     boolean isVisible() throws ServiceException {
         boolean isMailFolders = Provisioning.getInstance().getLocalServer().isImapDisplayMailFoldersOnly();
+        if(folder == null) {
+            ZimbraLog.imap.error("ImapFolderStore is null. This session was likely terminated.");
```
tmclane: Would this be more useful as a 'debug' level instead since it is likely to occur in production and isn't truly an 'error' per se?

store/src/java/com/zimbra/cs/imap/ImapServerListener.java
```diff
@@ -255,10 +338,9 @@ private synchronized void processAdminWaitSetResponse(AdminWaitSetResponse wsRes
         }
         if(wsResp.getCanceled() != null && wsResp.getCanceled().booleanValue() && wsID.equalsIgnoreCase(respWSId)) {
             //this waitset was canceled
-            //TODO: figure out what to do with listeners if they are still registered
-            wsID = null;
-            lastSequence.set(0);
-            ZimbraLog.imap.debug("AdminWaitSet %s was canceled", respWSId);
+            ZimbraLog.imap.warn("AdminWaitSet %s was cancelled", respWSId);
```
fatkudu: Is this an unexpected situation?  If not then debug logging may make more sense.

store/src/java/com/zimbra/cs/imap/ImapServerListener.java
```diff
@@ -185,11 +234,21 @@ private void initWaitSet(String accountId, boolean alreadyListening) throws Serv
...
+
         try {
+            ZimbraLog.imap.debug("Sending initial AdminWaitSetRequest. WaitSet ID: %s", wsID);
+            AdminWaitSetResponse wsResp = soapProv.invokeJaxbAsAdminWithRetry(waitSetReq, server);
             processAdminWaitSetResponse(wsResp);
-            continueWaitSet();
+        } catch (SoapFaultException e) {
+            if(AdminServiceException.NO_SUCH_WAITSET.equalsIgnoreCase(e.getCode())) {
+                //waitset is gone. Create a new one
+                ZimbraLog.imap.warn("AdminWaitSet %s does not exist anymore", wsID);
```
fatkudu: Looks below like this might not be that rare - debug level?

grishick: NO_SUCH_WAITSET is not expected under normal conditions. It will happen if remote mailboxd was restarted or if a sweeper on remote mailboxd killed this waitset. The latter may  indicate that some configuration settings should be updated.

store/src/java/com/zimbra/cs/imap/ImapServerListener.java
grishick: waitset gets cancelled before being deleted, so this situation may indicate that remote server should allow more waitsets.

store/src/java/com/zimbra/cs/imap/ImapPath.java
grishick: This scenario will happen when remote server restarts and it causes IMAP client session to be closed. So far I only tested this with Apple Mail and Apple Mail did not complain, it quietly created a new session. I suspect some IMAP clients may complain, so I would leave this as error until we get more data from testing in our hosting environment.

PR #112, opened 2017-05-04
ZCS-1286:Regression-ModifyOutgoingFilterRulesRequest fails

sneha-patil-synacor:
Removing changes related to bug ZCS-540

After changes of ZCS-540, the existence of the folder/tag is always checked and if not present, the filter doesn't get activated. 
But, the folder and tag get created at runtime (if they do not exist) for fileinto and tag action.
So, ZCS-540 does not need to be fixed.


sneha-patil-synacor:
Added Junit test

PR #115, opened 2017-05-05
ZCS-1281:NPE for delete folder action for user account having elsif c…

sneha-patil-synacor:
Issue:
NullPointerException for delete folder action for user account having elsif condition sieve script

Fix:
Added literal elsif to RULE_NODE_NAMES so that isRuleNode() returns true for elsif in SieveVisitor


rohan-ambasta:
I think you need to commit this to develop branch. Please confirm with @rupalid

PR #116, opened 2017-05-05
ZMS-227: ZMailbox::searchImap

ilyusha:
For clarity, I renamed this interface method from "search" to "searchImap" since it is only used by the IMAP server and only returns IMAP-compatible results.

### Client changes
 - A new interface _ZImapSearchHit_ is introduced. This interface combines _ZSearchHit_ (the base for all client-side search results) and _ZimbraQueryHit_ (returned by _ZimbraQueryHitResults_). Three search hit classes now implement _ZImapSearchHit_:
   - _ZMessageHit_ has new _imapUid_ and _modSeq_ fields that default to -1 unless _resultMode_ is set IMAP or MODSEQ.
   - _ZContactHit_ has new _imapUid_ and _modSeq_ fields like _ZMessageHit_, but always returns -1 for the parent ID. To get the _getFlagBitmask_ method to work, the _ZContact.Flag_ enum had to be updated with a method to convert the string to a bitmask, similar to the Flag class in zm-store.
   - _ZIdHit_ returns -1 or null for all _ZImapSearchHit_ methods except for _getItemId()_.
 - _ZSearchResult_ has a new field _imapHits_, exposed via _getImapHits()_, which is a list of _ZImapSearchHit_ objects returned in the search response.
 - A new class _ZRemoteQueryHitResults_ implements _ZimbraQueryHitResults_ by iterating over the imapHits described above. _Zmailbox::searchImap_ returns a _ZRemoteQueryHitResults_ instance.

### Server changes
In order for this to work, the server had to be updated to return the _modSeq_ and _imapUid_ values in the search response for messages and contacts. The following changes were made:
 - New MODSEQ bitmask in _PendingModifications.Change_
 - Updated _ToXML.NOTIFY_FIELDS_ bitmask to not include MODSEQ and IMAP_UID
 - Updated _SearchResponse_ class to add the MODSEQ and/or IMAP_UID masks to the _field_ bitmask when appropriate (both if _resultMode_ is IMAP; MODSEQ only if _resultMode_ is MODSEQ)
 - Added new _ToXML::encodeContact_ method signature to be able to pass in a field bitmask
 - Updated _ToXML::encodeContact_ and _ToXML::encodeMessageCommon_ to add the new fields when necessary

### Unit Tests
 - Added _testImapSearch_ and _testImapSearchContact_ tests in _TestZClient_. These tests compare the fields present in the response from _ZClient::searchImap_ to the values on the actual message/contact object. This is done for _resultMode_=IMAP, MODSEQ, and ID.
 - Re-activated the search unit tests in _TestRemoteImapShared_; they all pass.

### DIFF
--- a/client/src/java/com/zimbra/client/ZContactHit.java
+++ b/client/src/java/com/zimbra/client/ZContactHit.java
@@ -17,24 +17,25 @@
 
 package com.zimbra.client;
 
-import com.zimbra.common.mailbox.ContactConstants;
-import com.zimbra.common.service.ServiceException;
-import com.zimbra.common.soap.Element;
-import com.zimbra.common.soap.MailConstants;
-import com.zimbra.common.zclient.ZClientException;
-import com.zimbra.client.ZContact.Flag;
-import com.zimbra.client.ZContact.ZContactAttachmentInfo;
-import com.zimbra.client.event.ZModifyContactEvent;
-import com.zimbra.client.event.ZModifyEvent;
-import org.json.JSONException;
-
-import java.io.InputStream;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 
-public class ZContactHit implements ZSearchHit {
+import org.json.JSONException;
+
+import com.zimbra.client.ZContact.Flag;
+import com.zimbra.client.event.ZModifyContactEvent;
+import com.zimbra.client.event.ZModifyEvent;
+import com.zimbra.common.mailbox.ContactConstants;
+import com.zimbra.common.mailbox.ItemIdentifier;
+import com.zimbra.common.mailbox.MailItemType;
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.soap.Element;
+import com.zimbra.common.soap.MailConstants;
+import com.zimbra.common.zclient.ZClientException;
+
+public class ZContactHit implements ZImapSearchHit {
 
     private String mId;
     private String mFlags;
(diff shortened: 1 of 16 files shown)


client/src/java/com/zimbra/client/ZContactHit.java
```diff
@@ -351,4 +356,39 @@ private String get(Map<String,String> attrs, String key, String defaultValue) {
...
     }
+
+    @Override
+    public int getItemId() throws ServiceException {
+        return new ItemIdentifier(mId,  null).id;
+    }
+
+    @Override
+    public int getParentId() throws ServiceException {
+        return -1;
```
fatkudu: Seems unlikely to be correct?  Although it may not matter.  Worth a comment anyway.

ilyusha: Yeah I was actually wondering about this myself, but it seems like when you create a contact the parent is set to -1. There may be times when this is not the case, but for IMAP I don't think it really matters.

client/src/java/com/zimbra/client/ZContact.java
```diff
@@ -96,10 +97,12 @@ public long getLength() {
...
     /* Note for contacts, use a more limited set of flags than in ZItem */
     public enum Flag {
-        flagged('f'),
-        attachment('a');
+        flagged(-6, 'f'),
+        attachment(-2, 'a');
 
         private char mFlagChar;
+        final int id;
+        final int bitmask;
```
grishick: is this duplicating ZItem.Flag? There is already ZItem::Flag::toBitmask that mimics behavior of Flag class in zm-store.

ilyusha: Ah, looks like it is. I think when I looked for the existing code initially, I accidentally looked at at the ZItem class in the old zm-store repo, which doesn't have the toBitmask method.

ilyusha: I removed the ZContact.Flag commit and changed ZContactHit::getFlagBitmask to use ZItem::Flag::toBitmask

PR #121, opened 2017-05-08
avoid NPE when account does not exist

grishick:


store/src/java/com/zimbra/cs/imap/ImapServerListenerPool.java
```diff
@@ -58,10 +58,14 @@ public static ImapServerListenerPool getInstance() {
 
     public ImapServerListener get(ZMailbox zmbox) throws ServiceException {
         Account acct = Provisioning.getInstance().get(AccountBy.id, zmbox.getAccountId());
+        if(acct == null) {
+            ZimbraLog.imap.error("Cannot get ImapServerListener for %s. Account does not exist.", zmbox.getAccountId(), null);
```
ykomiyam: Do we need the null as the 2nd parameter?  If you just want to log an error event without the exception message, I think you can call error(String format, Object ... objects) method.

fatkudu:
Sorry, needs a rebase.

grishick:
rebased and addressed comment by @ykomiyam

PR #122, opened 2017-05-09
ZCS-375 Add envelope test to sieve soap api

rohan-ambasta:
[Story] -
Add support for envelope test in zimbra soap API

[Change] -
- Added EnvelopeTest that extends AddressTest 
- Add methods to convert envelope test soap object into sieve script
- Added sieve to soap visitor methods for envelope test

[Testing] -
- Manual testing for GetFilterRulesRequest and ModifyFilterRequests for envelope test with string
and value comparison
- Added junits for the same

### DIFF
--- a/soap/src/java/com/zimbra/soap/mail/type/FilterTest.java
+++ b/soap/src/java/com/zimbra/soap/mail/type/FilterTest.java
@@ -87,7 +87,7 @@ public class FilterTest {
     }
 
     @XmlAccessorType(XmlAccessType.NONE)
-    public static final class AddressTest extends FilterTest {
+    public static class AddressTest extends FilterTest {
 
         /**
          * @zm-api-field-tag comma-sep-header-names
@@ -210,6 +210,10 @@ public class FilterTest {
                 .add("value", value)
                 .toString();
         }
+    }
+
+    @XmlAccessorType(XmlAccessType.NONE)
+    public static class EnvelopeTest extends AddressTest {
     }
 
     @XmlAccessorType(XmlAccessType.NONE)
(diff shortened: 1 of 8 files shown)


soap/src/java/com/zimbra/soap/mail/type/FilterTest.java
```diff
@@ -213,6 +213,10 @@ public String toString() {
     }
 
     @XmlAccessorType(XmlAccessType.NONE)
+    public static class EnvelopeTest extends AddressTest {
```
sneha-patil-synacor: we can make this class final

store/src/java/com/zimbra/cs/filter/SieveVisitor.java
```diff
@@ -449,16 +459,28 @@ private void acceptTest(Node node, RuleProperties props) throws ServiceException
...
+                    if (valueComparison != null) {
+                        visitAddressTest(node, VisitPhase.begin, props, headers, part, valueComparison, isCount, value);
+                        accept(node, props);
+                        visitAddressTest(node, VisitPhase.end, props, headers, part, valueComparison, isCount, value);
+                    } else {
+                        visitAddressTest(node, VisitPhase.begin, props, headers, part, comparison, caseSensitive, value);
+                        accept(node, props);
+                        visitAddressTest(node, VisitPhase.end, props, headers, part, comparison, caseSensitive, value);
+                    }
+                } else if ("envelope".equalsIgnoreCase(nodeName)) {
```
ykomiyam: Why don't you use "else { ... } in this context?  In general, "else if { ... }" expects the "else { ... }", and it looks as if there might be another option of the nodeName than "address" and "envelope" which is none in this block.

PR #123, opened 2017-05-09
ZCS-1287:Regression-Variables not replaced in notify

sneha-patil-synacor:
Issue:
Header names as variables names if used in the script, were not being replaced with it's corresponding value in mime message. 

Fix:
If no variable with given header name is already declared(with 'set' action), the variable is now being replaced with header value.
If not present in mime header list as well, it should replace with empty string. If present in both, precedence is given to "set" action.

Added junit test for notify.
Executed existing automation MailClient/FilterRules/FilterNotifyReplyAction.xml

### DIFF
--- a/store/src/java/com/zimbra/cs/filter/FilterUtil.java
+++ b/store/src/java/com/zimbra/cs/filter/FilterUtil.java
@@ -375,9 +375,9 @@ public final class FilterUtil {
         }
         replyMsg.setSubject(subject, getCharset(account, subject));
 
-        Map<String, String> vars = getVarsMap(mailbox, parsedMessage, mimeMessage);
-        String body = StringUtil.fillTemplate(bodyTemplate, vars);
-        replyMsg.setText(body, getCharset(account, body));
+        //getVarsMap() result is now being passed to replaceVariables(), so following call is redundant
+        //Map<String, String> vars = getVarsMap(mailbox, parsedMessage, mimeMessage);
+        replyMsg.setText(bodyTemplate, getCharset(account, bodyTemplate));
 
         String origMsgId = mimeMessage.getMessageID();
         if (!StringUtil.isNullOrEmpty(origMsgId))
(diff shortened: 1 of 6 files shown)


store/src/java/com/zimbra/cs/filter/FilterUtil.java
```diff
@@ -893,7 +893,7 @@ public static String replaceVariables(ZimbraMailAdapter mailAdapter, String sour
      * @param sourceStr text string that may contain one or more than one "variable-ref"
      * @return Replaced text string
      */
-    public static String leastGreedyReplace(Map<String, String> variables, String sourceStr) {
+    public static String leastGreedyReplace(Map<String, String> variables, String sourceStr, Map<String, String> headerVars) {
```
chinmay15: Please overload leastGreedyReplace method being public static method

sneha-patil-synacor: Added

store/src/java-test/com/zimbra/cs/filter/NotifyTest.java
```diff
@@ -112,4 +112,37 @@ public void filterValidToField() {
...
 
+    @Test
+    public void testNotifyMailtoWithHeaderVariable() {
+        String sampleMsg = "from: abc@zimbra.com\n"
+                + "Subject: Hello\n"
+                + "to: test@zimbra.com\n";
+        String filterScript = "require [\"enotify\", \"variables\"];\n"
+            + "set \"to\" \"nick\";\n"
+            + "if anyof (header :contains [\"Subject\"] \"Hello\") {\n"
+            + "notify \"test2@zimbra.com\" \"\" \"${SUBJECT} ${to}\"\n"
```
ykomiyam: Please test the notify action (legacy version) using the body-template with the template-specific variables, such as ${SUBJECT}, ${TO}, ${CC}, and ${BODY} (see the docs/soap.txt for more details).  Those are predefined variables for the legacy notify action.

sneha-patil-synacor: Existing automation has been executed for notify action (legacy version). Also, the unit test has been added for legacy notify action.

ykomiyam: But this implementation brakes the backward compatibility.  In the context of the legacy notification action, the value of the ${SUBJECT}, ${TO}, ${CC}, ${BODY}, etc. should be referred from the triggering message.  I think the RFC 5229 complied variable should be applied only to the RFC compiled notification action, not to the legacy version.

chinmay15: Legacy version is not rfc specified. It is zimbra functionality. We have implemented rfc 5229 in a way where we do not break legacy zimbra functionality. Here, we are supporting both rfc and zimbra legacy functionality.

chinmay15: @Prashantsurana correct me if I am wrong

ykomiyam: The following use case breaks the backward compatibility; when the user migrates the filter rule which contains the legacy notify action with ${from}, then he/she adds the 'set' action on it to set some value to the variable name of "from".
But as talking with Rupali, and as considering the possibility of our current customers migrating the legacy notify action (which is low) to the 8.7.x, I was convinced that it can be handled by the well-warned documentation.

PR #133, opened 2017-05-12
ZCS-916:Implement soap API for elsif and else in sieve script

shrikant-prasad:
Implemented soap2sieve and sieve2soap api for elsif and else rules in sieve script
Added Junit tests for sieve2soap changes. Junits for soap2sieve will be sent in separate PR.

### DIFF
--- a/soap/src/java/com/zimbra/soap/mail/type/NestedRule.java
+++ b/soap/src/java/com/zimbra/soap/mail/type/NestedRule.java
@@ -31,21 +31,22 @@ import javax.xml.bind.annotation.XmlType;
 import org.codehaus.jackson.annotate.JsonPropertyOrder;
 
 import com.google.common.base.Objects;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.zimbra.common.soap.MailConstants;
 import com.zimbra.soap.json.jackson.annotate.ZimbraJsonArrayForWrapper;
 
 // JsonPropertyOrder added to make sure JaxbToJsonTest.bug65572_BooleanAndXmlElements passes
 @XmlAccessorType(XmlAccessType.NONE)
-@XmlType(propOrder = {"tests", "actions","child"})
-@JsonPropertyOrder({ "tests", "actions","child"})
-public final class NestedRule {
+@XmlType(propOrder = {"tests", "actions", "child", "elseRules"})
+@JsonPropertyOrder({ "tests", "actions", "child", "elseRules"})
+public class NestedRule {
 
     /**
      * @zm-api-field-description Filter tests
      */
     @XmlElement(name=MailConstants.E_FILTER_TESTS /* filterTests */, required=true)
-    private FilterTests tests;
+    protected FilterTests tests;
 
     /**
      * @zm-api-field-description Filter actions
@@ -69,20 +70,27 @@ public final class NestedRule {
         @XmlElement(name=MailConstants.E_ACTION_LOG /* actionLog */, type=FilterAction.LogAction.class)
     })
     // in nested rule case, actions could be null.
-    private List<FilterAction> actions;
-
+    protected List<FilterAction> actions;
+    
     // For Nested Rule
     /**
      * @zm-api-field-description NestedRule child
      */
     @XmlElement(name=MailConstants.E_NESTED_RULE /* nestedRule */)
-    private NestedRule child;
+    protected NestedRule child;
+
+    /**
+     * @zm-api-field-description Else Rule
+     */
+    @XmlElementWrapper(name=MailConstants.E_ELSE_RULES /* elseRules */, required=false)
+    @XmlElement(name=MailConstants.E_ELSE_RULE /* elseRule */, required=false)
+    protected List<NestedRule> elseRules = null;
 
     /**
      * no-argument constructor wanted by JAXB
      */
     @SuppressWarnings("unused")
-    private NestedRule() {
+    protected NestedRule() {
         this(null);
     }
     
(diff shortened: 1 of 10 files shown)


soap/src/java/com/zimbra/soap/mail/type/NestedRule.java
```diff
@@ -144,12 +152,48 @@ public void setChild(NestedRule nestedRule) {
         child = nestedRule;
     }
 
+    public void setElseRules(Iterable <NestedRule> elseRules) {
+        if(this.elseRules == null){
```
ykomiyam: add a space between "if" and "("

soap/src/java/com/zimbra/soap/mail/type/NestedRule.java
```diff
@@ -144,12 +152,48 @@ public void setChild(NestedRule nestedRule) {
...
+            this.elseRules.clear();
+        }
+
+        if (elseRules != null) {
+            Iterables.addAll(this.elseRules,elseRules);
+        }
+    }
+
+    public void addElseRule(NestedRule elseRule) {
+        if(this.elseRules == null){
```
ykomiyam: ditto

soap/src/java/com/zimbra/soap/mail/type/NestedRule.java
```diff
@@ -144,12 +152,48 @@ public void setChild(NestedRule nestedRule) {
...
+            this.elseRules = Lists.newArrayList();
+        }
+        this.elseRules.add(elseRule);
+    }
+
+    public void addElseRules(Iterable <NestedRule> elseRules) {
+        if(this.elseRules == null){
```
ykomiyam: ditto

store/src/java/com/zimbra/cs/filter/SoapToSieve.java
```diff
@@ -206,6 +240,25 @@ private String handleNest(String baseIndents, NestedRule currentNestedRule) thro
...
+        if(currentNestedRule.getElseRules() != null && currentNestedRule.getElseRules().size() > 0){
+            if(isElse) {
+                // elseRule can't have nested elseRule
+                throw ServiceException.INVALID_REQUEST("elseRule can't have nested elseRule", null);
+            }
+            boolean elseFound = false;
+            for (NestedRule elseRule : currentNestedRule.getElseRules()) {
+                if (elseFound) {
+                    // elsif or else MUST only follow an if or elsif
+                    throw ServiceException.INVALID_REQUEST("Else or elsif can't follow else", null);
```
ykomiyam: Perhaps the "Else or ..." may start with a lower-case letter as similar as the line 247?

store/src/java/com/zimbra/cs/filter/SoapToSieve.java
```diff
@@ -143,38 +143,72 @@ private void handleRule(FilterRule rule) throws ServiceException {
...
+        handleElseRules(rule.getElseRules());
+    }
+
+    private void handleElseRules(List<NestedRule> elseRules) throws ServiceException {
+        boolean elseFound = false;
+        if(elseRules != null && elseRules.size() > 0){
+            for (NestedRule elseRule : elseRules) {
+                if (elseFound) {
+                    // elsif or else MUST only follow an if or elsif
+                    throw ServiceException.INVALID_REQUEST("Else or elsif can't follow else", null);
```
ykomiyam: The "Else or ..." may start with a lower-case letter as "else or ..." ?

store/src/java/com/zimbra/cs/filter/SieveVisitor.java
```diff
@@ -304,13 +307,24 @@ private void accept(Node parent, RuleProperties props) throws ServiceException {
...
+                }
                 visitIfControl(node,VisitPhase.begin,newProps);
-                accept(node, newProps);
+                if (SieveConstants.ELSE.equalsIgnoreCase(nodeName)) {
+                    newProps.condition = null;
+                    visitRule(node, VisitPhase.begin, newProps);
+                    accept(node, newProps);
+                    visitRule(node, VisitPhase.end, newProps);
+                } else {
+                   accept(node, newProps);
```
ykomiyam: Please align the indent.

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesForElseRulesTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+    @Test
+    public void testIfElseIf() throws Exception {
+        Account acct = Provisioning.getInstance().get(Key.AccountBy.name, "test@zimbra.com");
+        RuleManager.clearCachedRules(acct);
+        String ss = "require [\"fileinto\", \"log\"];\n";
+        ss += "if anyof header :is \"Subject\" \"important\" {\n";
+        ss += "fileinto \"FromP2\"; log \"Move message to FromP2 folder\"; }\n";
+        ss += "elsif anyof (header :contains \"subject\" [\"automation\", \"nunit\"]) \n";
+        ss += "{ redirect \"qa-automation@zqa-380.eng.zimbra.com\"; log \"Forward message to qa-automation DL\"; }\n";
```
ykomiyam: Why don't you redirect the message to "redirect@zimbra.com" or something?
Although we can use arbitrary account and domain name for the unit test, I think it is convention that we use "@zimbra.com" as a domain name, and more general name than "qa" or "automation" as an account name.

soap/src/java/com/zimbra/soap/mail/type/NestedRule.java
shrikant-prasad: done at all occurrences.

store/src/java/com/zimbra/cs/filter/SoapToSieve.java
shrikant-prasad: done

store/src/java/com/zimbra/cs/filter/SoapToSieve.java
shrikant-prasad: done

store/src/java/com/zimbra/cs/filter/SieveVisitor.java
shrikant-prasad: done

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesForElseRulesTest.java
shrikant-prasad: done.

chinmay15:
I agree with @rohan-ambasta here. We can remove NestedRule/ElseRule and use a single FilterRule class to handle this. If we extend FilterAction on FilterRule, we probably can solve this problem easily by putting nested rules in action list. It might not be doable directly, but it is worth to attempt. 

Another approach, as rohan has mentioned, is using tree instead of stack, where child and parent relationship can be handled easily and in better way, further it will be easy to understand.

shrikant-prasad:
NestedRule and FilterRule classes existed before this solution and for elseRule no new class has been introduced. The scope and behaviour of if, nested if, elsif and else rules are different thats why different stacks are needed. If we want to handle all using a single tree, we can open separate spike bug for that and see if it can done or not. 

FilterAction extending FilterRule doesn't makes any sense as both these classes represent separate entities action and rule, respectively.

PR #139, opened 2017-05-15
ZMS-366: Remote IMAP Tag Rename Handling

ilyusha:
When a tag is renamed, two types of notifications are triggered on the server:
 - the renamed tag is notified
 - all non-empty folders are notified

Some changes had to be made in order for these changes to trigger a waitset:
#### Folder Interests
Prior to this change, a waitset was triggered on a folder when an action was taken on an item in the folder. When a tag is renamed, however, the folders themselves should be notified, as opposed to their parents. In order for this to be recognized by the waitset, the _folderInterest_ mechanism had to be updated to be triggered when a change occurs on the folder itself. To this end, the _PendingModifications_ now contains two integer sets:
 - _changedParentFolders_ replaced what used to be called _changedFolders_
 - _changedFolders_ now tracks IDs of folders in the scenario that the folder itself was somehow affected (such as during a tag rename operation)
Waitsets are triggered if a folder interest intersects with _either_ of the these sets.

#### Tag Modifications
Tag modifications have to be treated differently than other modifications because it is unknown which folder contains items that need to be re-tagged. To this end, the _ModifyNotification_ loop in _WaitSetRequest_ first iterates over all the changes and aggregates all tag changes in a single array of _ModifyTagNotifications_. It then iterates over the changes again, this time setting the tag changes on each non-empty folder that was notified. This way, the IMAP server can iterate over the messages in each folder and update their tags appropriately.

#### ZMailbox Changes
_ZMailbox::getFolderById_ had to be modified to rebuild the item cache from the known folder hierarchy in the event of a cache miss. The _getFolderById_ method looks for the folder in the item cache, which can be overwritten by the tag update.

#### Paged Folders
Finally, in order for this to work on paged folders, _ImapFolder_ and _RemoteImapMailboxStore_ classes had to be updated to do a null check when loading tags. This is necessary because when a paged folder is loaded, it attempts to restore the last-known item tags from the cache; however, renamed tags will no longer be there, leading to a null value.

#### Unit Tests
Two new unit tests have been added to _TestRemoteImapNotications_: _testRenameTagNotificationActiveFolder_ and _testRenameTagNotificationCachedFolder_.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/WaitSetRequest.java
+++ b/store/src/java/com/zimbra/cs/service/mail/WaitSetRequest.java
@@ -42,15 +42,16 @@ import com.zimbra.common.util.Constants;
 import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.cs.account.Account;
 import com.zimbra.cs.account.Provisioning;
+import com.zimbra.cs.mailbox.Folder;
 import com.zimbra.cs.mailbox.MailItem;
 import com.zimbra.cs.mailbox.MailboxManager;
+import com.zimbra.cs.mailbox.Tag;
 import com.zimbra.cs.service.admin.AdminServiceException;
 import com.zimbra.cs.service.util.SyncToken;
 import com.zimbra.cs.servlet.continuation.ResumeContinuationListener;
 import com.zimbra.cs.session.PendingModifications;
 import com.zimbra.cs.session.PendingModifications.Change;
 import com.zimbra.cs.session.PendingModifications.ModificationKey;
-import com.zimbra.cs.session.PendingRemoteModifications;
 import com.zimbra.cs.session.WaitSetAccount;
 import com.zimbra.cs.session.WaitSetCallback;
 import com.zimbra.cs.session.WaitSetError;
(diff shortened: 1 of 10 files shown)


store/src/java/com/zimbra/cs/service/mail/WaitSetRequest.java
```diff
@@ -308,30 +310,56 @@ public static void staticHandle(WaitSetReq req, Map<String, Object> context, Wai
...
+                                        PendingFolderModifications folderMods = getFolderMods(itemId, folderMap);
+                                        for (ModifyTagNotification modTag: tagMods) {
+                                            folderMods.addModifiedTag(modTag);
+                                        }
+                                    }
+                                } else if (!(itemInfo instanceof Tag)){
+                                    if(folderInterests != null && !folderInterests.contains(folderId)) {
+                                        continue;
+                                    }
+                                    getFolderMods(folderId, folderMap).addModifiedMsg(getModifiedItemSOAP(itemInfo, ((Change) mod).why));
```
grishick: I've moved getFolderMods to JaxbUtil, because it is now also used by SoapSession

PR #141, opened 2017-05-19
ZCS-1429: Append-related unit tests

ilyusha:
This is a continuation of the work started in ZMS-212.

Changes to SharedImapTests:

- _testAppend_ has been split into _testAppend_ and _testAppendFlags_
- new _getImapZMailbox_ method lets remote IMAP tests access the _ZMailbox_ instance used by the IMAP server. This is necessary for _testAppendTags_ unit test, as folders created using a non-IMAP client are not reflected on the IMAP server in certain scenarios (ZCS-1488 addresses this).
- all append-related remote unit tests have been activated
- _fetchMessage_ method has been updated to fetch flags

Other changes:

- _ZMailbox_ has a new static inner class _TagSpecifier_ that can be used by a new _addMessage_ method signature to specify tags either by name or ID
- _AppendMessage_ uses a _TagSpecifier_ to set the tags by name (other parts of the code use tag IDs)
- null dates are properly handled by _AppendMessage_ in the remote case
- _ImapMessage.getFlags()_ can return tags that are not registered with the mailbox

### DIFF
--- a/store/src/java/com/zimbra/qa/unittest/TestRemoteImapShared.java
+++ b/store/src/java/com/zimbra/qa/unittest/TestRemoteImapShared.java
@@ -1,15 +1,20 @@
 package com.zimbra.qa.unittest;
 
 import java.io.IOException;
+import java.util.Set;
 
 import org.dom4j.DocumentException;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Ignore;
 
+import com.zimbra.client.ZMailbox;
 import com.zimbra.common.localconfig.ConfigException;
 import com.zimbra.common.localconfig.LC;
 import com.zimbra.common.service.ServiceException;
+import com.zimbra.cs.imap.ImapRemoteSession;
+import com.zimbra.cs.imap.ImapServerListener;
+import com.zimbra.cs.imap.ImapServerListenerPool;
 
 /**
  * This is a shell test for Remote IMAP tests that does the necessary configuration to select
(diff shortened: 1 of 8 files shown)


store/src/java/com/zimbra/qa/unittest/TestRemoteImapShared.java
```diff
@@ -37,38 +42,42 @@ public void tearDown() throws ServiceException, DocumentException, ConfigExcepti
     }
 
     @Override
-    @Ignore ("failing on remote imap for now")
-    public void testMultiappendNoLiteralPlus() throws Exception {
-
+    protected ZMailbox getImapZMailbox() throws Exception {
```
fatkudu: Please add a comment on why this exists.  We may want to deprecate it at some point in the future if we get to a point where all ZMailboxes are in sync.

store/src/java/com/zimbra/qa/unittest/TestRemoteImapShared.java
```diff
@@ -37,38 +42,42 @@ public void tearDown() throws ServiceException, DocumentException, ConfigExcepti
...
-    @Ignore ("failing on remote imap for now")
-    public void testCatenateSimple() throws Exception {
+    public void testStoreTagsDirty() throws Exception {
 
     }
 
     @Override
     @Ignore ("failing on remote imap for now")
-    public void testAppendTags() throws Exception {
+    public void testStoreInvalidSystemFlag() throws Exception {
```
grishick: this comment by @codacy-bot seems ignorable. Comments inside "Ignore" explain why this method is empty.

PR #142, opened 2017-05-19
ZCS-1047:S/MIME certificate not seen in Contact Properties

sneha-patil-synacor:
Issue:
Configure AD as external GAL on a domain in ZCS. When ZCO sends SyncGalRequest and GetContactsRequest, the ZCS does not send the userCertificate in GetContactsResponse

Root cause:
AD sends 'userCertificate;binary' attribute as 'userCertificate' without appending ';binary' to it. So, the server is not able to pick the attribute from entry data sent by AD

Fix:
If the attribute is not found by 'userCertficate;binary' name, fallback to get attribute with 'userCertificate' name.

### DIFF
--- a/store/src/java/com/zimbra/cs/ldap/unboundid/UBIDAttributes.java
+++ b/store/src/java/com/zimbra/cs/ldap/unboundid/UBIDAttributes.java
@@ -23,7 +23,7 @@ import java.util.Set;
 import com.unboundid.ldap.sdk.Attribute;
 import com.unboundid.ldap.sdk.Entry;
 import com.unboundid.ldap.sdk.SearchResultEntry;
-
+import com.zimbra.common.mailbox.ContactConstants;
 import com.zimbra.common.util.ByteUtil;
 import com.zimbra.cs.account.AttributeManager;
 import com.zimbra.cs.ldap.LdapException;
@@ -93,7 +93,6 @@ public class UBIDAttributes extends ZAttributes {
     protected String getAttrString(String transferAttrName, boolean containsBinaryData) 
     throws LdapException {
         Attribute attr = entry.getAttribute(transferAttrName);
-        
         if (attr != null) {
             return getAttrStringInternal(attr, containsBinaryData);
         } else {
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/ldap/unboundid/UBIDAttributes.java
```diff
@@ -105,7 +104,11 @@ protected String getAttrString(String transferAttrName, boolean containsBinaryDa
     protected String[] getMultiAttrString(String transferAttrName, boolean containsBinaryData) 
     throws LdapException {
         Attribute attr = entry.getAttribute(transferAttrName);
-        
+        // AD sends 'userCertificate;binary' attribute as 'userCertificate' without appending ';binary' to it
+        if (transferAttrName.contains(ContactConstants.A_userCertificate) && attr == null) {
+            attr = entry.getAttribute(transferAttrName.replace(";binary", ""));
```
shrikant-prasad: no need of replace, can use ContactConstants.A_userCertificate directly as attribute name.

store/src/java/com/zimbra/cs/ldap/unboundid/UBIDAttributes.java
```diff
@@ -105,7 +104,11 @@ protected String getAttrString(String transferAttrName, boolean containsBinaryDa
     protected String[] getMultiAttrString(String transferAttrName, boolean containsBinaryData) 
     throws LdapException {
         Attribute attr = entry.getAttribute(transferAttrName);
-        
+        // AD sends 'userCertificate;binary' attribute as 'userCertificate' without appending ';binary' to it
```
rohan-ambasta: you can specify bug number for future reference

store/src/java/com/zimbra/cs/ldap/unboundid/UBIDAttributes.java
```diff
@@ -105,7 +104,11 @@ protected String getAttrString(String transferAttrName, boolean containsBinaryDa
...
+        if (transferAttrName.contains(ContactConstants.A_userCertificate) && attr == null) {
```
rohan-ambasta: instead of contains, may be you can use startsWith(String prefix)

PR #144, opened 2017-05-22
ZCS-541:Tasks causing slowness from ZWC and consuming CPU resources

shrikant-prasad:
when recurring task is imported, duration is calculated as (due date -start date), which exceeds the recurrence range. Due to this when task tab is opened, the search operation takes forever as it enters a while loop which goes on forever if the due date is after several years. Ideally such task should not be created, the client should handle that, but in this case task is imported, so we set the duration to 1day to avoid this problem.

Tested by importing ics files from customers and the tasks were shown properly in web client.

### DIFF
--- a/store/src/java-test/com/zimbra/cs/mailbox/RecurringTaskTest.java
+++ b/store/src/java-test/com/zimbra/cs/mailbox/RecurringTaskTest.java
@@ -1,0 +1,89 @@
+package com.zimbra.cs.mailbox;
+
+import static org.junit.Assert.assertEquals;
+
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.io.UnsupportedEncodingException;
+import java.util.List;
+
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import com.google.common.collect.Maps;
+import com.zimbra.common.account.Key;
+import com.zimbra.common.calendar.ZCalendar.ZCalendarBuilder;
+import com.zimbra.common.calendar.ZCalendar.ZVCalendar;
+import com.zimbra.common.mime.MimeConstants;
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.cs.account.Account;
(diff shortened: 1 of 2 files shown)


shrikant-prasad:
As per the standards, in this case duration property is not present, due and dtstart are present.Hence duration is calculated using due and dtstart (due - dtstart). 
Similar scenario can be there for calendar events but in our web client we don't allow user to create a recurring meeting in which duration of meeting exceeds the recurrence range. But in our web client  tasks do not support recurrence. The tasks in bug were created from external client. so we can reject such task while importing or just set the duration to 1 day.

store/src/java-test/com/zimbra/cs/mailbox/RecurringTaskTest.java
```diff
@@ -0,0 +1,89 @@
...
+                    + "TZOFFSETTO:+0200\n"
+                    + "TZOFFSETFROM:+0200\n"
+                    + "TZNAME:CAT\n"
+                    + "END:STANDARD\n"
+                    + "END:VTIMEZONE\n"
+                    + "BEGIN:VTODO\n"
+                    + "UID:93077c29_ab51_41ad_aaa8_f63a68f963a6_migwiz\n"
+                    + "RRULE:FREQ=MONTHLY;INTERVAL=1;BYDAY=MO,TU,WE,TH,FR;BYSETPOS=5\n"
+                    + "SUMMARY:Pay Cassim Salary\n"
+                    + "ATTENDEE;CN=Salim Jhavary;CUTYPE=RESOURCE;ROLE=NON-PARTICIPANT;PARTSTAT=NEEDS-ACTION:mailto:salim@sjcasa.co.za\n"
```
fatkudu: Please anonymize this data!  For instance, I assume this is a customer's email address - and github is a world readable system!!!!

shrikant-prasad: done.

PR #145, opened 2017-05-22
ZCS-1173:SIEVE:filter w/o anyof & allof causes NPE

ykomiyam:
ZCS-1173:SIEVE:filter w/o anyof & allof causes NPE

[Bug]
If anyof/allof test is not specified in condition within the sieve rule, NullPointerException occurs in mailbox.log when the rule is loaded on ZWC.

[Root cause]
An instance of the FilterRule or NestedRule which keeps a set of Sieve test and/or actions of each rule was created only while 'anyof' or 'allof' test was processed; if a filter does not have an 'anyof' or 'allof' in the 'if' condition, or even an action is not enclosed by the if-block, the member parameter the FilterRule or NestedRule in the com.zimbra.cs.filter.SieveToSoap class remained null.

[Fix]
If the member parameter FilterRule nor NestedRule are not initialized, create a FilterRule instance before being accessed.  By default, 'allof' test is set (As a result, the "filterTests" tag in the SOAP result has an 'allof' as a "condition" parameter).

[Manual/Unit test]
* Set a Sieve rule which has no 'anyof' or 'allof' test to the zimbraMailSieveScript, and accessed it from ZWC ==> PASS (No NPE on the server side, no network error no the client side, the filter is shown on the ZWC (only the supported Sieve commands were displayed))
* Unit test to verify the four patterns ==> PASS 
  (1) Not nested.  no 'allof' 'anyof' tests
  (2) Nested. no 'allof' 'anyof' tests
  (3) Nested. no 'allof' and 'anyof' test for the outer if block. 'anyof' test in the inner if block.
  (4) No if block. Only a Sieve action is specified.

### DIFF
--- a/store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
+++ b/store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
@@ -49,7 +49,7 @@ public class GetFilterRulesTest {
     }
 
     @Test
-    public void testIf_Without_Allof() throws Exception {
+    public void testIfWithoutAllof() throws Exception {
         // - no 'allof' 'anyof' tests
         String filterScript
                     = "require \"tag\";"
@@ -81,7 +81,7 @@ public class GetFilterRulesTest {
     }
 
     @Test
-    public void testNestedIf_Without_Allof() throws Exception {
+    public void testNestedIfWithoutAllof() throws Exception {
         // - no 'allof' 'anyof' tests
         // - nested if
         String filterScript
@@ -117,7 +117,7 @@ public class GetFilterRulesTest {
     }
 
     @Test
-    public void testNestedIf_AllofAnyof() throws Exception {
+    public void testNestedIfAllofAnyof() throws Exception {
         // - nested if
         // - no 'allof' and 'anyof' test for the outer if block
         // - 'anyof' test in the inner if block
@@ -160,7 +160,7 @@ public class GetFilterRulesTest {
     }
 
     @Test
-    public void testWithout_if() throws Exception {
+    public void testWithoutIf() throws Exception {
         // - no if block
         String filterScript
                     = "require \"tag\";"
--- a/store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
+++ b/store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
@@ -1,0 +1,191 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2017 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ * ***** END LICENSE BLOCK *****
+ */
+
+package com.zimbra.cs.filter;
+
+import java.util.HashMap;
+
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import com.zimbra.common.soap.Element;
+import com.zimbra.common.soap.MailConstants;
+
+import com.zimbra.cs.account.Account;
+import com.zimbra.cs.account.MockProvisioning;
+import com.zimbra.cs.account.Provisioning;
+import com.zimbra.cs.filter.RuleManager;
+import com.zimbra.cs.mailbox.MailboxTestUtil;
+import com.zimbra.cs.service.mail.GetFilterRules;
+import com.zimbra.cs.service.mail.ServiceTestUtil;
+import com.zimbra.cs.util.XMLDiffChecker;
+
+public class GetFilterRulesTest {
+    @BeforeClass
+    public static void init() throws Exception {
+        MailboxTestUtil.initServer();
+        Provisioning prov = Provisioning.getInstance();
+        prov.createAccount("test@zimbra.com", "secret", new HashMap<String, Object>());
+    }
+
+    @Before
+    public void setUp() throws Exception {
+        MailboxTestUtil.clearData();
+    }
+
+    @Test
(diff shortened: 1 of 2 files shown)


store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
```diff
@@ -0,0 +1,191 @@
...
+                      + "<actionFileInto folderPath=\"if-block\" index=\"0\"/>"
+                    + "</filterActions>"
+                  + "</filterRule>"
+                + "</filterRules>"
+              + "</GetFilterRulesResponse>";
+        XMLDiffChecker.assertXMLEquals(expectedSoapResponse, response.prettyPrint());
+    }
+
+    @Test
+    public void testNestedIf_Without_Allof() throws Exception {
```
ykomiyam: Fixed

ykomiyam: XMLDiffChecker.assertXMLEquals is called instead.  This method calls Assert.assertFalse() inside it. 
`

    public static void assertXMLEquals(String expected, String actual) {
        Diff myDiff = DiffBuilder.compare(Input.fromString(expected))
            .withTest(Input.fromString(actual)).ignoreWhitespace().build();
        Assert.assertFalse(myDiff.toString(), myDiff.hasDifferences());
    }
`

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
```diff
@@ -0,0 +1,191 @@
...
+                    + "</filterActions>"
+                  + "</nestedRule>"
+                + "</filterRule>"
+              + "</filterRules>"
+            + "</GetFilterRulesResponse>";
+        XMLDiffChecker.assertXMLEquals(expectedSoapResponse, response.prettyPrint());
+    }
+
+    @Test
+    public void testWithout_if() throws Exception {
```
ykomiyam: Fixed

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
```diff
@@ -0,0 +1,191 @@
...
+        prov.createAccount("test@zimbra.com", "secret", new HashMap<String, Object>());
+    }
+
+    @Before
+    public void setUp() throws Exception {
+        MailboxTestUtil.clearData();
+    }
+
+    @Test
+    public void testIf_Without_Allof() throws Exception {
```
ykomiyam: Fixed

ykomiyam: Ditto (XMLDiffChecker.assertXMLEquals is called instead).

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
ykomiyam: Ditto.

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
```diff
@@ -0,0 +1,191 @@
...
+                    + "<actionFileInto folderPath=\"nested-if-block\" index=\"0\"/>"
+                  + "</filterActions>"
+                + "</filterRule>"
+              + "</filterRules>"
+            + "</GetFilterRulesResponse>";
+        XMLDiffChecker.assertXMLEquals(expectedSoapResponse, response.prettyPrint());
+    }
+
+    @Test
+    public void testNestedIf_AllofAnyof() throws Exception {
```
ykomiyam: Fixed.

ykomiyam: Ditto (XMLDiffChecker.assertXMLEquals is called instead).

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
```diff
@@ -0,0 +1,191 @@
...
+                    + "</filterActions>"
+                  + "</nestedRule>"
+                + "</filterRule>"
+              + "</filterRules>"
+            + "</GetFilterRulesResponse>";
+        XMLDiffChecker.assertXMLEquals(expectedSoapResponse, response.prettyPrint());
+    }
+
+    @Test
+    public void testWithoutIf() throws Exception {
```
ykomiyam: Each test case does call the assertFalse() inside the XMLDiffChecker.assertXMLEquals().

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
```diff
@@ -0,0 +1,191 @@
...
+                      + "<actionFileInto folderPath=\"if-block\" index=\"0\"/>"
+                    + "</filterActions>"
+                  + "</filterRule>"
+                + "</filterRules>"
+              + "</GetFilterRulesResponse>";
+        XMLDiffChecker.assertXMLEquals(expectedSoapResponse, response.prettyPrint());
+    }
+
+    @Test
+    public void testNestedIfWithoutAllof() throws Exception {
```
ykomiyam: testNestedIfWithoutAllof() calls XMLDiffChecker.assertXMLEquals() to check the response of the GetFilterRule request and the expected string using Assert.assertFalse() method.

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
```diff
@@ -0,0 +1,191 @@
...
+                    + "<actionFileInto folderPath=\"nested-if-block\" index=\"0\"/>"
+                  + "</filterActions>"
+                + "</filterRule>"
+              + "</filterRules>"
+            + "</GetFilterRulesResponse>";
+        XMLDiffChecker.assertXMLEquals(expectedSoapResponse, response.prettyPrint());
+    }
+
+    @Test
+    public void testNestedIfAllofAnyof() throws Exception {
```
ykomiyam: ditto

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
```diff
@@ -0,0 +1,191 @@
...
+        prov.createAccount("test@zimbra.com", "secret", new HashMap<String, Object>());
+    }
+
+    @Before
+    public void setUp() throws Exception {
+        MailboxTestUtil.clearData();
+    }
+
+    @Test
+    public void testIfWithoutAllof() throws Exception {
```
ykomiyam: ditto

ykomiyam:
I found a bug around the nested rules.  Once I fix it, I'll reopen it.

PR #149, opened 2017-05-23
ZCS-1509:ImapMessage refactor getFlags

fatkudu:
PMD complaining about complexity of getFlags().
Refactored using new private class FlagsRepresentation

NEEDS TESTS before merging.


grishick:
@fatkudu is this PR still relevant?

fatkudu:
@grishick we could always drop the pull request but keep the branch around.  I think it needs a test to be sure of no regressions but the change would simplify things.

PR #151, opened 2017-05-24
feature/zms 180

grishick:
I added 'deletable' flag to Folder SOAP to be returned by default, because we currently do not have a way to request only specific attribute from GetFolderRequest. As far as I can see from the tests, this is not breaking any existing functionality. 
I also noticed that ZClient was not correctly picking up paths for top level folders. Instead ZClient was attempting to construct folder path based on internally cached folder tree. This results in ZClient returning incorrect paths when a folder is renamed outside of ZClient. I removed "TODO" comment from getPath and added implementation that relies on path returned from SOAP. This last change broke some tests that relied on old behavior where ZMailbox would construct Folder path based on internal folder tree. To address this, I added code that handles changes to folder path that are sent via context headers. I also added a test to verify how zclient handles renamed folders when folders are renamed outside of zclient and inside zclient.

### DIFF
--- a/client/src/java/com/zimbra/client/ZFolder.java
+++ b/client/src/java/com/zimbra/client/ZFolder.java
@@ -97,6 +97,7 @@ public class ZFolder implements ZItem, FolderStore, Comparable<Object>, ToZJSONO
     private final ZMailbox mMailbox;
     private RetentionPolicy mRetentionPolicy = new RetentionPolicy();
     private boolean mActiveSyncDisabled;
+    private boolean mDeletable;
     private int mImapRECENTCutoff;
 
 
@@ -280,6 +281,7 @@ public class ZFolder implements ZItem, FolderStore, Comparable<Object>, ToZJSONO
         mEffectivePerms = e.getAttribute(MailConstants.A_RIGHTS, null);
         mSize = e.getAttributeLong(MailConstants.A_SIZE, 0);
         mActiveSyncDisabled = e.getAttributeBool(MailConstants.A_ACTIVESYNC_DISABLED, false);
+        mDeletable = e.getAttributeBool(MailConstants.A_DELETABLE, true);
 
         mGrants = new ArrayList<ZGrant>();
         mSubFolders = new ArrayList<ZFolder>();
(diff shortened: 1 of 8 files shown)


store/src/java/com/zimbra/qa/unittest/TestZClient.java
```diff
@@ -1063,6 +1063,196 @@ public void testDelete() throws Exception {
...
+        zf = zmbox.getFolderRequestById(szFolderId);
+        assertNotNull("ZMailbox::getFolderRequestById should be able to get /rootNonDeletable folder by ID", zf);
+        assertFalse("zFolder::isDeletable should return false for 'immutable' folder", zf.isDeletable());
+        assertEquals("unexpected zfolder ID", szFolderId, zf.getId());
+        assertEquals("unexpected zfolder name", "rootNonDeletable", zf.getName());
+        assertEquals("unexpected zfolder path", "/rootNonDeletable", zf.getPath());
+    }
+
+    @Test
+    public void testVisibleFolders() throws Exception {
```
grishick: going to ignore @codacy-bot here, because IMHO this is fine for test code

client/src/java/com/zimbra/client/ZFolder.java
```diff
@@ -88,6 +89,7 @@
     private int mContentSequence;
     private int mImapUIDNEXT;
     private int mImapMODSEQ;
+    private String mAbsolutePath;
```
grishick: @codacy-bot picked the wrong line in the file to comment on. I've removed the unused import

soap/src/java/com/zimbra/soap/mail/type/Folder.java
```diff
@@ -279,6 +279,13 @@ public static View fromString(String name) {
     private String restUrl;
 
     /**
+     * @zm-api-field-tag deletable
+     * @zm-api-field-description whether this folder can be deleted
+     */
+    @XmlAttribute(name=MailConstants.A_DELETABLE /* deletable */, required=false)
+    private boolean deletable;
```
fatkudu: Please use ZmBoolean

tmclane: Check this commit if you want a cheatsheet: 6f28fd1a4b2f2070383a7d5171b269b7eb82428b

PR #153, opened 2017-05-24
feature/zcs-1494

gordyt:
The first commit in this change will allow the `zimbra-imapd` service to use remote connections independently of the *localconfig* setting `imap_always_use_remote_store`. 

The second commit will allow the `zimbra-imapd` service to start up, even if the server upon which it is running has not been added to `zimbraReverseProxyUpstreamImapServers`.  This will allow customers (or QA teams) to run *both* embedded and external IMAP servers at the same time.  

For a customer, it will allow them to have a more controlled transition from embedded to remote IMAP.  For a QA team, it will allow them to test embedded and remote IMAP services more easily.

The last commit is just to address some [PMD](https://pmd.github.io/) warnings.

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/EhcacheImapCache.java
+++ b/store/src/java/com/zimbra/cs/imap/EhcacheImapCache.java
@@ -47,7 +47,7 @@ final class EhcacheImapCache implements ImapSessionManager.Cache<String, ImapFol
     EhcacheImapCache(String name, boolean active) {
         // If running inside mailboxd, share mailboxd cache, else use separate imap cache.  This avoids issues when running
         // decoupled IMAP service on same host as mailbox.
-        EhcacheManager.Service service = System.getProperty(ImapDaemon.IMAP_SERVER_EMBEDDED, "true").equals("false")
+        EhcacheManager.Service service = !ImapDaemon.isRunningImapInsideMailboxd()
                 ? EhcacheManager.Service.IMAP
                 : EhcacheManager.Service.MAILBOX;
         ehcache = EhcacheManager.getInstance(service).getEhcache(name);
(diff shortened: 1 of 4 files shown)


gordyt:
@fatkudu You are exactly correct.  We can use the same set of tests to run against embedded/local, embeded/remote, and imapd.  

The `imap_always_use_remote_store` LC setting was added back in December to provide a convenient way to force ZClient use by the IMAP(S) servers and thus facilitate our testing. In the initial implementation of the new `zimbra-imapd` service, I used that same LC value to force the IMAP(S) servers it controlled to use ZClient, as it shares almost all the same code with embedded imap.  After our group discussion, which included Matt B, it was decided to change things up a bit to make it safer (or at least more predictable) for a customer to upgrade to using the new service.

Several changes were made to effect this change.  The `imap_always_use_remote_store` went back to controlling only the embedded imap(s) servers.  The `zimbra-imapd` service relies solely on the presence of a system property `imap.server.embedded` (which is unknown to mailboxd and which is set to "false" by the `ImapDaemon` prior to starting the imap(s) servers). This was from a suggestion made back in the PR for ZMS-38.

In addition, the `ImapDaemon` will now start its imap(s) servers, even if that server is not present in the list of `zimbraReverseProxyUpstreamImapServers`.  So one can safely run embedded imap and imapd, on the same server with no conflicts.  By default, embedded imap will use local connections if possible and imapd will still use remote connections.  If it is desired to test remote connections with embedded imap, then the LC setting may still be used.

I can add a bool function as per your suggestion and update all of the places in the code that currently read the `imap.server.embedded` property.  We should still leave the checks for the LC setting in place, just for testing purposes if for no other reason.

PR #156, opened 2017-05-26
Bugfix/zcs 1173 SIEVE:filter w/o anyof & allof causes NPE(reopen)

ykomiyam:
ZCS-1173:SIEVE:filter w/o anyof & allof causes NPE

[Bug]
If anyof/allof test is not specified in condition within the sieve rule, NullPointerException occurs in mailbox.log when the rule is loaded on ZWC.

[Root cause]
An instance of the FilterRule or NestedRule which keeps a set of Sieve test and/or actions of each rule was created only while 'anyof' or 'allof' test was processed; if a filter does not have an 'anyof' or 'allof' in the 'if' condition, or even an action is not enclosed by the if-block, the member parameter the FilterRule or NestedRule in the com.zimbra.cs.filter.SieveToSoap class remained null.

[Fix]
SieveVisitor.visitRule() method which initializes the simple/nested rule object was called when the 'allof' or 'anyof' test was processed. But it may still cause NPE when 'allof' or 'anyof' test is not specified.

- visitRule() is now called when the 'if' is started.
- Introduce visitAllOfTest() and visitAnyOfTest() methods to SieveVisitor class, and their implementations to SieveToSoap class. Each classe updates the condition of the rule condition (default: allof) when 'allof' or 'anyof' test is specified in the 'if'.

If the member parameter FilterRule nor NestedRule are not initialized, create a FilterRule instance before being accessed. By default, 'allof' test is set (As a result, the "filterTests" tag in the SOAP result has an 'allof' as a "condition" parameter).

[Manual/Unit test]

Set a Sieve rule which has no 'anyof' or 'allof' test to the zimbraMailSieveScript, and accessed it from ZWC ==> PASS (No NPE on the server side, no network error no the client side, the filter is shown on the ZWC (only the supported Sieve commands were displayed))
Unit test to verify the four patterns ==> PASS
(1) Not nested. no 'allof' 'anyof' tests
(2) Nested. no 'allof' 'anyof' tests
(3) Nested. no 'allof' and 'anyof' test for the outer if block. 'anyof' test in the inner if block.
(4) No if block. Only a Sieve action is specified.

### DIFF
--- a/store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
+++ b/store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
@@ -49,7 +49,7 @@ public class GetFilterRulesTest {
     }
 
     @Test
-    public void testIf_Without_Allof() throws Exception {
+    public void testIfWithoutAllof() throws Exception {
         // - no 'allof' 'anyof' tests
         String filterScript
                     = "require \"tag\";"
@@ -81,7 +81,7 @@ public class GetFilterRulesTest {
     }
 
     @Test
-    public void testNestedIf_Without_Allof() throws Exception {
+    public void testNestedIfWithoutAllof() throws Exception {
         // - no 'allof' 'anyof' tests
         // - nested if
         String filterScript
@@ -117,7 +117,7 @@ public class GetFilterRulesTest {
     }
 
     @Test
-    public void testNestedIf_AllofAnyof() throws Exception {
+    public void testNestedIfAllofAnyof() throws Exception {
         // - nested if
         // - no 'allof' and 'anyof' test for the outer if block
         // - 'anyof' test in the inner if block
@@ -160,7 +160,7 @@ public class GetFilterRulesTest {
     }
 
     @Test
-    public void testWithout_if() throws Exception {
+    public void testWithoutIf() throws Exception {
         // - no if block
         String filterScript
                     = "require \"tag\";"
--- a/store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
+++ b/store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
@@ -105,11 +105,15 @@ public class GetFilterRulesTest {
                 + "<filterRule active=\"1\">"
                   + "<filterTests condition=\"allof\">"
                     + "<headerTest stringComparison=\"matches\" header=\"Subject\" index=\"0\" value=\"*\"/>"
-                    + "<headerTest stringComparison=\"matches\" header=\"From\"    index=\"1\" value=\"*\"/>"
                   + "</filterTests>"
-                  + "<filterActions>"
-                    + "<actionFileInto folderPath=\"nested-if-block\" index=\"0\"/>"
-                  + "</filterActions>"
+                  + "<nestedRule>"
+                    + "<filterTests condition=\"allof\">"
+                      + "<headerTest stringComparison=\"matches\" header=\"From\" index=\"0\" value=\"*\"/>"
+                    + "</filterTests>"
+                    + "<filterActions>"
+                      + "<actionFileInto folderPath=\"nested-if-block\" index=\"0\"/>"
+                    + "</filterActions>"
+                  + "</nestedRule>"
                 + "</filterRule>"
               + "</filterRules>"
             + "</GetFilterRulesResponse>";
@@ -120,7 +124,7 @@ public class GetFilterRulesTest {
     public void testNestedIfAllofAnyof() throws Exception {
         // - nested if
         // - no 'allof' and 'anyof' test for the outer if block
-        // - 'anyof' test in the inner if block
+        // - 'anyof' test for the inner if block
         String filterScript
                     = "require \"tag\";"
                     + "if header :comparator \"i;ascii-casemap\" :matches \"Subject\" \"*\" {"
@@ -160,7 +164,7 @@ public class GetFilterRulesTest {
     }
 
     @Test
-    public void testWithoutIf() throws Exception {
+    public void testWithoutIfBlock1() throws Exception {
         // - no if block
         String filterScript
                     = "require \"tag\";"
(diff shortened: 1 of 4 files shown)


store/src/java/com/zimbra/cs/filter/SieveVisitor.java
```diff
@@ -284,14 +292,14 @@ private void acceptTest(Node node, RuleProperties props) throws ServiceException
         } else {
             if ("allof".equalsIgnoreCase(nodeName)) {
                 props.condition = Sieve.Condition.allof;
-                visitRule(node, VisitPhase.begin, props);
+                visitAllOfTest(node, VisitPhase.begin, props);
                 accept(node, props);
-                visitRule(node, VisitPhase.end, props);
+                visitAllOfTest(node, VisitPhase.end, props);
```
ykomiyam: Please see the fix description.   Source-code-wise, the method name explains its functionality .  Besides the the declaration of the visitAllOfTest() and visitAnyOfTest() have the same format as other visit*Test() methods.

store/src/java/com/zimbra/cs/filter/SieveVisitor.java
```diff
@@ -284,14 +292,14 @@ private void acceptTest(Node node, RuleProperties props) throws ServiceException
         } else {
             if ("allof".equalsIgnoreCase(nodeName)) {
                 props.condition = Sieve.Condition.allof;
-                visitRule(node, VisitPhase.begin, props);
```
ykomiyam: ditto.

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
```diff
@@ -0,0 +1,384 @@
...
+                      + "</filterActions>"
+                    + "</nestedRule>"
+                  + "</filterRule>"
+                + "</filterRules>"
+              + "</GetFilterRulesResponse>";
+        XMLDiffChecker.assertXMLEquals(expectedSoapResponse, response.prettyPrint());
+    }
+
+    @Test
+    public void testWithoutIfBlock4() throws Exception {
```
ykomiyam: Each test case does call the assertFalse() inside the XMLDiffChecker.assertXMLEquals().

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
```diff
@@ -0,0 +1,384 @@
...
+                      + "<actionFileInto folderPath=\"if-block\" index=\"0\"/>"
+                    + "</filterActions>"
+                  + "</filterRule>"
+                + "</filterRules>"
+              + "</GetFilterRulesResponse>";
+        XMLDiffChecker.assertXMLEquals(expectedSoapResponse, response.prettyPrint());
+    }
+
+    @Test
+    public void testNestedIfWithoutAllof() throws Exception {
```
ykomiyam: testNestedIfWithoutAllof() calls XMLDiffChecker.assertXMLEquals() to check the response of the GetFilterRule request and the expected string using Assert.assertFalse() method.

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
```diff
@@ -0,0 +1,384 @@
...
+                    + "</filterActions>"
+                  + "</nestedRule>"
+                + "</filterRule>"
+              + "</filterRules>"
+            + "</GetFilterRulesResponse>";
+        XMLDiffChecker.assertXMLEquals(expectedSoapResponse, response.prettyPrint());
+    }
+
+    @Test
+    public void testNestedIfAllofAnyof() throws Exception {
```
ykomiyam: ditto.

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
```diff
@@ -0,0 +1,384 @@
...
+                      + "<actionFileInto folderPath=\"no-if-block\" index=\"0\"/>"
+                    + "</filterActions>"
+                  + "</filterRule>"
+                + "</filterRules>"
+              + "</GetFilterRulesResponse>";
+        XMLDiffChecker.assertXMLEquals(expectedSoapResponse, response.prettyPrint());
+    }
+
+    @Test
+    public void testWithoutIfBlock5() throws Exception {
```
ykomiyam: ditto

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
```diff
@@ -0,0 +1,384 @@
...
+        prov.createAccount("test@zimbra.com", "secret", new HashMap<String, Object>());
+    }
+
+    @Before
+    public void setUp() throws Exception {
+        MailboxTestUtil.clearData();
+    }
+
+    @Test
+    public void testIfWithoutAllof() throws Exception {
```
ykomiyam: ditto

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
```diff
@@ -0,0 +1,384 @@
...
+                      + "<actionFileInto folderPath=\"no-if-block\" index=\"0\"/>"
+                    + "</filterActions>"
+                  + "</filterRule>"
+                + "</filterRules>"
+              + "</GetFilterRulesResponse>";
+        XMLDiffChecker.assertXMLEquals(expectedSoapResponse, response.prettyPrint());
+    }
+
+    @Test
+    public void testWithoutIfBlock2() throws Exception {
```
ykomiyam: ditto

store/src/java-test/com/zimbra/cs/filter/GetFilterRulesTest.java
```diff
@@ -0,0 +1,384 @@
...
+                    + "</filterActions>"
+                  + "</nestedRule>"
+                + "</filterRule>"
+              + "</filterRules>"
+            + "</GetFilterRulesResponse>";
+        XMLDiffChecker.assertXMLEquals(expectedSoapResponse, response.prettyPrint());
+    }
+
+    @Test
+    public void testWithoutIfBlock1() throws Exception {
```
ykomiyam: ditto

store/src/java/com/zimbra/cs/filter/SieveToSoap.java
```diff
@@ -358,6 +392,33 @@ protected void visitFlaggedTest(Node node, VisitPhase phase, RuleProperties prop
...
+    protected void visitAllOfTest(Node ruleNode, VisitPhase phase, RuleProperties props) {
+        visitLogicalTest(ruleNode, phase, props);
+    }
+
+    @Override
+    protected void visitAnyOfTest(Node ruleNode, VisitPhase phase, RuleProperties props) {
+        visitLogicalTest(ruleNode, phase, props);
+    }
+
+    private void visitLogicalTest(Node ruleNode, VisitPhase phase, RuleProperties props) {
```
ykomiyam: Save for the future use.

store/src/java/com/zimbra/cs/filter/SieveToSoap.java
```diff
@@ -81,26 +81,50 @@ protected void visitRule(Node ruleNode, VisitPhase phase, RuleProperties props)
...
     }
 
     private boolean isNestedRule(){
         // in non nested case, only one process is started but not done.
         if(nunOfIfProcessingStarted == numOfIfProcessingDone+1){
             return false;
         }
+        // if an action is specified outside of the 'if' clause,
+        // the number of Started becomes equal to that of Done.
+        if (nunOfIfProcessingStarted == numOfIfProcessingDone) {
```
ykomiyam: In this case, this looks intuitionally readable 'if' block because the length of the block is short (one line), there is the similar condition before it, and the source code comment is added.

ykomiyam:
As per discussion with Rupali et.al, the way of this fix is rejected because it may change the stored sieve rule.

PR #158, opened 2017-05-29
ZCS-1457: Run test target part of build.

Prashantsurana:


jiteshsojitra:
Yes, separate target for test and depends on better.

grishick:
I also don't want to have "publish" depend on "test". As our test suite grows it will be getting in the way of everyday tasks. We only must run "test" before creating a pull request.

Prashantsurana:
@fatkudu @grishick 
I agree, but since there is the interdependency publish local need to run while running the test. 

<ant dir="./common" target="test" inheritAll="true"/>   -- Work

<ant dir="./soap"   target="test" inheritAll="true"/>  -- Will not work without ./common publish-local

<ant dir="./client" target="test" inheritAll="true"/> -- Will not work without ./common and ./soap  publish-local

<ant dir="./store"  target="test" inheritAll="true"/> -- Will not work without ./common, ./client and ./soap  publish-local

silentsakky:
Something like this?

```
<target name="test-all" depends="publish-local-all">
	<ant dir="./native" target="test" inheritAll="true"/>
	<ant dir="./common" target="test" inheritAll="true"/>
	<ant dir="./soap"   target="test" inheritAll="true"/>
	<ant dir="./client" target="test" inheritAll="true"/>
	<ant dir="./store"  target="test" inheritAll="true"/>
</target>
```
And call test-all from at time of build?

By the way can we have a flag in build.pl to run test cases when creating build (default should be on), as a developer when creating dev builds i don't want to run unit tests

PR #160, opened 2017-05-30
ZCS-1515 ImapDaemon init ephemeral extension

gordyt:
Update the ImapDaemon to initialize the configured ephemeral
backend extension, based on the setting of
`zimbraEphemeralBackendURL`, if `ldap` is not the configured
backend.

### 2017-06-01 updates
* Implemented Gren's suggestion
* **Note:** This updated branch is based on top of [pull request 166](https://github.com/Zimbra/zm-mailbox/pull/166)

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/ImapDaemon.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapDaemon.java
@@ -25,6 +25,7 @@ import org.apache.log4j.PropertyConfigurator;
 import com.zimbra.common.localconfig.LC;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.cs.account.Provisioning;
+import com.zimbra.cs.extension.ExtensionUtil;
 import com.zimbra.common.util.ZimbraLog;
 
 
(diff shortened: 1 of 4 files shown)


store/src/java/com/zimbra/cs/imap/ImapDaemon.java
```diff
@@ -148,6 +152,26 @@ private static boolean isZimbraImapEnabled() throws ServiceException {
         return false;
     }
 
+    private static void maybeInitEphemeralBackendExtension() throws ServiceException {
```
fatkudu: I wonder if we could do with a shared method in ExtensionUtil.  Maybe some crossover with AttributeMigration.initEphemeralBackendExtension?

gordyt: Not a bad thought.  I'll take a look at that and see if it makes sense.

store/src/java/com/zimbra/cs/imap/ImapDaemon.java
```diff
@@ -148,6 +150,17 @@ private static boolean isZimbraImapEnabled() throws ServiceException {
...
+        String url = Provisioning.getInstance().getConfig().getEphemeralBackendURL();
+        if(url != null) {
+            String[] tokens = url.split(":");
+            String backendName = tokens[0];
```
rupalid: we need to check if token is non null before assigning to backendName, otherwise it may cause a NPE

PR #169, opened 2017-06-01
ZCS-126 PasswordLockoutFailureTime is updated for accounts in lockout

rohan-ambasta:
[Bug]:
In case of failed login, zimbraPasswordLockoutFailureTime is updated for accounts in lockout state

[Fix]:
If the account is in lockout state, return without updating zimbraPasswordLockoutFailureTime.
This fix has been applied from the patch provided by Prashant for bug #106408

[Testing]:
Mock provisioning does not support authentication. Hence, did manual testing.
Automation needs to be done.


rupalid:
Please explore if Mock Provisioning can be extended to support authentication

rohan-ambasta:
Created the following jira ticket for extending MockProvisioning capabilities

PR #170, opened 2017-06-01
ZCS-1335:Fix the encodeCertificate method call to fix automation failure

shrikant-prasad:
both failing automations passed after the fix.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/ToXML.java
+++ b/store/src/java/com/zimbra/cs/service/mail/ToXML.java
@@ -841,7 +841,7 @@ public final class ToXML {
                     if (Contact.isUrlField(name)) {
                         v = DefangFilter.sanitize(v, true);
                     } else if (Contact.isSMIMECertField(name)) {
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/service/mail/ToXML.java
```diff
@@ -841,7 +841,7 @@ private static void encodeContactAttr(Element elem, String name, String value,
                     if (Contact.isUrlField(name)) {
                         v = DefangFilter.sanitize(v, true);
                     } else if (Contact.isSMIMECertField(name)) {
-                        encodeCertificate(octxt, elem, name, value, contact.getEmailAddresses(), returnCertInfo);
+                        encodeCertificate(octxt, elem, name, v, contact.getEmailAddresses(), returnCertInfo);
```
rohan-ambasta: Can we cover this in junit?

shrikant-prasad: added

PR #176, opened 2017-06-08
ZCS-1668 TestImapVia* subclasses

fatkudu:


fatkudu:
Gordon, I don't think that does any harm and if we move to using proxy, may save time in the future.

PR #177, opened 2017-06-08
ZMS-675 ensure session cache is updated

tmclane:
The in-memory cache should ALWAYS be updated even if the operation is
a permanent operation. Otherwise the two views get out of sync.

Allows the zmsoap test to complete succesfully:
`zimbra@zcs-dev-one:/home/zimbra/git/zm-mailbox/store$ zmsoap -z RunUnitTestsRequest test=TestImapClient#testDelete`

    <RunUnitTestsResponse numFailed="0" numSkipped="0" numExecuted="1" xmlns="urn:zimbraAdmin">
      <results>
        <completed name="testDelete" class="com.zimbra.qa.unittest.TestImapClient" execSeconds="0.95"/>
      </results>
    </RunUnitTestsResponse>

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/ImapHandler.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapHandler.java
@@ -3975,11 +3975,10 @@ public abstract class ImapHandler {
                                         ZimbraLog.imap.info("IMAP client has flagged the item with id %d to be Deleted altertag", msg.msgId);
                                     }
                                     mbox.alterTag(getContext(), itemIds, i4flag.mName, add);
-                                } else {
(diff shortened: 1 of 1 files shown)


ilyusha:
I know this is merged already, but I'm not sure that this really fixes the underlying problem. I'm still getting test failures on testDelete, which are still caused by the timing issue of waitsets not triggering in time for the expunge command to be aware of the modification. In fact, the `ImapMessage::setSessionFlags` method gets short-circuited right away if the passed-in flag doesn't intersect the MUTABLE_SESSION_FLAGS bitmask, which does not include `\Deleted`.

PR #178, opened 2017-06-09
ZCS-1649 Disable extensions if zimbraNetworkModuleNGEnabled is TRUE

rohan-ambasta:
1. Add service exception for zimbraNetworkModulesNGEnabled
2. In ExtensionUtil.java added ServiceException in catch clause because some extensions like HSM and Backup extend the deprecated ZimbraNetworkExtension
3. In GetAdminExtensionZimlets disable zimlets if zimbraNetworkModulesNGEnabled is true

### DIFF
--- a/store/src/java/com/zimbra/cs/service/admin/GetAdminExtensionZimlets.java
+++ b/store/src/java/com/zimbra/cs/service/admin/GetAdminExtensionZimlets.java
@@ -24,6 +24,7 @@ import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.soap.AccountConstants;
 import com.zimbra.common.soap.AdminConstants;
 import com.zimbra.common.soap.Element;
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/service/admin/GetAdminExtensionZimlets.java
```diff
@@ -61,8 +62,23 @@ private void doExtensionZimlets(ZimbraSoapContext zsc, Map<String, Object> conte
 			
 			if (z.isExtension()) {
 			    boolean include = true;
+			    boolean isNGEnabled = true;
+			    try {
+			        isNGEnabled = Provisioning.getInstance().getLocalServer().isNetworkModulesNGEnabled();
+                } catch (ServiceException e) {
```
plobbes: nit: it appears there are some indentation/whitespace issues here and below - maybe tabs vs spaces...

PR #180, opened 2017-06-09
ZCS-1580 - Append command does not work with Remote IMAP/IMAPS

gordyt:
### Overview
 
Updates to enable `APPEND` command to function properly with both embedded and remote IMAP(S)
 
### Notes
 
In the parsing of the `APPEND` command...
 
    NioImapRequest.parseCommand/1 ->
    Literal.newInstance/2 ->
    BlobLiteral/1 ->
    StoreManager.getInstance().getBlobBuilder()
 
...we end up initializing the `StoreManager`.
 
`StoreManager.getInstance/0` was using the `zimbra_class_store` LC value.  It will still use that now, if not running from the `zimbra-imapd` process.  If it is, is uses a new LC value `imapd_class_store`.  The interesting part is that, at least for the `APPEND` command, it doesn't actually store anything on disk.
 
The new `ImapTransientStoreManager` is now what is being initialized for the remote IMAP(S) servers to use.  As the name implies, it really is for transient blobs that do not need to be replicated.  We may decide at some point in the future to optionally spool large transient information to disk and this would be a suitable candidate for handling that.
 
As to why adding a new `StoreManager`at all: The default configured one is `FileBlobStore`.  This requires access to a `DbPool`.  This is not a problem when the `zimbra-imapd` server is running on a mailbox server node, where MariaDB is available, but is a problem when running as a stand-alone service.

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/ImapDaemon.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapDaemon.java
@@ -18,6 +18,7 @@
 package com.zimbra.cs.imap;
 
 import java.io.FileInputStream;
+import java.io.IOException;
 import java.util.Properties;
 
 import org.apache.log4j.PropertyConfigurator;
@@ -26,6 +27,7 @@ import com.zimbra.common.localconfig.LC;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.cs.account.Provisioning;
 import com.zimbra.cs.extension.ExtensionUtil;
+import com.zimbra.cs.store.StoreManager;
 import com.zimbra.common.util.ZimbraLog;
 
 
@@ -92,6 +94,13 @@ public class ImapDaemon {
             props.load(new FileInputStream(IMAPD_LOG4J_CONFIG));
             PropertyConfigurator.configure(props);
             maybeInitEphemeralBackendExtension();
+
+            String imapdClassStore=LC.imapd_class_store.value();
+            try {
+                StoreManager.getInstance(imapdClassStore).startup();
+            } catch (IOException e) {
+                throw ServiceException.FAILURE(String.format("Unable to initialize StoreManager: %s", imapdClassStore), e);
+            }
 
             if(isZimbraImapEnabled()) {
                 ImapDaemon daemon = new ImapDaemon();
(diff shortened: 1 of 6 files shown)


store/src/java/com/zimbra/cs/imap/ImapDaemon.java
```diff
@@ -93,6 +95,12 @@ public static void main(String[] args) {
             PropertyConfigurator.configure(props);
             maybeInitEphemeralBackendExtension();
 
+            try {
+                StoreManager.getInstance(LC.imapd_class_store.value()).startup();
+            } catch (IOException e) {
+                throw ServiceException.FAILURE("Unable to initialize StoreManager.", e);
```
fatkudu: Worth logging the LC value that causes a problem?

store/src/java/com/zimbra/cs/imap/NioImapRequest.java
```diff
@@ -53,6 +53,9 @@ private void parseLiteral(byte[] b) throws IOException {
 
     private void parseCommand(String line) throws IOException, ProtocolDecoderException {
         addPart(line);
+        if (line.endsWith("\r\n")) {
+            line = line.substring(0, line.lastIndexOf('\r'));
```
fatkudu: What's the problem being addressed here?  The comment says the same as the code - what is being done but not why.  https://tools.ietf.org/html/rfc3501#section-2 says this:

> All interactions transmitted by client and server are in the form of lines, that is, strings that end with a CRLF.

So this feels wrong to me.  I'm also rather puzzled why this change and the previous change affect imapd but apparently don't affect either of the embedded variants?

gordyt:
@fatkudu I addressed both of your questions and pushed up a rebased branch.  Had checked with Travis about this:

    commit f0b4530edce51104fab57c50a5a6652281069ce9
    Author: Travis McLane <travis.mclane@synacor.com>
    Date:   Mon Jun 5 17:45:02 2017 -0500
    
        ZCS-1580 trim \r\n from literal

After we talked about it I decided to rebase it out and retest.  I think the problem may have been elsewhere because I am not seeing the issue with the updated code.

PR #181, opened 2017-06-09
ZMS-632 update description of ip throttling parameters

tmclane:


store/conf/attrs/zimbra-attrs.xml
```diff
@@ -3295,16 +3295,16 @@ TODO - add support for multi-line values in globalConfigValue and defaultCOSValu
...
-    from that IP address will be throttled.  Therefore, all users from
+    from that IP address will be throttled for that protocol.  Therefore, all users from
     the same IP will contribute to (and be affected by) this counter.
     Logins using all protocols (POP3/POP3S/IMAP/IMAPS) will affect
-    this counter (the counter is aggregate for all protocols, *not*
-    separate).  If this value is set to 0, then no limiting will take
-    place for any IP.</desc>
+    this counter.  Each protocol will have it's own limit counter.
+    If this value is set to 0, then no limiting will take
+    place for any IP unless it is overridden with zimbraReverseProxyIPLoginPop3Limit or zimbraReverseProxyIPLoginImapLimit.</desc>
```
fatkudu: My expectation is that `zimbraReverseProxyIPLoginLimit` would be completely ignored if either `zimbraReverseProxyIPLoginImapLimit` or `zimbraReverseProxyIPLoginPop3Limit` was set to a non-zero value and that it is important that that is made clear here.

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -3295,16 +3295,16 @@ TODO - add support for multi-line values in globalConfigValue and defaultCOSValu
...
 </attr>
 
 <attr id="623" name="zimbraReverseProxyIPLoginLimitTime" type="integer" min="0" cardinality="single" optionalIn="globalConfig" since="5.0.3">
```
fatkudu: I'm guessing `zimbraReverseProxyIPLoginLimitTime` will be ignored for IMAP and POP3 protocols (does it get used for SOAP/other?) if either `zimbraReverseProxyIPLoginImapLimitTime` or `zimbraReverseProxyIPLoginPop3LimitTime` is set?  If so, that should be made clear here.

store/conf/attrs/zimbra-attrs.xml
tmclane: Ok. I'll do some more word-smithing on it.

store/conf/attrs/zimbra-attrs.xml
tmclane: The PR was updated. Please let me know if it's clear enough now or if perhaps it should be rewritten more.

PR #182, opened 2017-06-13
ZMS-1536

tmclane:
Ensure that ip addresses that are negative numbers are processed as if they are positive.
This ensures that the index is within [ 0, pool.size() )

### DIFF
--- a/store/src/java-test/com/zimbra/cs/imap/ImapLoadBalancingMechanismTest.java
+++ b/store/src/java-test/com/zimbra/cs/imap/ImapLoadBalancingMechanismTest.java
@@ -84,6 +84,10 @@ public final class ImapLoadBalancingMechanismTest {
         Assert.assertEquals(s1, mech.getImapServerFromPool(req, pool));
         headers.put(ImapLoadBalancingMechanism.ClientIpHashMechanism.CLIENT_IP, "127.0.0.4");
         Assert.assertEquals(s2, mech.getImapServerFromPool(req, pool));
+        headers.put(ImapLoadBalancingMechanism.ClientIpHashMechanism.CLIENT_IP, "192.168.56.1");
+        req = new MockHttpServletRequest(null, null, null, 123, "127.0.0.1", headers);
+        Assert.assertEquals("Expected address 192.168.56.1 to hash to s1", s1, mech.getImapServerFromPool(req, pool));
+
 
 		/* Verify we get the same results using IPV6 */
         headers.put(ImapLoadBalancingMechanism.ClientIpHashMechanism.CLIENT_IP, "::FFFF:127.0.0.2");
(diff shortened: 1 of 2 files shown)


tmclane:
Removed usage of Math.abs since Codacy doesn't want java.lang to be used.

store/src/java/com/zimbra/cs/imap/ImapLoadBalancingMechanism.java
```diff
@@ -182,6 +182,8 @@ public Server getImapServerFromPool(HttpServletRequest httpReq, List<Server> poo
                 pool.sort(serverComparator);
                 int clientIpHash = InetAddress.getByName(httpReq.getHeader(CLIENT_IP)).hashCode();
                 int serverPoolIdx = clientIpHash % pool.size();
+                if (serverPoolIdx < 0)
+                    serverPoolIdx = -serverPoolIdx;
```
gordyt: Trav is it better get absolute value  of `serverPoolIdx` or `clientIpHash`? Or does it matter?

tmclane:
Swapped negativity test to the clientIpHash per Gordon's comment (that I couldn't see).

store/src/java-test/com/zimbra/cs/imap/ImapLoadBalancingMechanismTest.java
```diff
@@ -84,6 +84,9 @@ public void ClientIpHashMechanismIpHashFromPool()
         Assert.assertEquals(s1, mech.getImapServerFromPool(req, pool));
         headers.put(ImapLoadBalancingMechanism.ClientIpHashMechanism.CLIENT_IP, "127.0.0.4");
         Assert.assertEquals(s2, mech.getImapServerFromPool(req, pool));
+        headers.put(ImapLoadBalancingMechanism.ClientIpHashMechanism.CLIENT_IP, "192.168.56.1");
+        req = new MockHttpServletRequest(null, null, null, 123, "127.0.0.1", headers);
+        Assert.assertEquals(s1, mech.getImapServerFromPool(req, pool));
```
fatkudu: General minor comment that it can save time looking at failed test results if there is a clear one line reason for the failure, so I prefer variants of assertEquals where the first arg is a string description.

tmclane: That is a good point.  Would we want to start doing this kind of update anytime a test file is updated?  Could be a lot of work to update all of the tests in the file in this fashion.

fatkudu: I tend to do a mix but would be good for new stuff to do it I think.

tmclane: I have updated the unit test to use the form with a description of what is failing.

tmclane: Just for the new tests I added at this time.

PR #183, opened 2017-06-14
ZCS-1716:build.xml target all->publish-local-all

fatkudu:
builds broken due to all running tests which are failing.
Back out that change so that we can create builds again.
ZCS-1715 opened for a more complete addressing of this problem.


Prashantsurana:
./build.pl --ant-options "-DskipTests=true"   is the option.

fatkudu:
Had a chat with Prashant and we agreed that we would leave things as they are (with test running disabled by default) until we agree a way forward.

PR #184, opened 2017-06-14
ZCS-1033 external account shows 500 error when password restrictions not met

rohan-ambasta:
[Issue]:
This bug occurred because ServletException was thrown in case of any exception while creating the virtual  account. 
[Fix]:
Handle the exception and send the error code and error message to extuserprov.jsp. I will create a separate JIRA ticket to handle the UI change
[Testing]:
Manual testing

### DIFF
--- a/store/src/java/com/zimbra/cs/service/ExternalUserProvServlet.java
+++ b/store/src/java/com/zimbra/cs/service/ExternalUserProvServlet.java
@@ -47,6 +47,7 @@ import com.zimbra.common.util.LogFactory;
 import com.zimbra.common.util.StringUtil;
 import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.cs.account.Account;
+import com.zimbra.cs.account.AccountServiceException;
 import com.zimbra.cs.account.AuthToken;
 import com.zimbra.cs.account.AuthTokenException;
 import com.zimbra.cs.account.Domain;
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/service/ExternalUserProvServlet.java
```diff
@@ -362,8 +416,10 @@ private static void provisionVirtualAccountAndRedirect(HttpServletRequest req, H
             enableAppFeatures(grantee, viewTypes);
 
             setCookieAndRedirect(req, resp, grantee);
-        } catch (Exception e) {
-            throw new ServletException(e);
+        } catch (ServiceException e) {
+            throw e;
+        }  catch (Exception e) {
+            throw ServiceException.TEMPORARILY_UNAVAILABLE();
```
rupalid: Log the exception at debug level

rohan-ambasta: done

store/src/java/com/zimbra/cs/service/ExternalUserProvServlet.java
```diff
@@ -277,19 +278,70 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws S
...
+            sendHtmlResponse(resp, reqHeaders);
+        } else {
+            for (Map.Entry<String, String> entry : errorAttrs.entrySet()) {
+                req.setAttribute(entry.getKey(), entry.getValue());
+            }
+            forward(req, resp);
+        }
+    }
+
+    private void forward(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
```
shrikant-prasad: similar code is there at several places in this class, better to make the jsp page as argument of the method and use this method at all these places to remove duplicate code. same for sendHtmlResposne method.

rohan-ambasta: done

PR #192, opened 2017-06-15
ZCS-1726 Resolve only xercesImpl from red hat

rupalid:
Fix the xerces resolver to use redhat repo only for xerces#xercesImpl;2.9.1-patch-01 dependency

### DIFF
--- a/build-ivysettings.xml
+++ b/build-ivysettings.xml
@@ -27,6 +27,6 @@
     </filesystem>
   </resolvers>
      <modules>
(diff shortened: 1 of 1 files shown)


build-ivysettings.xml
```diff
@@ -27,6 +27,6 @@
     </filesystem>
   </resolvers>
      <modules>
-      <module organisation="xerces" resolver="maven-redhat"/>
+      <module organisation="xerces" name="xercesImpl" resolver="maven-redhat"/>
```
fatkudu: As elsewhere.  We should be consistent and match by version as well

PR #196, opened 2017-06-15
ZMS-672 add upstream imap servers to whitelist

tmclane:
This treats all servers listed in `zimbraReverseProxyUpstreamImapServers` as 'trusted' servers which are automatically ignored by the DosFilter in the mailboxd.

Note:  The accessors to the `_whitelist` member of the `org.eclipse.jetty.servlets.DosFilter` class do not prevent duplicates from being included in the whitelist.  Likely not to cause any performance issues in the real world since most clusters are not all that large.  Just something I thought I would bring up since I discovered this fact during debugging.

I should also note that in my correspondence with Karl, Mark, and others that Mark believes this is a safe thing to do for all installation configurations he is aware of at this time.

### DIFF
--- a/store/src/java/com/zimbra/cs/servlet/DoSFilter.java
+++ b/store/src/java/com/zimbra/cs/servlet/DoSFilter.java
@@ -55,6 +55,16 @@ public class DoSFilter extends org.eclipse.jetty.servlets.DoSFilter {
             for (String addr : addrs) {
                 addWhitelistAddress(addr);
             }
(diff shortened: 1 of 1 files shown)


tmclane:
@plobbes Thanks for the input here as well!  I am going to wait on actually merging this change.
I think this needs some further thought on the best way forward and if this feature should even remain in the product. @wdomburg Also chimed in yesterday with some good points.

tmclane:
I am going to backburner this.  It seems the best option at this time is to add some documentation surrounding the Remote IMAP installation to let the installer know that the new server will need to be whitelisted in order to bypass the DoSFilter.  Servers shouldn't be added/subtracted all that often so the additional configuration burden should be relatively light.

PR #197, opened 2017-06-16
ZCS-128: auto accept/decline not sent from resource on appointment update

shrikant-prasad:
When appointment was updated, auto response from resource was not sent as to send the auto response, RSVP should be true for the resource attendee. The rsvp check was added as fix for bug 39404. 
But in fix for bug 84830, check was added to set RSVP true only if the parstat for attendee is needs-action. 
I have modified the above check to set RSVP true for resource irrespective of the parstat value.
Verified that for resource attendee, accept response is correctly displayed in sierra calendar app configured through CalDAV(in bug 84830 this was not working).


Victor-D:
Hello guys

Is it the same bus as

shrikant-prasad:
Hi Victor, 

It's not same as bug 107702. Need more details about the scenarios in bug 107702. 

Regards, 
Shrikant

PR #199, opened 2017-06-19
ZCS-1036 Publishing store-test

rupalid:
Adding task to publish zimbrastore-test.jar to .zcs-deps.

### DIFF
--- a/store/build.xml
+++ b/store/build.xml
@@ -298,6 +298,9 @@
       <param name="jar.build.dir" value="${server.dir}/build/test-classes/"/>
     </antcall>
   </target>
+  <target name="publish-store-test" depends="test-jar">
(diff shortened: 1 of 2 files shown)


store/build.xml
```diff
@@ -298,6 +298,9 @@
       <param name="jar.build.dir" value="${server.dir}/build/test-classes/"/>
     </antcall>
   </target>
+  <target name="publish-store-test" depends="test-jar">
+      <copy file="${build.dir}/zimbrastore-test.jar" tofile="${user.home}/.zcs-deps/zimbrastore-test-1.0.jar"/>
```
fatkudu: build-ivysettings.xml refers to ${dev.home}/.zcs-deps - suggest using that for consistency.

grishick: yes, this is actually quiet important when development environment co-exists with production and user.home will point to /opt/zimbra

rupalid: fixed

PR #200, opened 2017-06-19
ZCS-1670: IMAP notifications in response headers

ilyusha:
The current remote IMAP notification mechanism introduces race conditions that result in bugs. The race condition occurs when an IMAP operation potentially returns before the corresponding changes are propagated back to the originating IMAP server via waitsets; this causes problems when a subsequent IMAP operation sees a stale view of the IMAP data. [ZCS-1482](https://jira.corp.synacor.com/browse/ZCS-1482) is an example of this type of bug.

This PR introduces two changes to IMAP notifications:
1. All IMAP notifications are returned in the SOAP response headers.
_WaitSetRequest_ and _SoapSession_ now encode pending modifications in an identical way, using common logic. This code is now contained in the new _PendingModifications::encodeFolderModifications_ method. This method accepts an optional _folderInterests_ argument; _WaitSetRequest_ makes use of this argument, while _SoapSession_ does not.
2. A waitset is _not_ triggered if the same pending notifications are to be returned in the response headers. This is done so that IMAP servers don't receive two copies of the notifications through the two mechanisms. 
SOAP requests issued by _ZMailbox_ can include an optional waitset ID in the header. This is set/unset by the the new _ZMailbox::setCurWaitSetID_ and _ZMailbox::unsetCurWaitSetID_ methods. When _ImapServerListener_ adds an _ImapRemoteSession_ listener, it sets the current waitset ID on that remote session's ZMailbox instance. If a waitset is initialized, the new ID is propagated to all ZMailboxes in the sessionMap. Likewise, when a waitset is destroyed, the field is unset on all ZMailboxes.
On the mailbox server, this ID is  included in _ZimbraSoapContext_ and exposed by _SoapSession::getCurWaitSetID_. Since SoapSessions are cached, _DocumentHandler::getSession_ sets the waitset ID directly on the cached session if its waitset ID doesn't match that of the request. Lastly, _WaitSetSession::notifyPendingChanges_ checks whether this ID matches the ID of its waitset, and if does, does not call _signalDataReady_. In other words, an IMAP server doesn't receive waitset notifications for operations _it_ triggered, instead getting all changes in the response headers. This guarantees that all changes from an IMAP operation have taken effect by the time the operation returns.

With these changes, tests in _TestRemoteImapNotifications_ no longer have to repeatedly sleep and check if the changes have been applied. However, the tests had to be changed to make use of the same ZMailbox instance that is used by the IMAP listener, as only that instance has the ZEventHandler registered that can process IMAP notifications. This instance is accessed by the new _TestRemoteImapNotifications_:: getImapZMailboxForFolder_ method.

Finally, _ImapResponse::readUntagged_ has been updated to not throw an exception when receiving an NOOP response. This makes it easier to debug a running server, as the _ImapHandler_ sends a NOOP after 15 seconds of inactivity, which previously led to a _ParseException_.

This fix resolves [ZCS-1482](https://jira.corp.synacor.com/browse/ZCS-1482) and [ZMS-675](https://jira.corp.synacor.com/browse/ZMS-675), and potentially others.

### DIFF
--- a/client/src/java/com/zimbra/client/ZMailbox.java
+++ b/client/src/java/com/zimbra/client/ZMailbox.java
@@ -553,6 +553,7 @@ public class ZMailbox implements ToZJSONObject, MailboxStore {
     private final ZMailboxLock lock;
     private int lastChangeId = 0;
     private NotificationFormat mNotificationFormat = NotificationFormat.DEFAULT;
+    private String mCurWaitSetID = null;
 
     private final List<ZEventHandler> mHandlers = new ArrayList<ZEventHandler>();
 
@@ -863,7 +864,7 @@ public class ZMailbox implements ToZJSONObject, MailboxStore {
                 if(nosession) {
                     return mTransport.invoke(request, false, nosession, requestedAccountId);
                 } else {
-                    return mTransport.invoke(request, nosession, requestedAccountId, this.mNotificationFormat);
+                    return mTransport.invoke(request, nosession, requestedAccountId, this.mNotificationFormat, this.mCurWaitSetID);
                 }
             } catch (SoapFaultException e) {
                 throw e; // for now, later, try to map to more specific exception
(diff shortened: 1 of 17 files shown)


grishick:
Great summary.
One small note on the summary: s/TestImapRemoteNotifications/TestRemoteImapNotifications

ilyusha:
@grishick very good point.

ilyusha:
The latest set of commits refactor _TestRemoteImapNotifications_ to support testing notifications that come through both waitsets and response headers. _TestRemoteImapNotifications_ is now an abstract class that implements the test methods; _TestRemoteImapNotificationsViaWaitsets_ and _TestRemoteImapNotificationsViaMailbox_ check the two notification mechanisms.

ilyusha:
The last commit should make codacy happy

PR #201, opened 2017-06-19
ZCS-1547: IMAP notifications for unlisted tags

ilyusha:
Prior to this change, unlisted tags were not returned in the notification response headers. This resulted in IMAP clients not being notified of tag changes if the tags were not explicitly created on the mailbox first. This change adds a new boolean field _Tag.isImapVisible_ that is set to _true_ if the tag is created by _NormalizedTags_, provided the new _imapVisible_ constructor argument is used. This only happens via the _Mailbox::alterTag_ and _Mailbox::setTags_ methods.

one of the assertions in SharedImapTests.testStoreTags() has been updated to check the absence of the target tag as opposed to checking that the flag list is empty, since the \Seen flag can be set on the message.


ilyusha:
This branch is based off the unmerged bugfix/zcs-1670 branch, and will be updated to be based off feature/imap when zcs-1670 is merged.

PR #202, opened 2017-06-19
ZMS-358: updated tag deletion IMAP tests

ilyusha:
Enabled IMAP notification tests for tag deletion, as well as updated them to check the presence/absence of the tag on messages themselves in addition to the folder.

### DIFF
--- a/store/src/java/com/zimbra/qa/unittest/TestRemoteImapNotifications.java
+++ b/store/src/java/com/zimbra/qa/unittest/TestRemoteImapNotifications.java
@@ -257,7 +257,6 @@ public class TestRemoteImapNotifications {
         assertEquals("Size of map returned by fetch 2", 1, mdMap.size());
     }
 
-    @Ignore("TODO - support for delete tag notifications")
     @Test
     public void testDeleteTagNotificationActiveFolder() throws Exception {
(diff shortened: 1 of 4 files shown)


store/src/java/com/zimbra/qa/unittest/TestRemoteImapNotifications.java
```diff
@@ -275,15 +274,21 @@ public void testDeleteTagNotificationActiveFolder() throws Exception {
 
         Flags flags = info.getPermanentFlags();
         assertTrue(flags.contains(new Atom(tagName)));
+        Map<Long, MessageData> mdMap = connection.fetch("1:*", "(FLAGS)");
+        Flags msgFlags = mdMap.get(1L).getFlags();
+        assertTrue(msgFlags.contains(new Atom(tagName)));
```
fatkudu: I'd prefer the assert variants which leave a clear message in the reports when they fail.
Perhaps we need to discuss as a team whether I should carry on saying this or whether it isn't important enough?

ilyusha: No, I think you're right. I would prefer to use the message variants as well.

PR #203, opened 2017-06-19
ZMS-357: local/embedded/remote IMAP notification tests

ilyusha:
While tests exist for checking remote IMAP notifications of deleted folders, this PR moves all IMAP notification tests into a framework similar to _SharedImapTests_ where test are defined in an abstract class and server configuration is moved to subclasses. This way, we can test IMAP notifications for the local, embedded remote, and IMAP daemon modes of operation. Since there is a lot of common code between this and _SharedImapTests_, a new class _ImapTestBase_ has been introduced as a base class for both _SharedImapTests_ and the new _SharedImapNotificationTests_ class.


store/src/java/com/zimbra/qa/unittest/TestImapNotificationsViaEmbeddedLocal.java
```diff
@@ -0,0 +1,47 @@
...
+import com.zimbra.cs.imap.ImapServerListenerPool;
+
+public class TestImapNotificationsViaEmbeddedLocal extends SharedImapNotificationTests {
+
+    @Before
+    public void setUp() throws ServiceException, IOException, DocumentException, ConfigException  {
+        saveImapConfigSettings();
+        TestUtil.setLCValue(LC.imap_always_use_remote_store, String.valueOf(false));
+        imapServer.setReverseProxyUpstreamImapServers(new String[] {});
+        super.sharedSetUp();
```
grishick: This test should use assumptions to check that embedded local IMAP is enabled. If you install remote IMAP and disable local one, all tests in this suite will fail.

store/src/java/com/zimbra/qa/unittest/TestImapNotificationsViaEmbeddedRemote.java
```diff
@@ -0,0 +1,50 @@
...
+import com.zimbra.cs.imap.ImapServerListener;
+import com.zimbra.cs.imap.ImapServerListenerPool;
+
+public class TestImapNotificationsViaEmbeddedRemote extends SharedImapNotificationTests {
+    @Before
+    public void setUp() throws ServiceException, IOException, DocumentException, ConfigException  {
+        saveImapConfigSettings();
+        TestUtil.setLCValue(LC.imap_always_use_remote_store, String.valueOf(true));
+        imapServer.setReverseProxyUpstreamImapServers(new String[] {});
+        super.sharedSetUp();
```
grishick: this test should also check that embedded IMAP is enabled

PR #205, opened 2017-06-20
zcs-1530

grishick:
Replace strings and Element with JAXB classes in the test.

### DIFF
--- a/store/src/java-test/com/zimbra/cs/service/mail/GetMsgTest.java
+++ b/store/src/java-test/com/zimbra/cs/service/mail/GetMsgTest.java
@@ -55,8 +55,6 @@ import com.zimbra.cs.mailbox.MailboxTestUtil;
 import com.zimbra.cs.mailbox.calendar.Invite;
 import com.zimbra.cs.mailbox.calendar.ZOrganizer;
 
-
-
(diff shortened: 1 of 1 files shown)


store/src/java-test/com/zimbra/cs/service/mail/GetMsgTest.java
```diff
@@ -54,11 +53,16 @@
...
+import com.zimbra.soap.mail.message.CreateMountpointRequest;
+import com.zimbra.soap.mail.message.FolderActionRequest;
+import com.zimbra.soap.mail.message.GetMsgRequest;
+import com.zimbra.soap.mail.type.FolderActionSelector;
+import com.zimbra.soap.mail.type.MsgSpec;
+import com.zimbra.soap.mail.type.NewMountpointSpec;
 
 public class GetMsgTest {
-
+    private static final String desc = "The following is a new meeting " + "request";
```
tmclane: Any reason this isn't just a single String?  Unsure why it needs a + in there.  I see the original line had one.

grishick: heh, must be the automatic formatting that Eclipse does.

PR #218, opened 2017-06-24
zcs 1556 imap share copy

fatkudu:
NOT ready for merge - posting for safety and perhaps early review.
* Local copy should now work from a share.
* Share SELECT with remote should now work (diff Jira ticket exists I think but due to ZCS project deletion, don't know which)
* Remote copy from a share NOT working yet (Needs ItemActionRequest enhancements).
* debug code present.

### DIFF
--- a/client/src/java/com/zimbra/client/ZMailbox.java
+++ b/client/src/java/com/zimbra/client/ZMailbox.java
@@ -6046,10 +6046,13 @@ public class ZMailbox implements ToZJSONObject, MailboxStore {
      * @param targetFolder - Destination folder
      */
     @Override
-    public void copyItemAction(OpContext ctxt, String authenticatedAcctId, ItemIdentifier targetFolder,
+    public List<String> copyItemAction(OpContext ctxt, String authenticatedAcctId, ItemIdentifier targetFolder,
             List<Integer> idlist)
     throws ServiceException {
-        copyItemAction(targetFolder.id, idlist);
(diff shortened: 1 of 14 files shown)


grishick:
I don't think we have to instantiate ZMailbox for a sharer's mailbox when performing a copy between owned and shared folders, because ItemActionHelper already takes care of sorting this out. Before we made changes for remote mailbox, ImapHandler already had the logic to use ZMailbox if source and target folders are not in the same mailbox. I assume that logic was there specifically to avoid instantiating local Mailbox object for the sharer's mailbox, which may reside on another mail store. The code that has gone missing from before remote IMAP changes is this:

                    ```
for (String target : op.getCreatedIds()) {
                        createdList.add(new ItemId(target, selectedFolder.getAuthenticatedAccountId()).getId());
                    }
```
wouldn't we be able to fix this bug by simply adding this back?

fatkudu:
@grishick I think this pull request is mostly a false start.  The code you recommend is essentially part of this pull request but the thing I believe I'm missing is that when the sharer's mailbox is remote we were using ImapProxy before and now we are not - I was trying to use ZMailbox and to get that to work but using ImapProxy makes much more sense.  Just tried it on the develop branch with 2 servers and IMAP commands are seen on the remote machine where appropriate, using the X-ZIMBRA mechanism to login as the sharee working on behalf of the sharer.

PR #221, opened 2017-06-26
ZCS-1789: removed IMAP daemon notification tests

ilyusha:
_TestImapDaemonNotificationsViaMailbox_ and _TestImapDaemonNotificationsViaWaitsets_ didn't work because they assumed access to the IMAP listeners, like the embedded remote tests. Since these are not available, the only way to test notifications is to repeatedly check/wait for the expected response to be propagated to the server. However, at this point this is indistinguishable from a regular IMAP test. In other words, testing the different IMAP notification mechanisms doesn't really make sense for the daemon case.


ilyusha:
Well, the defining feature of notification tests, at least to me, was that the notification tests actually grabbed the ImapServerListener instances and acted on them directly, be it checking the sequence number (for waitset notifications) or running requests through the listener's ZMailbox instance (for mailbox notifications). We can't really do this for the daemon, so we're left with simply running IMAP commands and waiting for them to finish. That said, I might be overthinking this - it's probably worthwhile to do just that, as much as we usually try to avoid "check-sleep" loops in unit tests.

ilyusha:
Closing this in lieu of https://github.com/Zimbra/zm-mailbox/pull/225

PR #225, opened 2017-06-28
ZCS-1818: TestImapDaemonNotifications

ilyusha:
Replaced the previous (broken) _TestImapDaemonNotificationsViaWaitsets_ and _TestImapDaemonNotificationsViaMailbox_ with a single class _TestImapDaemonNotifications_. This  class runs the tests in _SharedImapNotificationTests_ by polling the server for expected results for up to six seconds.
To get this to work, the _MailboxOperation_ inner helper class is extended with a new _checkResult_ method. All assertions performed by the tests are moved into this new method. _TestImapDaemonNotifications::runOp_ calls _checkResult_, catching  _AssertionErrors_ and retrying for up to the allotted time.
The other subclasses of _SharedImapNotificationTests_ have been updated to make use of the _checkResult_ method as well.

I encountered an issue where _testNotificationsEmptyActiveFolder_  would fail with a newly-restarted mailboxd. I tracked this down to imapd hitting the _imap_max_consecutive_error_ LC value of 5, closing the connection. Increasing the sleep time in _runOp()_ from 500ms to 1000ms in the retry loop solved the issue, since the change is applied on the daemon before the command has a chance to fail five times in a row. I attempted to increase the value of _imap_max_consecutive_error_ for the duration of the test, but this did not work because there is currently no way to force _ImapDaemon_ to refresh its LC instance.


store/src/java/com/zimbra/qa/unittest/TestImapDaemonNotifications.java
```diff
@@ -32,4 +34,23 @@ public void tearDown() throws ServiceException, DocumentException, ConfigExcepti
...
+        op.run(mbox);
+        AssertionError saved = null;
+        int timeout = 6000;
+        while(timeout > 0) {
+            try {
+                op.checkResult();
+                return;
+            } catch (AssertionError e) {
+                saved = e;
+                Thread.sleep(500);
```
fatkudu: I don't think my comments in the previous pull request have been addressed.  The fact that you are needing to introduce waits to get these tests to pass suggests one of two things to me:
* The tests are a waste of time because they are testing something that isn't relevant to the IMAP daemon.
* The IMAP daemon is fundamentally broken because it doesn't have access to notifications it needs to have access to and therefore will not function correctly as an IMAP server.
Either way, I don't see this pull request being useful.

fatkudu:
Hmm.  The pull request view seems to be hiding my concerns because I associated them with some slightly out of date code but I think they are still applicable, so including them again here:
I don't think my comments in the previous pull request have been addressed.  The fact that wait loops need to be introduced to get these tests to pass suggests one of two things to me:
* The tests are a waste of time because they are testing something that isn't relevant to the IMAP daemon.
* The IMAP daemon is fundamentally broken because it doesn't have access to notifications it needs to have access to and therefore will not function correctly as an IMAP server.
Either way, I don't see this pull request being useful.

ilyusha:
Gren you're absolutely right, I got my wires crossed a bit. It seems like this is set up to test the scenario where a _different_ client makes changes that the daemon picks up via waitsets. These tests should be able to pass with no wait loops.

ilyusha:
Actually, nevermind. I think this actually is an appropriate test of waitset notifications for the daemon, since the changes in the _MailboxOperation.runOp()_ method aren't being made against the  daemon itself via IMAP commands but via a _ZMailbox_ client. The daemon picks them up asynchronously via waitsets, hence the wait loops. If the commands were run via _ImapConnection_ like in _SharedImapTests_, there would be no need to wait.

fatkudu:
@ilyusha so, if I understand correctly, these are changes made by other clients, so it doesn't matter if the daemon has a slightly inconsistent view of the mailbox - because the daemon's view IS consistent with changes made via its IMAP connection.  Approving on that basis.

ilyusha:
@fatkudu exactly, we are testing that the daemon catches up to the mailbox via waitsets.

PR #226, opened 2017-06-28
Bugfix/zcs 67

tmclane:
Adds a new LDAP attribute `zimbraReverseProxyUpstreamFairShmSize` to allow user control of the `upstream_fair_shm_size` Nginx configuration value.

Note: Additional PR will be forthcoming that leverages these changes.


tmclane:
@fatkudu I do think those changes were removed before. I feel like it would be easier to just accept a positive number and let the ProxyConfGen add the designator to the output when generating the configuration files. We could instead note that this means 'pages' and compute the actual number of bytes to be output. Would that be better? I feel like using the raw number as the 'k' value will give the administrator more control.

fatkudu:
@tmclane - happy with whatever you decide on that.

PR #228, opened 2017-06-29
ZCS-1597: child folder path updated after parent rename

ilyusha:
This bug was caused by the _ZMailbox_ not updating subfolder paths in its folder cache when receiving a _ZModifyFolderEvent_. More specifically, _ZFolder.getPath()_ returns the _mAbsolutePath_ field if it is set, falling back to determining the path based on the parent folder otherwise. To fix this, I added a method _ZFolder.refreshAbsolutePath()_ that recalculates the absolute path based on the parent folder; this also happens recursively for all subfolders. This method is called in _ZFolder.modifyNotification()_, only if the folder name is modified.

A new test method _testRenameParentFolder_ has been added to _SharedImapTests_. The test creates a parent folder with two layers of children, renames the parent folder, and verifies that the children have been renamed accordingly.

EDIT:

renamed _refreshAbsolutePath_ to _updateAbsolutePaths_ and added a method _updateAbsolutePathOfSubfolders_ for clarity. As per a suggestion by Travis, the name of the parent folder is now passed to these methods on the child folder so that nested folders don't have to keep re-calculating the parent folder names.

### DIFF
--- a/client/src/java/com/zimbra/client/ZFolder.java
+++ b/client/src/java/com/zimbra/client/ZFolder.java
@@ -406,6 +406,7 @@ public class ZFolder implements ZItem, FolderStore, Comparable<Object>, ToZJSONO
     public void modifyNotification(ZModifyEvent event) throws ServiceException {
         if (event instanceof ZModifyFolderEvent) {
             ZModifyFolderEvent fevent = (ZModifyFolderEvent) event;
+            boolean nameChange = !mName.equals(fevent.getName(mName));
             mName = fevent.getName(mName);
             mParentId = fevent.getParentId(mParentId);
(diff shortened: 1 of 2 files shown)


tmclane:
Have you thought about passing the parent path down to the children so they don't have to recompute it over and over again?  I doubt that this happens all that often so it is probably not something to worry about too much.

ilyusha:
@tmclane you're right that it's not heavily used code and I wouldn't expect the recursion to ever go very deep, but it's a good suggestion anyways.

trantor:
Well, AFAICS these changes messed up both the output of zmmailbox getAllFolders / getFolder _and_ the access to shared folders in the HTML web client. Nice chain effect.

PR #229, opened 2017-06-29
Bugfix/zms 666

tmclane:
* Adds a new LDAP attribute `zimbraReverseProxyUpstreamFairShmSize` to control the 'upstream_fair_shm_size' configuration value in the Nginx template files.
* Updates ProxyConfGen.java to process 'upstream.fair.shm.size' and replace the variable with
nothing if not present in LDAP, or with `upstream_fair_shm_size <value>k` 
   Example: `upstream_fair_shm_size 32k`

If an invalid value is placed in the attribute it is replaced with the minimum which is equal to 32k.

Note: This is based on `feature/imap`. I am unsure which branch i.e. release should be targeted at this time. If this is the wrong one per PM direction I will rebase and regenerate the getter/setters after fixing the attribute id if necessary.

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9526,5 +9526,10 @@ TODO: delete them permanently from here
   <desc>port number on which the remote IMAP SSL server should listen</desc>
 </attr>
 
+<attr id="3017" name="zimbraReverseProxyUpstreamFairShmSize" type="integer" min="32" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="8.8.1" requiresRestart="nginxproxy">
+  <globalConfigValue>32</globalConfigValue>
+  <desc>Configures the 'upstream_fair_shm_size' value used by Nginx to set the size of shared memory for storing information about the busy-ness of backends. Values are in kilobytes.</desc>
+</attr>
+
 </attrs>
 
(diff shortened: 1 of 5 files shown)


store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9526,5 +9526,10 @@ TODO: delete them permanently from here
   <desc>port number on which the remote IMAP SSL server should listen</desc>
 </attr>
 
+<attr id="3017" name="zimbraReverseProxyUpstreamFairShmSize" type="integer" min="32" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="8.8.1" requiresRestart="nginxproxy">
```
rohan-ambasta: I think instead of using type="integer", you should use type="long" because can also contain human readable memory unit values: ex 1KB => 1024 bytes

tmclane: I chose integer on purpose since I am unsure if anything other than 'k' is allowed.
This removes all ambiguity and prevents me from having to compute a value for 'k' from whatever unit the user specified.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
```diff
@@ -488,6 +488,40 @@ public void update() {
     }
 }
 
+class ProxyFairShmVar extends ProxyConfVar {
+
+    public ProxyFairShmVar() {
+        super("upstream.fair.shm.size", "zimbraReverseProxyUpstreamFairShmSize", "",
+                ProxyConfValueType.CUSTOM, ProxyConfOverride.CONFIG,
+                "Proxy IMAP banner message (contains build version if " +
```
grishick: This description seems to have been copied from somewhere else

tmclane: You are correct. I'll fix it.

tmclane:
@grishick Updated the description.

PR #236, opened 2017-07-05
ZCS-1556 IMAP mountpoint and /home/?/? support

fatkudu:


gordyt:
Gren does the test procedure documented in ZCS-1556 work for you with this change?

PR #239, opened 2017-07-06
ZCS-777 : Fix possible XXE [CWE-611] Issues

chinmay15:
Problem : possible XXE

Fix : applied fixes on factory classes as per XXE prevention cheat sheet (Ref. : https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Prevention_Cheat_Sheet)

Testing done : 
1) Unit tests to verify applied changes
2) ant test in zm-wsdl-test repo(no failures because of current changes)
3) Verified EWS.wsdl and ZimbraService.wsdl, none of them are using any includes or any external entities

Testing to be done by QA :
1) full soap/ews automation run and sanity testing around category and outbound messages in EWS

### DIFF
--- a/common/src/java/com/zimbra/common/soap/W3cDomUtil.java
+++ b/common/src/java/com/zimbra/common/soap/W3cDomUtil.java
@@ -34,6 +34,7 @@ import javax.xml.parsers.SAXParserFactory;
 import javax.xml.transform.OutputKeys;
 import javax.xml.transform.Source;
 import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerConfigurationException;
 import javax.xml.transform.TransformerException;
 import javax.xml.transform.TransformerFactory;
 import javax.xml.transform.TransformerFactoryConfigurationError;
@@ -57,6 +58,7 @@ import com.google.common.base.Strings;
 import com.google.common.io.Closeables;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.soap.Element.ElementFactory;
+import com.zimbra.common.util.Constants;
 import com.zimbra.common.util.Log;
 import com.zimbra.common.util.ZimbraLog;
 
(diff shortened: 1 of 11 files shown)


common/src/java/com/zimbra/common/soap/W3cDomUtil.java
```diff
@@ -91,29 +81,58 @@ private W3cDomUtil() {}
...
+        dbf.setIgnoringComments(true);
+        // protect against recursive entity expansion DOS attack and perhaps other things
+        dbf.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+        try {
+            // XXE attack prevention
+            dbf.setFeature(Constants.DISALLOW_DOCTYPE_DECL, true);
+            dbf.setFeature(Constants.EXTERNAL_GENERAL_ENTITIES, false);
+            dbf.setFeature(Constants.EXTERNAL_PARAMETER_ENTITIES, false);
+            dbf.setFeature(Constants.LOAD_EXTERNAL_DTD, false);
+            dbf.setXIncludeAware(false);
```
fatkudu: I know our WSDL makes use of includes.  Worth checking how some JAXB stuff works out and tests from the `zm-wsdl-test` repository.  What testing have you done?

chinmay15: I've done testing by verifying features on factory object in unit tests. Can you guide me on how to test this?

fatkudu: I've checked out the test results that Chinmay sent to me and I think the fact that a significant number of them pass is sufficient to be sure that this change is not the root of the remaining failures.

fatkudu:
Sorry to ask more but I do think that it is important to do call path analysis to be reasonably confident that the methods which now support reduced functionality are not being called under circumstances where that reduced functionality matters - for instance, I know WSDL is used in other parts of the project, for instance EWS - does that WSDL use includes and if so, are these routines used to process them?

chinmay15:
Checked in EWS.wsdl and ZimbraService.wsdl, none of them are using any includes or any external entities.

chinmay15:
Merged using git flow

commit e859950cea04f93ea101a6bdcc7ac4912efcef8e
Date:   Thu Jul 6 19:08:44 2017 +0530

    ZCS-777 : Fix possible XXE [CWE-611] Issues

    Removing unused import

common/src/java/com/zimbra/common/soap/W3cDomUtil.java
```diff
@@ -91,29 +80,58 @@ private W3cDomUtil() {}
...
+            dbf.setExpandEntityReferences(false);
+        } catch (IllegalArgumentException iae) {
+            ZimbraLog.misc.debug("Disabling doctype-decl not supported", iae);
```
plobbes: (Sorry for the late review) A few questions here...
1. Should we allow the disabling of doctype-decl failure to fall through and return dbf still?  That seems ill advised, but perhaps I'm missing something.
2. In addition, is the IllegalArgumentException specific to doctype-decl only? I'm guessing no and perhaps it's more just any setFeature or setXIncludeAware request which can throw this, and we didn't update the debug message appropriately.
3. Does the order of lines 91-95 matter?  Per Q1 above, If we're allowing a failure to successfully follow through, shouldn't we try each setFeature in a separate try/catch block in case only one of those attempts fails but the others can still succeed? Or is the assumption if one fails, they will all fail?

chinmay15: Please have a look at https://github.com/Zimbra/zm-mailbox/pull/242

1) IllegalArgumentException is removed
2) ParserConfigurationException is already added as throws, which is the only exception thrown by setFeature() method.
3) Reply to point 3 - Googled around the order and try-catching separately but i did not see try-catching separately anywhere.

PR #241, opened 2017-07-07
ZCS-1169:mandate "require" for Sieve extension

ykomiyam:
[Related issue]
ZCS-168

[Fix]
* Before executing the Sieve extension test or action, ZCS will check whether the corresponding capability string is declared in the "require" control.  If it is declared, such Sieve extension will be executed.  If it is not declared, the entire sieve filter will be failed and no filter will be applied to the message (message will be stored in the default folder (Inbox or Sent)).
* This ticket covers the following Sieve extension:
  - "fileinto" (RFC 5228) : "fileinto" action
  - "copy" (RFC 3894) : "fileinto" and "redirect" actions
  - "variables" (RFC 5229) : "set" action, ":encodeurl" argument for the notify action.
      (note: if "variables" is not declared, ${n} will be appeared as is)
  - "enotify" (RFC 5435) : "notify" action, "notify_method_capability" test, and "valid_notify_method" test.
* For the backward compatibility "zimbraSieveRequireControlRFCCompliant" attribute (account,cos,domain) is introduced.
  - FALSE (default) - Backword compatibility mode. The "require" control is optional. Sieve extension can be used without "require" control.
  - TRUE - RFC compliant mode. The "require" control is mandatory for the Sieve extension.

[test]
* All unit tests under /store/src/java-test/com/zimbra/cs/filter/ are PASS

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9496,5 +9496,10 @@ TODO: delete them permanently from here
   <globalConfigValue>TRUE</globalConfigValue>
   <desc>Whether to enable zimbra network next generation modules.</desc>
 </attr>
+
+<attr id="2118" name="zimbraSieveRequireControlRFCCompliant" type="boolean" cardinality="single" optionalIn="account,cos,domain" flags="accountCosDomainInherited" since="8.7.8">
+<defaultCOSValue>FALSE</defaultCOSValue>
+<desc>Whether the declaration of the Sieve extension feature is mandatory by the 'require' control. If TRUE, before ZCS evaluates a Sieve extension test or action, it checks the corresponding capability string at 'require' control; and if the capability string is not declared in the 'require', the entire Sieve filter execution will be failed. If FALSE, any Sieve extensions can be used without declaring the capability string in the 'require' control.</desc>
+</attr>
 </attrs>
 
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9497,8 +9497,8 @@ TODO: delete them permanently from here
   <desc>Whether to enable zimbra network next generation modules.</desc>
 </attr>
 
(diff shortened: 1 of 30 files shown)


rupalid:
please fix the review comment,it mentions "zimbraSieveRejectMailEnabled"

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9496,5 +9496,10 @@ TODO: delete them permanently from here
   <globalConfigValue>TRUE</globalConfigValue>
   <desc>Whether to enable zimbra network next generation modules.</desc>
 </attr>
+
+<attr id="2118" name="zimbraSieveRequireControlRFCCompliant" type="boolean" cardinality="single" optionalIn="account,cos,domain" flags="accountCosDomainInherited" since="8.7.8">
```
rupalid: Can we name the attribute "zimbraSieveRequireControlEnabled"

rupalid: I think we should make the attribute TRUE by default otherwise it will never get used.

store/src/java/com/zimbra/cs/filter/FilterUtil.java
```diff
@@ -857,6 +860,14 @@ public static String replaceVariables(ZimbraMailAdapter mailAdapter, String sour
             return sourceStr;
         }
         validateVariableIndex(sourceStr);
+        
+        try {
+            Require.checkCapability(mailAdapter, CAPABILITY_VARIABLES);
+        } catch (SyntaxException e) {
+            ZimbraLog.filter.info("\"variables\" capability is not declared. No variables won't be replaced");
```
rupalid: No variables will be replaced

chinmay15: No variables will be replaced or Variables won't be replaced

ykomiyam: Fixed.

ykomiyam: Fixed.

store/conf/attrs/zimbra-attrs.xml
ykomiyam: Name changed.  Default value is now TRUE.

PR #246, opened 2017-07-11
ZCS-1845:editheader throws ParseException

ykomiyam:
SIEVE: editheader throws ParseException exception if original mail's mime header violates RFC.

[Bug]
When the triggering (original) message contains any MIME header which is not compliant with RFC, the editheader action (addheader/deleteheader/replaceheader) throws the exception of "javax.mail.internet.ParseException", and entire filter execution is canceled.

For example, the Content-Type is sometimes minorly malformed, like a value contains multiple semi-colon characters at the end of the value "Content-Type: text/plain;;". When a message with such malformed Content-Type is edited, even the editheader does not edit the Content-Type header, the ParseException is thrown during the execution of the action, and the message is delivered to the Inbox without editing.

Because the addheader, deleteheader, and replaceheader actions are particularly used for the spam handling, if the filter execution is canceled due to the malformed message, end user may receive more spam messages without the SPAM label (and it may cause another security problems).

[Root Cause]
After modifying the message header, and before writing down it to the blob file, ZCS checks the consistency of the MIME headers by MimeMessage.saveChanges().  This method checks MIME headers, and if one of them is malformed, the method throws the ParseException exception.

[Fix]
* Added to call MimeMessage.saveMessages() before editing the headers to verify the format of the original message.  If the MimeMessage.saveMessages() is failed due to the ParseException, just logged it (INFO level) and continue to execute the action.
* Added the following MIME headers immutable by default so that the editheader can't change the headers which may cause any inconsistency around the MIME structure.
 - Content-Type
 - Content-Transfer-Encoding
 - Content-Disposition
 - Mime-Version
* For "addheader" command, added to check if the adding header is an immutable one.
* (Optimization) Added a flag to update the blob file only if the message was actually altered.

[Unit test]
Sieve related unit test: PASS
 - AddHeaderTest
 - DeleteHeaderTest
 - ReplaceHeaderTest

### DIFF
--- a/store/src/java/com/zimbra/cs/filter/ZimbraMailAdapter.java
+++ b/store/src/java/com/zimbra/cs/filter/ZimbraMailAdapter.java
@@ -130,6 +130,12 @@ public class ZimbraMailAdapter implements MailAdapter, EnvelopeAccessors {
     private LmtpEnvelope envelope = null;
 
     /**
+     * Parse result of the triggering message (editheader actions)
+     */
+    public enum PARSESTATUS { UNKNOWN, TOLERABLE, MIMEMALFORMED };
+    private PARSESTATUS eheParseStatus = PARSESTATUS.UNKNOWN;
+
+    /**
      * List of capability strings declared by "require" control.
      */
     private List<String> capabilities = new ArrayList<String>();
@@ -958,5 +964,13 @@ public class ZimbraMailAdapter implements MailAdapter, EnvelopeAccessors {
 
     public boolean isImplicitKeep () {
         return fieldImplicitKeep;
+    }
+
+    public PARSESTATUS getEheParseStatus() {
+        return eheParseStatus;
+    }
+
+    public void setEheParseStatus(PARSESTATUS eheParseStatus) {
+        this.eheParseStatus = eheParseStatus;
     }
 }
--- a/store/src/java/com/zimbra/cs/filter/ZimbraMailAdapter.java
+++ b/store/src/java/com/zimbra/cs/filter/ZimbraMailAdapter.java
@@ -966,11 +966,11 @@ public class ZimbraMailAdapter implements MailAdapter, EnvelopeAccessors {
         return fieldImplicitKeep;
     }
 
-    public PARSESTATUS getEheParseStatus() {
+    public PARSESTATUS getEditHeaderParseStatus() {
         return eheParseStatus;
     }
 
-    public void setEheParseStatus(PARSESTATUS eheParseStatus) {
+    public void setEditHeaderParseStatus(PARSESTATUS eheParseStatus) {
         this.eheParseStatus = eheParseStatus;
     }
 }
--- a/store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
+++ b/store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
@@ -568,9 +568,9 @@ public class EditHeaderExtension {
     }
 
     static public boolean isTolerableFormedMessage(ZimbraMailAdapter zma, String actionName, MimeMessage mm) {
-        if (zma.getEheParseStatus() == PARSESTATUS.UNKNOWN) {
+        if (zma.getEditHeaderParseStatus() == PARSESTATUS.UNKNOWN) {
             return saveChanges(zma, actionName, mm);
-        } else if (zma.getEheParseStatus() == PARSESTATUS.MIMEMALFORMED) {
+        } else if (zma.getEditHeaderParseStatus() == PARSESTATUS.MIMEMALFORMED) {
             ZimbraLog.filter.info(actionName + ": Triggering message is malformed MIME");
             return false;
         }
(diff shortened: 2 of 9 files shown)


store/src/java/com/zimbra/cs/filter/ZimbraMailAdapter.java
```diff
@@ -959,4 +965,12 @@ public boolean isCapable(String capability) {
     public boolean isImplicitKeep () {
         return fieldImplicitKeep;
     }
+
+    public PARSESTATUS getEheParseStatus() {
+        return eheParseStatus;
+    }
+
+    public void setEheParseStatus(PARSESTATUS eheParseStatus) {
```
rupalid: can we call this setEdtHdrParseStatus for clarity

store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
```diff
@@ -564,4 +566,35 @@ private boolean matchValue(String regex, String value) {
...
+        String ct = "";
+        try {
+            ct = mm.getContentType();
+        } catch (MessagingException e) {
+            ct = "(unknown)";
+        }
+
+        try {
+            mm.saveChanges();
+        } catch (ParseException pe) {
```
rupalid: Should this be 
 the other way around for ParseException log and return false
For exception just return false

store/src/java/com/zimbra/cs/filter/ZimbraMailAdapter.java
ykomiyam: Okay.  I'll change it to setEditHeaderParseStatus and getEditHeaderParseStatus respectively.

store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
ykomiyam: No, in this context, ParseException is the one which I want to treat as a non-error status.  In other words, the EditHeaderException.saveChanges() should pretend to be fine (and return true), if and only if the ParseException occurs; because the it was caused by the formatting error found in the unrelated headers.

store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
```diff
@@ -564,4 +566,35 @@ private boolean matchValue(String regex, String value) {
...
+            return saveChanges(zma, actionName, mm);
+        } else if (zma.getEheParseStatus() == PARSESTATUS.MIMEMALFORMED) {
+            ZimbraLog.filter.info(actionName + ": Triggering message is malformed MIME");
+            return false;
+        }
+        return true;
+    }
+
+    static public boolean saveChanges(ZimbraMailAdapter zma, String actionName, MimeMessage mm) {
+        String ct = "";
```
shrikant-prasad: whats the use of ct? the parse error could be caused because of other malformed header value too other than contentType.

store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
```diff
@@ -564,4 +566,35 @@ private boolean matchValue(String regex, String value) {
         Matcher matcher = pattern.matcher(value);
         return matcher.matches();
     }
+
+    static public boolean isTolerableFormedMessage(ZimbraMailAdapter zma, String actionName, MimeMessage mm) {
```
shrikant-prasad: saveChanges is called twice now, once during malformed mime check and again after editing the mime. instead can't we just call the save changes after editing the mime as it was done earlier and just catch the exceptions like it is being done in saveChanges() method below.

store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
ykomiyam: Okay.  I'll remove it.

store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
ykomiyam: The reason I called it twice is MimeMessage.saveChanges() may be failed due to some error other than ParseException, and in such case, we should not proceed evaluations.

store/src/java/com/zimbra/cs/filter/ZimbraMailAdapter.java
ykomiyam: Done.

store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
ykomiyam: Done

store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
```diff
@@ -564,4 +566,28 @@ private boolean matchValue(String regex, String value) {
...
+        if (zma.getEditHeaderParseStatus() == PARSESTATUS.UNKNOWN) {
+            return saveChanges(zma, actionName, mm);
+        } else if (zma.getEditHeaderParseStatus() == PARSESTATUS.MIMEMALFORMED) {
+            ZimbraLog.filter.info(actionName + ": Triggering message is malformed MIME");
```
shrikant-prasad: should change this log level to debug level as for this mime, it has already been logged that it's malformed in saveChanges method. this would prevent multiple logging at info level for the same mime message.

store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
shrikant-prasad: if saveChanges fails due to some error other than ParseException, the mime gets updated or not? if not, then calling saveChanges once after making changes should be sufficient.

store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
ykomiyam: fair enough.  I'll change.

store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
ykomiyam: Okay.  I'll change to call saveChanges() only once after editing.

store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
ykomiyam: Done.

store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
ykomiyam: Done

shrikant-prasad: this method is not used anymore, so can be removed. In the edit header classes like AddHeader, you should check if parseStatus is equal to PARSESTATUS.MIMEMALFORMED, then return without performing any operation as we already know the mime is malformed.

PR #250, opened 2017-07-12
ZCS-1790: Ability to flush cache / reload LC on imapd servers

ilyusha:
This PR adds two new zimbra-only IMAP operations and updates the server to make use of those operations where necessary.

#### FLUSHCACHE 
The `FLUSHCACHE` imap command is used to flush the specified cache. It is equivalent to _FlushCacheRequest_, except that it does not support an _allServers_ parameter. The syntax is as follows:
  - `FLUSHCACHE <cacheTypes>` flushes the entire cache of the specified comma-separated cache types, such as `FLUSHCACHE account` or `FLUSHCACHE server,config`.
  - `FLUSHCACHE <cacheTypes> (<id|name> <entry id or name> ... )` flushes the cache entries specified in the parameters list. For example, a single account can be flushed with `FLUSHCACHE account (id <accountId>)`

#### RELOADLC
The `RELOADLC` imap command forces the imapd server to reload its LocalConfig instance. It is equivalent to _ReloadLocalConfigRequest_. This command has no other parameters.

Both `FLUSHCACHE` and `RELOADLC` _must_ be run from a session authenticated with the X-ZIMBRA auth mechanism. In other words, only a client that can generate a valid auth token can use these commands. An error will be thrown if these commands are used without proper authentication.

Since these commands are issued by a ZCS server to an imapd server, these commands are authenticated with an admin auth token for the internal _zimbra_ account. In order for this to work, several small changes had to be made to make the zimbra user compatible with the imap server:
 - The _ZimbraAuthenticator_ class now supports admin users
 - _ZimbraAuthenticator_ doesn't call _authorize()_ when dealing with the _zimbra_ user. The _Authenticator::authorize_ method doesn't currently fetch admin accounts, but if it did, it would succeed in this case because both the _authAccount_ and _username_ params are the _zimbra_ user.
 - ImapHandler::startSession doesn't begin tracking IMAP if the user is _zimbra_. This is because the _zimbra_ user doesn't have a server associated with it, and to avoid unnecessary work.
  
### Server changes

1. The _FlushCache_ SOAP handler has several new methods:
    - _flushCacheOnImapDaemon_ issues a FLUSHCACHE imap command to the specified server for the given cache types and entries, using the provided auth token.
    - _flushCacheOnAllImapDaemons_ calls the above method for all servers that have the "imapd" service enabled. Note that this may include servers that are not part of the _zimbraReverseProxyUpstreamImapServers_ pool, but given that this attribute is both server- and config-level, this seemed more straightforward than trying to aggregate the exact list of current imap servers.
    - If a _FlushCacheRequest_ has _allServers=true_, the handler will call _flushCacheOnAllImapDaemons_ in addition to forwarding the SOAP request to other mailbox servers.

2. The _ReloadLocalConfig_ SOAP handler sends out a `RELOADLC` command to all imapd servers. Both _ReloadLocalConfig_ and _FlushCache_ handlers make use of the new static utility method _ImapConnection::getAdminConnection_.

3. In order to clear imapd caches when changing _zimbraEphemeralBackendURL_, the auth token used to authenticate the IMAP session needs to be registered with the *previous* ephemeral backend. This is because the token is generated on a server which is aware of the new backend, but validated on a server which is not. To this end:
    - A new LDAP attribute zimbraPreviousEphemeralBackendURL keeps track of the previous ephemeral store. This attribute is immutable, and is changed by the callback attached to _zimbraEphemeralBackendURL_.
    - _EphemeralStore_ has a new enum _BackendType_ {primary, previous} and a new method _getNewFactory_ that can return a new _Factory_ instance for a given BackendType, without setting it as the singleton. This value of _BackendType_ is also set on the factory itself; the new protected method _getURL()_ uses this to return either the current or previous ephemeral backend URL. Subclasses of _EphemeralStore_ need to use _getURL()_ to access the backend configuration instead of assuming _zimbraEphemeralBackendURL_.
    - _ZimbraAuthToken_ has a new method _registerWithEphemeralStore_ that bypasses the normal _ZAttrAccount::addAuthTokens_ code path and registers the auth token with the provided backend.
    - _AttributeMigration::clearConfigCacheOnAllServers_ flushes the cache on all imapd servers asynchronously. Prior to doing so, the auth token is registered with the previous ephemeral backend so that imapd can validate it. Note that this may not be necessary - this method is also called before/after attribute migration, at which point imapd will already be aware of the new backend.
    - _EphemeralStore_ loads the ephemeral backend extensions as needed, instead of relying on the extension being initialized previously. imapd used to only initialize the extension associated with the current backend URL, which caused problems when the backend is changed.

### Testing

 - New tests _testClearDaemonCache_, _testReloadLocalConfig_, and _testClearDaemonCacheWrongAuthenticator_ in _TestImapViaImapDaemon_
 - New test _testPreviousEphemeralBackendURL_ in _TestChangeEphemeralStore_.
 - _TestUtil_ has a new method _flushImapDaemonCache_. This method is added to the _setUp_ and _tearDown_ methods of _TestImapViaImapDaemon_ and _TestImapDaemonNotifications_, as well as to _TestImapClient::cleanup_. 
 - new empty method _SharedImapTests::flushCacheIfNecessary_ that is overridden by _TestImapViaImapDaemon_. This is used by _testMailfoldersOnlyList_ to force the imapd to pick up the change to _zimbraImapDisplayMailFoldersOnly_.

The changes to _TestUtil_ allow ImapDaemon tests to run consecutively without the LOGIN errors seen before, since the account cache is now flushed before each test run. It also fixes the test failures caused by imapd not picking up attribute changes. On my environment, all tests in _TestImapClient_, _TestImapVia*_, and _TestImap*Notifications_ passed.

I also tested changing the ephemeral backend and running an attribute migration while logged into imapd:
 - If an IMAP command is issued after a change to the ephemeral backend but prior to migration, the session is disconnected since the auth token is no longer valid.
 - If an attribute migration is run first, the session continues normally.

#### Updates from Feedback:
 - The new imap commands have been renamed to `X-ZIMBRA-FLUSHCACHE` and `X-ZIMBRA-RELOADLC`
 - removed the special-case logic in _ZimbraAuthenticator_ in lieu of updating _Authenticator::authenticate_ method to handle admin accounts
 - _FlushCache_ SOAP handler uses the the same account name and auth token to flush imapd caches as was used to authenticate the request
 - ImapHandler checks whether the authenticated user has admin rights to flush the cache
 - The new IMAP commands require _ZimbraAuthenticator_ to be authenticated with an admin auth token; added new test _testZimbraCommandsNonAdminUser_ to verify this
 - Changed `X-ZIMBRA-FLUSHCACHE` command signature to specify multiple cache types in parentheses instead of CSV

### DIFF
--- a/store/src/java/com/zimbra/cs/account/callback/EphemeralBackendCheck.java
+++ b/store/src/java/com/zimbra/cs/account/callback/EphemeralBackendCheck.java
@@ -2,12 +2,15 @@ package com.zimbra.cs.account.callback;
 
 import java.util.Map;
 
+import com.google.common.base.Strings;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.util.Log.Level;
 import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.cs.account.AttributeCallback;
+import com.zimbra.cs.account.Config;
 import com.zimbra.cs.account.Entry;
 import com.zimbra.cs.account.Provisioning;
+import com.zimbra.cs.account.callback.CallbackContext.DataKey;
 import com.zimbra.cs.ephemeral.EphemeralStore;
 import com.zimbra.cs.ephemeral.EphemeralStore.Factory;
 import com.zimbra.cs.ephemeral.migrate.AttributeMigration;
@@ -25,6 +28,7 @@ public class EphemeralBackendCheck extends AttributeCallback {
             if (tokens != null && tokens.length > 0) {
                 String backend = tokens[0];
                 if (backend.equalsIgnoreCase("ldap")) {
+                    savePreviousUrl(context, entry);
                     EphemeralStore.clearFactory();
                     return;
                 }
@@ -58,11 +62,25 @@ public class EphemeralBackendCheck extends AttributeCallback {
                 throw ServiceException.FAILURE(String.format(
                         "unable to modify %s; no ephemeral backend specified", attrName), null);
             }
+            savePreviousUrl(context, entry);
+        }
+    }
+
+    private void savePreviousUrl(CallbackContext context, Entry entry) {
+        String prevUrl = ((Config) entry).getEphemeralBackendURL();
+        if (!Strings.isNullOrEmpty(prevUrl)) {
+            context.setData(DataKey.PREV_EPHEMERAL_BACKEND_URL, prevUrl);
         }
     }
 
     @Override
     public void postModify(CallbackContext context, String attrName, Entry entry) {
+        String prevUrl = context.getData(DataKey.PREV_EPHEMERAL_BACKEND_URL);
+        try {
+            ((Config) entry).setPreviousEphemeralBackendURL(prevUrl);
+        } catch (ServiceException e) {
+            ZimbraLog.ephemeral.error("unable to set zimbraPreviousEphemeralBackendURL to %s", prevUrl, e);
+        }
         AttributeMigration.clearConfigCacheOnAllServers(false);
     }
 
--- a/store/src/java/com/zimbra/cs/ephemeral/EphemeralStore.java
+++ b/store/src/java/com/zimbra/cs/ephemeral/EphemeralStore.java
@@ -15,6 +15,7 @@ import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.util.CliUtil;
 import com.zimbra.common.util.Log.Level;
 import com.zimbra.common.util.ZimbraLog;
+import com.zimbra.cs.account.Config;
 import com.zimbra.cs.account.Provisioning;
 import com.zimbra.cs.ephemeral.migrate.AttributeMigration;
 import com.zimbra.cs.ephemeral.migrate.AttributeMigration.MigrationFlag;
@@ -204,7 +205,7 @@ public abstract class EphemeralStore {
             factory.startup();
             ZimbraLog.ephemeral.debug("using ephemeral store factory %s", factoryClass.getDeclaringClass().getSimpleName());
         } catch (InstantiationException | IllegalAccessException e) {
-            handleFailure(onFailure, String.format("unable to initialize EphemeralsStore factory %s", factoryClass.getDeclaringClass().getSimpleName()), e);
+            handleFailure(onFailure, String.format("unable to initialize EphemeralStore factory %s", factoryClass.getDeclaringClass().getSimpleName()), e);
         }
     }
 
@@ -228,6 +229,52 @@ public abstract class EphemeralStore {
         return encoder.decode(key, value);
     }
 
+    private static String getFactoryClassName(String backendName) throws ServiceException {
+        String factoryClassName = factories.get(backendName);
+        if (factoryClassName == null) {
+            //perhaps the extension hasn't been loaded - try to find it and check again
+            ExtensionUtil.initEphemeralBackendExtension(backendName);
+            factoryClassName = factories.get(backendName);
+        }
+        return factoryClassName;
+    }
+
+    public static Factory getNewFactory(BackendType backendType) throws ServiceException {
+        Config config = Provisioning.getInstance().getConfig();
+        String factoryClassName = null;
+        String url = backendType == BackendType.primary ?
+                config.getEphemeralBackendURL() : config.getPreviousEphemeralBackendURL();
+        if (url != null) {
+            String[] tokens = url.split(":");
+            if (tokens != null && tokens.length > 0) {
+                factoryClassName = getFactoryClassName(tokens[0]);
+            }
+        } else {
+            throw ServiceException.FAILURE("no ephemeral backend URL specified", null);
+        }
+        if (factoryClassName == null) {
+            throw ServiceException.FAILURE("no EphemeralStore.Factory class specified", null);
+        }
+        Class<? extends Factory> factoryClass = null;
+        try {
+            factoryClass = Class.forName(factoryClassName).asSubclass(Factory.class);
+        } catch (ClassNotFoundException cnfe) {
+            try {
+                factoryClass = ExtensionUtil.findClass(factoryClassName).asSubclass(Factory.class);
+            } catch (ClassNotFoundException cnfe2) {
+                throw ServiceException.FAILURE(String.format("Unable to find EphemeralStore factory %s", factoryClassName), cnfe2);
+            }
+        }
+        try {
+            Factory newFactory = factoryClass.newInstance();
+            newFactory.setBackendType(backendType);
+            newFactory.startup();
+            return newFactory;
+        } catch (InstantiationException | IllegalAccessException e) {
+            throw ServiceException.FAILURE(String.format("unable to initialize EphemeralsStore factory %s", factoryClass.getDeclaringClass().getSimpleName()), e);
+        }
+    }
+
     public static Factory getFactory(FailureMode onFailure) throws ServiceException {
         if (factory == null) {
             String factoryClass = null;
@@ -235,7 +282,7 @@ public abstract class EphemeralStore {
             if (url != null) {
                 String[] tokens = url.split(":");
                 if (tokens != null && tokens.length > 0) {
-                    factoryClass = factories.get(tokens[0]);
+                    factoryClass = getFactoryClassName(tokens[0]);
                 }
             } else {
                 factoryClass = factories.get("ldap");
@@ -354,11 +401,28 @@ public abstract class EphemeralStore {
         }
     }
 
-    public static interface Factory {
+    public static abstract class Factory {
 
-        EphemeralStore getStore();
-        void startup();
-        void shutdown();
+        private BackendType backendType = BackendType.primary;
+
+        public void setBackendType(BackendType type) {
+            backendType = type;
+        }
+
+        public BackendType getBackendType() {
+            return backendType;
+        }
+
+        protected String getURL() throws ServiceException {
+            Config config = Provisioning.getInstance().getConfig();
+            String backendURL = backendType == BackendType.primary ?
+                    config.getEphemeralBackendURL() : config.getPreviousEphemeralBackendURL();
+            return backendURL;
+        }
+
+        public abstract EphemeralStore getStore();
+        public abstract void startup();
+        public abstract void shutdown();
 
         /**
          * Validate the given URL for the EphemeralStore implementation.
@@ -367,10 +431,14 @@ public abstract class EphemeralStore {
          * @param url
          * @throws ServiceException
          */
-        void test(String url) throws ServiceException;
+        public abstract void test(String url) throws ServiceException;
     }
 
     public static enum FailureMode {
         halt, safe;
+    }
+
+    public static enum BackendType {
+        primary, previous;
     }
 }
--- a/store/src/java/com/zimbra/cs/imap/ImapHandler.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapHandler.java
@@ -45,6 +45,8 @@ import java.util.regex.Pattern;
 import javax.mail.MessagingException;
 import javax.mail.internet.MimeMessage;
 
+import org.dom4j.DocumentException;
+
 import com.google.common.base.Charsets;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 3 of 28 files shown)


store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -515,6 +520,11 @@ boolean executeRequest(ImapRequest req) throws IOException, ImapException {
                     }
                     checkEOF(tag, req);
                     return isProxied ? imapProxy.proxy(req) : doFETCH(tag, sequence, attributes, parts, byUID, modseq);
+                } else if (command.equals("FLUSHCACHE")) {
```
grishick: seems unlikely that IMAP will be extended with 'FLUSHCACHE' or 'RELOADLC' commands, but to avoid any potential namespace conflict, could you rename these commands to something like X-ZIMBRA-FLUSHCACHE and X-ZIMBRA-RELOADLC?

store/src/java/com/zimbra/cs/security/sasl/ZimbraAuthenticator.java
```diff
@@ -92,24 +93,29 @@ public ZimbraAuthenticator(AuthenticatorUser user) {
...
         // if necessary, check that the authenticated user can authorize as the target user
-        Account targetAcct = authorize(authAccount, username, AuthToken.isAnyAdmin(at));
-        if (targetAcct != null)
+        if (at.isAdmin() && username.equals(authenticateId) && username.equals(LC.zimbra_ldap_user.value())) {
             prov.accountAuthed(authAccount);
-        return targetAcct;
+            return authAccount;
+        } else {
+            Account targetAcct = authorize(authAccount, username, AuthToken.isAnyAdmin(at));
+            if (targetAcct != null)
```
grishick: please add {
}

store/src/java/com/zimbra/cs/security/sasl/ZimbraAuthenticator.java
```diff
@@ -92,24 +93,29 @@ public ZimbraAuthenticator(AuthenticatorUser user) {
...
         // make sure the protocol is enabled for the user
         if (!isProtocolEnabled(authAccount, protocol)) {
             ZimbraLog.account.info("Authentication failed - %s not enabled for %s", protocol, authAccount.getName());
             return null;
         }
 
         // if necessary, check that the authenticated user can authorize as the target user
-        Account targetAcct = authorize(authAccount, username, AuthToken.isAnyAdmin(at));
-        if (targetAcct != null)
+        if (at.isAdmin() && username.equals(authenticateId) && username.equals(LC.zimbra_ldap_user.value())) {
```
grishick: I think that instead of restricting this to LC.zimbra_ldap_user, this code should perform the same permission check as the one performed by FlushCache SOAP handler. Also, FlushCache SOAp handler should pass on the token of whoever is calling FlushCache. This way we have a continuous audit trail of who flushed the cache.

store/src/java/com/zimbra/cs/security/sasl/ZimbraAuthenticator.java
ilyusha: It was like that before I touched it, but probably a good idea nonetheless 👍

store/src/java/com/zimbra/cs/imap/ImapHandler.java
ilyusha: 👍

store/src/java/com/zimbra/cs/security/sasl/ZimbraAuthenticator.java
ilyusha: Yeah, in fact this made me realize that this bypassing of the authorize() method means that other admin accounts won't be able to use this command.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
plobbes: RFC specifically includes this in the ABNF (ref: [https://tools.ietf.org/search/rfc3501#page-91](https://tools.ietf.org/search/rfc3501#page-91)):

`x-command       = "X" atom <experimental command arguments>`

FWIW, Gmail seems to use X-GM-... these days (ref: [https://developers.google.com/gmail/imap/imap-extensions](https://developers.google.com/gmail/imap/imap-extensions)).

store/src/java/com/zimbra/cs/imap/ImapRequest.java
```diff
@@ -610,6 +614,51 @@ String readSequence() throws ImapParseException {
         return validateSequence(readContent(SEQUENCE_CHARS), true);
     }
 
+    CacheEntryType[] readCacheEntryTypes() throws IOException, ImapParseException {
+        String[] cacheStrTypes = readAstring().split(",");
```
plobbes: nit: Would it make more sense (be more IMAP like) to not use a CSV and instead use a parenthesized list?
- () :: default to all
- (cache1 cache2 ...)

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -877,6 +886,14 @@ boolean executeRequest(ImapRequest req) throws IOException, ImapException {
                     req.skipSpace();  Set<String> patterns = Collections.singleton(req.readFolderPattern());
                     checkEOF(tag, req);
                     return doLIST(tag, base, patterns, (byte) 0, RETURN_XLIST, (byte) 0);
+                } else if (command.equals("X-ZIMBRA-FLUSHCACHE")) {
+                    req.skipSpace();
+                    CacheEntryType[] cacheTypes = req.readCacheEntryTypes();
+                    CacheEntry[] entries = req.readCacheEntries();
+                    return doFLUSHCACHE(tag, cacheTypes, entries);
```
plobbes: Seems like a checkEOF(tag, req) is appropriate here just in case extra text is left in req after the list entries (if any).

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -1330,6 +1347,56 @@ boolean doLOGOUT(String tag) throws IOException {
...
+    }
+
+    boolean doRELOADLC(String tag) throws IOException {
+        if (!checkState(tag, State.AUTHENTICATED)) {
+            return true;
+        } else if (!checkZimbraAuth()) {
+            sendNO(tag, "must be authenticated with X-ZIMBRA auth mechanism");
+            return true;
+        }
+        try {
```
plobbes: Should we use the same flushCache checkRight for this capability/command?

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -1330,6 +1347,56 @@ boolean doLOGOUT(String tag) throws IOException {
...
+            sendNO(tag, "must be authenticated with X-ZIMBRA auth mechanism");
+            return true;
+        }
+        for (CacheEntryType type: types) {
+            try {
+                AuthToken authToken = ((ZimbraAuthenticator) authenticator).getAuthToken();
+                AdminAccessControl aac = AdminAccessControl.getAdminAccessControl(authToken);
+                Server localServer = Provisioning.getInstance().getLocalServer();
+                try {
+                    aac.checkRight(localServer, Admin.R_flushCache);
```
plobbes: nit: for efficiency reasons it might be nice to perform the static checkRight outside of the for loop.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -1462,11 +1529,12 @@ private ImapCredentials startSession(Account account, EnabledHack hack, String t
...
         setCredentials(new ImapCredentials(account, hack));
-        credentials.getImapMailboxStore().beginTrackingImap();
-        ZimbraLog.addAccountNameToContext(credentials.getUsername());
-        ZimbraLog.imap.info("user %s authenticated, mechanism=%s%s",
-                credentials.getUsername(), mechanism == null ? "LOGIN" : mechanism, startedTLS ? " [TLS]" : "");
-
+        if (!account.getName().equalsIgnoreCase(LC.zimbra_ldap_user.value())) {
+            credentials.getImapMailboxStore().beginTrackingImap();
+            ZimbraLog.addAccountNameToContext(credentials.getUsername());
+            ZimbraLog.imap.info("user %s authenticated, mechanism=%s%s",
```
plobbes: I think we'd still want to log this even for case where account is zimbra_ldap_user.  Can we move this outside of the if and/or remove special case handling completely?

store/src/java/com/zimbra/cs/imap/ImapRequest.java
ilyusha: This is a good idea. I originally made it a CSV to match the format used by FlushCacheRequest, but this makes more sense.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
ilyusha: Oh, good catch - silly mistake on my part.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
ilyusha: There's actually no corresponding rights check for ReloadLocalConfigRequest

store/src/java/com/zimbra/cs/imap/ImapHandler.java
ilyusha: Yeah, good catch here as well.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
plobbes: Agreed. Since there is no equivalent right, the requirement then should be that the operation has to be performed by LC.zimbra_ldap_user.value (user zimbra) only, perhaps?  I'd be OK calling the flushCache right as "close enough" for this, although perhaps just making it an official/proper option to flushCache is even better?

As-is, other non-admin users could potentially authenticate via the X-Zimbra mechanism, and then be able to perform this action, no?  (I didn't go look at the code to verify that)

store/src/java/com/zimbra/cs/account/callback/EphemeralBackendCheck.java
```diff
@@ -58,11 +62,25 @@ public void preModify(CallbackContext context, String attrName,
...
         }
     }
 
     @Override
     public void postModify(CallbackContext context, String attrName, Entry entry) {
+        String prevUrl = context.getData(DataKey.PREV_EPHEMERAL_BACKEND_URL);
+        try {
+            ((Config) entry).setPreviousEphemeralBackendURL(prevUrl);
+        } catch (ServiceException e) {
+            ZimbraLog.ephemeral.error(String.format("unable to set zimbraPreviousEphemeralBackendURL to %s", prevUrl), e);
```
fatkudu: You don't actually need the String.format here - that is already done by the native method.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
ilyusha: Yes, as-is, non-admin users would be able to do this. Perhaps I should update doRELOADLC so that only admin users can run this command? That would put it more in line with how ReloadLocalConfigRequest works.

store/src/java/com/zimbra/cs/ephemeral/EphemeralStore.java
```diff
@@ -228,14 +229,60 @@ protected EphemeralKeyValuePair decode(String key, String value) throws ServiceE
...
+                throw ServiceException.FAILURE(String.format("Unable to find EphemeralStore factory %s", factoryClassName), cnfe2);
+            }
+        }
+        try {
+            Factory newFactory = factoryClass.newInstance();
+            newFactory.setBackendType(backendType);
+            newFactory.startup();
+            return newFactory;
+        } catch (InstantiationException | IllegalAccessException e) {
+            throw ServiceException.FAILURE(String.format("unable to initialize EphemeralsStore factory %s", factoryClass.getDeclaringClass().getSimpleName()), e);
```
fatkudu: s/EphemeralsStore/EphemeralStore/

store/src/java/com/zimbra/cs/imap/ImapHandler.java
plobbes: Yes.  Please update.

store/src/java/com/zimbra/cs/account/callback/EphemeralBackendCheck.java
ilyusha: Ah thanks - for some reason I often forget this

store/src/java/com/zimbra/cs/imap/ImapRequest.java
ilyusha: I implemented this change, with the exception that a blank list is not permitted; this is because FlushCacheRequest requires at least one cache type, and I figured this would be consistent.

PR #251, opened 2017-07-13
ZCS-1922 Fix travis script

rohan-ambasta:
Came across the failures on Travis CI while working on ZCS-1922
Fixed travis script so that the unit tests pass on Travis CI

We can remove the dependency for zm-timezones later. But at this point at least the script will run to give us correct output

Testing:
The script is passing on my fork
https://travis-ci.org/rohan-ambasta/zm-mailbox


Prashantsurana:
Had discussion Rohan , we need resolve path and interdependecy issues and nice to have to remove zm-timezone dependency to run zm-mailbox test. The testdata for one repository should not depend on some other repository.

PR #255, opened 2017-07-14
ZCS-1571 ZMailbox::getFolderById exception with invalid id

gordyt:
Part of the work done for ZCS-1571 changed the behavior of the
`getFolderById` method. Prior to the changes done for that ticket
the `getFolderById` function would return `null` if called
with an argument that was not a folder id; e.g., "INBOX".

More aggressive cache checking created a new behavior that would
instead result in a ServiceException when the method was
called in a similar fashion.

### DIFF
--- a/client/src/java/com/zimbra/client/ZMailbox.java
+++ b/client/src/java/com/zimbra/client/ZMailbox.java
@@ -2956,9 +2956,16 @@ public class ZMailbox implements ToZJSONObject, MailboxStore {
             /* sometimes when working on behalf of, we're getting just by item id but the
              * cache has fully qualified keys.
              */
-            String ident = new ItemIdentifier(id, this.getAccountId()).toString();
-            if (!id.equals(ident)) {
-                item = mItemCache.getById(ident);
+            try {
+                String ident = new ItemIdentifier(id, this.getAccountId()).toString();
+                if (!id.equals(ident)) {
+                    item = mItemCache.getById(ident);
+                }
+            } catch (ServiceException e) {
+                // This can be raised when constructing an ItemIdentifier with a folder name; e.g., "INBOX"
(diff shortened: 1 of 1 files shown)


client/src/java/com/zimbra/client/ZMailbox.java
```diff
@@ -2953,9 +2953,16 @@ public ZFolder getFolderById(String id) throws ServiceException {
...
-                item = mItemCache.getById(ident);
+            try {
+                String ident = new ItemIdentifier(id, this.getAccountId()).toString();
+                if (!id.equals(ident)) {
+                    item = mItemCache.getById(ident);
+                }
+            } catch (ServiceException e) {
+                // This can be raised when constructing an ItemIdentifier with a folder name; e.g., "INBOX"
+                if (!(e.getCause() instanceof NumberFormatException)) {
+                    throw e;
```
grishick: so if this happens, wouldn't item remain NULL and cause an NPE several lines down?

gordyt: That situation is covered by this bit that follows:

        if (!(item instanceof ZFolder)) {
            return null;
        }

`null` isn't a `ZFolder` so the function returns `null`.

grishick: so if a caller passes "INBOX" instead of an number, the caller gets NULL. Does this mean that the caller is misusing this method or that INBOX does not exist?

grishick: nm. 
zmmailbox -v -z -m user1@zimbra-dev.local gf "INBOX" 
works.

gordyt:
Gren I agree with everything you write.  It was the "risky change in behavior" part that guided the fix.  One way to look at it is that there does not exist a folder with the ID "INBOX" so a `null` response is appropriate.

PR #256, opened 2017-07-16
ZCS-1841:Add attribute to turn enable/disable edit header functionality

sneha-patil-synacor:
Added a ldap attribute to enabled or disable sieve edit header feature at account-cos-domain level.
If the attribute value is false, all the edit header commands in admin scripts will be ignored.
If the attribute value is true, all the edit header commands in admin scripts will be executed.
The attribute value is by default false.

Regardless if the attribute value, user script can not have edit header commands and if they appear in the script, the entire user script will be ignored.

Added junit test for add/delete/replace header for admin as well as user script.
Since the user script can not have the edit header commands now, the existing unit tests needed to be fixed. Modified them to be in admin script.

### DIFF
--- a/store/src/java/com/zimbra/cs/filter/JsieveConfigMapHandler.java
+++ b/store/src/java/com/zimbra/cs/filter/JsieveConfigMapHandler.java
@@ -39,6 +39,7 @@ public class JsieveConfigMapHandler {
     public static final String CAPABILITY_EREJECT    = "ereject";
     public static final String CAPABILITY_REJECT     = "reject";
     public static final String CAPABILITY_ENOTIFY    = "enotify";
+    public static final String CAPABILITY_EDITHEADER = "editheader";
 
     /*
      * jSieve's command map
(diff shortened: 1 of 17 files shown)


store/src/java/com/zimbra/cs/filter/JsieveConfigMapHandler.java
```diff
@@ -39,6 +39,7 @@
     public static final String CAPABILITY_EREJECT    = "ereject";
     public static final String CAPABILITY_REJECT     = "reject";
     public static final String CAPABILITY_ENOTIFY    = "enotify";
+    public static final String CAPABILITY_EDITHEADER = "editheader";
```
ykomiyam: Please use CAPABILITY_EDITHEADER in the declaration of the "mCommandMap" parameter (line 65-66)

sneha-patil-synacor: Its already there. Just replaced it with constant variable now.

store/src/java/com/zimbra/cs/filter/RuleManager.java
```diff
@@ -414,9 +415,37 @@ private static void setXMLRules(Account account,List<FilterRule> rules, String s
...
-
                 if (applyRules) {
-                    SIEVE_FACTORY.evaluate(mailAdapter, node);
+                    if (filter.equals(FILTER_RULES_CACHE_KEY)) {
+                        mailAdapter.setUserScriptExecuting(true);
+                    }
+                    try {
+                        SIEVE_FACTORY.evaluate(mailAdapter, node);
+                    } catch (SieveException e) {
+                        if (editHeaderUserScriptError.equals(e.getMessage())) {
```
shrikant-prasad: lines 425 to 448 can be moved to a common method as same code is repeated at line 497 to 526.

sneha-patil-synacor: Done

PR #257, opened 2017-07-17
ZCS-884:deleteheader/replaceheader operation should be written to mai…

sneha-patil-synacor:
Added consistent log statements for delete header and replace header operations.
No Junits as its only log change.


store/src/java/com/zimbra/cs/filter/jsieve/ReplaceHeader.java
```diff
@@ -125,6 +125,9 @@ protected Object executeBasic(MailAdapter mail, Arguments arguments,
                                     newHeaderValue = header.getValue();
                                 }
                                 ZimbraLog.filter.debug("replaceheader: header after processing\n%s: %s", newHeaderName, newHeaderValue);
```
shrikant-prasad: can remove this logging at debug level as same info is now logged at info level.

sneha-patil-synacor: Done

PR #259, opened 2017-07-17
ZCS-1922 Remove file dependency issues for test-all target

rohan-ambasta:
- In the test target we are now passing the path for sever directory as a JVM argument. So, running ant test from zm-mailbox/store or running ant test-all from zm-mailbox does not result in file not found exception
- Instead of using the zm-timezone repo, I have copied the timezone.ics file to the java-test directory. Thus the travis script will not have to clone zm-timezones repo
- Added method getZimbraServerDir in MailboxTestUtil so that the correct server directory path is formed
- Used getZimbraServerDir in all other tests


store/src/java-test/com/zimbra/cs/db/HSQLDB.java
```diff
@@ -1,7 +1,7 @@
 /*
  * ***** BEGIN LICENSE BLOCK *****
  * Zimbra Collaboration Suite Server
- * Copyright (C) 2011, 2012, 2013, 2014, 2015, 2016 Synacor, Inc.
+ * Copyright (C) 2011, 2012, 2013, 2014, 2015, 2016, 2017 Synacor, Inc.
```
gordyt: Would recommend putting changes that are not related to the commit in separate commit(s).

fatkudu:
Does the zm-mailbox package not depend on the zip file produced from zm-timezones?  If it does, then I would prefer us to work with that (presumably we can do that in the same way that we can work with zm-mailbox depending on 3rd party libraries via ivy.xml)  That way we won't have to choose between constantly updating the copy of timezones.ics and working with an out of date version of it?

store/src/java-test/com/zimbra/cs/db/HSQLDB.java
rohan-ambasta: I think we should update this information whenever we change the file, unless we have a separate task which does this after every release.

rohan-ambasta:
@fatkudu Right now, I am using a copy of timezone.ics file. But we can have a separate task to use a basic version of timezone.ics as per the needs of the junits.

store/src/java-test/com/zimbra/cs/mailbox/MailboxTestUtil.java
```diff
@@ -85,33 +85,35 @@ public static void initProvisioning() throws Exception {
...
-        System.setProperty("zimbra.config", zimbraServerDir + pathPrefix + "src/java-test/localconfig-test.xml");
-        LC.reload();
-        if (Strings.isNullOrEmpty(zimbraServerDir)) {
-            zimbraServerDir = System.getProperty("user.dir");
+    zimbraServerDir = getZimbraServerDir(zimbraServerDir);
+    System.setProperty("log4j.configuration", "log4j-test.properties");
+    System.setProperty("zimbra.config", zimbraServerDir + "src/java-test/localconfig-test.xml");
+    LC.reload();
+
+    String timezonefilePath = zimbraServerDir + "src/java-test/timezones.ics";
```
rupalid: Can we call this timezones-test.ics, that we are very clear this is just for unit test purpose.

PR #265, opened 2017-07-18
ZCS-2150: restrict flushing cache types on imapd

ilyusha:
This bug was caused by `X-ZIMBRA-FLUSHCACHE` routing all cache types to _Provisioning.flushCache_, which is incorrect since some cache types, such as _license_ require special handling.

This is fixed with the following changes:

1. Restrict which cache types are forwarded by the _FlushCache_ SOAP handler to imapd servers. Not all cache types are applicable to imapd servers, so it makes sense to only forward relevant ones to avoid unnecessary requests. The relevant caches are listed in the new public static set _ImapHander.IMAP_CACHE_TYPES_, which current consists of _account, cos, domain, server, config_, and _all_. If there are other cache types that should be included here, please comment and I will update this set.

2. Update _ImapHandler_ to ignore other cache types even if they are encountered. While this may be somewhat unnecessary since irrelevant cache types should be filtered out by step 1, it makes testing easier and prevents edge cases where caches like _uistrings_ try to be flushed on imapd, which may lead to complications due to further request forwarding.

3. Instead of simplly calling _Provisioning::flushCache_ from _ImapHandler_, use the same code path as the FlushCache SOAP handler, which allows for additional code to be executed for each cache type. To this end, _FlushCache::doFlush_ has been made public so it can be called from _ImapHandler::doFLUSHCACHE_.

A new test _TestImapViaImapDaemon.testFlushNonImapCacheTypes()_ is added to test that attempting to flush cache types not listed in _IMAP_CACHE_TYPES_ does not result in errors.

#### Updates
 - _FlushCache_ and _ReloadLocalConfig_ SOAP handlers forward their respective IMAP commands only to those servers listed in the _zimbraReverseProxyUpstreamImapServers_ attribute for the current server. For _FlushCache_, this happens regardless of the value of the _allServers_ attribute. The idea here is that we want to flush the cache on all imapd servers that serve the mailbox server receiving the _FlushCacheRequest_.
 - _FlushCacheRequest_ has a new request attribute _imapServers_ that defaults to true. If set to _false_, `X-ZIMBRA-FLUSHCACHE` will _not_ be forwarded to _zimbraReverseProxyUpstreamImapServers_. This is used by _AttributeMigration_, as it issues its own custom `X-ZIMBRA-FLUSHCACHE` commands, and needs a way to disable _FlushCache_ from doing so.
 - Autoloading of ephemeral extensions can be disabled on _EphemeralStore_. `zmprov` does this when using _SoapProvisioning_, to prevent the process from unnecessarily loading extensions and outputting logging noise to the console. This is a follow-up to work that was done for ZCS-1790, which introduced the autoloading mechanism.

### DIFF
--- a/common/src/java/com/zimbra/common/soap/AdminConstants.java
+++ b/common/src/java/com/zimbra/common/soap/AdminConstants.java
@@ -1391,6 +1391,7 @@ public final class AdminConstants {
 
     // flush cache
     public static final String A_ALLSERVERS = "allServers";
+    public static final String A_IMAPSERVERS = "imapServers";
 
     public static final String A_SYNCHRONOUS = "synchronous";
 
--- a/soap/src/java/com/zimbra/soap/admin/type/CacheSelector.java
+++ b/soap/src/java/com/zimbra/soap/admin/type/CacheSelector.java
@@ -55,6 +55,19 @@ public final class CacheSelector {
     @XmlAttribute(name=AdminConstants.A_ALLSERVERS /* allServers */, required=false)
     private ZmBoolean allServers;
 
+
+    /**
+     * @zm-api-field-tag imap-servers
+     * @zm-api-field-description
+     * <table>
+     * <tr> <td> <b>0 (false)</b> </td> <td> don't issue X-ZIMBRA-FLUSHCACHE IMAP command to upstream IMAP servers </td> </tr>
+     * <tr> <td> <b>1 (true) [default]</b> </td>
+     *      <td> flush cache on servers listed in zimbraReverseProxyUpstreamImapServers for the current server via X-ZIMBRA-FLUSHCACHE</td> </tr>
+     * </table>
+     */
+    @XmlAttribute(name=AdminConstants.A_IMAPSERVERS /* imapServers */, required=false)
+    private ZmBoolean imapServers;
+
     /**
      * @zm-api-field-description Cache entry selectors
      */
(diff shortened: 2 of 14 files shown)


grishick:
From what I can see if no IMAPd servers are running and all mailboxd servers are running IMAP as embedded, this code will still try to send flush cache requests to these mailboxd servers via IMAP. The code should send IMAP flushcache requests only to IMAPd servers and skip embedded IMAP servers.

ilyusha:
@grishick I was thinking about that when working on ZCS-1790. I agree that the current approach may send unnecessary IMAP flush requests to embedded servers, but I think that this should be addressed in a separate bug.

Is there a reliable way to determine which servers are running imap daemons and which are running embedded? I assume it would involve zimbraReverseProxyUpstreamImapServers, but it's both a server and config-level attribute, so I'm not sure how to calculate a comprehensive list. For now, I erred on the side of sending extra requests - even if some of the recipients are embedded servers.

grishick:
@ilyusha I think there are 2 components to this: 1 - check zimraRemoteImapServerEnabled, 2 - send requests to remote IMAP port rather than via proxy.

gordyt:
To determine all of the servers where the imapd daemon is enabled could we construct this list from the servers whose attribute `zimbraServiceEnabled` contains `imapd`?

grishick:
Ideally, this code will use the same logic for determining whether to send an IMAP request somewhere as the logic used by zmcontrol, zmimapdctl and ImapDaemon to determine whether to start up. 
When a given mailboxd is deciding whether to send FlushCache to an IMAPd node, it should check the value of zimbraReverseProxyUpstreamImapServers for it's own host. Since zimbraReverseProxyUpstreamImapServers is both GC and Server, any Server record that has zimbraReverseProxyUpstreamImapServers==NULL will default to GC value and there is no need to check GC separately.

PR #266, opened 2017-07-19
ZCS-1937 shared sub folder support.

fatkudu:
Fix various problems with selecting sub folders.
Also problems fetching from shared folders (both top level and sub folders)

### DIFF
--- a/client/ivy.xml
+++ b/client/ivy.xml
@@ -14,6 +14,7 @@
   <dependency org="org.json" name="json" rev="20090211" />
   <dependency org="javax.servlet" name="javax.servlet-api" rev="3.1.0" />
   <dependency org="commons-httpclient" name="commons-httpclient" rev="3.1" />
+  <dependency org="commons-lang" name="commons-lang" rev="2.6" />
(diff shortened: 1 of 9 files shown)


fatkudu:
Note.  Just gone to the bug to point at this pull request and spotted that Rohit was reporting a problem with STATUS as well.  Just done a quick hack to the tests to try that out.  Remote is still failing for mountpoints :(

fatkudu:
@gordyt which do you mean, SELECT or STATUS?  I'm expecting STATUS to work for both a top level and a subfolder but STATUS is an outstanding problem.

gordyt:
That was a typo.  Have updated the comment.  `STATUS` is failing for the top-level `Manage` folder (that was shared from user 1 to user 2) but working with the `Manage2` (sub-folder of `Manage`) by user 2 as per step 6 in the description.

fatkudu:
@gordyt.  My turn for typos!  This Jira is targeting `SELECT` not `STATUS`.  I've created some sketchy SOAP test cases for STATUS.  I'll create a Jira for that tomorrow and probably start working on it unless there is something obviously more urgent.

PR #269, opened 2017-07-20
ZCS-296:Wildcard :matches does not work if target string contains CRLF

shrikant-prasad:
Wildcard :matches does not work if target string contains CRLF and is base64 encoded

### DIFF
--- a/store/src/java/com/zimbra/cs/filter/ZimbraAsciiCasemap.java
+++ b/store/src/java/com/zimbra/cs/filter/ZimbraAsciiCasemap.java
@@ -24,14 +24,14 @@ import static com.zimbra.cs.filter.jsieve.MatchRelationalOperators.LT_OP;
 import static com.zimbra.cs.filter.jsieve.MatchRelationalOperators.NE_OP;
 
 import java.util.List;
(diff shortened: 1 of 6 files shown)


store/src/java/com/zimbra/cs/filter/ZimbraAsciiCasemap.java
```diff
@@ -69,8 +69,13 @@ public boolean counts(String operator, List<String> lhs, String rhs)
      */
     public boolean matches(String string, String glob)
             throws SievePatternException {
-        return ZimbraComparatorUtils
-                .matches(string.toUpperCase(), glob.toUpperCase());
+        try {
+            String regex = FilterUtil.sieveToJavaRegex(glob.toUpperCase());
```
chinmay15: use 
ComparatorUtils.sieveToJavaRegex(regex);

shrikant-prasad: ZimbraComparatorUtils.matches internally uses FilterUtil.sieveToJavaRegex. thats why used this.

PR #278, opened 2017-07-25
ZCS-1842:Edit header should modify the headers only for a user with EditHeader Sieve script

shrikant-prasad:
EditHeader script rule changes should be applied only to recipient who has the rule configured, not to other recipients of the same mail.

Before mime is updated from Edit header classes, parsed message will be cloned if mail is for multiple recipients and all rule operation will be done on cloned message. After updating mime, updated blob is stored in deliveryContext. If the mime is shared, updated blob will be stored in  a map with key as mailbox id, otherwise incoming blob of the context will be updated.
when message is added to mail store at that time, we will check if there is updated blob for this user, if so, link will be created with updated blob, else link will be created with incoming blob.

### DIFF
--- a/store/src/java/com/zimbra/cs/filter/ZimbraMailAdapter.java
+++ b/store/src/java/com/zimbra/cs/filter/ZimbraMailAdapter.java
@@ -99,6 +99,7 @@ public class ZimbraMailAdapter implements MailAdapter, EnvelopeAccessors {
     private boolean allowFilterToMountpoint = true;
     private Map<String, String> variables = new HashMap<String, String>();
     private List<String> matchedValues = new ArrayList<String>();
+    private boolean parsedMessageCloned = false;
 
     public enum VARIABLEFEATURETYPE { UNKNOWN, OFF, AVAILABLE};
     private VARIABLEFEATURETYPE variablesExtAvailable = VARIABLEFEATURETYPE.UNKNOWN;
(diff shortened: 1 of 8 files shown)


store/src/java/com/zimbra/cs/filter/ZimbraMailAdapter.java
```diff
@@ -1025,4 +1039,20 @@ public boolean isUserScriptExecuting() {
...
+        if (ctxt != null && ctxt.getShared() && !parsedMessageCloned && handler instanceof IncomingMessageHandler) {
+            try {
+                MimeMessage cloneMM = new MimeMessage(getMimeMessage());
+                ParsedMessage pm = handler.getParsedMessage();
+                ParsedMessage clonePM = new ParsedMessage(cloneMM, pm.isAttachmentIndexingEnabled());
+                ((IncomingMessageHandler) handler).setParsedMessage(clonePM);
+                parsedMessageCloned = true;
+                ZimbraLog.filter.debug("cloned ParsedMessage");
+            } catch (MessagingException | ServiceException e) {
+                ZimbraLog.filter.debug("failed to clone parsed message", e);
```
sneha-patil-synacor: Should we log it at warn level too.

shrikant-prasad: done.

rupalid:
Can you add a brief description of how the bug was fixed.

PR #286, opened 2017-07-28
Bugfix/zcs 2272

grishick:
This pull requests includes multiple changes that together enable updates to be passed to multiple IMAP clients connected to the same IMAPd. 
Major fixes:
 1. Update cached folder data when NOOP or STATUS commands are called. This fixes some of the failures of IMAP compliance tests that expect to get updated message counts in response to NOOP and STATUS commands on additional IMAP clients.
 2. When a notification comes to IMAPd via SOAP headers, pass it through ImapServerListener in order to notify any other ImapListener that is subscribed to changes for the affected folders
 3. In order to avoid applying the same pending modification twice, use lastChangeID to track the last applied change. Per-mailbox last change ID is now passed in (Admin)WaitSetResponse and SOAP headers. Previously it was passed only in SOAP headers and was ignored by IMAPd.
4. Fix unread count returned by IMAPd when a folder has been updated after an IMAP session started. Folder SOAP will contain only 'unread' and will not contain 'imap unread' when both values are the same. Therefore, fall back to 'unread' when updating 'imap unread'

I ran IMAP compliance suite with these changes and got 18 failed test groups with IMAPd (vs 21 on 'develop' branch). I did not get any new failures on local IMAP. Local IMAP still gets 13 failed test groups. 

I am not adding any more unit tests for 2 reasons:
1) @tmclane has already added a SOAP test for checking that 2d IMAP connection gets notifications
2) IMAP compliance tests also provide coverage for behaviors fixed in this pull request

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/ImapHandler.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapHandler.java
@@ -1198,6 +1198,19 @@ public abstract class ImapHandler {
     }
 
     boolean doNOOP(String tag) throws IOException {
+        ImapListener i4selected = getCurrentImapListener();
+        if(i4selected != null) {
+            MailboxStore mbox = i4selected.getMailbox();
+            if(mbox != null) {
+                try {
+                    mbox.noOp();
+                } catch (ServiceException e) {
+                    ZimbraLog.imap.error("NOOP failed with error %s", e.getMessage(), e);
+                    sendNO(tag, "NOOP failed");
+                    return canContinue(e);
+                }
+            }
+        }
         sendNotifications(true, false);
         sendOK(tag, "NOOP completed");
         return true;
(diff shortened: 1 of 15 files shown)


store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -2475,7 +2488,12 @@ boolean doSTATUS(String tag, ImapPath path, byte status) throws ImapException, I
                 sendNO(tag, "STATUS failed");
                 return true;
             }
-
+            if(credentials != null) {
+                MailboxStore mbox = credentials.getMailbox();
+                if(mbox != null) {
+                    mbox.noOp();
```
fatkudu: Given that doNOOP uses try/catch and canContinue round this, is it worth having that here?

grishick: this block is already wrapped in try/catch with canContinue - see lines 2498-2506 ;)

PR #290, opened 2017-08-01
ZCS-2369 encoding test fails in ZMimeBodyPartTest

rohan-ambasta:
Set UTF-8 encoding in build-common.xml

The encoding JUnit test passes after this fix


rohan-ambasta:
@rupalid this test is not part of exclusion list. It only fails on our build machine

PR #291, opened 2017-08-01
ZCS-2195: run attribute migration prior to changing ephemeral backend

ilyusha:
This PR introduces a significant change to how data is migrated from LDAP to an ephemeral backend. 

### Motivation
Prior to this change, the migration had to be run *after* the backend URL was changed, and would pull data from LDAP to the new backend. This meant that the migration had to be run immediately after the backend change, or users would encounter authentication errors as auth tokens would not be found in the new ephemeral store.
A better approach is to run the migration _first_, and then change the actual backend URL when all data has been migrated. Combined with a mechanism to forward newly-incoming ephemeral attributes to the destination store until _zimbraEphemeralBackendURL_ is switched to point to the new backend, this provides a much safer approach to attribute migration.

### MigrationInfo
A new class _MigrationInfo_ is introduced to encapsulate information about the last known attribute migration. It is an abstract class; the implementation _ZimbraMigrationInfo_ uses the new immutable LDAP attribute _zimbraAttributeMigrationInfo_ to persist the information. _InMemoryMigrationInfo_ is also provided to use with dry runs and for testing.
The data stored by migration info is:
 - The URL of the ephemeral store the data is being migrated to
 - The current status of the migration. This is one of NONE (if no record exists), IN_PROGRESS, COMPLETED, or FAILED.
 - The migration start time

### ForwardingEphemeralStore
Prior to this change, the server switched to using the _FallbackEphemeralStore_ backend for the duration of the migration. This backend looked for attributes in both the current store and in LDAP, since the given attribute may or may not have been migrated yet. With this change, this has been replaced with the _ForwardingEphemeralStore_. This is the reverse of the fallback backend: all ephemeral attribute modifications are forwarded to the backend that the migration is happening for, so that new data is not left behind. Errors encountered when forwarding data to the new backend are swallowed as to not interrupt normal operation in the event that the future backend is taken down.
One crucial difference is that the forwarding backend needs to remain in place even after the migration finishes. This is because there is likely to be a gap between the end of the migration and when the value of _zimbraEphemeralBackendURL_ is updated to the new value. The forwarding store remains in place to keep the future store up to date with LDAP. This is done by instantiating _FallbackEphemeralStore_ when the status in _AttributeInfo_ is IN_PROGRESS or COMPLETED.

### Rules governing changing _zimbraEphemeralBackendURL_

In general, the most common use case _should_ be:
1. `zmmigrateattrs [destination]` is ran successfully
2. `zmprov mcf zimbraEphemeralBackendURL [destination]` is ran to flip the switch.

However, logic is introduced to handle nonstandard scenarios. One case disallows the change entirely, while several others allow it but log a warning message:
 - If a migration is currently in progress, an exception will be thrown notifying the user that the backend cannot be modified
 - If no migration info is available, a warning is logged that data may not have been migrated to the new backend
 - If the migration info has a status of FAILED for the provided new backend, a warning is logged that  data may not have been migrated
 - If the new value of _zimbraEphemeralBackendURL_ does not match the URL of the known migration, a warning is logged that data may not have been migrated to this URL.

If the new URL value matches the URL of the known migration, the migration info is reset and the config cache is cleared, meaning that mailbox servers will cease to use the forwarding store and switch over to using the backend corresponding to the new URL. However, in the event that the new URL value does _not_ match the one used during migration, the migration info is not reset. This leaves the forwarding backend in use, keeping it up to date with new data in the event that the backend is changed to the "proper" URL in the future.

At any point, the migration info can be reset by running `zmmigrateattrs --clear`, which will turn off the forwarding store on all servers and allow the backend URL to be changed or for a migration to be restarted.

### CLI Changes
 - The URL of the destination backend now has to be provided as the first argument to `zmmigrateattrs`. The following arguments, if available, are interpreted as the attribute names to migrate.
 - The `--status` or `-s` option displays the status of the current migration, if available
 - The `--clear` or `-c` option resets the migration info
 - The `--keep-old` flag has been removed, since migration no longer deletes original LDAP values
 - the `--set-flag` and `--unset-flag` options have been removed, as the system no longer uses a boolean migration flag
 - A new migration cannot be started if the current migration status is IN_PROGRESS.

### Other changes
- _ExtensionUtil::initEphemeralBackendExtension_ does not instantiate an _EphemeralStore_ after loading the factory. This is because the URL of the store can come from one of three places (current backend, previous backend, migration backend) and this method is not aware of which one should be used.
 - _EphemereralStore::getFactory_ has new signature allowing the ability to disable autoloading of extensions. This is used by _ExtensionUtil::initEphemeralBackendExtension_ to avoid circular dependencies, as EphemeralStore may call _initEphemeralBackendExtension_ with autoloading enabled.
 - _EphemeralStore::getFactoryClassName_ silences extension/ephemeral log messages less than _error_ in severity to avoid unnecessary clutter from CLI utilities. This also removes the need for _ProvUtil_ to explicitly disable autoloading of extensions to avoid noise.
 - The logic governing changing _zimbraEphemeralBackendURL_ in the _EphemeralBackendCheck_ callback has been factored out into an inner class _EphemeralBackendMigrationRules_, which in turn relies on an implementation of a _MigrationStateHelper_ helper interface. This allows for testing of this logic without actually modifying the underlying LDAP attribute

### Testing
Several new unit and SOAP tests have been added:
 -  _MigrateAttributesTest.testMigrationInfo()_ tests that _AttributeMigration_ properly sets the status on the _MigrationInfo_ object.
 - _ChangeEphemeralBackendTest_ is a new test class containing tests checking the various scenarios that can be encountered when changing the backend URL
 - _TestChangeEphemeralStore.testChangeURLDuringMigration()_ is a SOAP test that verifies that the backend URL cannot be changed when MigrationInfo has an IN_PROGRESS status

Additionally, unit tests under _MigrateAttributesTest_ have been updated to reflect the fact that ephemeral attributes are no longer deleted from the source backend.

#### A note on deleting old data
It is mentioned above that LDAP data is no longer deleted as part of the migration process. This is because LDAP will still be used as the backend until _zimbraEphemeralBackendURL_ is changed, requiring the "old" data to still be there. Since the presence of old ephemeral data in LDAP has no adverse effects aside from taking up space, I think we should open a new ticket to provide the ability to delete this data at a later time. This came up in a conversation with John Hurley; he suggested that it would be useful for admins to have the options of deleting old LDAP data at their discretion. This is a sufficiently separate issue that I did not include it as part of this bug.


store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9539,4 +9539,9 @@ TODO: delete them permanently from here
 <attr id="3018" name="zimbraPreviousEphemeralBackendURL" type="string" cardinality="single" optionalIn="globalConfig" immutable="1" since="8.8.1">
   <desc>URL of the previous ephemeral storage backend</desc>
 </attr>
+
+<attr id="3019" name="zimbraAttributeMigrationInfo" type="string" cardinality="single" optionalIn="globalConfig" immutable="1" since="8.8.1">
```
fatkudu: type `enum` seems more appropriate.
As an aside, some SOAP tests are currently failing because of attempts to modify immutable attributes.  Is there any danger that adding a new one is making that problem worse?  (i.e. are there circumstances where we try to do "get all attrs on an object/update them/put them back" where having some immutable attributes among them causes a problem.

store/src/java/com/zimbra/cs/account/callback/EphemeralBackendCheck.java
```diff
@@ -76,12 +106,121 @@ private void savePreviousUrl(CallbackContext context, Entry entry) {
...
+             */
+            helper.setURL(newURL);
+            MigrationInfo info = AttributeMigration.getMigrationInfo();
+            Status curStatus = info.getStatus();
+            String migrationURL = info.getURL();
+            if (curStatus == Status.IN_PROGRESS) {
+                helper.deny();
+            } else if (curStatus == Status.NONE) {
+                helper.warn(Reason.NO_MIGRATION);
+            } else if (curStatus == Status.FAILED && newURL.equals(migrationURL)) {
```
fatkudu: I do wonder if this shouldn't result in deny - perhaps with a "-f" override.

store/src/java/com/zimbra/cs/ephemeral/EphemeralStore.java
```diff
@@ -186,27 +189,36 @@ public static final void setFactory(Class<? extends Factory> factoryClass) {
...
-            boolean useFallback = false;
+            boolean useForwarding = false;
             Factory fac = factoryClass.newInstance();
-            if (!(fac instanceof LdapEphemeralStore.Factory)) {
-                EphemeralStore store = fac.getStore();
-                MigrationFlag flag = AttributeMigration.getMigrationFlag(store);
+            try {
+                MigrationInfo info = AttributeMigration.getMigrationInfo();
+                Status curStatus = info.getStatus();
+                if (!Strings.isNullOrEmpty(info.getURL()) && (curStatus == Status.COMPLETED || curStatus == Status.IN_PROGRESS)) {
```
fatkudu: What about Status.FAILED?  Does that necessarily mean things are completely broken or might forwarding perhaps still be useful?

store/src/java/com/zimbra/cs/ephemeral/EphemeralStore.java
```diff
@@ -302,7 +324,17 @@ public static Factory getFactory() throws ServiceException {
     }
 
     public static Factory getFactory(String backendName) {
-        String factoryClassName = factories.get(backendName);
+        return getFactory(backendName, false);
```
fatkudu: The call to this from `ZimbraMigrationCallback` doesn't do a null check on the result..

store/src/java/com/zimbra/cs/ephemeral/migrate/AttributeMigrationUtil.java
```diff
@@ -86,83 +96,98 @@ public static void main(String[] args) throws Exception {
...
+                backendName = tokens[0];
+                if (backendName.equalsIgnoreCase("ldap")) {
+                    ZimbraLog.ephemeral.info("migrating to LDAP is not supported");
+                    return;
                 }
-            } else {
-                Zimbra.halt("no ephemeral backend specified");
+            }
+            if (Provisioning.getInstance().getConfig().getEphemeralBackendURL().equalsIgnoreCase(destURL)) {
+                ZimbraLog.ephemeral.info("destination URL cannot be the same as the currenet ephemeral backend URL");
```
fatkudu: typo s/currenet/current/

store/src/java/com/zimbra/cs/ephemeral/migrate/ZimbraMigrationInfo.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    private void decode(String encoded) {
+        if (Strings.isNullOrEmpty(encoded)) {
+            //no migration info present
+            return;
+        }
+        String[] tokens = encoded.split("\\|");
+        URL = tokens[0];
+        status = Status.valueOf(tokens[1]);
```
fatkudu: Worth having error checking on number of tokens?

store/conf/attrs/zimbra-attrs.xml
ilyusha: This can't really be an enum because it's an encoded string containing the status, destination URL, and timestamp.
Regarding immutability: If there are SOAP tests that fail due to the presence of immutable attributes, that seems like an underlying issue with the tests themselves. AFAIK, attributes should be made immutable when they store some internal server state, which is the case here.
That said, I'm happy to remove the immutable constraint for the time being if it's likely to cause problems.

store/src/java/com/zimbra/cs/account/callback/EphemeralBackendCheck.java
ilyusha: I was actually debating whether these cases should result in warnings or errors. An -f override is a good alternative, especially because running zmprov without the "-l" option doesn't actually output the warnings to the console but just hides them in the logs. I'll look into what it takes to propagate a new flag down to the attribute callbacks.

store/src/java/com/zimbra/cs/ephemeral/EphemeralStore.java
ilyusha: If the migration failed, the assumption is that it will have to be rerun. In that case, I don't think forwarding is not necessary.

store/src/java/com/zimbra/cs/ephemeral/EphemeralStore.java
ilyusha: Ah good catch - this causes an NPE if the destination set from zmmigrateattrs is invalid.

store/conf/attrs/zimbra-attrs.xml
fatkudu: Sorry, didn't read the full description properly.  Clearly can't be an enum.  On the immutable issue.  I think there really is a bug but it can probably be addresses separately.  Just creating a COS after a SOAP test run barfs.  I'm guessing because we clone the default COS and if it has 2 factor auth attributes in it, we just blindly copy them over even though they probably don't make sense.

store/src/java/com/zimbra/cs/ephemeral/migrate/ZimbraMigrationInfo.java
ilyusha: Yep, and validation as well. I had this as a TODO and forgot about it.

store/conf/attrs/zimbra-attrs.xml
ilyusha: If that's the case, it seems like that may be fixed by setting _checkImmutable_ to _False_ in the right place :)

grishick:
looks like this PR needs to be rebased

store/src/java/com/zimbra/cs/ephemeral/EphemeralStore.java
```diff
@@ -234,31 +246,41 @@ public static void setAutoloadExtensions(boolean bool) {
         autoloadExtensions = bool;
     }
 
-    private static String getFactoryClassName(String backendName) throws ServiceException {
```
grishick: I don't understand the logic for having a static autoloadExtensions field and then an overriding parameter disableAutoload. Also, why do we need to try and find again - what may change after the application tries loading extensions and before this code executes?

ilyusha: The reason it tries to load the extension if it's not registered is because non-mailboxd processes may not have the proper extension initialized when the backend URL is changed. For example, if imapd is started when LDAP is the backend, and then the backend is changed to SSDB, the SSDB extension may not be loaded since only the current backend was being initialized. The alternative is to load all ephemeral backend extensions preemptively, which is possible, but requires updating all CLI utilities that may touch authentication, and isn't really future-proof. Autoloading just seemed like an easier solution.

Regarding the boolean field: think I can actually remove the static autoloadExtensions field. I recently introduced it in order to disable unnecessary extension loading by ProvUtil and other scripts, but as part of this PR I also silenced those log messages when calling initEphemeralBackendExtension from getFactoryClassName.

grishick: ok, yes, I think autoloadExtensions should be removed. I do not see it being set anywhere except in declaration and you are removing the only call to setAutoloadExtensions in this PR.

PR #294, opened 2017-08-02
ZCS-2395 FolderActionResponse for trash operation has empty id

fatkudu:


tmclane:
Looks good to me.  Removing the unused variable like codacy says would be a good idea.

PR #295, opened 2017-08-02
Bugfix/zcs 2242 Optimize the BeginTrackingIMAPRequest SOAP call

tmclane:
Currently the `BeginTrackingIMAPRequest` is issued for every:
- session creation
- folder selection

This PR:
- removes the folder selection tracking request entirely
- piggy-backs the ability to notify the caller using the GetInfoResponse of the tracking state of the user with the 'isTrackingIMAP' tag.

ZMailbox has been updated to only issue the `BeginTrackingIMAPRequest` If and only if the `isTrackingIMAP` value is '0'.

### DIFF
--- a/soap/src/java/com/zimbra/soap/account/message/GetInfoResponse.java
+++ b/soap/src/java/com/zimbra/soap/account/message/GetInfoResponse.java
@@ -69,11 +69,11 @@ import com.zimbra.soap.type.ZmBoolean;
 @XmlAccessorType(XmlAccessType.NONE)
 @XmlRootElement(name=AccountConstants.E_GET_INFO_RESPONSE)
 @XmlType(propOrder = {"version", "accountId", "accountName", "crumb", "lifetime", "adminDelegated", "restUrl",
-        "quotaUsed", "previousSessionTime", "lastWriteAccessTime", "recentMessageCount", "cos", "prefs", "attrs",
+        "quotaUsed", "isTrackingIMAP", "previousSessionTime", "lastWriteAccessTime", "recentMessageCount", "cos", "prefs", "attrs",
         "zimlets", "props", "identities", "signatures", "dataSources", "childAccounts", "discoveredRights",
         "soapURL", "publicURL", "changePasswordURL", "license", "adminURL", "boshURL"})
 @JsonPropertyOrder({"version", "id", "name", "crumb", "lifetime", "adminDelegated", "docSizeLimit", "attSizeLimit",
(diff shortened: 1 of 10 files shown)


tmclane:
Looking into the CI failure now.

soap/src/java/com/zimbra/soap/account/message/GetInfoResponse.java
```diff
@@ -504,6 +512,14 @@ public String getSoapURL() {
         return Prop.toMultimap(props, userPropKey);
     }
 
+    public Boolean getIsTrackingIMAP() {
```
fatkudu: I would change this to a boolean as toBool specifying a default will never return null.

store/src/java/com/zimbra/cs/service/account/GetInfo.java
```diff
@@ -173,6 +173,7 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
             try {
                 Mailbox mbox = getRequestedMailbox(zsc);
                 response.addAttribute(AccountConstants.E_QUOTA_USED, mbox.getSize(), Element.Disposition.CONTENT);
+                response.addAttribute("isTrackingIMAP", mbox.isTrackingImap() ? "1" : "0", Element.Disposition.CONTENT);
```
fatkudu: No, this will break the JSON variant which expects booleans to have values true/false.
Just use `response.addAttribute("isTrackingIMAP", mbox.isTrackingImap(), Element.Disposition.CONTENT)`
and that will do the right thing - i.e. use 0/1 for XML and true/false for JSON.

soap/src/java/com/zimbra/soap/account/message/GetInfoResponse.java
tmclane: Roger that.

tmclane: Don't we want to return 'false' in the case of the value not being returned from the server?
The only side-effect would be a perhaps unneeded 'BeginTrackingImapRequest' being submitted.

fatkudu: I don't see a way around that.  Not set is the equivalent of false.

tmclane:
@fatkudu Updated to address your last concern + fix an issue with the JSON JAXB test.

tmclane:
Rebased.

PR #300, opened 2017-08-04
ZCS-2462 - Build script improvements

tmclane:
The intention of this PR is to remove reliance on `ant-contrib` being externally installed on the system. 

The following was done:
- rewrite the ivy2 download section to not rely upon `antcontrib:if`
- add `antcontrib` to the `ivy.xml` files for each of the sub-projects
- add new target `taskdefs` which makes the `antcontrib` namespace available (although not without the namespacing)
- update all targets that use `if` to depend on `taskdefs` and to prefix use of `if` with `antcontrib:`
- move propertyregex statements inside the target that requires them so as to utilize the `taskdefs` dependency along with adding the `antcontrib:` prefix.

**Verification:**
`zm-build/build.pl` verification was done along with an install of the resulting installer.

Also ran all of the various targets that are used in normal development to ensure all ran without errors.
- clean
- compile
- publish-local
- deploy-foss
- test
- test-all

### DIFF
--- a/build-common.xml
+++ b/build-common.xml
@@ -1,4 +1,4 @@
-<project name="build-common" xmlns:ivy="antlib:org.apache.ivy.ant">
+<project name="build-common" xmlns:ivy="antlib:org.apache.ivy.ant" xmlns:antcontrib="antlib:net.sf.antcontrib">
     <!-- Set properties that can be used in all projects. -->
     <dirname property="zimbra.root.dir" file="${ant.file.build-common}/../"/>
     <dirname property="zm-mailbox.basedir" file="${ant.file.build-common}"/>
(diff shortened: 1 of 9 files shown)


tmclane:
Working through fixing the 'deploy-foss' command.

tmclane:
It appears that Travis CI doesn't like the new scripts.  More investigation is needed it appears. Unsure what is different in that environment compared to my test machines.

build-common.xml
```diff
@@ -37,15 +37,15 @@
...
     </target>
     <!-- common resolve target that should work to resolve dependencies based on ivy.xml for most projects -->
     <target name="resolve" depends="init-ivy" description="resolve dependencies">
+      <ivy:settings id="dev.settings" file="../build-ivysettings.xml"/>
       <ivy:resolve settingsRef="dev.settings" />
-      <ivy:cachepath pathid="class.path" />
     </target>
 
-    <target name="taskdefs"  description="Declare 3rd party ANT tasks" depends="resolve">
+    <target name="taskdefs"  description="Declare 3rd party ANT tasks">
```
fatkudu: `taskdefs` doesn't intuitively shout "tasks provided by antcontrib" to me.  How about renaming `def-antcontrib-tasks`?  Happy to be over-ridden on this if others thing that is too wordy.

store/build.xml
```diff
@@ -35,15 +35,16 @@
...
       <available file="${zimbra.home.dir}/bin/zmimapdctl" type="file"/>
-      <then>
-        <exec executable="zmimapdctl">
-          <arg value="${action}"/>
-        </exec>
-      </then>
-    </if>
+    </condition>
+  </target>
+  <target name="imapd-control" unless="imapd.installed" depends="test.if.imapd.available">
```
fatkudu: I'm guessing should be run `if="imapd.available"` instead of `unless="imapd.installed"`?

fatkudu:
I imagine we will need a new task at some point to do similar changes for all the other repos.

build-common.xml
tmclane: Doesn't bother me either way although the idea was 'task defines' (of which others could be included here if necessary in the future)

store/build.xml
tmclane: Updated this.

grishick:
@tmclane did you try making RHEL and Ubuntu builds with build.pl and with these changes applied?

tmclane:
@grishick I built ubuntu.  I'll attempt a RHEL one now.

tmclane:
Just finished CentOS 7 verification of build, install, login using imap.   There appears to be an issue with postfix starting up which I am investigating.

tmclane:
Looks like the postfix startup issue may be caused by the system having one installed and activated by default and not an issue with the installer. Is that something we have documented as far as CentOS installation prerequisites?

Just confirmed both the installer built with these changes and one built using the original build files install and run on the same CentOS install.

tmclane:
Believe it's now safe to push in.  Does anyone disagree?

grishick:
what's that Travis-ci/push failure about?

gordyt:
Greg it looks like there was a test failure:

    BUILD FAILED
    /home/travis/build/Zimbra/zm-mailbox/build.xml:32: The following error occurred while executing this line:
    /home/travis/build/Zimbra/zm-mailbox/build-common.xml:304: Test com.zimbra.cs.imap.ImapHandlerTest failed
    Total time: 2 minutes 24 seconds
    The command "ant -Dzimbra.buildinfo.version=1.1.1_GA_1000 test-all" exited with 1.

tmclane:
I agree.  I think it may have something to do with my earlier changeset: ZCS-2408

PR #303, opened 2017-08-10
ZCS-2337:Modify REST API for Get Contacts to support LDIF

sneha-patil-synacor:
Added support for LDIF format for Contacts REST API
Added Junit test

### DIFF
--- a/store/src/java/com/zimbra/cs/service/formatter/LdifFormatter.java
+++ b/store/src/java/com/zimbra/cs/service/formatter/LdifFormatter.java
@@ -1,0 +1,162 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2017 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ * ***** END LICENSE BLOCK *****
+ */
+package com.zimbra.cs.service.formatter;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+import javax.mail.MessagingException;
+import javax.mail.Part;
+import javax.servlet.ServletException;
+
+import org.apache.commons.codec.binary.Base64;
+
+import com.zimbra.common.mailbox.ContactConstants;
+import com.zimbra.common.mime.MimeConstants;
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.util.HttpUtil;
+import com.zimbra.cs.mailbox.MailItem;
+import com.zimbra.cs.service.UserServletContext;
+import com.zimbra.cs.service.UserServletException;
+import com.zimbra.cs.service.formatter.FormatterFactory.FormatType;
+import com.zimbra.cs.service.util.UserServletUtil;
+
+public class LdifFormatter extends Formatter {
+
+    private static final String NEW_LINE = "\r\n";
+
+    @Override
+    public FormatType getType() {
+        return FormatType.LDIF;
+    }
+
+    @Override
+    public void formatCallback(UserServletContext context) throws UserServletException,
+        ServiceException, IOException, ServletException, MessagingException {
+        Iterator<? extends MailItem> contacts = null;
+        StringBuilder sb = new StringBuilder();
+        sb.append("version: 1"); // ldif version
+        sb.append(NEW_LINE);
+        try {
+            contacts = getMailItems(context, -1, -1, Integer.MAX_VALUE);
+            ArrayList<Map<String, String>> allContacts = new ArrayList<Map<String, String>>();
+            HashSet<String> fields = new HashSet<String>();
+            UserServletUtil.populateContactFields(contacts, context.targetMailbox,
+                context.opContext, allContacts, fields);
+            ArrayList<String> allFields = new ArrayList<String>();
+            allFields.addAll(fields);
+            Collections.sort(allFields);
+            for (Map<String, String> contactMap : allContacts) {
+                toLDIFContact(allFields, contactMap, sb);
+                sb.append(NEW_LINE);
+            }
+        } finally {
+            if (contacts instanceof QueryResultIterator)
+                ((QueryResultIterator) contacts).finished();
+        }
+        String filename = context.itemPath;
+        if (filename == null || filename.length() == 0)
+            filename = "contacts";
+        if (!filename.toLowerCase().endsWith(".ldif")) {
+            filename = filename + ".ldif";
+        }
+        String cd = HttpUtil.createContentDisposition(context.req, Part.ATTACHMENT, filename);
+        context.resp.addHeader("Content-Disposition", cd);
+        context.resp.setCharacterEncoding(context.getCharset().name());
+        context.resp.setContentType(MimeConstants.CT_TEXT_LDIF);
+        context.resp.getWriter().print(sb.toString());
+    }
+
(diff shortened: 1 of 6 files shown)


grishick:
I think this change needs an automated integration/SOAP test that creates a user, creates a bunch of contacts, exports them in LDIF and verifies the downloaded file.

store/src/java/com/zimbra/cs/service/formatter/LdifFormatter.java
```diff
@@ -0,0 +1,125 @@
...
+            filename = filename + ".ldif";
+        }
+        String cd = HttpUtil.createContentDisposition(context.req, Part.ATTACHMENT, filename);
+        context.resp.addHeader("Content-Disposition", cd);
+        context.resp.setCharacterEncoding(context.getCharset().name());
+        context.resp.setContentType(MimeConstants.CT_TEXT_LDIF);
+        context.resp.getWriter().print(sb.toString());
+    }
+
+    private void toLDIFContact(List<String> fields, Map<String, String> contact, StringBuilder sb) {
```
plobbes: ref: https://stackoverflow.com/questions/33433453/conversion-of-csv-file-to-ldif-file-using-java/45622057#45622057

The solution offered seems incomplete when it comes to dealing with any possible data that may show up. If you look at [RFC2849](https://tools.ietf.org/html/rfc2849) you'll find recommendations like (both on Page 5):

```
 4)  ... Any
     value that contains characters other than those defined as
     "SAFE-CHAR", or begins with a character other than those
     defined as "SAFE-INIT-CHAR", above, MUST be base-64 encoded.
     Other values MAY be base-64 encoded.
```
where SAFE-INIT-CHAR is defined as

```
SAFE-INIT-CHAR       = %x01-09 / %x0B-0C / %x0E-1F /
                       %x21-39 / %x3B / %x3D-7F
                       ; any value <= 127 except NUL, LF, CR,
                       ; SPACE, colon (":", ASCII 58 decimal)
                       ; and less-than ("<" , ASCII 60 decimal)
```
and

```
 8)  Values or distinguished names that end with SPACE SHOULD be
     base-64 encoded.
```

Unless I'm missing something, the code should be updated to handle encoding properly if you want a more complete and robust solution.

store/src/java/com/zimbra/cs/service/formatter/LdifFormatter.java
```diff
@@ -0,0 +1,125 @@
...
+     * time is reached. Given that we don't send a content-length down to the
+     * browser for archive responses, we have to close the socket to tell the
+     * browser its done. Since we have to do that.. leaving this endpoint
+     * without a timeout is safe. If the connection was being reused (ie
+     * keep-alive) this could have issues, but its not in this case.
+     * 
+     * @throws IOException
+     */
+    private void disableJettyTimeout(UserServletContext context) {
+        if (LC.zimbra_csv_formatter_disable_timeout.booleanValue()) {
```
plobbes: I would certainly hope that disabling the timeout is unnecessary.  With reasonable limits on the number of allowed entries (zimbraContactMaxNumEntries) and a relatively efficient conversion process that would seem like overkill for most use cases.

Potential RFE: It might be nice to have the ability to disable the timeout only for admin users (and only if they explicitly ask for that), but that perhaps complicates logic too much?

sneha-patil-synacor: Disabling jetty timeout was actually introduced as part of following bug:
https://bugzilla.zimbra.com/show_bug.cgi?id=77682
This is already being done for other formatters like csv.

store/src/java/com/zimbra/cs/service/formatter/LdifFormatter.java
```diff
@@ -0,0 +1,124 @@
...
+            filename = filename + ".ldif";
+        }
+        String cd = HttpUtil.createContentDisposition(context.req, Part.ATTACHMENT, filename);
+        context.resp.addHeader("Content-Disposition", cd);
+        context.resp.setCharacterEncoding(context.getCharset().name());
+        context.resp.setContentType(MimeConstants.CT_TEXT_LDIF);
+        context.resp.getWriter().print(sb.toString());
+    }
+
+    private void toLDIFContact(List<String> fields, Map<String, String> contact, StringBuilder sb) {
```
rupalid: We could make this method public and test whether LDIF conversion happens correctly for various test data.

store/src/java/com/zimbra/cs/service/formatter/LdifFormatter.java
sneha-patil-synacor: Thanks. Updated the toLDIFContact() method to handle the specified encoding.
Added the encoding test cases in Junit.

store/src/java/com/zimbra/cs/service/formatter/LdifFormatter.java
sneha-patil-synacor: Done.

store/src/java/com/zimbra/cs/service/formatter/LdifFormatter.java
plobbes: Right, I'm aware of that.  Note the disabling of the timeout was specifically for *import* of CSV and not for the export (per the bug).  It seems like we're only handling export here, and *in theory* export should be quite a bit faster than import (and thus hopefully no need to disable the timeout).  Being concerned about disabling the timeout is still valid and the RFE ideas also seem reasonable to me still.

store/src/java/com/zimbra/cs/service/formatter/LdifFormatter.java
```diff
@@ -0,0 +1,192 @@
...
+            filename = filename + ".ldif";
+        }
+        String cd = HttpUtil.createContentDisposition(context.req, Part.ATTACHMENT, filename);
+        context.resp.addHeader("Content-Disposition", cd);
+        context.resp.setCharacterEncoding(context.getCharset().name());
+        context.resp.setContentType(MimeConstants.CT_TEXT_LDIF);
+        context.resp.getWriter().print(sb.toString());
+    }
+
+    public void toLDIFContact(List<String> fields, Map<String, String> contact, StringBuilder sb) {
```
plobbes: Should we consider using com.unboundid.ldif.LDIFWriter (seems like we already ship with it) instead of reinventing the wheel here?  An added bonus would be that we also end up doing proper wrapping as recommended by the RFC for the LDIF output too (with this last proposed change the base64 writer isn't doing "chunking" as recommended).

sneha-patil-synacor: Although this API is returning contacts in LDIF format, we are not directly reading it from LDAP. Rather, the contacts are fetched from mysql and then converted to LDIF. 
Since, we are not talking to LDAP directly here, I think LDIFWriter would not be useful for this use case.

store/src/java/com/zimbra/cs/service/formatter/LdifFormatter.java
rupalid: I think lets remove it from LdifFormatter as it is used for export only

sneha-patil-synacor: Removed. The API will be tested for max number of contacts.

store/src/java/com/zimbra/cs/service/formatter/LdifFormatter.java
plobbes: Actually using an LDIF writer that is specifically designed to create LDIF (https://docs.ldap.com/ldap-sdk/docs/javadoc/com/unboundid/ldif/LDIFWriter.html) would be a good idea. As I mentioned, the implementation we have created doesn't seem to cover the requirement that base64 encoded lines be no more than 76 characters long (https://tools.ietf.org/html/rfc2045 as hinted at in https://tools.ietf.org/html/rfc2849).

sneha-patil-synacor: 1) Yes, we understand that LDIFWriter would be an appropriate choice. But, for that we would have to convert contact mail-item to an LDAP attribute record and then feed it to LDIF writer.
2) Just clarifying that RFC 2849 makes an exception to max 76 characters requirement.
Quoting the RFC:
"the encoding rules specified in [5] are used with the following exceptions:  a) The requirement that base64 output streams must be represented as lines of no more than 76 characters **is removed.**"

plobbes: Indeed wrapping is optional. In my experience, most tend to wrap (OpenLDAP in particular, which we ship with) but that doesn't make it required and OpenLDAP added support a little while back to no longer wrap if requested.  Leaving out wrapping support certainly minimizes the advantage of using an external module too.

Thanks for the follow up.

PR #305, opened 2017-08-10
Fixing literal apostrophe characters not escaped in format strings

zx-fulvio:
Same issue detailed at length [here](https://github.com/Zimbra/zm-web-client/pull/115)


zx-fulvio:
I don't think the build failure has anything to do with my commit.

silentsakky:
Dropping this PR as mentioned in https://github.com/Zimbra/zm-web-client/pull/115#discussion_r133938247

PR #307, opened 2017-08-15
ZCS-1417

mihaipopahtp:
- adding Vietnamese translation


store-conf/conf/msgs/ZsMsgRights_vi.properties
```diff
@@ -0,0 +1,463 @@
+#
+#
+# DO NOT MODIFY - generated by RightManager
```
silentsakky: why we have this message to not edit this file? are we generating it from somewhere?

silentsakky: every other language file is having this, so ignore it for now, will check later

PR #310, opened 2017-08-15
Bugfix/zcs 1825

grishick:
There are 2 reasons why when a folder is deleted outside of IMAP connection, the connection is not being dropped. Bot are addressed in this PR. 

1. Notifications that are sent via WaitSets and SOAP headers when a folder is deleted are sent within context of the parent of the deleted folder. At the same time, ImapListener is registered to listen for modifications only on the selected folder and does not get notified of changes in the parents of the selected folder.

2. Remote IMAP does not register to listen for modification notifications until IMAP client selects a folder. If IMAP client issues DELETE command in non-selected state, remote IMAP ignores modification notifications sent by mailbox server via SOAP headers.

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/ImapCredentials.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapCredentials.java
@@ -26,6 +26,7 @@ import java.util.Set;
 
 import com.google.common.base.Objects;
 import com.zimbra.client.ZMailbox;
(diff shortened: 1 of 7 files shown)


grishick:
@fatkudu as far as I can see, ImapPath calls ZMailbox.getMailbox for shared folders. This may be why remote IMAP is not being notified about changes in shared folders.

PR #313, opened 2017-08-16
WIP: ZCS:2486 - Separated jetty.base and jetty.home, removed jetty-<version>

shri314:
- Zimbra's Jetty Distribution (9.3.5.v20151012) is installed in /opt/zimbra/common/jetty_home/ via a separate package.
- Zimbra specific webapps and other custom jetty configuration files will be installed in /opt/zimbra/jetty_base/.
- Now /opt/zimbra/mailboxd and /opt/zimbra/jetty are symbolic links that point to /opt/zimbra/jetty_base/.

### DIFF
--- a/pkg-builder.pl
+++ b/pkg-builder.pl
@@ -19,9 +19,6 @@ sub parse_defines()
 {
    Die("wrong commandline options")
      if ( !GetOptions( "defines=s" => \%DEFINES ) );
-
-   Die("jetty.distro not specified (--define jetty.distro=<xxxx>)")
-     if ( !exists $DEFINES{'jetty.distro'} );
 }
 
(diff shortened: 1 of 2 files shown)


pkg-builder.pl
```diff
@@ -20,8 +20,8 @@ ()
    Die("wrong commandline options")
      if ( !GetOptions( "defines=s" => \%DEFINES ) );
 
-   Die("jetty.distro not specified (--define jetty.distro=<xxxx>)")
-     if ( !exists $DEFINES{'jetty.distro'} );
+#   Die("abc.def not specified (--define abc.def=<xxxx>)")
```
fatkudu: Again.  I think this just clutters up the source.

PR #314, opened 2017-08-16
ZCS-2320:Undeclared extension error is seen when executing few automa…

sneha-patil-synacor:
Issue:
"Undeclared extension" error is seen when executing filters manually from web client preferences

Fix:
Add the require statement at the beginning when fetching the partial rule from script

Added junit test.

### DIFF
--- a/store/src/java/com/zimbra/cs/filter/RuleManager.java
+++ b/store/src/java/com/zimbra/cs/filter/RuleManager.java
@@ -309,6 +309,10 @@ public final class RuleManager {
 
         try {
             while ((line = reader.readLine()) != null) {
+                if (line.startsWith("require")) {
+                    buf.append(line).append("\r\n");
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/filter/RuleManager.java
```diff
@@ -309,6 +309,9 @@ public static String getRuleByName(String script, String ruleName) {
 
         try {
             while ((line = reader.readLine()) != null) {
+                if (line.startsWith("require")) {
+                    buf.append(line).append("\r\n");
+                }
```
chinmay15: Adding continue; will not go for matching when require statement is line.

sneha-patil-synacor: Done

PR #316, opened 2017-08-17
ZCS-2410 : Support JWT in AuthRequest for authentication

chinmay15:
Change : 
1)  support jwt in AuthRequest
2) validate jwt received in AuthRequest
3) create new jwt if requested jwt is valid
4) send created jwt in response

Testing :
Added positive and negative test cases in unit tests

Testing to be done by QA :
1) Automate positive and negative test cases for jwt in AuthRequest
2) Test all scenarios related to jwt in AuthRequest

### DIFF
--- a/common/src/java/com/zimbra/common/service/ServiceException.java
+++ b/common/src/java/com/zimbra/common/service/ServiceException.java
@@ -347,7 +347,11 @@ public class ServiceException extends Exception {
     }
 
     public static ServiceException AUTH_REQUIRED() {
-        return new ServiceException("no valid authtoken present", AUTH_REQUIRED, SENDERS_FAULT);
+        return AUTH_REQUIRED("no valid authtoken present", null);
+    }
+
+    public static ServiceException AUTH_REQUIRED(String message, Throwable cause) {
+        return new ServiceException(message, AUTH_REQUIRED, SENDERS_FAULT, cause);
     }
 
     public static ServiceException CANNOT_DISABLE_TWO_FACTOR_AUTH() {
--- a/common/src/java/com/zimbra/common/service/ServiceException.java
+++ b/common/src/java/com/zimbra/common/service/ServiceException.java
@@ -347,11 +347,7 @@ public class ServiceException extends Exception {
     }
 
     public static ServiceException AUTH_REQUIRED() {
-        return AUTH_REQUIRED("no valid authtoken present", null);
-    }
-
-    public static ServiceException AUTH_REQUIRED(String message, Throwable cause) {
-        return new ServiceException(message, AUTH_REQUIRED, SENDERS_FAULT, cause);
+        return new ServiceException("no valid authtoken present", AUTH_REQUIRED, SENDERS_FAULT);
     }
 
     public static ServiceException CANNOT_DISABLE_TWO_FACTOR_AUTH() {
--- a/store/src/java/com/zimbra/cs/service/account/Auth.java
+++ b/store/src/java/com/zimbra/cs/service/account/Auth.java
@@ -150,12 +150,6 @@ public class Auth extends AccountDocumentHandler {
             acct = prov.getAccountById(claims.getSubject());
             acctAutoProvisioned = true;
         }
-        // generate and return jwt only if tokenType is JWT
-        if (TokenType.JWT.equals(tokenType) && claims != null) {
-            AuthToken at = AuthProvider.getAuthToken(acct, TokenType.JWT);
-            ZimbraLog.account.debug("auth: generated JWT based on authToken Element");
-            return doResponse(request, at, zsc, context, acct, csrfSupport, trustedToken, newDeviceId);
-        }
 
         if (authTokenEl != null) {
             boolean verifyAccount = authTokenEl.getAttributeBool(AccountConstants.A_VERIFY_ACCOUNT, false);
--- a/store/src/java/com/zimbra/cs/service/account/Auth.java
+++ b/store/src/java/com/zimbra/cs/service/account/Auth.java
@@ -20,8 +20,6 @@
  */
 package com.zimbra.cs.service.account;
 
-import java.nio.charset.StandardCharsets;
-import java.util.Date;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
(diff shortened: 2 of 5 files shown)


common/src/java/com/zimbra/common/service/ServiceException.java
```diff
@@ -347,7 +347,11 @@ public static ServiceException DEFEND_CALENDAR_RESOURCE_HARVEST(String calendarR
     }
 
     public static ServiceException AUTH_REQUIRED() {
-        return new ServiceException("no valid authtoken present", AUTH_REQUIRED, SENDERS_FAULT);
+        return AUTH_REQUIRED("no valid authtoken present", null);
+    }
+
+    public static ServiceException AUTH_REQUIRED(String message, Throwable cause) {
```
fatkudu: I dislike this historical overloading of ServiceException instead of extending it.  This means that if we want to catch only exceptions related to auth problems, we have to catch ServiceException and introspect it instead of catching, say, AuthException.  I'd consider introducing an AuthRequiredException which extends ServiceException.

store/src/java/com/zimbra/cs/service/account/Auth.java
```diff
@@ -508,4 +528,36 @@ public static void addAccountToLogContextByAuthToken(Provisioning prov, AuthToke
         if (aid != null && !aid.equals(id))
             AccountUtil.addAccountToLogContext(prov, aid, ZimbraLog.C_ANAME, ZimbraLog.C_AID, null);
     }
+
+    public static AuthToken validateAndCreateNewJwtToken(String jwt, Account acct, Provisioning prov) throws ServiceException {
+        if (StringUtil.isNullOrEmpty(jwt)) {
+            throw ServiceException.AUTH_REQUIRED();
+        }
+        String tokenKey = acct.getJWTKey();
```
Prashantsurana: can Storing JWT secret key on account level satisfy the requirement of multiple user sessions requirement? 
In case multiple users logged in from the same browser how will server understand which secret key should be used to validate the token. 

Is there any change in the requirement from client side? 

@shrikant-prasad  already raised this point.

store/src/java/com/zimbra/cs/service/account/Auth.java
```diff
@@ -508,4 +528,36 @@ public static void addAccountToLogContextByAuthToken(Provisioning prov, AuthToke
...
+        if (StringUtil.isNullOrEmpty(tokenKey)) {
+            throw ServiceException.AUTH_REQUIRED();
+        }
+        java.security.Key key = new SecretKeySpec(tokenKey.getBytes(StandardCharsets.UTF_8), SignatureAlgorithm.HS512.getJcaName());
+        Claims claims = null;
+        try {
+            claims = Jwts.parser().setSigningKey(key).parseClaimsJws(jwt).getBody();
+        } catch(ExpiredJwtException eje) {
+            throw ServiceException.AUTH_EXPIRED(eje.getMessage());
+        } catch(SignatureException se) {
```
plobbes: For ( UnsupportedJwtException | MalformedJwtException | IllegalArgumentException | SignatureException ) (and possibly the generic exception too)I think we should probably be throwing INVALID_REQUEST() and not AUTH_REQUIRED().

store/src/java/com/zimbra/cs/service/account/Auth.java
shrikant-prasad: JWT key will not be stored per account. I will change this implementation as part of ZCS-2390, bug I am currently working on. So for this ticket, we can allow it.

store/src/java/com/zimbra/cs/service/account/Auth.java
shrikant-prasad: I think separate exception should be  thrown for SignatureException which indicates that signature is not valid.

store/src/java/com/zimbra/cs/service/account/Auth.java
shrikant-prasad: Did you test this flow with soap request as the account object will be null here.
Moreover JWT key won't be stored per account now.

chinmay15: Rest of the changes are done but this change is done under https://github.com/Zimbra/zm-mailbox/pull/317 by @shrikant-prasad 
Will update pr once this pr is closed, as the changes are dependent.

store/src/java/com/zimbra/cs/service/account/Auth.java
```diff
@@ -508,4 +528,36 @@ public static void addAccountToLogContextByAuthToken(Provisioning prov, AuthToke
...
+            throw ServiceException.AUTH_REQUIRED("Signature verification failed", se);
```
shrikant-prasad: Instead of Auth Required, should use code like JWT_VALIDATION_FAILED.

common/src/java/com/zimbra/common/service/ServiceException.java
chinmay15: Used existing AuthFailedServiceException.AUTH_FAILED

store/src/java/com/zimbra/cs/service/account/Auth.java
chinmay15: Used existing AuthFailedServiceException.AUTH_FAILED

store/src/java/com/zimbra/cs/service/account/Auth.java
chinmay15: used existing AuthFailedServiceException.AUTH_FAILED

store/src/java/com/zimbra/cs/service/account/Auth.java
chinmay15: This has been taken care under zcs-2398 and tested basic(sending tokenType="JWT" and jwtAuthToken in req) scenario through soap.

chinmay15:
Changed jwtAutToken to jwtToken and kept tokenType as an optional field.

store/src/java/com/zimbra/cs/service/account/Auth.java
```diff
@@ -137,19 +137,26 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
             String jwt = jwtTokenElement.getText();
-            Claims claims = validateJwtToken(jwt);
-            Account authTokenAcct = prov.getAccountById(claims.getSubject());
-            AuthToken at = AuthProvider.getAuthToken(authTokenAcct, TokenType.JWT);
+            claims = validateJwtToken(jwt);
+            acct = prov.getAccountById(claims.getSubject());
+            acctAutoProvisioned = true;
+        }
+        // generate and return jwt only if tokenType is JWT
+        if (TokenType.JWT.equals(tokenType) && claims != null) {
```
shrikant-prasad: no need of this code.

chinmay15: done

PR #317, opened 2017-08-18
ZCS-2389: change per account key for jwt to global key

shrikant-prasad:
changing per account jwt secret key to global key. Reused auth key.

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9381,11 +9381,7 @@ TODO: delete them permanently from here
   <desc>Whether to enable zimbra network new generation mobile sync module.</desc>
 </attr>
 
-<attr id="2119" name="zimbraJWTKey" type="ostring" max="128" immutable="1" cardinality="single" optionalIn="account"  flags="domainAdminModifiable" since="8.8.0">
-  <desc>JWT secret key</desc>
-</attr>
-
-<attr id="2120" name="zimbraJWTLifetime" type="duration"  max="3600" cardinality="single" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" callback="PositiveTimeInterval" since="8.8.0">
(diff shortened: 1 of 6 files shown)


store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9381,11 +9381,7 @@ TODO: delete them permanently from here
   <desc>Whether to enable zimbra network new generation mobile sync module.</desc>
 </attr>
 
-<attr id="2119" name="zimbraJWTKey" type="ostring" max="128" immutable="1" cardinality="single" optionalIn="account"  flags="domainAdminModifiable" since="8.8.0">
-  <desc>JWT secret key</desc>
-</attr>
-
-<attr id="2120" name="zimbraJWTLifetime" type="duration"  max="3600" cardinality="single" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" callback="PositiveTimeInterval" since="8.8.0">
+<attr id="2119" name="zimbraJWTLifetime" type="duration"  max="3600" cardinality="single" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" callback="PositiveTimeInterval" since="8.8.0">
```
chinmay15: Can we use zimbraAuthTokenLifetime, instead of introducing new attribute?

shrikant-prasad: auth token and jwt can have different lifetime, so can't use auth token lifetime.

Prashantsurana: We should not maintain both the lifetime. 
The session timeout should be controlled and maintained by one attribute, client using auth token and or JWT is irrelevant to the end user. If session timeout is say 30 mins it should be similar across all webclients irrespective of clients using JWT or authtoken.

shrikant-prasad: session timeout is not controlled by jwt life time. the client might request for new jwt just before the jwt is about to expire. in this way session time would still be 30 mins however jwt lifetime might only be 5 mins or so.

shrikant-prasad:
updated JWTBasedAuthTest

store/conf/attrs/zimbra-attrs.xml
chinmay15: Let's discuss this on a separate jira ticket and unblock this one.

PR #320, opened 2017-08-18
ZMS-710: migration doesn't check pre-existing data in SSDB

ilyusha:
The attribute migration process used to check whether each attribute already had a value in SSDB prior to storing it; this was put in place to prevent old last-logon timestamps in LDAP from overwriting newer timestamps in SSDB. Since migration now happens prior to changing the backend, this is no longer a concern. And since SSDB setters are idempotent, running a migration process twice wouldn't really have any adverse effects; the data would simply be overwritten with an identical copy.

Removing this logic has two benefits:
1. The number of requests to SSDB is halved, improving migration efficiency
2. The output CSV can no longer have rows with zeroes for the number of attributes migrated, yet show "migration completed" as the status, which is potentially confusing to the user.

In the code, this is done by changing the return type of _MigrationCallback::setEphemeralData_  from boolean to void, so that the caller doesn't care whether the callback actually performed the operation or not.

### DIFF
--- a/store/src/java-test/com/zimbra/cs/ephemeral/migrate/MigrateAttributesTest.java
+++ b/store/src/java-test/com/zimbra/cs/ephemeral/migrate/MigrateAttributesTest.java
@@ -284,7 +284,7 @@ public class MigrateAttributesTest {
         }
 
         @Override
-        public boolean setEphemeralData(EphemeralInput input,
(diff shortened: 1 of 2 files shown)


ilyusha:
CI caught that I forgot to update MigrateAttributesTest accordingly

PR #323, opened 2017-08-21
ZCS-2390: support JWT in soap requests for authentication

shrikant-prasad:
Added logic to look for jwtToken in soap context and if not found, look in http request in Authorization header. Once jwtToken is found, validate it and pass the auth token created from encoded string stored inside jwt to dispatch the request.
Added junit test.

### DIFF
--- a/store/src/java/com/zimbra/cs/account/ZimbraAuthToken.java
+++ b/store/src/java/com/zimbra/cs/account/ZimbraAuthToken.java
@@ -83,6 +83,7 @@ public class ZimbraAuthToken extends AuthToken implements Cloneable {
     private static final String C_VALIDITY_VALUE  = "vv";
     private static final String C_AUTH_MECH = "am";
     private static final String C_USAGE = "u";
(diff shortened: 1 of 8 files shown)


store/src/java/com/zimbra/cs/account/ZimbraAuthToken.java
```diff
@@ -215,6 +216,12 @@ protected ZimbraAuthToken(String encoded) throws AuthTokenException {
...
             }
+            String tokenTypeCode = (String)map.get(C_TOKEN_TYPE);
+            if (tokenTypeCode != null) {
+                tokenType = TokenType.fromCode(tokenTypeCode);
+            } else {
+                tokenType = TokenType.AUTH;
+            }
             externalUserEmail = (String)map.get(C_EXTERNAL_USER_EMAIL);
             digest = (String)map.get(C_DIGEST);
             String vv = (String)map.get(C_VALIDITY_VALUE);
```
rupalid: Keep consistent fieldNames

PR #328, opened 2017-08-23
[ZCS-1957]  [NG NE] Allow a better transition for Delegated Admin Modules	

DinoPuller:
Allow old admin NE and admin NG to coexist until Delegated Admin rights migration is done.

### DIFF
--- a/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
+++ b/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
@@ -11485,6 +11485,14 @@ public class ZAttrProvisioning {
     public static final String A_zimbraNetworkActivation = "zimbraNetworkActivation";
 
     /**
+     * Whether to enable old zimbra network admin module.
+     *
+     * @since ZCS 8.8.0
+     */
+    @ZAttr(id=2119)
+    public static final String A_zimbraNetworkAdminEnabled = "zimbraNetworkAdminEnabled";
+
+    /**
      * Contents of a signed Zimbra license key - an XML string.
      */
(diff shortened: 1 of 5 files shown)


common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
```diff
@@ -11485,6 +11485,14 @@ public static TwoFactorAuthSecretEncoding fromString(String s) throws ServiceExc
     public static final String A_zimbraNetworkActivation = "zimbraNetworkActivation";
 
     /**
+     * Whether to enable old zimbra network admin module.
+     *
+     * @since ZCS 8.8.0
```
grishick: this should say since 8.8.2 at this point, because the next release is 8.8.2

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9381,6 +9381,11 @@ TODO: delete them permanently from here
   <desc>Whether to enable zimbra network new generation mobile sync module.</desc>
 </attr>
 
+<attr id="2119" name="zimbraNetworkAdminEnabled" type="boolean" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" requiresRestart="mailbox" since="8.8.0">
```
grishick: same comment about "since" as above

DinoPuller:
@grishick I've made the requested change. Is it possible to merge into the next version 17.37 ? (by Thursday should be better) 
Tnx

grishick:
@DinoPuller we need two approvals and then can assign this item to someone for merging.

PR #330, opened 2017-08-25
zcs-2473:Sieve filter's error handling about :matches

ykomiyam:
[bug]
The way of handling the error case regarding the ":matches"
operation has been altered, any mal-encoded value cancels
the filter execution with UnsupportedEncodingException,
and the message is delivered to the Inbox.

Previously, when the header value was mal-encoded, the
:matches compared the "key" to the header value "as is"
(the encoded value).

[root cause]
Each header value is decoded and unfolded before the match
operation is attempted.  When MimeUtility.decodeText() tries
to decode the header value, and if it finds that the given
value is mal-encoded, it throws the UnsupportedEncodingException.
In reality, as the header would contain the un-decodable
value, it is totally fine to ignore such exception.

This issue affects:
 * header test
 * address test
 * addheader action
 * deleteheader action
 * replaceheader action

[fix]
When the ":matches" operation needs to handle a mal-encoded
header value, it will no longer throw the UnsupportedEncodingException
and it will not abort the filter execution any more.
The mal-encoded header value will be compared with the
keys "as is" (as a not-encoded-just-a-plain-ascii string).

* unit test
  - removed an old test cases (HeaderTest.ttt())
  - Fix to add an initialization of the filter rule (HeaderTest.doTest())

### DIFF
--- a/store/src/java-test/com/zimbra/cs/filter/ReplaceHeaderTest.java
+++ b/store/src/java-test/com/zimbra/cs/filter/ReplaceHeaderTest.java
@@ -34,6 +34,7 @@ import org.junit.Test;
 import com.google.common.collect.Maps;
 import com.zimbra.common.account.Key;
 import com.zimbra.cs.account.Account;
+import com.zimbra.cs.account.MockProvisioning;
 import com.zimbra.cs.account.Provisioning;
 import com.zimbra.cs.filter.jsieve.EditHeaderExtension;
 import com.zimbra.cs.mailbox.DeliveryContext;
@@ -1676,6 +1677,31 @@ public class ReplaceHeaderTest {
             Assert.assertEquals("[Test]line 1 CRLF\r\n line 2", newSubject);
         } catch (Exception e) {
             fail("No exception should be thrown: " + e.getMessage());
+        }
+    }
+
+    @Test
+    public void testMalencodedHeader() throws Exception {
+        String script = "require [\"editheader\", \"variables\"];\n"
+                    + "replaceheader :newvalue \"[test]${1}\" :matches \"X-Mal-Encoded-Header\" \"*\";";
+        try {
+            Account account = Provisioning.getInstance().get(Key.AccountBy.name, "test@zimbra.com");
+            Mailbox mbox = MailboxManager.getInstance().getMailboxByAccount(account);
+            RuleManager.clearCachedRules(account);
+            account.setSieveEditHeaderEnabled(true);
+            account.setAdminSieveScriptBefore(script);
+
+            List<ItemId> ids = RuleManager.applyRulesToIncomingMessage(new OperationContext(mbox),
+                    mbox, new ParsedMessage("X-Mal-Encoded-Header: =?ABC?A?GyRCJFskMhsoQg==?=".getBytes(), false), 0,
+                    account.getName(), new DeliveryContext(), Mailbox.ID_FOLDER_INBOX, true);
+            Message message = mbox.getMessageById(null, ids.get(0).getId());
+            String[] headers = message.getMimeMessage().getHeader("X-Mal-Encoded-Header");
+            Assert.assertNotNull(headers);
+            Assert.assertNotSame(0, headers.length);
+            Assert.assertEquals("[test]=?ABC?A?GyRCJFskMhsoQg==?=", headers[0]);
+        } catch (Exception e) {
+            e.printStackTrace();
+            fail("No exception should be thrown" + e);
         }
     }
 }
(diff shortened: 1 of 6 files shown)


store/src/java/com/zimbra/cs/filter/jsieve/EditHeaderExtension.java
```diff
@@ -406,14 +406,12 @@ public void setupEditHeaderData(Arguments arguments, AbstractCommand ac) throws
...
         boolean matchFound = false;
-        String unfoldedAndDecodedHeaderValue = "";
+        String unfoldedAndDecodedHeaderValue = MimeUtility.unfold(header.getValue());
         try {
-            unfoldedAndDecodedHeaderValue =  MimeUtility.decodeText(MimeUtility.unfold(header.getValue()));
-            ZimbraLog.filter.debug("Header value before unfolding and decoding: %s", header.getValue());
+            unfoldedAndDecodedHeaderValue =  MimeUtility.decodeText(unfoldedAndDecodedHeaderValue);
             ZimbraLog.filter.debug("Header value after unfolding and decoding: %s", unfoldedAndDecodedHeaderValue);
         } catch (UnsupportedEncodingException uee) {
-            ZimbraLog.filter.debug("Failed to decode \"%s\"", MimeUtility.unfold(header.getValue()));
```
shrikant-prasad: should not remove the log.

ykomiyam: Yeh, I did think so first, but on second thought, I reconsidered this log may be less informative because of the following thoughts:
(1) because it is a debug log, system admin can't use this log for the production's root cause analysis.
(2) for the debug purpose, the original target header value has been logged before this operation (it is not un-folded though), and the resulting header value is also logged after this operation.  By comparing the original one and the result one, we can recognize what's happened on the decoding process.

That's the reason I removed this log event.

store/src/java/com/zimbra/cs/filter/jsieve/HeaderTest.java
```diff
@@ -348,11 +348,13 @@ public static void evaluateVarExp(ZimbraMailAdapter mailAdapter, List<String> so
                 }
                 List<String> decodedValues = new ArrayList<>();
                 for (String value : values) {
+                    value = MimeUtility.unfold(value);
                     try {
-                        decodedValues.add(MimeUtility.decodeText(MimeUtility.unfold(value)));
+                        value = MimeUtility.decodeText(value);
                     } catch (UnsupportedEncodingException e) {
-                        throw new SieveMailException("Exception occured while decoding header value", e);
+                        // "value" would contain the undecoded value, fine
```
rupalid: Log a message regarding the exception

store/src/java-test/com/zimbra/cs/filter/ReplaceHeaderTest.java
```diff
@@ -1678,4 +1679,29 @@ public void testMatchedStringForReplaceHeaderUsingWildcardMatchesToControlChars(
...
+            List<ItemId> ids = RuleManager.applyRulesToIncomingMessage(new OperationContext(mbox),
+                    mbox, new ParsedMessage("X-Mal-Encoded-Header: =?ABC?A?GyRCJFskMhsoQg==?=".getBytes(), false), 0,
+                    account.getName(), new DeliveryContext(), Mailbox.ID_FOLDER_INBOX, true);
+            Message message = mbox.getMessageById(null, ids.get(0).getId());
+            String[] headers = message.getMimeMessage().getHeader("X-Mal-Encoded-Header");
+            Assert.assertNotNull(headers);
+            Assert.assertNotSame(0, headers.length);
+            Assert.assertEquals("[test]=?ABC?A?GyRCJFskMhsoQg==?=", headers[0]);
+        } catch (Exception e) {
+            e.printStackTrace();
```
rupalid: Do you want keep this?

store/src/java/com/zimbra/cs/filter/jsieve/HeaderTest.java
ykomiyam: As in the return comment on the Shrikant, the result status of the decoding is detectable by other log event.  Besides, most of other places where the UnsupportedEncodingException is cached don't log the exception.

store/src/java-test/com/zimbra/cs/filter/ReplaceHeaderTest.java
ykomiyam: will remove the e.printStackTrace()

PR #331, opened 2017-08-29
zcs-2520:replaceheader may break the message structure

ykomiyam:
[Bug]
Performing replaceheader to the header with encoded line-breaks may break the message structure.  When the replaceheader generates the following style of the header, any undesired headers may be added or the MIME structure of the resulting message may be broken.
The resulting header value after replaceheader...
 - contains one or more than one line break characters (0x0A or 0x0D)
 - consists of only printable US-ASCII characters (0x21 - 0x7E), space (0x20), or tab (0x09)
 - the first character of the second line or later starts at the beginning of the line.

[Fix]
If the modified header value meets the following condition, that modified header value will be Base64 encoded:
 - the text consists of only 0x21 - 0x7E, 0x20, 0x09, 0x0A, and 0x0D
 - the 0x0A and/or 0x0D is in the middle of the text.

[Affected area]
  replaceheader
  addheader

### DIFF
--- a/store/src/java/com/zimbra/cs/mime/MimeUtil.java
+++ b/store/src/java/com/zimbra/cs/mime/MimeUtil.java
@@ -1,0 +1,179 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2017 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ * ***** END LICENSE BLOCK *****
+ */
+package com.zimbra.cs.mime;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.UnsupportedEncodingException;
+
+import javax.mail.internet.MimeUtility;
+
+import com.sun.mail.util.BEncoderStream;
+import com.sun.mail.util.PropUtil;
+import com.sun.mail.util.QEncoderStream;
+
+public class MimeUtil {
+    /** (non-Javadoc)
+     * Encode <tt>string</tt> no matter what characters are contained in it.
+     * @param string text to be encoded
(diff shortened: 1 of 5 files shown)


store/src/java/com/zimbra/cs/mime/MimeUtil.java
```diff
@@ -0,0 +1,179 @@
...
+     * @see  javax.mail.internet.MimeUtility.encodeWord(String string, String charset, String encoding, boolean encodingWord)
+     */
+    public static String encodeWord(String string, String charset,
+            String encoding, boolean encodingWord) throws UnsupportedEncodingException {
+        if (!string.contains("\r") && !string.contains("\n")){
+            String result = string;
+            try {
+                result = MimeUtility.encodeText(string);
+            } catch (UnsupportedEncodingException e) {
+                // ignore the exception and return the source string as is.
```
chinmay15: Do we need to log a message here? At lease at debug level.

store/src/java/com/zimbra/cs/mime/MimeUtil.java
```diff
@@ -0,0 +1,179 @@
...
+            return result;
+        }
+
+        // Else, apply the specified charset conversion.
+        String jcharset;
+        if (charset == null) { // use default charset
+            jcharset = MimeUtility.getDefaultJavaCharset(); // the java charset
+            charset = getDefaultMIMECharset(); // the MIME equivalent
+        } else // MIME charset -> java charset
+            jcharset = MimeUtility.javaCharset(charset);
```
chinmay15: please add {} for else block

store/src/java/com/zimbra/cs/mime/MimeUtil.java
```diff
@@ -0,0 +1,179 @@
...
+                eos = new BEncoderStream(os);
+            } else {
+                // "Q" encoding
+                eos = new QEncoderStream(os, encodingWord);
+            }
+
+            try { // do the encoding
+                eos.write(bytes);
+                eos.close();
+            } catch (IOException ioex) { }
```
chinmay15: log or throw exception here?

store/src/java/com/zimbra/cs/mime/MimeUtil.java
ykomiyam: fixed

store/src/java/com/zimbra/cs/mime/MimeUtil.java
ykomiyam: I don't think any log message is necessary; if the encoding is failed, I will see the un-encoded string in the resulting message.

store/src/java/com/zimbra/cs/mime/MimeUtil.java
ykomiyam: The "eos" (as wellas "os") virtually does not throw the IOException; the write() method could throw the NullPointerException or IndexOutOfBoundsException, but the root cause of both exceptions has been taken care before.  So, in this code, that is just a formality to catch such exception.

PR #332, opened 2017-08-29
ZCS-2388 : Editheader actions should not be supported in zimbraMail namespace

chinmay15:
Problem : editheader actions should not be supported in user script using api

Fix : Allowed editheader actions only in admin scripts using api

Testing done :
1) api tested manually
2) added unit tests to verify the same


chinmay15:
Merged in sieve_ver2

commit 4b833596c74a9f8034f968e5166b71e0da8893d1
Date:   Tue Aug 29 15:21:00 2017 +0530

    ZCS-2388 : Editheader actions should not be supported in zimbraMail namespace

PR #333, opened 2017-08-29
Bugfix/zcs 2638

grishick:
This PR addresses the scenarios where remote IMAP server uses outdated list of folders when processing IMAP requests. The two scenarios that I found so far are 
1) a remote IMAP session in a non-selected state does not pick up folder changes from SOAP notifications
2) a remote IMAP session finds cached folder in IMAP session cache after that folder has been deleted and re-created with the same path

These changes bring the number of IMAP compliance test failures on remote IMAP down to 13 (the same as local IMAP and older ZCS versions)


tmclane:
Looks great!

PR #336, opened 2017-08-31
ZCS-2388 : Possible NPE solved

chinmay15:
Possible NPE solved

Added unit test for same

### DIFF
--- a/soap/src/java/com/zimbra/soap/mail/type/FilterAction.java
+++ b/soap/src/java/com/zimbra/soap/mail/type/FilterAction.java
@@ -749,6 +749,9 @@ public class FilterAction {
             if (last != null && last && offset == null) {
                 throw ServiceException.PARSE_ERROR(":index <offset> must be provided with :last", null);
             }
+            if (test == null) {
+                throw ServiceException.PARSE_ERROR("<test> is mandatory in action", null);
+            }
             test.validateEditheaderTest();
         }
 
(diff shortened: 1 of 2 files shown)


chinmay15:
Merged in sive_ver2 branch

commit a5db37c1e3576cfdee27e16d9fdb98c304199ce5
Date:   Thu Aug 31 11:50:34 2017 +0530

    ZCS-2388 : Possible NPE solved

soap/src/java/com/zimbra/soap/mail/type/FilterAction.java
```diff
@@ -749,6 +749,9 @@ public void validateDeleteheaderAction() throws ServiceException {
             if (last != null && last && offset == null) {
                 throw ServiceException.PARSE_ERROR(":index <offset> must be provided with :last", null);
             }
+            if (test == null) {
+                throw ServiceException.PARSE_ERROR("<test> is mandatory in action", null);
+            }
```
grishick: This NULL check seems to contradict the JAXB contract that is being defined earlier in this file: @XmlElement(name=AdminConstants.E_TEST /* test */, required=false)
if "test" is required, then the annotation should state so and this NPE will be avoided without the explicit NULL check. Otherwise, the JAXB annotation is effectively invalid.

chinmay15: agree. will remove null check and add required=true.

PR #339, opened 2017-09-05
ZIMBRA-93: Prompt for saved search folder when searching 3+ times

ilyusha:
This PR introduces Search History to the ZCS server

### Interface
A new _SearchHistory_ interface has been added. The interface provides methods to:
 - check whether a given search string exists in the history
 - get the number of times a given search string has been searched
 - register a new search string with history
 - add an instance of a search string to the history
 - search the search history for entries matching a prefix
 - purge history older than a given lifetime
 - delete all search history
 - return a _SavedSearchPromptLog_ instance (see below)

A second interface _SavedSearchPromptLog_ has been added. This provides the ability to get/set the _SavedSearchStatus_ for a given search string. _SavedSearchStatus_ can be one of `NOT_PROMPTED`, `PROMPTED`, `CREATED`, or `REJECTED`; this is used to ensure that a user is only prompted to create a saved search folder once per search string.

The primary implementation of _SearchHistory_ is the _ZimbraSearchHistory_ class. While this is a concrete class, it delegates to three other interfaces: 
 - _SearchHistoryIndex_ can be queried for the IDs of unique search strings that match a given query. 
- _SearchHistoryMetadata_ stores information about items in the search history, and can be queried for the count or list of searches. This query can include an optional time cutoff, result limit, or ID subset provided by the index.
 - _SearchHistoryConfig_ is a helper interface that exposes a getMaxAge() method. The primary implementation gets this value from LDAP, but an in-memory implementation was used during development.

### Database
_DbSearchHistoryMetadata_ is the SQL-based implementation of _SearchHistoryMetadata_. It relies on two new database tables:
 - The `searches` table stores information about unique search strings in the history. It stores the mailbox id, search string id unique to mailbox, the string itself, last search date, and prompt status.
 - The `search_history` table stores a running log of searches with associated timestamps, with the search id as a foreign key.
 - The `zimbra.mailbox` table has been updated with a new column `search_id_checkpoint` that mimics `item_id_checkpoint` functionality.
An in-memory implementation is provided as well; this was mainly used for initial development.

### Indexing
Since searching through search history is beyond the scope of this ticket, true indexing is not included in this PR. _InMemorySearchHistoryIndex_ is provided as a dummy implementation; a Lucene-based implementation will be added as part of ZIMBRA-92.

### SOAP endpoints
 - _SearchRequest_ has been updated with an optional boolean `logSearch` attribute. If set to false, the  query is not added to search history. If not specified, defaults to true.
 - _SearchResponse_ has been updated with an optional boolean `saveSearchPrompt` attribute, which is set to true if the search count for the query reaches the necessary threshold within the search history duration window
 - _RejectSaveSearchPrompt_ sets the _SavedSearchStatus_ to `REJECTED` for the given search string. If the provided string is not currently in `PROMPTED` state, a warning is logged. The response for this request is empty in either case.
 - _ClearSearchHistoryRequest_ clears all search history and returns an empty response.

### New LDAP attributes
 - _zimbraFeatureSearchHistoryEnabled_: Account/COS level toggle to enable the new feature; defaults to TRUE
 - _zimbraNumSearchesForSavedSearchPrompt_ The number of searches required to trigger a saved search prompt; defaults to 3
 - zimbraSearchHistoryDuration: Lifetime of search history. Searches older than this will not count towards the count, and will be permanently deleted during account purging

### Other Notes:
 - Mailbox purging has been updated to purge search history entries logged outside the _zimbraSearchHistoryDuration_ window. This means that while searches older than this cutoff do not count towards _zimbraNumSearchesForSavedSearchPrompt_, the will remain in the database until a purge thread activates.
 - Operations on search history are wired into the Mailbox transaction mechanism. Redolog entries have been added for updating, purging, and deleting search history.
 - Not all calls to SearchRequest make sense to record in the history. For example, the client issues an automated "inid:" search when the UI is loaded. Navigating between folders uses "in:" queries. Paginating search results uses SearchRequest with an offset. Currently, these rules are hard-coded in the static _SearchHistory::shouldSaveInHistory_ method. However, the `logSearch` SearchRequest attribute has been added so that the client can eventually set it for only those searches that make sense to log.
 - While the SavedSearchStatus is set to CREATED during a CreateSearchFolderRequest, it is NOT unset if the search folder is deleted.  Presumably, if the user deletes a search folder, the prompt should not come up again.
 - The `zmmailbox` utility has been updated to support search history operations.
 - Unit tests have been added for the individual indexing/metadata components, as well as for the system as a whole; a _TestSavedSearchPrompt_ SOAP test is included as well.
 - One piece that is currently missing, but not required for this initial development, is caching: it is likely worthwhile to introduce a caching layer to avoid excessive DB calls to fetch search history information.


client/src/java/com/zimbra/client/ZMailbox.java
```diff
@@ -6450,6 +6452,21 @@ public void beginTrackingImap() throws ServiceException {
...
 
+    /**
+     * Clear the search history for this mailbox
+     */
+    public void clearSearchHistory() throws ServiceException {
+        invokeJaxb(new ClearSearchHistoryRequest());
+    }
+
+    /**
+     * Notifiy the server that the client rejected a prompt to save a search
```
tmclane: nit: Notify

grishick:
@ilyusha would it make sense to add your collaborators on the search V team to this PR?

ilyusha:
@grishick yes, but I don't think Varun has been added to this group yet. I'll work on getting him added.

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9563,4 +9563,19 @@ TODO: delete them permanently from here
...
+  <desc>Whether to enable Search History</desc>
+  <defaultCOSValue>TRUE</defaultCOSValue>
+</attr>
+
+<attr id="3021" name="zimbraNumSearchesForSavedSearchPrompt" type="integer" min="0" cardinality="single" optionalIn="account,cos" flags="accountInherited" since="8.8.1">
+  <desc>Searching for the same query this many times will result in a prompt offering the user to save this query as a search folder. This will only happen once per query. A value of 0 disables this feature.</desc>
+  <defaultCOSValue>3</defaultCOSValue>
+</attr>
+
+<attr id="3022" name="zimbraSearchHistoryDuration" type="duration" cardinality="single" optionalIn="account,cos" flags="accountInherited" since="8.8.1">
```
grishick: minor: perhaps "Age" instead of "Duration"?

ilyusha: yeah that would be more clear, especially because the parameter name used by the api is "maxAgeMillis"

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -10422,4 +10453,103 @@ public void noOp() throws ServiceException {
...
+            success = true;
+        } finally {
+            endTransaction(success);
+        }
+    }
+
+    public List<String> getSearchHistory(OperationContext octxt, SearchHistoryParams params) throws ServiceException {
+        boolean success = false;
+        try {
+            beginReadTransaction("getSearchHistory", octxt);
```
grishick: why do we need read and write to search history be inside a transaction?

ilyusha: I thought it was necessary for consistency (the searches table has an FK constraint on mailbox_id) and for redologs. I suppose reading doesn't have to be in a read transaction... but I worried about adding DB code that references mailboxes outside of the existing transaction mechanism.

grishick:
how is search_history table currently created in MariaDB?

ilyusha:
create_database.sql creates these tables. For existing databases, the zm-build and zm-db-conf repos have PRs for the DB schema upgrade.

grishick:
Not sure if this is relevant at this point, since you are changing DB_VERSION you may have to add an upgrade procedure to MailboxUpgrade.java

ilyusha:
Hm, I need to look into this a bit more, but I'm not sure that this mechanism is used anymore

grishick:
Please make sure com.zimbra.cs.filter.EnvelopeTest failure is not related to these changes.

ilyusha:
@grishick with the latest rebase, com.zimbra.cs.filter.EnvelopeTest passes on my dev machine.

PR #340, opened 2017-09-08
ZCS-1010:Add an attribute for comparator in case of valueComparison

shrikant-prasad:
1. added support for caseSensitivity attribute when valueComparison is used in header/address/envelope test.
2. added valueComparisonComparator attribute to explicitly specify the comparator to use for valueComparison

added junit tests.

### DIFF
--- a/store/src/java/com/zimbra/cs/filter/SieveToSoap.java
+++ b/store/src/java/com/zimbra/cs/filter/SieveToSoap.java
@@ -288,7 +288,7 @@ public final class SieveToSoap extends SieveVisitor {
 
     @Override
     protected void visitHeaderTest(Node node, VisitPhase phase, RuleProperties props,
-            List<String> headers, Sieve.ValueComparison comparison, boolean isCount, String value) {
+            List<String> headers, Sieve.ValueComparison comparison, Sieve.Comparator comparator, boolean isCount, String value) {
         if (phase == VisitPhase.begin) {
             FilterTest.HeaderTest test = addTest(new FilterTest.HeaderTest(), props);
             test.setHeaders(Joiner.on(',').join(headers));
@@ -296,6 +296,12 @@ public final class SieveToSoap extends SieveVisitor {
                test.setCountComparison(comparison.toString());
             } else {
                test.setValueComparison(comparison.toString());
+               if (comparator != null) {
+                   test.setValueComparisonComparator(comparator.toString());
+                   if (Sieve.Comparator.ioctet.equals(comparator)) {
+                       test.setCaseSensitive(true);
+                   }
+               }
             }
             test.setValue(value);
         }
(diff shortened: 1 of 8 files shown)


store/src/java/com/zimbra/cs/filter/SieveToSoap.java
```diff
@@ -288,14 +288,20 @@ protected void visitHeaderTest(Node node, VisitPhase phase, RuleProperties props
...
         if (phase == VisitPhase.begin) {
             FilterTest.HeaderTest test = addTest(new FilterTest.HeaderTest(), props);
             test.setHeaders(Joiner.on(',').join(headers));
             if (isCount) {
                test.setCountComparison(comparison.toString());
             } else {
                test.setValueComparison(comparison.toString());
+               if (comparator != null) {
+                   test.setValueComparisonComparator(comparator.toString());
+                   if (Sieve.Comparator.ioctet.equals(comparator)) {
```
ykomiyam: Is "comparator" guaranteed to be lower-case string?  If not, isn't it safer to be "Sieve.Comparator.ioctet.equalsIgnoreCase(comparator)" ?

store/src/java/com/zimbra/cs/filter/SieveToSoap.java
```diff
@@ -353,6 +359,12 @@ private void visitAddress(FilterTest.AddressTest test, VisitPhase phase, RulePro
                 test.setCountComparison(comparison.toString());
              } else {
                 test.setValueComparison(comparison.toString());
+                if (comparator != null) {
+                    test.setValueComparisonComparator(comparator.toString());
+                    if (Sieve.Comparator.ioctet.equals(comparator)) {
```
ykomiyam: ditto

store/src/java/com/zimbra/cs/filter/SoapToSieve.java
```diff
@@ -412,20 +413,29 @@ private static String toSieve(String name, String header, Sieve.StringComparison
...
+            boolean numeric = true;
+            try {
+                Integer.parseInt(value);
+            } catch (NumberFormatException e) {
+                numeric = false;
+            }
+            //for :count, iasciinumeric comparator will be used always
+            //for :value, if comparator value is not set in input, iasciinumeric comparator will be used if value is numeric
+            //else iasciicasemap will be used for case-insensitive and ioctet for case-sensitive 
+            comparator= Sieve.Comparator.iasciinumeric;
```
ykomiyam: put a space before '=' in this line and line 432 and 434.

store/src/java/com/zimbra/cs/filter/SieveToSoap.java
shrikant-prasad: the comparison done here checks  enum reference equality rather than value. So if the comparators are supposed to be case insensitive (couldn't find in rfc), then Sieve.Comparator enum fromString method should do case insensitive comparison. currently it's case sensitive.

store/src/java/com/zimbra/cs/filter/SoapToSieve.java
```diff
@@ -412,20 +413,29 @@ private static String toSieve(String name, String header, Sieve.StringComparison
...
+            } catch (NumberFormatException e) {
+                numeric = false;
+            }
+            //for :count, iasciinumeric comparator will be used always
+            //for :value, if comparator value is not set in input, iasciinumeric comparator will be used if value is numeric
+            //else iasciicasemap will be used for case-insensitive and ioctet for case-sensitive 
+            comparator = Sieve.Comparator.iasciinumeric;
+            if (!numeric && !isCount) {
+                if (caseSensitive) {
+                    comparator= Sieve.Comparator.ioctet;
```
ykomiyam: put a space here too

store/src/java/com/zimbra/cs/filter/SoapToSieve.java
```diff
@@ -412,20 +413,29 @@ private static String toSieve(String name, String header, Sieve.StringComparison
...
+                } else {
+                    comparator= Sieve.Comparator.iasciicasemap;
```
ykomiyam: And put a space here as well.

PR #343, opened 2017-09-08
Updated schema version

sbluhm:
The tag async-supported only exists from schema version 3.0 onwards. So the XML version was upped and both schema location changed to the correct ones


fatkudu:
Committed INSTALL-DEV-VMWARE-GUEST rename to develop as 1646704fbb

fatkudu:
Merged to develop: See:
merge commit 77e2ce671b 
for 060593478d and cd9c06bb41

PR #351, opened 2017-09-12
ZCS-1870:Multi-byte text in a long line gets garbled

ykomiyam:
[Problem]
Currently when the mail server receives the SendMsg request
whose message's mime parts have following characteristics,
the mail server transfers the outgoing message to the SMTP
server with the header "Content-Transfer-Encoding: binary".
 - If that mime part is not an attachment or its Content-Type's primary type is "text".
 - If that mime part contains non-ASCII characters more than 25% of the part size.
 - If that mime part has a line which is longer than MAX_LINE_OCTETS (900).
Because the MTA (Postfix) does not support the "Content-
Transfer-Encoding: binary", Postfix treats such mime part as
"7bit", and folds a long line into 998 characters without
consideration of multi-byte characters.  As a result, the
multi-byte character which CRLF is inserted in the middle of
the byte-string gets garbled.

[History]
This bug was introduced by the Bug 86352 (a redirected message
was garbled when the original message consisted of the multi-byte
characters but didn't have a Content-Transfer-Encoding header).
During the fix of Bug 86352, when a mime-part did not have CTE
header, and the mime contents matched the above conditions, the
"Content-Transfer-Encoding: binary" was inserted. Correctly, it
should have 7bit or 8bit, because the content is still formatted
in the line-oriented mime-part.

[Fix]
(1) If the mime-part is matched with the condition above, and if
it is sent via SnedMail, ZCS will put "Content-Transfer-Encoding:
base64".
(2) When ZCS redirects such a message, and if the original message
does not have any Content-Transfer-Encoding header, then ZCS will
add "Content-Transfer-Encoding: 8bit" before redirecting.

Although 7bit is a default value of the CTE defined in the RFC,
8bit is chosen for the pattern (2), because:
 - Most of the MTA servers, including Postfix, support 8BITMIME extension.
 - 8bit is a super-set of the 7bit.

----
Manual test
* Compose a long line multi-byte text message on the ZWC, and send to userA --> Verify that message is encoded in base64 and not garbled on the userA's ZWC.
* Send a raw Shift-JIS text message via LMTP to userB; the message is single mime part without Content-Transfer-Encoding header.  UserB has a Sieve filter which redirects messages to userC --> Verify that userC received a redirected message which is not garbled on the userC's ZWC.
- Ran automation script zm-qa/data/soapvalidator/MailClient/Mail/Bugs/bug50507.xml  (similar issues to this bug and Bug 86352)

### DIFF
--- a/common/src/java/com/zimbra/common/zmime/ZMimeBodyPart.java
+++ b/common/src/java/com/zimbra/common/zmime/ZMimeBodyPart.java
@@ -1,7 +1,7 @@
 /*
  * ***** BEGIN LICENSE BLOCK *****
  * Zimbra Collaboration Suite Server
- * Copyright (C) 2011, 2012, 2013, 2014, 2016 Synacor, Inc.
+ * Copyright (C) 2011, 2012, 2013, 2014, 2016, 2017 Synacor, Inc.
  *
  * This program is free software: you can redistribute it and/or modify it under
  * the terms of the GNU General Public License as published by the Free Software Foundation,
@@ -412,7 +412,7 @@ public class ZMimeBodyPart extends MimeBodyPart implements ZMimePart {
             } else if (toolong == 0) {
                 return "8bit"; //section 6.2 of RFC2045
             } else {
-                return "binary";
+                return "base64";
             }
         }
     }
--- a/common/src/java-test/com/zimbra/common/zmime/ZMimeBodyPartTest.java
+++ b/common/src/java-test/com/zimbra/common/zmime/ZMimeBodyPartTest.java
@@ -1,7 +1,7 @@
 /*
  * ***** BEGIN LICENSE BLOCK *****
  * Zimbra Collaboration Suite Server
- * Copyright (C) 2011, 2012, 2013, 2014, 2016 Synacor, Inc.
+ * Copyright (C) 2011, 2012, 2013, 2014, 2016, 2017 Synacor, Inc.
  *
  * This program is free software: you can redistribute it and/or modify it under
  * the terms of the GNU General Public License as published by the Free Software Foundation,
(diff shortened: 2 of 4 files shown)


common/src/java/com/zimbra/common/zmime/ZMimeBodyPart.java
```diff
@@ -412,7 +412,7 @@ String getEncoding(boolean isAttachment, boolean isText) {
             } else if (toolong == 0) {
                 return "8bit"; //section 6.2 of RFC2045
             } else {
-                return "binary";
+                return "base64";
```
chinmay15: LGTM.
@Prashantsurana can you please review this one?

ykomiyam:
Hi @Prashantsurana , Here is the result of the unit tests for Bug 86352
- com.zimbra.cs.mime.FixedMimeMessageTest --> PASS
- com.zimbra.common.zmime.ZMimeBodyPartTest --> PASS
      (This fix also added extra test cases to cover some patters with multi-byte characters text)

fatkudu:
This is probably a silly question but I assume that this fix causes base64 encoding to actually be done on the content?  It would be bad to claim a body part was base64 encoded if it wasn't.

ykomiyam:
Hi @fatkudu , 
That's right; before Bug 86352, the redirect message was sent as a base64 encoded MIME part but actually it was a plain multi-byte text.  To fix this, Bug 86352 changed the ZMimePart.EncodingTester.getEncoding() to return "binary" for the CTE so that the redirect message could send as binary data.  The problem was, this method is primarily used for determining how to encode the outgoing message requested via SendMsg API; ZCS converts the message body based on the return value of getEncoding().
So, I backed out the fix of 86352 so that getEncoding() would return the base64 again; that's the same behaviour as the 8.6.0. And explicitly put "8bit" if the redirect message didn't have a CTE header.

PR #359, opened 2017-09-17
ZCS-2746 Reworked the fix

rupalid:
Reworked the fix to a new location for failing test cases.
Added unit test.
Verified that SOAP automation is passing.


rupalid:
Working on failing unit test cases as part of ZCS-2772. The failing unit test cases are not because of changes in this bug.

PR #363, opened 2017-09-18
ZCS-2730:Adding resolve request parameter

sneha-patil-synacor:
Adding one more parameter in RestoreContactRequest called 'resolve' to support all 4 options (replace, ignore, reset and modify) in import API

### DIFF
--- a/soap/src/java/com/zimbra/soap/mail/message/RestoreContactsRequest.java
+++ b/soap/src/java/com/zimbra/soap/mail/message/RestoreContactsRequest.java
@@ -16,12 +16,16 @@
  */
 package com.zimbra.soap.mail.message;
 
+import java.util.Arrays;
+
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlEnum;
 import javax.xml.bind.annotation.XmlRootElement;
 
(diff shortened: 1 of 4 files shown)


shrikant-prasad:
Please update soap.txt also.

sneha-patil-synacor:
soap.txt updated.

soap/src/java/com/zimbra/soap/mail/message/RestoreContactsRequest.java
```diff
@@ -35,6 +39,38 @@
     @XmlAttribute(name = MailConstants.A_CONTACTS_BACKUP_FILE_NAME, required = true)
     private String contactsBackupFileName;
 
+    /**
+     * @zm-api-field-tag resolve
+     * @zm-api-field-description restore resolve action
```
fatkudu: Please state which is the default here - that is useful for the HTML SOAP docs.
Also, could you document what the various options mean please. I'm guessing ignore means if a contact with the same ID (what identifies a contact?) already exists, skip this one. modify means merge?? replace is fairly clear, no idea what reset means.  Although updating the enum with the docs makes most sense from a source perspective, currently here would be the best place to ensure the information ends up in the SOAP docs.

sneha-patil-synacor: Updated the description. Please review https://github.com/Zimbra/zm-mailbox/pull/364

PR #365, opened 2017-09-18
Feature/zcs 2836

grishick:
I had to abandon the idea of adding a completely generic 'custom' DataSource type, because it was turning out to be too convoluted and seemed like an premature optimization. Instead, I added 'oauth' type similar to 'imap', 'pop3' and others except that it allows a custom DataImport implementation. 

One optimization I added is to stop using special objectClass value for this new kind of datasource. 

Going forward, we may have to generalize SOAP handlers to handle all possible dataSource attributes so that new dataSource implementations don't have to add anything to zm-soap or zm-client libraries. For now, new zm-soap and zm-client classes cover oauth-based datasources.

### DIFF
--- a/client/src/java/com/zimbra/client/ZPop3DataSource.java
+++ b/client/src/java/com/zimbra/client/ZPop3DataSource.java
@@ -66,7 +66,7 @@ public class ZPop3DataSource extends ZDataSource implements ToZJSONObject {
         src.addAttribute(MailConstants.A_DS_PASSWORD, data.getPassword());
         src.addAttribute(MailConstants.A_FOLDER, data.getFolderId());
         src.addAttribute(MailConstants.A_DS_CONNECTION_TYPE, data.getConnectionType().name());
-        src.addAttribute(MailConstants.A_DS_LEAVE_ON_SERVER, getData().isLeaveOnServer());
+        src.addAttribute(MailConstants.A_DS_LEAVE_ON_SERVER, leaveOnServer());
         ZimbraLog.test.info("XXX bburtin: " + src.prettyPrint());
         return src;
     }
@@ -89,7 +89,7 @@ public class ZPop3DataSource extends ZDataSource implements ToZJSONObject {
         jaxbObject.setPassword(data.getPassword());
         jaxbObject.setFolderId(data.getFolderId());
         jaxbObject.setConnectionType(data.getConnectionType());
-        jaxbObject.setLeaveOnServer(getData().isLeaveOnServer());
+        jaxbObject.setLeaveOnServer(leaveOnServer());
         jaxbObject.setEnabled(data.isEnabled());
         return jaxbObject;
     }
--- a/client/src/java/com/zimbra/client/ZPop3DataSource.java
+++ b/client/src/java/com/zimbra/client/ZPop3DataSource.java
@@ -143,7 +143,7 @@ public class ZPop3DataSource extends ZDataSource implements ToZJSONObject {
         zjo.put("username", data.getUsername());
         zjo.put("folderId", data.getFolderId());
         zjo.put("connectionType", data.getConnectionType().toString());
-        zjo.put("leaveOnServer", getData().isLeaveOnServer());
+        zjo.put("leaveOnServer", leaveOnServer());
         return zjo;
     }
 
--- a/client/src/java/com/zimbra/client/ZPop3DataSource.java
+++ b/client/src/java/com/zimbra/client/ZPop3DataSource.java
@@ -35,8 +35,6 @@ import com.zimbra.soap.type.Pop3DataSource;
 
 public class ZPop3DataSource extends ZDataSource implements ToZJSONObject {
 
-    private Pop3DataSource data;
-
     public ZPop3DataSource(Pop3DataSource data) {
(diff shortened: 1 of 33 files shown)


client/src/java/com/zimbra/client/ZPop3DataSource.java
```diff
@@ -68,7 +66,7 @@ public Element toElement(Element parent) {
         src.addAttribute(MailConstants.A_DS_PASSWORD, data.getPassword());
         src.addAttribute(MailConstants.A_FOLDER, data.getFolderId());
         src.addAttribute(MailConstants.A_DS_CONNECTION_TYPE, data.getConnectionType().name());
-        src.addAttribute(MailConstants.A_DS_LEAVE_ON_SERVER, data.isLeaveOnServer());
+        src.addAttribute(MailConstants.A_DS_LEAVE_ON_SERVER, getData().isLeaveOnServer());
```
fatkudu: The contents of the `leaveOnServer` method below imply that the last arg in this can be null which will lead to an NPE.  Suggest just using the `leaveOnServer` method which handles that.

client/src/java/com/zimbra/client/ZPop3DataSource.java
```diff
@@ -150,7 +143,7 @@ public ZJSONObject toZJSONObject() throws JSONException {
         zjo.put("username", data.getUsername());
         zjo.put("folderId", data.getFolderId());
         zjo.put("connectionType", data.getConnectionType().toString());
-        zjo.put("leaveOnServer", data.isLeaveOnServer());
+        zjo.put("leaveOnServer", getData().isLeaveOnServer());
```
fatkudu: use `leaveOnServer` instead to avoid needing to handle nulls?

soap/src/java/com/zimbra/soap/mail/message/GetDataSourcesResponse.java
```diff
@@ -56,6 +57,7 @@
         @XmlElement(name=MailConstants.E_DS_RSS /* rss */, type=MailRssDataSource.class),
         @XmlElement(name=MailConstants.E_DS_GAL /* gal */, type=MailGalDataSource.class),
         @XmlElement(name=MailConstants.E_DS_CAL /* cal */, type=MailCalDataSource.class),
+        @XmlElement(name=MailConstants.E_DS_OAUTH /* custom */, type=MailOAuthDataSource.class),
```
fatkudu: s/custom/oauth/

store/src/java/com/zimbra/cs/account/ldap/entry/LdapDataSource.java
```diff
@@ -62,10 +62,8 @@ public static String getObjectClass(DataSourceType type) {
                 return AttributeClass.OC_zimbraRssDataSource;
             case gal:
                 return AttributeClass.OC_zimbraGalDataSource;
-            case custom:
-                return AttributeClass.OC_zimbraDataSource;
             default: 
-                return null;
+                return AttributeClass.OC_zimbraDataSource;
```
fatkudu: Nothing for `oauth`?

store/src/java/com/zimbra/cs/datasource/DataSourceManager.java
```diff
@@ -220,7 +220,7 @@ public DataImport getDataImport(DataSource ds, boolean test) throws ServiceExcep
             } catch (Exception x) {
                 ZimbraLog.datasource.warn("Failed instantiating xsync class: %s", ds, x);
             }
-        case custom:
+        case oauth:
```
fatkudu: There are references to `custom` in this case block which should be changed to `oauth`

store/src/java/com/zimbra/cs/account/ldap/entry/LdapDataSource.java
grishick: correct, oauth and everything else should return "dataSource" objectClass. What I am trying to accomplish here is to avoid using special LDAP objectClass for every kind of data source and use a generic dataSource LDAP class while moving al the specifics into implementation. I think that all the subclasses of LDAP objectClass "dataSource" are a design flaw and shouldn't have been added.

fatkudu: Could you add a comment on your reasoning there please?  To avoid someone going in the opposite direction in the future :)

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -8944,19 +8944,19 @@ TODO: delete them permanently from here
   <desc>Subject prefix for the spam training messages used to sent to the zimbraSpamIsSpamAccount/zimbraSpamIsNotSpamAccount account.</desc>
 </attr>
 
-<attr id="2021" name="zimbraDataSourceOAuthClientId" type="string" cardinality="single" optionalIn="imapDataSource" since="8.7.0,9.0.0">
+<attr id="2021" name="zimbraDataSourceOAuthClientId" type="string" cardinality="single" optionalIn="dataSource,imapDataSource" since="8.7.0,9.0.0">
```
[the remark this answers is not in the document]
grishick: done

PR #366, opened 2017-09-19
ZCS-2287 IMAP and folder IDs

fatkudu:
Changes related to exploring whether we had any remaining issues related to using only ints for folder IDs in IMAP.

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/ImapPath.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapPath.java
@@ -19,6 +19,7 @@ package com.zimbra.cs.imap;
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.zimbra.client.ZFolder;
 import com.zimbra.client.ZMailbox;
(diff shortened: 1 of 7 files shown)


store/src/java/com/zimbra/cs/imap/ImapPath.java
```diff
@@ -422,7 +424,8 @@ protected boolean useReferent() throws ServiceException {
      * @return If the folder is a mountpoint (i.e. an accepted share), may return an ImapPath representing
      *         that, otherwise, the value is this.
      */
-    protected ImapPath getReferent() throws ServiceException {
+    @VisibleForTesting
+    public ImapPath getReferent() throws ServiceException {
```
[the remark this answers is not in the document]
tmclane: How does it compute these complexity values? Sheesh that's so big it's useless.

store/src/java/com/zimbra/cs/imap/RemoteImapMailboxStore.java
```diff
@@ -107,7 +101,10 @@ public void deleteMessages(OperationContext octxt, List<Integer> ids) {
         }
     }
 
-    /** @return List of IMAP UIDs */
+    /**
+     * MUST only be called when the source items and target folder are in the same mailbox
+     * @return List of IMAP UIDs
+     */
     @Override
```
tmclane: Would it be smart to add a runtime check to validate this requirement such that it fails nicely?
(Perhaps it's done inside zMailbox.imapCopy ? )

fatkudu: Well, in some senses the requirements will always be met - because the API only talks about IDs in terms of a single mailbox.  The comment is really a warning about that limitation - if it were to be called in the wrong circumstances, odd things would happen.  The only "fails nicely" bit would be if some of the IDs from the wrong mailbox happened not to exist in the one being used then there would be failures.

tmclane: Thanks for the explanation 👍

PR #367, opened 2017-09-19
ZIMBRA-92: Search suggestions based on search history

ilyusha:
This PR expands on search history functionality in ZIMBRA-93 to introduce search suggestion funcationality.

### Search History Indexing
Search history is indexed using the new class _LuceneSearchHistoryIndex_.  Index documents for search history entries have four fields:
 - `sh_exact` stores the whole search string without any tokenization, for exact matches and prefix matches that act on the entire string
 - `sh_terms` stores the tokenized search string, for term and prefix matching
 - `search_id` stores the integer ID of the search string
 - `type` stores the string _sh_ for every search history entry. This is only used to make it easy to delete all search history entries from the index using during a _SearchHistory.deleteHistory()_ method call. It is possible that reliance on this field will be removed during the transition to SOLR.

While the `sh_terms` field is indexed using the default _UniversalAnalyzer_, A new _SearchHistoryQueryAnalyzer_ is used to tokenize the provided prefix string. The only difference between _UniversalAnalyzer_ and _SearchHistoryQueryAnalyzer_ is that the latter doesn't drop stopwords from the token stream; this is necessary to make it possible to make suggestions for a prefix like "a". 

As per the _SearchHistoryIndex_ interface, _LuceneSearchHistoryIndex::search_ returns a list of search history IDs taken from the `search_id` lucene field.

### New SOAP endpoints
_GetSearchHistoryRequest_ returns recent search history entries, in reverse chronological order
_SearchSuggestRequest_ returns entries that match the given prefix, sorted by relevance

### zmmailbox updates
 - `searchSuggest {query} [limit]` returns search suggestions for the given prefix
 - `getSearchHistory [limit]` returns chronological list of search history
 
### Autocompletion Relevance
Ideally, the match precedence should be:
 1. Exact match
 2. Term match ("apple" matches "apple juice" or "green apple")
 3. Edge match ("apple" matches "applesauce")
 4. prefix (n-gram) match ("apple" matches "green apples")

 - suffix/infix matches should  NOT returned ("apple" should not match "cranapple"),
 - if the prefix is on a word boundary like "apple " (note the space), don't suggest prefix matches like "applesauce"

### Other Changes
 - _DbSearchHistory_ doesn't return results sorted by search date or apply a result limit if a list of search is provided. This is because results will be returned to the client in order of relevance provided by the index.
 - The _SearchHistoryMetadata::search_ method returns a list of _SearchHistoryEntry_ objects instead of Strings, which has fields for the String itself as well as the id of the entry.

### Testing
 - _LuceneSearchHistoryIndexTest_ contains unit tests for indexing and search functionality. While search tests do not check the exact order of expected results (this is hard to define when sorting by relevance), they broadly check the ordering of match types (for example, edge matches should outrank n-gram matches).
 - _SearchHistoryTest_ has been updated with _testQuerySearchHistory_ and _testGetSearchHistory_ test methods.
 - _TestSearchHistory_ includes SOAP tests for _GetSearchHistoryRequest_ and _SearchSuggestRequest_

### Issues and notes
 - If no prefix is provided as part of _SearchSuggestRequest_, the response will be equivalent to that of _GetSearchHistoryResponse_, as this is the best guess as to what to suggest to the user.

 - It turns out that the `search_history` DB table introduced with ZIMBRA-93 currently doesn't store timestamps with sub-second resolution. This means that _GetSearchHistoryResponse_ cannot be guaranteed to return searches in chronological order for entires registered within <1 sec of each other. I do not think that this is a problem from the user's perspective, as this is outside the bounds of user behavior, but unfortunately it does require _GetSearchHistoryRequest_ SOAP tests to wait 1sec between each call. This should ideally be addressed in the future, but is not a priority right now.

 - As part of ZIMBRA-105, _LuceneSearchHistoryIndex_ will be rewritten to make use of the SOLR indexing backend, making some of this code obsolete. I chose to create this PR regardless, prior to the switch to SOLR, as it gives reviewers a clear picture of the search suggestion logic.


ilyusha:
Not sure why LuceneSearchHistoryIndexTest is failing here, it was passing for me. I'll look into this.

store/src/java/com/zimbra/cs/index/history/LuceneSearchHistoryIndex.java
```diff
@@ -0,0 +1,172 @@
...
+
+        Analyzer analyzer = new SearchHistoryQueryAnalyzer();
+        try {
+            Reader reader = new HalfwidthKanaVoicedMappingFilter(new StringReader(searchString));
+            TokenStream tokenStream = analyzer.tokenStream(LuceneFields.L_SEARCH_TERMS, reader);
+            CharTermAttribute termAttr = tokenStream.addAttribute(CharTermAttribute.class);
+            try {
+                tokenStream.reset();
+                while (tokenStream.incrementToken()) {
+                    termMatch.add(new TermQuery(new Term(LuceneFields.L_SEARCH_TERMS, termAttr.toString())), Occur.MUST);
```
fatkudu: Any value in only doing `termAttr.toString()` once?

store/src/java/com/zimbra/cs/index/history/LuceneSearchHistoryIndex.java
```diff
@@ -0,0 +1,172 @@
...
+        } catch (IOException e) {
+            ZimbraLog.search.error("unable to search search history for prefix '%s'", searchString, e);
+            return Collections.emptyList();
+        }
+    }
+
+    @Override
+    public void delete(Collection<Integer> ids) throws ServiceException {
+        List<Integer> idList = new ArrayList<Integer>(ids);
+        try {
```
fatkudu: For less verbose code, could take advantage of Indexer's closable nature?

        try (Indexer indexer = index.openIndexer()) {
            indexer.deleteDocument(idList, LuceneFields.L_SEARCH_ID);
        } catch (IOException e) {
            ZimbraLog.search.error("unable to delete %s search history docs from the index", ids.size(), e);
        }

store/src/java/com/zimbra/cs/index/history/ZimbraSearchHistoryFactory.java
```diff
@@ -11,17 +8,9 @@
...
     @Override
     public SearchHistoryIndex getIndex(Account acct) throws ServiceException {
-        String key = acct.getId();
-        SearchHistoryIndex index = indexCache.get(key);
-        if (index == null) {
-            index = new InMemorySearchHistoryIndex();
-            indexCache.put(key, index);
-        }
-        return index;
+        return new LuceneSearchHistoryIndex(acct);
```
fatkudu: Shouldn't this support different search backends?

store/src/java/com/zimbra/cs/index/history/LuceneSearchHistoryIndex.java
ilyusha: Ah yes, I forgot about this syntax.

store/src/java-test/com/zimbra/cs/index/history/InMemorySearchHistoryTest.java
```diff
@@ -162,31 +163,31 @@ public void testInMemorySearchHistoryMetadata() throws Exception {
         assertEquals("search1 count should be 1 within 1500ms", 1, md.getCount("search1", 1500));
 
         SearchHistoryMetadataParams params = new SearchHistoryMetadataParams();
-        List<String> results = md.search(params);
+        List<SearchHistoryEntry> results = md.search(params);
```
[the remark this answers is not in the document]
ilyusha: Maybe :) I'll take a look.

store/src/java/com/zimbra/cs/db/DbSearchHistory.java
```diff
@@ -197,11 +198,14 @@ public boolean isRegistered(DbConnection conn, String searchString) throws Servi
...
                 }
             } else {
-                sb.append(DebugConfig.disableMailboxGroups ? "" : "mailbox_id = ?");
+                sb.append(DebugConfig.disableMailboxGroups ? "" : "mailbox_id = ? ");
             }
-            sb.append(" ORDER BY last_search_date DESC ");
-            if (params.getNumResults() > 0 && Db.supports(Capability.LIMIT_CLAUSE)) {
-                sb.append(Db.getInstance().limit(params.getNumResults()));
+            if (!params.hasIds()) {
+                //no need to sort/limit otherwise, since results will be re-sorted according to relevance
```
grishick: I don't understand what's happening here. If the results will be reordered according to relevance and the comment is saying that there is no need to sort, why add "ORDER BY date"?

ilyusha: that might be a case of unclear commenting - it says "no need to sort/limit _otherwise_", referring to the case when the params _do_ have IDs. If we are in this _if_ block, then we did not have to go the index, meaning we need to sort chronologically.

store/src/java/com/zimbra/cs/index/history/LuceneSearchHistoryIndex.java
```diff
@@ -0,0 +1,172 @@
...
+            return Collections.emptyList();
+        }
+    }
+
+    @Override
+    public void delete(Collection<Integer> ids) throws ServiceException {
+        List<Integer> idList = new ArrayList<Integer>(ids);
+        try (Indexer indexer = index.openIndexer()) {
+            indexer.deleteDocument(idList, LuceneFields.L_SEARCH_ID);
+            indexer.close();
```
fatkudu: "try with resources" calls `close` anyway - so this line is redundant - might even be harmful.
See http://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html for a discussion of the advantages of the newish mechanism (for instance, exceptions from close are suppressed)
Note that the `add` method above uses similar logic - and if an exception was thrown before the close in that, the indexer wouldn't get closed.

ilyusha: Ah, I rushed this and forgot to remove the original `indexer.close()` line. Will fix.

PR #369, opened 2017-09-20
copied ZimbraHttpClientManager from perforce/main

grishick:
"main" branch in perforce already has an implementation of HTTP Client manager that uses the newer HTTP Components library. I simply copied the implementation from perforce.


grishick:
oops. Forgot that I already did that earlier this year.

PR #373, opened 2017-09-22
Bugfix/zimbra 113 event logging interface

varunrisbud:
Added the event logging interface and defined what an event would look like. This is the first ticket in the series of tickets to provide the analytics needed for machine learning.

- **EventLogger**: An abstract class to provide a framework for event logger.
- **Event**: Class that defines an event. Event.EventType enum defines the type of events. Event.EventContextField defines fields that can be added for event specific data.
- **InMemoryEventLogger**: In memory implementation of the event logger. Will be useful in testing when we add Solr event logger implementation.
- **InMemoryEventLoggerTest**: Unit tests the InMemoryEventLogger.
- **ZimbraLog**: Added new category "zimbra.event" to log events to a separate file which might be useful in debugging issues related to the event logger.

### DIFF
--- a/common/src/java/com/zimbra/common/util/ZimbraLog.java
+++ b/common/src/java/com/zimbra/common/util/ZimbraLog.java
@@ -461,6 +461,11 @@ public final class ZimbraLog {
     public static final Log ephemeral = LogFactory.getLog("zimbra.ephemeral");
 
     /**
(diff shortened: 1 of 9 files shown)


ilyusha:
I forgot to mention one thing on your original PR - the new package should probably just be `com/zimbra/cs/event`; no need to place it under the `index` hierarchy.

PR #384, opened 2017-09-29
ZCS-3093:error during tgz import results in endless loop and memory leak

sneha-patil-synacor:
Issue:
Truncated contacts tgz import results in endless loop and memory leak

Fix:
Handling EOFException thrown by  ArchiveFormatter.readArchiveEntry() which will throw the ServiceException and the loop will not continue.

Testing done:
Manual testing. Added Junit test.


store/src/java-test/com/zimbra/cs/service/ModifyPrefsTest.java
```diff
@@ -36,12 +36,10 @@
 import com.zimbra.common.util.L10nUtil;
 import com.zimbra.cs.account.Account;
 import com.zimbra.cs.account.Provisioning;
-import com.zimbra.cs.mailbox.MailItem;
```
shrikant-prasad: these changes have already been done in ZCS2948.

sneha-patil-synacor: Reverted

store/src/java-test/com/zimbra/cs/mailbox/ContactTest.java
```diff
@@ -293,4 +303,22 @@ public void testEncodeContact() throws Exception {
         ToXML.encodeContact(response, new ItemIdFormatter(), new OperationContext(acct), contact, true, null);
         Assert.assertEquals(response.getElement("cn").getElement("a").getText(), "Cert1149638887753217");
     }
+
+    @Test
+    public void testTruncatedContactsTgzImport() throws IOException {
+        File file = new File("src/java-test/Truncated.tgz");
```
ykomiyam: Truncated.tgz is missing; need to commit the file as well.

store/src/java/com/zimbra/cs/service/formatter/ArchiveFormatter.java
```diff
@@ -957,6 +958,8 @@ public void saveCallback(UserServletContext context, String contentType, Folder
                         }
                         try {
                             id = new ItemData(readArchiveEntry(ais, aie));
+                        } catch (EOFException e) {
+                            throw ServiceException.FAILURE("Error reading file", e);
```
ykomiyam: Is there any particular reason to catch not the IOException but only the EOFException?  The "readArchiveEntry" may throw some other subclass of IOException which would also cause some endless loop.

store/src/java-test/com/zimbra/cs/mailbox/ContactTest.java
sneha-patil-synacor: Done

store/src/java/com/zimbra/cs/service/formatter/ArchiveFormatter.java
sneha-patil-synacor: Done

PR #385, opened 2017-09-29
zcs-2352:Bug108145:Windows-unfriendly files in zip

ykomiyam:
File name with a Japanese diacritic sign generated by Mac
is garbled in zip format when "download all attachments" is
performed from the Windows.

[bug]
When an attachment file is created on Mac environment, and
the name of the attachment contains some "dakuten (full-phonetic-voiced)"
symbols, and the message with such attachment is received
and downloaded in zip format from Windows PC, then the file
name of the expanded attachment file gets garbled.

[root cause]
Mac OS represents the file name in Normalization Form D (NFD:
Canonical Decomposition), whereas Windows OS in Normalization
Form C (NFC: Canonical Decomposition, followed by Canonical
Composition). The attachment file name is specified in the
MIME header (Content-Disposition and/or Content-Type). When
the email is composed on the Mac OS, it is represented in the
NFD style.  When the recipient user downloads a set of attachment
files in zip format, the file name in the zip archive is taken
from the MIME header.  If the file name is specified in NFD,
the file name in the zip format is also composed in NFD.
Windows can't display the file name in NFD properly, and the file
name in the zip file is displayed garbled on the Windows OS.

Note: Because Mac OS can handle the file name in both NFD and NFC,
the files created and attached on the Windows OS can be opened
on the Mac OS without any garbled file name.

[fix]
If the file name is represented in NFD, it will be converted to NFC.

----
No unit test case.
Automation script will cover the output of the zip file.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/formatter/ArchiveFormatter.java
+++ b/store/src/java/com/zimbra/cs/service/formatter/ArchiveFormatter.java
@@ -26,6 +26,7 @@ import java.io.StringWriter;
 import java.nio.charset.Charset;
 import java.nio.charset.CharsetEncoder;
 import java.text.DateFormat;
+import java.text.Normalizer;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


fatkudu:
It would be nice to have some unit tests for this.

PR #389, opened 2017-10-02
Bug 98011: Enable UseGCLogFileRotation per default

mss:
Right now the file `/opt/zimbra/log/gc.log` grows slowly but steadily
and without bounds until the mailbox process is restarted.

This was added in [bug 85857](https://bugzilla.zimbra.com/show_bug.cgi?id=85857) but it looks like the option was accidently
kept to off.

This patch also increases the rotation limit from 4M to 10M to keep the
file from being rotated too often.

Since the file `/opt/zimbra/log/gc.log.0.current` is created as root, the
permissions have to be fixed up as they are done for `/opt/zimbra/log/gc.log`
without rotation. The required change can be found in zimbra/zm-launcher#2

### DIFF
--- a/common/src/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/java/com/zimbra/common/localconfig/LC.java
@@ -605,15 +605,15 @@ public final class LC {
             " -Dorg.apache.jasper.compiler.disablejsr199=true" +
             " -XX:+UseConcMarkSweepGC" +
             " -XX:SoftRefLRUPolicyMSPerMB=1" +
+            " -XX:-OmitStackTraceInFastThrow" +
             " -verbose:gc" +
             " -XX:+PrintGCDetails" +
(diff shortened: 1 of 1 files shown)


grishick:
merged into develop

mss:
Without the changes in zimbra/zm-launcher#2 this PR alone might cause issues (or not).

mss:
Hmmm… according to GitHub was this change not merged to `develop` yet.

gordyt:
@mss - Check out the commits on the `develop` branch.  The commit from your PR is in there.

mss:
@gordyt Yes, I googled a bit and it looks like the GH UI only recognizes merged commits if you use the merge button in the UI, not the command line tools. :roll_eyes:

fatkudu:
I've now merged zimbra/zm-launcher#2
Also opened a Jira ticket ZMS-721 to track this change as may be worth release noting.

PR #390, opened 2017-10-03
adjust Jaxb for TestDataSourceResponse, fix ZMailbox::testDataSource and add SOAP test

grishick:
Jaxb classes for TestDataSourceResponse do not match what the response looks like and therefore ZMailbox::testDataSource does not properly report errors when DataImport::test method fails.
There was no test that validated ZMailbox::testDataSource for negative behavior - add this test.

This change is part of ZCS-2976, because this problem blocks me from creating fully tested Data Source implementation for ZCS-2976

### DIFF
--- a/soap/src/java/com/zimbra/soap/mail/message/TestDataSourceResponse.java
+++ b/soap/src/java/com/zimbra/soap/mail/message/TestDataSourceResponse.java
@@ -17,53 +17,62 @@
 
 package com.zimbra.soap.mail.message;
 
-import com.google.common.base.Objects;
+import java.util.Collections;
+import java.util.List;
+
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
-import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlElement;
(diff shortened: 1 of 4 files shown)


soap/src/java/com/zimbra/soap/mail/message/TestDataSourceResponse.java
```diff
@@ -17,53 +17,62 @@
...
     }
 
-    public TestDataSourceResponse(boolean success) {
-        this.success = ZmBoolean.fromBool(success);
+    public TestDataSourceResponse addDataSource(TestDataSource dataSource) {
+        this.dataSources.add(dataSource);
+        return this;
     }
 
-    public void setError(String error) { this.error = error; }
```
fatkudu: Isn't this an API change?  Will no longer have `error` and `success`

grishick: no, it's a fix to the JAXB to make JAXB match the API. The API already behaves differently (and probably always had), but the JAXB class does not correspond to how API actually behaves.

fatkudu: Ok. That makes sense.

PR #391, opened 2017-10-03
ZCS-2349 - Ensure Remote IMAP sees fresh view of account

tmclane:
When the IMAP code is executed as a separate process the shared `LdapProvisioning` object is no longer available.

This causes an issue since the in-memory cached view of the Account is not updated automatically when the mailbox processes the `ChangePassword` request.
This update forces an account reload if and only if we are operating inside the remote imapd process.
In the case of the embedded imap server there is no reason to do the update since LdapProvisioning contains a `write-through` cache of account objects.


grishick:
This change may turn out to also address this issue: https://bugzilla.zimbra.com/show_bug.cgi?id=95983

tmclane:
I like the idea of just grabbing the relevant details needed.

tmclane:
Will be reopening this when the approach has been changed.

PR #396, opened 2017-10-09
Fix JAXB classes to allow external implementations of DataImport class

grishick:
This is a change similar to the previous one where I replaced "Undefined***" Jaxb subclasses with base ones. Undefined*** subclasses are equivalent to the base ones and JAXB correctly assigns types with base classes being present in the same list as subclasses.
I also added tests for all DataSource JAXB classes and fixed a typo in MdsConnectionType.java. The typo rendered ImapDataSource JAXBs broken, but that wasn't showing up in any tests, because all old DataSource code is using XML Element parsing instead of JAXB


fatkudu:
Could perhaps have a test to demonstrate that when creating JAXB from XML, you end up with the expected classes.  Also, would be nice to have some of the text in this pull request description in the commit comment text.

grishick:
@desouzas could you take a look to check if this may break anything for you?

desouzas:
This shouldn't affect anything looks fine.

PR #399, opened 2017-10-11
Feature/zcs 3177

tmclane:
Updates to ProxyConfGen to do the following:
- replace all usage of "zimbra*" options with the equivalent Provisioning.A_zimbra*
- allow strict server name enforcement based on `zimbraReverseProxyStrictServerName` LDAP attribute
- extract and make available the ip addresses used in `zimbraVirtualHostName` attributes of individual domains so they may be added to the strict server block.
- allow non-dns resolvable virtual hostnames (logged with warning)

Updates zimbra-attrs:
- add new attribute: zimbraReverseProxyStrictServerName
- getters / setters

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9609,4 +9609,11 @@ TODO: delete them permanently from here
   <desc>Information about the latest run of zmmigrateattrs. Includes the URL of the destination ephemeral store and the state of the migration (in progress, completed, failed)</desc>
 </attr>
 
+<attr id="3020" name="zimbraReverseProxyStrictServerNameEnabled" type="boolean" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="8.8.6" requiresRestart="nginxproxy">
+  <globalConfigValue>TRUE</globalConfigValue>
+  <desc>Configure the default server block in 'nginx.conf.web.https?.default.template' to return a default HTTP response
+    for all unconfigured host names. See also related attributes 'zimbraVirtualHostname' and 'zimbraVirtualIPAddress'.
+  </desc>
+</attr>
+
 </attrs>
--- a/store/src/java/com/zimbra/cs/util/ProxyConfGen.java
+++ b/store/src/java/com/zimbra/cs/util/ProxyConfGen.java
@@ -336,7 +336,7 @@ class ProxyConfVar
         int timeout = server.getIntAttr(
                 Provisioning.A_zimbraMailProxyReconnectTimeout, 60);
         String version = server.getAttr(Provisioning.A_zimbraServerVersion, "");
-        int maxFails = server.getIntAttr("zimbraMailProxyMaxFails", 1);
+        int maxFails = server.getIntAttr(Provisioning.A_zimbraMailProxyMaxFails, 1);
         if (maxFails != 1 && version != "") {
             return String.format("%s:%d fail_timeout=%ds max_fails=%d version=%s", serverName, serverPort,
                     timeout, maxFails, version);
@@ -491,7 +491,7 @@ class IPv6OnlyEnablerVar extends IPModeEnablerVar {
 class ProxyFairShmVar extends ProxyConfVar {
 
     public ProxyFairShmVar() {
-        super("upstream.fair.shm.size", "zimbraReverseProxyUpstreamFairShmSize", "",
+        super("upstream.fair.shm.size", Provisioning.A_zimbraReverseProxyUpstreamFairShmSize, "",
                 ProxyConfValueType.CUSTOM, ProxyConfOverride.CONFIG,
                 "Controls the 'upstream_fair_shm_size' configuration in the proxy configuration file: nginx.conf.web.template.");
     }
@@ -524,15 +524,15 @@ class ProxyFairShmVar extends ProxyConfVar {
 class Pop3GreetingVar extends ProxyConfVar {
 
     public Pop3GreetingVar() {
-        super("mail.pop3.greeting", "zimbraReverseProxyPop3ExposeVersionOnBanner", "",
+        super("mail.pop3.greeting", Provisioning.A_zimbraReverseProxyPop3ExposeVersionOnBanner, "",
                 ProxyConfValueType.STRING, ProxyConfOverride.CONFIG,
                 "Proxy IMAP banner message (contains build version if " +
-                "zimbraReverseProxyImapExposeVersionOnBanner is true)");
+                Provisioning.A_zimbraReverseProxyImapExposeVersionOnBanner + " is true)");
     }
 
     @Override
     public void update() {
-        if (serverSource.getBooleanAttr("zimbraReverseProxyPop3ExposeVersionOnBanner", false)) {
+        if (serverSource.getBooleanAttr(Provisioning.A_zimbraReverseProxyPop3ExposeVersionOnBanner, false)) {
             mValue = "+OK " + "Zimbra " + BuildInfo.VERSION + " POP3 ready";
         } else {
             mValue = "";
@@ -543,15 +543,15 @@ class Pop3GreetingVar extends ProxyConfVar {
 class ImapGreetingVar extends ProxyConfVar {
 
     public ImapGreetingVar() {
-        super("mail.imap.greeting", "zimbraReverseProxyImapExposeVersionOnBanner", "",
+        super("mail.imap.greeting", Provisioning.A_zimbraReverseProxyImapExposeVersionOnBanner, "",
                 ProxyConfValueType.STRING, ProxyConfOverride.CONFIG,
                 "Proxy IMAP banner message (contains build version if " +
-                "zimbraReverseProxyImapExposeVersionOnBanner is true)");
+                Provisioning.A_zimbraReverseProxyImapExposeVersionOnBanner + " is true)");
     }
 
     @Override
     public void update() {
-        if (serverSource.getBooleanAttr("zimbraReverseProxyImapExposeVersionOnBanner",false)) {
+        if (serverSource.getBooleanAttr(Provisioning.A_zimbraReverseProxyImapExposeVersionOnBanner, false)) {
             mValue = "* OK " + "Zimbra " + BuildInfo.VERSION + " IMAP4 ready";
         } else {
             mValue = "";
@@ -1121,7 +1121,7 @@ class ImapCapaVar extends ProxyConfVar {
 
         ArrayList<String> capabilities = new ArrayList<String>();
         String[] capabilityNames =
-            serverSource.getMultiAttr("zimbraReverseProxyImapEnabledCapability");
+            serverSource.getMultiAttr(Provisioning.A_zimbraReverseProxyImapEnabledCapability);
         for (String c:capabilityNames)
         {
             capabilities.add(c);
@@ -1170,7 +1170,7 @@ class Pop3CapaVar extends ProxyConfVar {
 
         ArrayList<String> capabilities = new ArrayList<String>();
         String[] capabilityNames = serverSource
-                .getMultiAttr("zimbraReverseProxyPop3EnabledCapability");
+                .getMultiAttr(Provisioning.A_zimbraReverseProxyPop3EnabledCapability);
         for (String c : capabilityNames) {
             capabilities.add(c);
         }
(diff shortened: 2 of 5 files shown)


tmclane:
This does not yet allow IP address based access when strict enforcement is enabled.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
```diff
@@ -1196,10 +1196,42 @@ public String format(Object o) {
...
+
+    @Override
+    public String format(Object o) throws ProxyConfException {
+       Set<String> addresses = (Set<String>)o;
+        if (addresses.size() == 0) {
+            return "";
+        }
+        StringBuilder sb = new StringBuilder();
+        for (String addr: addresses) {
+            sb.append(String.format("${web.strict.servername}    listen%s",
```
plobbes: I'd prefer an explicit space between listen and %s.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
```diff
@@ -2407,7 +2407,9 @@ private static void fillVarsWithDomainAttrs(DomainAttrItem item)
                 vip = InetAddress.getByName(item.virtualIPAddress);
             }
         } catch (UnknownHostException e) {
-            throw new ProxyConfException("virtual host name \"" + item.virtualHostname + "\" is not resolvable", e);
+            mLog.warn("virtual host name \"" + item.virtualHostname + "\" is not resolvable");
```
gordyt: A minor nit (but one that Gren has pointed out to me before).. Instead of appending multiple strings to compose the message, use something like this:

     String.format("virtual host name \"%s\" is not resolvable", item.virtualHostname)

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
tmclane: I did this on purpose since I wanted to right justify it by the correct number of spaces.
The second format string accomplishes this.

tmclane: This keeps the layout of the file as similar as possible.

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9609,4 +9609,9 @@ TODO: delete them permanently from here
   <desc>Information about the latest run of zmmigrateattrs. Includes the URL of the destination ephemeral store and the state of the migration (in progress, completed, failed)</desc>
 </attr>
 
+<attr id="3020" name="zimbraReverseProxyStrictServerName" type="boolean" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="8.8.1" requiresRestart="nginxproxy">
+  <globalConfigValue>FALSE</globalConfigValue>
+  <desc>Controls if the default server block in 'nginx.conf.web.https.default.template' is includedq returning a '400' status to all unknown host names.</desc>
```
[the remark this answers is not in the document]
tmclane: I could make an additional attribute that becomes the actual return code. 
What do you think of that?

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
```diff
@@ -1890,6 +1890,24 @@ public void update() {
...
+class WebStrictServerName extends WebEnablerVar {
+
+    public WebStrictServerName() {
+        super("web.strict.servername", "#",
+                "Indicates whether the default server block is generated returning a '400' response to all unknown hostnames");
+    }
+
+    @Override
+    public String format(Object o)  {
+        boolean strictServerName = serverSource.getBooleanAttr(Provisioning.A_zimbraReverseProxyStrictServerName,", false);
```
gordyt: Extra `"` in function call.

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9609,4 +9609,9 @@ TODO: delete them permanently from here
   <desc>Information about the latest run of zmmigrateattrs. Includes the URL of the destination ephemeral store and the state of the migration (in progress, completed, failed)</desc>
 </attr>
 
+<attr id="3020" name="zimbraReverseProxyStrictServerName" type="boolean" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="8.8.1" requiresRestart="nginxproxy">
```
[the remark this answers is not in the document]
tmclane: Will make this change.

store/conf/attrs/zimbra-attrs.xml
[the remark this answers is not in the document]
tmclane: I will fix the typo and update the wording.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
tmclane: Will make this change.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
plobbes: Unless I'm missing something, there is no explicit space.  There's the hope that we'll never have an address that is 26 characters in length, but it's easily possible to create a valid IPv6 address that is that long or even longer.
Potentially useful reference: https://stackoverflow.com/questions/166132/maximum-length-of-the-textual-representation-of-an-ipv6-address/
- I still vote for an explicit space :-)

store/conf/attrs/zimbra-attrs.xml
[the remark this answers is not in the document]
plobbes: I'd prefer to avoid creating yet another attribute/config var at this point, but others are welcome to chime in.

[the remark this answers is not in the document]
tmclane: I could remove the status code portion and just specify that it will enable the nginx configuration block that handles all unknown hostnames and returns a specified response.

tmclane:
If you have viewed this already the last commit removes a restriction that the count of entries for `zimbraVIrtualHostName` match the count for `zimbraVirtualIPAddress` for each domain.
This situation prevents multiple hostnames from sharing the same IP Address setting since the backend wouldn't actually set multiple identical entries for the ip address in my testing.

If you wish to create an IP Address based server block you merely add it as follows:

    zmprov md zcs-dev.test +zimbraVirtualHostName zcs.dev +zimbraVirtualHostName 1.2.3.4 +zimbraVirtualIPAddress 1.2.3.4

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9609,4 +9609,11 @@ TODO: delete them permanently from here
   <desc>Information about the latest run of zmmigrateattrs. Includes the URL of the destination ephemeral store and the state of the migration (in progress, completed, failed)</desc>
 </attr>
 
+<attr id="3020" name="zimbraReverseProxyStrictServerNameEnabled" type="boolean" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="8.8.1" requiresRestart="nginxproxy">
+  <globalConfigValue>FALSE</globalConfigValue>
```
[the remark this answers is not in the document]
tmclane: That was my intention. I was thinking that the installer was the one that set the value one way or the other and was taking that approach to setting the proper value.

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9609,4 +9609,11 @@ TODO: delete them permanently from here
   <desc>Information about the latest run of zmmigrateattrs. Includes the URL of the destination ephemeral store and the state of the migration (in progress, completed, failed)</desc>
 </attr>
 
+<attr id="3020" name="zimbraReverseProxyStrictServerNameEnabled" type="boolean" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="8.8.1" requiresRestart="nginxproxy">
```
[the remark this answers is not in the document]
tmclane: Fixed.

store/conf/attrs/zimbra-attrs.xml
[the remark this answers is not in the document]
tmclane: The actual setting either way is conducted here: https://github.com/Zimbra/zm-build/pull/64

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
```diff
@@ -1196,10 +1196,41 @@ public String format(Object o) {
...
+    public String format(Object o) throws ProxyConfException {
+       Set<String> addresses = (Set<String>)o;
+        if (addresses.size() == 0) {
+            return "";
+        }
+        StringBuilder sb = new StringBuilder();
+        for (String addr: addresses) {
+            sb.append(String.format("${web.strict.servername}    listen                  %s:${web.https.port} default_server;\n", addr));
+        }
+        sb.setLength(sb.length() - 1); //trim the last space
```
fatkudu: That's going to be a newline, not a space isn't it?

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
```diff
@@ -1922,6 +1977,7 @@ public void update() {
     private static Map<String, String> mVars = new HashMap<String, String>();
     private static Map<String, ProxyConfVar> mDomainConfVars = new HashMap<String, ProxyConfVar>();
     static List<DomainAttrItem> mDomainReverseProxyAttrs;
+    static Set<String> mListenAddresses = new HashSet<>();
```
fatkudu: Shouldn't the RHS be "new HashSet<String>()"? or "Sets.newHashSet()"?

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
```diff
@@ -2389,7 +2444,9 @@ private static void fillVarsWithDomainAttrs(DomainAttrItem item)
                 vip = InetAddress.getByName(item.virtualIPAddress);
             }
         } catch (UnknownHostException e) {
-            throw new ProxyConfException("virtual host name \"" + item.virtualHostname + "\" is not resolvable", e);
+            mLog.warn("virtual host name \"" + item.virtualHostname + "\" is not resolvable");
+            // TODO: Allow strict enforcement as an argument
+            //throw new ProxyConfException("virtual host name \"" + item.virtualHostname + "\" is not resolvable", e);
```
fatkudu: Does this TODO need to be addressed before committing this change?  Seems likely as no longer throwing an exception.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
tmclane: I have to be honest I copy and pasted this section and didn't actually pay attention to the comment.
It does appear to format properly in the file without adding an extra newline to the last item.

tmclane: I will update the comment accordingly.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
tmclane: This form compiles fine so I suppose it's up to whatever style we wish to follow.
The compiler is filling in the appropriate type to the right hand side.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
tmclane: I am unsure if we want to enforce this any longer.  The issue is if addresses are placed in say /etc/hosts they will not necessarily resolve on the proxy server but should still be handled by the 'Host' header value.
If the value fails to resolve and throws the exception then it prevents successful completion of the configuration generation process.

tmclane: I mean the client may expect them to be handled properly on the server even without dns resolution. 
In reality it seemed the DNS lookup was used merely to specify the 'listen' address for the server block.
Loosening this up doesn't seem to cause great issues because if you want to make it work the old way you simply make some DNS entries for the hostnames.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
grishick: Perhaps also need to specify the type in new HashSet<String>(); like it is specified in the preceding lines.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
```diff
@@ -1890,6 +1921,30 @@ public void update() {
...
+class WebStrictServerName extends WebEnablerVar {
+
+    public WebStrictServerName() {
+        super("web.strict.servername", "#",
+                "Indicates whether the default server block is generated returning a default HTTP response to all unknown hostnames");
+    }
+
+    @Override
+    public String format(Object o)  {
+        boolean strictServerName = isStrictEnforcementEnabled();
```
grishick: nit pick: since you are only calling this once, there seems to be no need for a local boolean variable

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
tmclane: @plobbes Didn't want me to add yet another LDAP attribute to control this resolution enforcement.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
tmclane: Done.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
tmclane: Updated to add a `-f` `--force-dns-resolution` command-line argument to enable the old behavior.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
```diff
@@ -336,7 +336,7 @@ String generateServerDirective(Server server, String serverName, String portName
         int timeout = server.getIntAttr(
                 Provisioning.A_zimbraMailProxyReconnectTimeout, 60);
         String version = server.getAttr(Provisioning.A_zimbraServerVersion, "");
-        int maxFails = server.getIntAttr("zimbraMailProxyMaxFails", 1);
+        int maxFails = server.getIntAttr(Provisioning.A_zimbraMailProxyMaxFails,1);
```
plobbes: nit: space after comma

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
```diff
@@ -543,15 +543,15 @@ public void update() {
...
                 ProxyConfValueType.STRING, ProxyConfOverride.CONFIG,
                 "Proxy IMAP banner message (contains build version if " +
-                "zimbraReverseProxyImapExposeVersionOnBanner is true)");
+                Provisioning.A_zimbraReverseProxyImapExposeVersionOnBanner + " is true)");
     }
 
     @Override
     public void update() {
-        if (serverSource.getBooleanAttr("zimbraReverseProxyImapExposeVersionOnBanner",false)) {
+        if (serverSource.getBooleanAttr(Provisioning.A_zimbraReverseProxyImapExposeVersionOnBanner,false)) {
```
plobbes: nit: space after comma

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
```diff
@@ -2752,6 +2813,11 @@ public static void updateDefaultDomainVars ()
         }
     }
 
+    public static void updateListenAddresses () throws  ProxyConfException
```
plobbes: nit: extra space after throws

tmclane:
@plobbes Thanks for pointing out the formatting mistakes!

PR #401, opened 2017-10-13
ZIMBRA-116: event processing engine and "sent" event logging

ilyusha:
This PR introduces the asynchronous event processing system, as well as "sent" event logging.
While ZIMBRA-116 was originally meant to only include logging of "sent" events, it ended up encompassing the event processing framework itself out of necessity.

### Event processing
Event processing now follows an observer pattern. When events are logged to the _EventLogger_ singleton, they are appended to a processing queue. This queue is drained by _EventNotifier_ threads. Each thread has a list of its own _EventLogHandler_ instances, which do the actual processing. As of this PR there is only an _InMemoryEventLogHandler_, but logfile and Solr handlers are in development.

Handler classes are created through factories implementing _EventLogHandler.Factory_. These factories must be registered by name with the _EventLogger_ class. Processing threads are instantiated with a multimap of factory names mapped to one or more configuration strings; one handler is then created for each configuration.

_EventLogger_ now has methods to startup/shutdown/restart the executor, as well as to shut down and wait for the queue to be drained. 

_EventLogger_ configuration is handled by an inner _ConfigProvider_ interface, which specifies the number of processing threads and a map of handler names and configurations. The primary implementation gets these values from LDAP, but unit tests inject their own mock classes.

#### New LDAP attributes
A new multivalued attribute _zimbraEventLoggingBackends_ controls which event log handlers are used. Each value must be of the form `name:configuration`; the name must correspond to the label of an _EventLogHandler.Factory_ registered with _EventLogger_. The configuration will be passed to the `Factory::createHandler` method. Duplicate name/configuration pairs are not allowed.

The integer attribute _zimbraEventLoggingNumThreads_ governs the number of event processing threads.

If either of these attributes are modified, the _EventLogger_ executor is restarted, which causes the newly created threads to pick up the new configuration.

### SENT event logging
When the SentMsg SOAP handler is invoked, a SENT event is dispatched for each recipient. This event contains the following data:
 - account ID
 - SENT type
 - timestamp
 - email of recipient
 - email of sender

### Testing
The following unit tests are provided:
 - _EventLoggerTest_ for testing EventLogHandler invocation and queue draining
 - _InMemoryEventLogHandlerTest_ for testing the in-memory log handler
 - _SendMsgTest.testSendMailEventLoggerTest()_ for testing that SendMsgRequest triggers SENT events.

It should be noted that as of this PR, tests in _SendMsgTest_ fail due to its reliance on a broken _EmbeddedSolrIndex_. However, this will be resolved a part of ZIMBRA-114, which includes changes to the IndexStore API that get around this issue. With those changes in place, all event-related unit tests pass.

### Update 10/25
Added a new inner _MessageCallback_ interface to _Mailbox_ as well as a  _MessageCallbackContext_ static class. _DeliveryOptions_ now has an optional _MessageCallbackContext_ field that tells the mailbox to execute a specific callback. If set, the callback is executed at the end of the _addMessageInternal_ method.

As of this PR, the only implementation is _SentMessageCallback_, which generates and logs "sent" events.

_MessageCallbackContext_ carries additional info that can be used by the callback's _execute_ method. As of right now this only includes an optional datasource ID value, but it will be expanded in future PRs.

All this is done to avoid various parts of the codebase having to explicitly call the EventLogger class. Instead, if a message addition should be accompanied by an event, the caller just has to set the appropriate _MessageCallbackContext_. 

Lastly, both _ArchiveFormatter_ and _MailItemImport_ classes have been updated to trigger _SentMessageCallback_ when importing messages, so that events are generated upon mailitem import.


store/src/java/com/zimbra/cs/event/logger/EventLogger.java
```diff
@@ -1,25 +1,299 @@
...
 
-    public static EventLogger getEventLogger() {
-        return eventLogger;
+    private Map<String, Collection<String>> getConfigMap() {
+        Map<String, Collection<String>> configMap = config.getHandlerConfig();
+        configMap.keySet().retainAll(factoryMap.keySet());
+        return configMap;
+    }
+
+    public void startupEventNotifierExecutor() {
```
tmclane: Stylistic question:  Why is this section spaced out so much more than the rest?

store/src/java/com/zimbra/cs/event/logger/EventLoggerUtils.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class EventLoggerUtils {
+    public static List<Event> getEventForEachRecipient(Event event) {
+        List<Event> events = new ArrayList<>();
+        String receiver = (String) event.getContextField(Event.EventContextField.RECEIVER);
+        String[] recipients = receiver.split(Event.MULTI_VALUE_SEPARATOR);
+        if(recipients.length > 1) {
```
tmclane: nit: space missing between 'if' and '("

store/src/java/com/zimbra/cs/event/logger/EventLogger.java
ilyusha: Varun and I collaborated on this file and I guess we have different spacing habits :)

store/src/java/com/zimbra/cs/event/logger/EventLogger.java
```diff
@@ -1,25 +1,299 @@
...
+        }
+
+        ZimbraLog.event.warn("Shutdown called for Event Notifier Executor; initiating shutdown sequence...");
+        executorService.shutdownNow();
+        try {
+            executorService.awaitTermination(30, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            executorService.shutdownNow();
+            Thread.currentThread().interrupt();
+        }
```
plobbes: nit: finally on same line as brace?

store/src/java/com/zimbra/cs/event/logger/EventLogger.java
```diff
@@ -1,25 +1,299 @@
...
+            executorService.execute(new EventNotifier(factoryMap, configMap));
+        }
+        executorServiceRunning.set(true);
+    }
+
+    /**
+     * This method shuts down all the event notifier threads. The event queue is left intact.
+     */
+    public void shutdownEventNotifierExecutor() {
+        if(!executorServiceRunning.get()) {
```
plobbes: nit: space after the if (before open parenthesis)

store/src/java/com/zimbra/cs/account/callback/EventLoggerCallback.java
```diff
@@ -0,0 +1,58 @@
...
+                    throw ServiceException.FAILURE(msg,  null);
+                }
+            }
+        }
+    }
+
+    @Override
+    public void postModify(CallbackContext context, String attrName, Entry entry) {
+        if (attrName.equalsIgnoreCase(Provisioning.A_zimbraEventLoggingBackends) ||
+                attrName.equalsIgnoreCase(Provisioning.A_zimbraEventLoggingNumThreads)) {
```
plobbes: nit: attrNames on both lines should line up (at least that's how I interpret our style guide)

store/src/java/com/zimbra/cs/event/logger/EventLogger.java
plobbes: It might make sense to update http://wiki.eng.zimbra.com/index.php/Zimbra_Server_Code_Style and/or refer to that to keep code style consistency (more nits noted below).  Consistency is good.

store/src/java/com/zimbra/cs/event/logger/EventLogger.java
```diff
@@ -1,25 +1,299 @@
...
+        public void run() {
+            try {
+                while (!Thread.currentThread().isInterrupted()) {
+                    consume(eventQueue);
+                }
+            } catch (InterruptedException e) {
+                ZimbraLog.event.debug("%s was interrupted, Shutting it down", Thread.currentThread().getName(), e);
+                Thread.currentThread().interrupt();
+            } finally {
+                if(drainQueueBeforeShutdown.get()) {
```
plobbes: another nit

store/src/java/com/zimbra/cs/event/logger/EventLogger.java
varunrisbud: Done, reduced the spacing.

store/src/java-test/com/zimbra/cs/service/mail/SendMsgTest.java
```diff
@@ -255,4 +263,60 @@ public void sendUpload() throws Exception {
...
+        Assert.assertNotNull(captor.getAllValues());
+        Assert.assertEquals(2, captor.getAllValues().size());
+
+        Assert.assertEquals(Event.EventType.SENT, captor.getAllValues().get(0).getEventType());
+        Assert.assertEquals(Event.EventType.SENT, captor.getAllValues().get(1).getEventType());
+
+        Assert.assertEquals("test@zimbra.com", captor.getAllValues().get(0).getContextField(Event.EventContextField.SENDER));
+        Assert.assertEquals("test@zimbra.com", captor.getAllValues().get(1).getContextField(Event.EventContextField.SENDER));
+
+        Assert.assertEquals("rcpt1@zimbra.com", captor.getAllValues().get(0).getContextField(Event.EventContextField.RECEIVER));
```
grishick: Just wondering if the ordering here is guaranteed. Will recipients always be logged in the same order as they are listed in the message header?

store/src/java-test/com/zimbra/cs/service/mail/SendMsgTest.java
```diff
@@ -255,4 +263,60 @@ public void sendUpload() throws Exception {
...
+        Mockito.doReturn(mockConfigMap.asMap()).when(mockConfigProvider).getHandlerConfig();
+
+        Mockito.doReturn(1).when(mockConfigProvider).getNumThreads(); //ensures sequential event processing
+
+        EventLogger.getEventLogger(mockConfigProvider).startupEventNotifierExecutor();
+
+        Account acct = Provisioning.getInstance().getAccountByName("test@zimbra.com");
+        Mailbox mbox = MailboxManager.getInstance().getMailboxByAccount(acct);
+
+        // first, add draft message
```
grishick: Why do you need to create a draft to test logging of SENT events?

ilyusha:
@grishick We have not tested SMTP - is there cmdline utility for doing that?

store/src/java-test/com/zimbra/cs/service/mail/SendMsgTest.java
ilyusha: They will be logged in the same order here because there is only 1 consumer thread:
`Mockito.doReturn(1).when(mockConfigProvider).getNumThreads()`

store/src/java-test/com/zimbra/cs/service/mail/SendMsgTest.java
ilyusha: Honestly, we had a hard time figuring out how to construct a proper SendMsgRequest otherwise, so this just copies send-from-draft logic from another test.

store/src/java/com/zimbra/cs/event/logger/EventLoggerUtils.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class EventLoggerUtils {
+    public static List<Event> getEventForEachRecipient(Event event) {
+        List<Event> events = new ArrayList<>();
+        String receiver = (String) event.getContextField(Event.EventContextField.RECEIVER);
+        String[] recipients = receiver.split(Event.MULTI_VALUE_SEPARATOR);
+        if (recipients.length > 1) {
```
fatkudu: Not >= 1?

store/src/java/com/zimbra/cs/account/callback/EventLoggerCallback.java
```diff
@@ -0,0 +1,58 @@
...
+import com.zimbra.cs.account.Server;
+import com.zimbra.cs.event.logger.EventLogger;
+
+public class EventLoggerCallback extends AttributeCallback {
+
+    @Override
+    public void preModify(CallbackContext context, String attrName, Object attrValue, Map attrsToModify, Entry entry) throws ServiceException {
+        if (attrName.equalsIgnoreCase(Provisioning.A_zimbraEventLoggingBackends)) {
+            String newBackend = (String) attrValue;
+            String[] tokens = newBackend.split(":", 2);
```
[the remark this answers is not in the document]
ilyusha: I actually fixed this logic as part of another bug that I don't have a PR for yet, but I will add that fix here.

store/src/java/com/zimbra/cs/account/callback/EventLoggerCallback.java
```diff
@@ -0,0 +1,58 @@
...
+            String newBackendName = tokens[0];
+            String newBackendConfig = tokens[1];
```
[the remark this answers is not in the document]
ilyusha: If not, it should return a meaningful error.

store/src/java/com/zimbra/cs/event/logger/EventLoggerUtils.java
ilyusha: The else block handles the case when there is only one recipient.

fatkudu: Ah.  Sorry, daftness.

store/src/java/com/zimbra/cs/account/callback/EventLoggerCallback.java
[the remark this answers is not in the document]
fatkudu: Ah - it looks like you've fixed the potential NPE here in the latest code.

PR #404, opened 2017-10-17
ZCS-3350 TestLog SOAP test

fatkudu:


store/src/java/com/zimbra/qa/unittest/TestLog.java
```diff
@@ -1,7 +1,7 @@
 /*
  * ***** BEGIN LICENSE BLOCK *****
  * Zimbra Collaboration Suite Server
- * Copyright (C) 2010, 2013, 2014, 2016 Synacor, Inc.
+ * Copyright (C) 2010, 2013, 2014, 2016, 2017 Synacor, Inc.
```
tmclane: Question: Should we start replacing these with an inclusive line?  Seems odd for example that 2011, 2012, and 2015 are missing.

fatkudu: I think a range is OK but so should the above at least in theory.  My understanding is that copyright should be asserted for any changes, so if no changes are made in a particular year, the copyright message doesn't need to have that year in it.  Copyright rules are a bit weird though, so I may be wrong.

PR #405, opened 2017-10-18
ZCS-3227:zmmetadump NPE

fatkudu:
* ZimbraPerf - ignore attempt to add a callback if initialization not
               done.
* MetadataDump - declare some stings in variables for re-use.
* TestMetadataDump - new tests which execute zmmetadump - hopefully will
  help catch any breakage in the tool early in future.
* ZimbraSuite - add `TestMetadataDump`

### DIFF
--- a/store/src/java/com/zimbra/cs/stats/ZimbraPerf.java
+++ b/store/src/java/com/zimbra/cs/stats/ZimbraPerf.java
@@ -432,6 +432,14 @@ public class ZimbraPerf {
      * during realtime stats collection.
      */
     public static void addStatsCallback(RealtimeStatsCallback callback) {
+        if (realtimeStats == null) {
+            ZimbraLog.perf.debug("Call to addStatsCallback when realtimeStats has not been initialized\n%s",
+                    ZimbraLog.getStackTrace(15));
+            /* This probably happens inside a commandline tool like zmmetadump where it doesn't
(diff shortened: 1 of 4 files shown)


store/src/java/com/zimbra/cs/stats/ZimbraPerf.java
```diff
@@ -432,6 +432,14 @@ public static void incrementPrepareCount() {
...
     public static void addStatsCallback(RealtimeStatsCallback callback) {
+        if (realtimeStats == null) {
+            ZimbraLog.perf.debug("Call to addStatsCallback when realtimeStats has not been initialized\n%s",
+                    ZimbraLog.getStackTrace(15));
+            /* This probably happens inside a commandline tool like zmmetadump where it doesn't
+             * make sense to mix in stats with those of the main Zimbra process.
+             */
+            return;
+        }
         realtimeStats.addCallback(callback);
```
tmclane: Is this the location of the NPE?

fatkudu: Yes.  The assumption of the new code is that this method has been called in error and can be safely ignored.  Could have tried to fix the root cause but quite likely that new code would trip up again as people findling with DB Pool initialization (in this case) may not think of all the use cases.

PR #409, opened 2017-10-20
Bugfix/zcs 2349 - Handle password updates in IMAPD transparently

tmclane:
This is a completely different approach to the issue of the Account's password being cached in the Remote IMAPD service (or any other service which is decoupled and running in a separate address space from the mailbox).

This changeset does the following:
- implements a function in LDAPProvisioning to allow the cached credentials for a specific Account to be refreshed (only including items related to authentication)
- calls this refreshing scheme ONLY in the case when the cached password fails to match.

Note: SoapProvisioning DOES NOT implement this new feature at this time and was modified to return a UnsupportedOperationException

### DIFF
--- a/store/src/java/com/zimbra/cs/account/Provisioning.java
+++ b/store/src/java/com/zimbra/cs/account/Provisioning.java
@@ -182,6 +182,15 @@ public abstract class Provisioning extends ZAttrProvisioning {
     public static final String MAIL_FORWARDREPLY_FORMAT_SAME = "same";
 
     /**
+     * Updates the values of the following attributes in the provided account argument:
+     * - userPassword
+     * - zimbraAuthTokens
+     * - zimbraAuthTokenValidityValue
+     * @param account Account instance who's credentials are to be refreshed
+     */
+    public abstract void refreshUserCredentials(Account account) throws ServiceException;
+
+    /**
      * Possible values for zimbraMailMode and ZimbraReverseProxyMailMode. "mixed"
      * means web server should authenticate in HTTPS and redirect to HTTP (useful
      * if all clients are on the intranet and you want only do authentication in
--- a/store/src/java/com/zimbra/cs/account/auth/AuthMechanism.java
+++ b/store/src/java/com/zimbra/cs/account/auth/AuthMechanism.java
@@ -226,6 +226,14 @@ public abstract class AuthMechanism {
             super(authMech);
         }
 
+        protected boolean isEncodedPassword(String encodedPassword) {
+            return PasswordUtil.SSHA512.isSSHA512(encodedPassword) || PasswordUtil.SSHA.isSSHA(encodedPassword);
+        }
+
+        protected boolean isValidEncodedPassword(String encodedPassword, String password) {
+            return PasswordUtil.SSHA512.verifySSHA512(encodedPassword, password) || PasswordUtil.SSHA.verifySSHA(encodedPassword, password);
+        }
+
         @Override
         public boolean isZimbraAuth() {
             return true;
@@ -235,30 +243,32 @@ public abstract class AuthMechanism {
         public void doAuth(LdapProv prov, Domain domain, Account acct, String password,
                 Map<String, Object> authCtxt) throws ServiceException {
 
-            String encodedPassword = acct.getAttr(Provisioning.A_userPassword);
             if (AuthMechanism.doTwoFactorAuth(acct, password, authCtxt)) {
                 return;
             }
 
+            String encodedPassword = acct.getAttr(Provisioning.A_userPassword);
             if (encodedPassword == null) {
                 throw AuthFailedServiceException.AUTH_FAILED(acct.getName(),
                         namePassedIn(authCtxt), "missing "+Provisioning.A_userPassword);
             }
 
-            if (PasswordUtil.SSHA512.isSSHA512(encodedPassword)) {
-                if (PasswordUtil.SSHA512.verifySSHA512(encodedPassword, password)) {
-                    return; // good password, RETURN
-                }  else {
-                    throw AuthFailedServiceException.AUTH_FAILED(acct.getName(),
-                            namePassedIn(authCtxt), "invalid password");
-                }
-            } else if (PasswordUtil.SSHA.isSSHA(encodedPassword)) {
-                    if (PasswordUtil.SSHA.verifySSHA(encodedPassword, password)) {
-                        return; // good password, RETURN
-                    }  else {
+            if (isEncodedPassword(encodedPassword)) {
+                if (isValidEncodedPassword(encodedPassword, password)) {
+                    return;
+                } else {
+                    acct.refreshUserCredentials();
+                    String refreshedPassword = acct.getAttr(Provisioning.A_userPassword);
+                    if (!isEncodedPassword(refreshedPassword)) {
+                        doAuth(prov, domain, acct, password, authCtxt);
+                    }
+                    if (!isValidEncodedPassword(encodedPassword, refreshedPassword)) {
                         throw AuthFailedServiceException.AUTH_FAILED(acct.getName(),
                                 namePassedIn(authCtxt), "invalid password");
                     }
+                    return;
+                }
+
             } else if (acct instanceof LdapEntry) {
                 // not SSHA/SSHA512, authenticate to Zimbra LDAP
                 prov.zimbraLdapAuthenticate(acct, password, authCtxt);
@@ -437,4 +447,3 @@ public abstract class AuthMechanism {
 
     }
 }
-
--- a/store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
+++ b/store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
@@ -10721,4 +10721,22 @@ public class LdapProvisioning extends LdapProv implements CacheAwareProvisioning
         return useCache;
     }
 
+    @Override
+    public void refreshUserCredentials(Account account) throws ServiceException {
+        ZLdapContext zlc = null;
+
+        try {
+            zlc = LdapClient.getContext(LdapServerType.REPLICA, LdapUsage.GET_ENTRY);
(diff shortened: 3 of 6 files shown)


store/src/java/com/zimbra/cs/account/Provisioning.java
```diff
@@ -188,6 +188,14 @@
     public static final String MAIL_FORWARDREPLY_FORMAT_SAME = "same";
 
     /**
+     * Updates the cached values on an Account:
```
fatkudu: At first, I thought this list was a mistaken list of method arguments.
I suggest a clearer first sentence would be "Updates the values of the following attributes in the provided account argument:"

store/src/java/com/zimbra/cs/account/auth/AuthMechanism.java
```diff
@@ -249,13 +249,21 @@ public void doAuth(LdapProv prov, Domain domain, Account acct, String password,
                 if (PasswordUtil.SSHA512.verifySSHA512(encodedPassword, password)) {
                     return; // good password, RETURN
                 }  else {
+                    acct.refreshUserCredentials();
+                    if (PasswordUtil.SSHA512.verifySSHA512(acct.getAttr(Provisioning.A_userPassword), password)) {
```
fatkudu: Can the password change whether is is SSHA or SSHA512?  If so, we should be able to cope with that.

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
```diff
@@ -10721,4 +10721,26 @@ public boolean isCacheEnabled() {
...
+
+        try {
+            zlc = LdapClient.getContext(LdapServerType.MASTER, LdapUsage.GET_ENTRY);
+            String[] returnAttrs = {"userPassword", "zimbraAuthTokens", "zimbraAuthTokenValidityValue"};
+            ZAttributes attrs = helper.getAttributes(zlc, ((LdapEntry) account).getDN(), returnAttrs);
+            Map<String, Object> finalAttrs = account.getAttrs(false, false);
+            finalAttrs.putAll(attrs.getAttrs());
+            account.setAttrs(finalAttrs);
+            extendLifeInCacheOrFlush(account);  // Put this back into the cache
+        } catch (LdapException e) {
```
fatkudu: Can combine these 2 catch blocks - `catch (LdapException | AccountServiceException e)`

store/src/java/com/zimbra/cs/account/Provisioning.java
tmclane: Alright I will update this to make it more clear.

store/src/java/com/zimbra/cs/account/auth/AuthMechanism.java
tmclane: I don't think it can in between deciding it's of type SSHA512 and the user updating it in the background.

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
tmclane: roger that.

store/src/java/com/zimbra/cs/account/auth/AuthMechanism.java
fatkudu: Perhaps someone who knows more can say definitively.  My suspicion is that passwords generated by old versions of Zimbra might theoretically still be SSHA but new passwords would always be SSHA512.

tmclane: Ok.  Wouldn't the second case be more likely to run into this situation?
i.e. the old password was of type SSHA and the new updated password is of type SSHA512?

fatkudu: Yeah, I guess so.

plobbes: As Gren noted, this is off (and apparently has been for a while).  There's no no guarantee the new password shares the same encoding with the old one.  This applies for the SSHA512 and SSHA cases.  Heck, the new password could be PLAIN... sigh.  So, it feels like we've over optimized, and can fail in a less than ideal way.

For each of the else cases, we should probably just do ```acct.refreshUserCredentials();``` if we're optimistic that our cache is stale.  But, even then, perhaps we should fall through to authenticating to LDAP as per the non-optimized case and the next time a request comes in we'll have the updated cache.

If we're not optimistic, perhaps we don't bother doing a manual refresh of the cache, but just let it age out normally ~15m and just always fall through regardless on the case where we fail to positively match the hashes.

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
```diff
@@ -10721,4 +10721,22 @@ public boolean isCacheEnabled() {
         return useCache;
     }
 
+    @Override
+    public void refreshUserCredentials(Account account) throws ServiceException {
+        ZLdapContext zlc = null;
+
+        try {
+            zlc = LdapClient.getContext(LdapServerType.MASTER, LdapUsage.GET_ENTRY);
```
plobbes: It isn't clear we need to burden the master with this request.  If LDAP replication is working properly there should be very little delay between LDAP servers.  Generally speaking I don't think we're trying to force a conversation with the master unless we're doing writes.

tmclane: So you are saying I should swap the type of LdapServerType to what?
You are correct I don't care which LDAP server services the request I just know at this point that we likely have a stale password hash cached in the Provisioning instance (specifically in the imapd service in this case)

tmclane:
I'll reorganize the code so it can validate the type of the hash after the refresh.

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
plobbes: It appears REPLICA is the only other option, but I get the feeling we shouldn't need to make that decision here. Perhaps we should just call a helper?  Maybe something that uses refreshEntry()/reload() or similar to avoid only getting a partially updated user entry (and then extending the lifetime of that potentially partially updated object).  Doing a partial update and then extending the cache time could lead to less desirable behavior (e.g. change in COS or more, that is expected to be updated because they occurred around the same time).

store/src/java/com/zimbra/cs/account/auth/AuthMechanism.java
```diff
@@ -235,30 +243,32 @@ public boolean isZimbraAuth() {
...
-                        return; // good password, RETURN
-                    }  else {
+            if (isEncodedPassword(encodedPassword)) {
+                if (isValidEncodedPassword(encodedPassword, password)) {
+                    return;
+                } else {
+                    acct.refreshUserCredentials();
+                    String refreshedPassword = acct.getAttr(Provisioning.A_userPassword);
+                    if (!isEncodedPassword(refreshedPassword)) {
+                        doAuth(prov, domain, acct, password, authCtxt);
```
plobbes: This line bugs me.  Should we be returning here (e.g. ```return doAuth(...)```) or do we really intend to fall back through the code that follows after we make it through the subsequent call to doauth()?

tmclane: The old code did that very thing only throwing an exception in the case of a failure.  I was merely trying to maintain it's basic flow.

tmclane: The thought crossed my mind to add a 'return doAuth' there but was unsure if that was 'proper form' for Java considering it's a void return function.

store/src/java/com/zimbra/cs/account/soap/SoapProvisioning.java
```diff
@@ -2938,4 +2938,9 @@ public void autoProvControl(String action) throws ServiceException {
         return result;
     }
 
+    @Override
+    public void refreshUserCredentials(Account account) {
+        // FIXME: Currently no way to refresh required attributes over SOAP
```
fatkudu: Hmm.  Tricky.  I'd normally throw a NotImplementedException in these cases.  Tempted to think should do that if this is called things won't work as expected.

tmclane: Fair enough.  At least in this case we don't/can't use SoapProvisioning as a direct swap-in replacement for LdapProvisioning because of this and other 'features' such as the ephemeral data piece. 

Throwing an exception here would make this even more obvious since it would fail if it ever reached this point in the code during runtime.

Will make this update.

tmclane: Do you mean this one? sun.reflect.generics.reflectiveObjects.NotImplementedException;

I am getting a warning telling me it is an internal exception and shouldn't be used. 

     warning: NotImplementedException is internal proprietary API and may be removed in a future release

tmclane: Would adding a new code to ServiceException be the correct course of action?

grishick: new code for ServiceException is a good idea. As far as I can see, we throw generic exceptions in similar cases elsewhere. So, you could just throw  ServiceException.FAILURE("Currently no way to refresh required attributes over SOAP")

fatkudu: Actually - I meant UnsupportedOperationException which is a RuntimeException - that is used in some other places.
Personally, I think ServiceException is a bad idea as it it is treated as something we can deal with in a lot of cases whereas if we're calling this code we're doing something we shouldn't be doing and we need that shouted from the rooftops.

tmclane: Thank you for the update.  updated to throw UnsupportedOperationException

PR #410, opened 2017-10-24
ZCS-3269 Load Balancing Algorithm

fatkudu:


store/src/java/com/zimbra/cs/account/Provisioning.java
```diff
@@ -79,8 +79,8 @@
 
     // The public versions of TRUE and FALSE were moved to ProvisioningConstants.
     // These are used by ZAttr*.
-    static final String TRUE  = "TRUE";
-    static final String FALSE = "FALSE";
+    protected static final String TRUE  = "TRUE";
```
grishick: why not use 
ProvisioningConstants.TRUE and ProvisioningConstants.FALSE everywhere instead of these two?

tmclane: Honestly no idea.  I'll change it 👍

tmclane: Oops not my PR. Brain fart

fatkudu: Will push this in a commit soon.

PR #413, opened 2017-10-26
ZCS-2730:Using zm-client for internal rest call

sneha-patil-synacor:
Replacing the http client code with zm-client post call.
Removed httpmime jar dependency
Removed Junit, since we are using existing zm-client post call instead of explicitly making one.

Testing done: 
Manual testing + Automation


store/src/java/com/zimbra/cs/service/mail/RestoreContacts.java
```diff
@@ -68,85 +59,55 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
+                    zoptions.setTargetAccount(account.getId());
+                    zoptions.setTargetAccountBy(AccountBy.id);
+                    ZMailbox zmbx = ZMailbox.getMailbox(zoptions);
+                    try {
+                        InputStream is = new FileInputStream(file);
+                        zmbx.postRESTResource(CONTACT_RES_URL, is, true, file.length(), null,
+                            LC.httpclient_internal_connmgr_so_timeout.intValue());
+                    } catch (FileNotFoundException e) {
+                        throw ServiceException
+                            .INVALID_REQUEST("File not found: " + contactBackupFileName, e);
```
shrikant-prasad: better to use string constant for the string: "File not found".  same can be used in line 84.

store/src/java/com/zimbra/cs/service/mail/RestoreContacts.java
```diff
@@ -68,85 +59,55 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
+                    ZimbraAuthToken token = new ZimbraAuthToken(account);
+                    ZMailbox.Options zoptions = new ZMailbox.Options(token.toZAuthToken(),
+                        AccountUtil.getSoapUri(account));
+                    zoptions.setNoSession(true);
+                    zoptions.setTargetAccount(account.getId());
+                    zoptions.setTargetAccountBy(AccountBy.id);
+                    ZMailbox zmbx = ZMailbox.getMailbox(zoptions);
+                    try {
+                        InputStream is = new FileInputStream(file);
+                        zmbx.postRESTResource(CONTACT_RES_URL, is, true, file.length(), null,
```
shrikant-prasad: postRESTResource closes the is? if not should close the stream in finally block.

sneha-patil-synacor: Yes, postRESTResource closes input stream. Third parameter is set to true for that.

store/src/java/com/zimbra/cs/service/mail/RestoreContacts.java
sneha-patil-synacor: Done

sneha-patil-synacor:
Does restore overwrites all contact folders?
- The ContactRestoreRequest has one optional parameter 'resolve'. It defines 4 values - reset, ignore, modify and replace. Whether to overwrite contact or not will be based on this parameter. By default its reset.

Does it create the folder if not present?
Yes

Does is backup shared contact folders or not?
No

PR #414, opened 2017-10-26
ZIMBRA-121: File event logger

ilyusha:
This is the default event logger that serializes events into a logfile. Events are logged to `opt/zimbra/log/event.log` with the same rollover configuration as the other logs.

This is done using a new `eventlog` logger category. This is distinct from the existing `event` category, which is used for general logging of the event system. 

The _EventFormatter_ class is used for serializing/deserializing an event object. The deserialization functionality will be used in the future for re-indexing events in solr.

### DIFF
--- a/store-conf/conf/log4j.properties.production
+++ b/store-conf/conf/log4j.properties.production
@@ -103,6 +103,16 @@ log4j.appender.EWS.File=/opt/zimbra/log/ews.log
 log4j.appender.EWS.layout=com.zimbra.common.util.ZimbraPatternLayout
 log4j.appender.EWS.layout.ConversionPattern=%d %-5p [%t] [%z] %c{1} - %m%n
 
+#Appender EVENTS writes to the file "event.log".
+log4j.appender.EVENTS=org.apache.log4j.rolling.RollingFileAppender
+log4j.appender.EVENTS.RollingPolicy=org.apache.log4j.rolling.TimeBasedRollingPolicy
+log4j.appender.EVENTS.RollingPolicy.FileNamePattern=/opt/zimbra/log/event.log.%d{yyyy-MM-dd}
+log4j.appender.EVENTS.File=/opt/zimbra/log/event.log
+log4j.appender.EVENTS.layout=com.zimbra.common.util.ZimbraPatternLayout
+#The conversion pattern is simpler because event log entries are already serialized in a
+#meaningful way, and can be deserialized back into Event objects.
+log4j.appender.EVENTS.layout.ConversionPattern=%m%n
+
 # HttpMethodBase spews out too many WARN on the badly formatted cookies.
 log4j.logger.org.apache.commons.httpclient.HttpMethodBase=ERROR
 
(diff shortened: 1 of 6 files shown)


store/src/java/com/zimbra/cs/event/logger/FileEventLogHandler.java
```diff
@@ -0,0 +1,140 @@
...
+        private static final String KEY_TIMESTAMP = "timestamp";
+        private static final String KEY_ACCT_ID = "id";
+
+        @Override
+        public String toLogString(Event event) {
+            String accountId = event.getAccountId();
+            String type = event.getEventType().toString();
+            String timestamp = String.valueOf(event.getTimestamp());
+            ObjectMapper mapper = new ObjectMapper();
+            Map<String, Object> eventMap = new HashMap<>();
```
fatkudu: Thanks for teaching me something new!  I haven't noticed the diamond operator in `new HashMap<>()` before where `<>` on the RHS is a shorthand for what is between `<` and `>` on the LHS.  I think I have a new preferred way of initializing where generics are involved (-: !

store-conf/conf/log4j.properties.production
```diff
@@ -103,6 +103,14 @@ log4j.appender.EWS.File=/opt/zimbra/log/ews.log
 log4j.appender.EWS.layout=com.zimbra.common.util.ZimbraPatternLayout
 log4j.appender.EWS.layout.ConversionPattern=%d %-5p [%t] [%z] %c{1} - %m%n
 
+#Appender EVENTS writes to the file "event.log".
+log4j.appender.EVENTS=org.apache.log4j.rolling.RollingFileAppender
+log4j.appender.EVENTS.RollingPolicy=org.apache.log4j.rolling.TimeBasedRollingPolicy
+log4j.appender.EVENTS.RollingPolicy.FileNamePattern=/opt/zimbra/log/event.log.%d{yyyy-MM-dd}
+log4j.appender.EVENTS.File=/opt/zimbra/log/event.log
+log4j.appender.EVENTS.layout=com.zimbra.common.util.ZimbraPatternLayout
+log4j.appender.EVENTS.layout.ConversionPattern=%m%n
```
fatkudu: Please add a comment explaining why this conversion pattern differs from the ones for most other appenders.
If nothing else, may save bad future copy/pastes.

store/src/java/com/zimbra/cs/event/logger/FileEventLogHandler.java
tmclane: FYI: I did this in one of my PRs and was told not to at that time.  Do we have consensus then to move to this style for future updates?

tmclane: I certainly like this style better!

fatkudu: I suspect it may have been me who complained - I probably didn't see the "<>" and assumed it was absent.

store-conf/conf/log4j.properties.production
ilyusha: Good idea. I changed the conversion pattern to simplify the log output and to potentially make it easier to parse the logs back into Event objects in the future.

ilyusha:
@fatkudu the solr error is thrown by the disfunctional EmbeddedSolrIndex, which currently fails when initIndex() is called. That said, the PR right after this one removes initIndex() from the IndexStore interface, which resolves these test errors.

store/src/java/com/zimbra/cs/event/logger/FileEventLogHandler.java
ilyusha: I actually picked that up from @varunrisbud recently - I wasn't aware of that syntax either. Definitely saves keystrokes!

PR #415, opened 2017-10-27
Bugfix/zcs 2862

tmclane:
I believe that this particular synchronized keyword is superfluous and is at least one cause of the thread-lock that was seen in this case.

Nowhere else in the code is 'pendingRequest' actually used except in this function.
The variable is used to hold the Future and is only assigned once per asynchronous request.
It depends on a callback firing when the asynchronous SOAP request is issued to handle the response when it happens.

There are two main objects being locked:
- ImapListener instance
- ImapServerListener instance

This function is forcing synchronization on the entire object when each of them call each other respectively which causes each thread to lock one lock and attempt to acquire the lock in the other class.  Sometimes the timing works out and the first lock is released before the second one in either class is hit, and sometimes it isn't.  

I can't get this lock contention to happen on demand so far.  I am going to keep banging on it and seeing if refinement to my IMAP stress utility will eventually trigger it in a reliable way.

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/ImapServerListener.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapServerListener.java
@@ -347,8 +347,8 @@ public class ImapServerListener {
     }
 
     private void cancelPendingRequest() {
-        ZimbraLog.imap.debug("Canceling pending AdminWaitSetRequest for waitset %s. Sequence %s", wsID, lastSequence.toString());
         if(pendingRequest != null && !(pendingRequest.isCancelled() || pendingRequest.isDone())) {
+            ZimbraLog.imap.debug("Canceling pending AdminWaitSetRequest for waitset %s. Sequence %s", wsID, lastSequence.toString());
             pendingRequest.cancel(true);
             pendingRequest = null;
         }
--- a/store/src/java/com/zimbra/cs/imap/ImapServerListener.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapServerListener.java
@@ -346,10 +346,11 @@ public class ImapServerListener {
         return wsID;
     }
 
-    private synchronized void cancelPendingRequest() {
+    private void cancelPendingRequest() {
         ZimbraLog.imap.debug("Canceling pending AdminWaitSetRequest for waitset %s. Sequence %s", wsID, lastSequence.toString());
         if(pendingRequest != null && !(pendingRequest.isCancelled() || pendingRequest.isDone())) {
             pendingRequest.cancel(true);
+            pendingRequest = null;
         }
     }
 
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/imap/ImapServerListener.java
```diff
@@ -346,10 +346,11 @@ public String getWSId() {
...
     }
 
-    private synchronized void cancelPendingRequest() {
-        ZimbraLog.imap.debug("Canceling pending AdminWaitSetRequest for waitset %s. Sequence %s", wsID, lastSequence.toString());
-        if(pendingRequest != null && !(pendingRequest.isCancelled() || pendingRequest.isDone())) {
+    private void cancelPendingRequest() {
+        if (pendingRequest != null && !(pendingRequest.isCancelled() || pendingRequest.isDone())) {
+            ZimbraLog.imap.debug("Canceling pending AdminWaitSetRequest for waitset %s. Sequence %s", wsID, lastSequence.toString());
             pendingRequest.cancel(true);
+            pendingRequest = null;
```
fatkudu: Is it possible for cancelPendingRequest to be called from more than one thread?  If it is, there are possibilities for NPEs here.  Having said that, can probably fix that by assigning it to another variable or guard it with an atomic perhaps.

tmclane: The only time i EVER see it used is when it is assigned and that usage IS NOT protected.
Does it need to be?

tmclane: Well the lock on this class itself is called from more than one thread.  This is the lock that was deadlocked in the crash dump.  

Would you be happy with the 'synchronized' removal if I remove the assignment?
This would prevent any NPEs since this function is the only thing that actually operates on the 'pendingRequest' variable directly.

fatkudu: @grishick added the synchronized keyword in change  14f483194e72b7edf2d5557c1bca4909519ea67f
Perhaps he would be best to explain the impact?

grishick: I see how this synchronization is a problem. The reason for having the keyword there is to avoid trying to double-cancel the same Future. If you are removing the keyword, it probably makes sense to surround pendingRequest.cancel(true); with try-catch block and swallow the exception when Future has already been cancelled.

tmclane: It appears based on the documentation that it simply returns 'false' in the case that it is already canceled.  

https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html#cancel(boolean)

Does reality not match the documentation?

tmclane: I'll test it.

PR #417, opened 2017-10-30
ZCS-3424: change default value of zimbraFeatureContactBackupFrequency to 0

shrikant-prasad:
change default value of zimbraFeatureContactBackupFrequency to 0, to disable contact backup feature by default.
Also, changed the logger in ContactBackupFeature class to be in sync with logger used everywhere for this feature.


shrikant-prasad:
also changed default value of zimbraFeatureContactBackupEnabled to False.

PR #419, opened 2017-11-01
ZIMBRA-114: Solr Event Backend

ilyusha:
Functionality for indexing events in `SolrCloud`, along with other changes in the solr / event system

#### Solr Event Logger
Logging events in solr is based on an abstract _BatchingEventLogger_. This class batches events on a per-account basis using an expiring guava cache. When a cache entry is removed, either due to a cache timeout or due to the max batch size being reached, the _BatchedEventCallback::execute_ method is called. Submitting events to Solr is done via _SolrEventCallback_, which converts the list of batched events to a solr update document and indexing it.

#### Lazy Solr Core Instantiation
Previously, every Solr call first performed a check to see if the core is initialized using the _IndexStore::indexExists_, and created the core with _IndexStore::initIndex_ if necessary. This was inefficient, as this requires a network call and will return `true` for all but few cases. This PR removes those two _IndexStore_ methods, replacing them with a new generic _RetryUtil_ class that provides the ability to execute a command, and upon catching a known error, execute some other code prior to retrying the command again. The methods _SolrUtils::executeRequestWithRetry_ and  _SolrUtils::executeCloudRequestWithRetry_ uses this to create a core if the initial request fails.

#### SolrRequestHelper classes
In anticipation of multiple NextGen search features being built with Solr, two new helper classes are introduced. 
 - _SolrRequestHelper_ is an abstract class that has methods to execute requests, delete indexes, and perform other common operations; subclasses are _SolrCloudHelper_ and _StandaloneSolrHelper_. Some of its methods are delegated to _SolrUtils_.
 - _SolrCollectionLocator_ is an abstract class used by _SolrRequestHelper_ to allow data to be indexed in a single joint index (via _JointCollectionLocator_) or in multiple account-level indexes (via _AccountCollectionLocator_). Load testing and further research will determine which configuration is better for which solr-backed feature.

#### Data Source Handling
Events for messages from data sources now have the data source ID added as a field. This allows us to delete events from the solr backend if the underlying data source is deleted (see below).

#### EventStore
_EventStore_ is a new interface that allows querying and manipulation of event data. This is distinct from _EventLogger_, which strictly deals with writing events. The only methods currently in _EventStore_ are for deleting events upon account or data source deletion, but will be updated with query methods in the near future.

#### Event Updates
_EventType_ enum values now have to be instantiated with two values. 
The first is a _UniqueOn_ value, which is the scope in which the event is unique. For example, only one SENT event can exist per message, so the value is _UniqueOn.MESSAGE_. While normal operation should not result in duplicate events,  redo operations and the upcoming ability to re-index events from the file log may make this necessary in the future.  _SolrEventDocument_ uses the _UniqueOn_ value to determine how to construct the document ID; the solr update document is indexed using the `skipexisiting` update processor chain, which silently ignores documents for which the ID already exists in the index.
The second (optional) _EventType_ attribute marks the type as "internal". An internal event represents an action that affects other events - for example, the new event types DELETE_ACCOUNT and DELETE_DATASOURCE are internal because deleting an account/datasource results in events being deleted from store. When we implement event re-indexing from the file log, these events will be used to delete newly-reindexed events.
Internal events can only be indexed by EventLogHandlers for which the new interface method _acceptsInternalEvents_ is _true_, which is currently only the _FileEventLogHandler_.

#### New LDAP Attributes
 - `zimbraEventBatchMaxSize` determines how many events are batched per account before they are indexed in solr.
 - `zimbraEventBatchLifetime` determines how long an event batch can remain unmodified before being flushed to solr.
 - `zimbraEventSolrIndexType` determines whether events are indexed in a single joint index for all accounts or in individual account-level indexes. The two possible values are "account" or "combined".
 - `zimbraEventLoggingEnabled` determines whether the event logging system is enabled for the server
 - `zimbraEventBackendURL` specifies the URL of the _EventStore_ backend to be used for querying event data. Since _SolrEventStore_ is the only current implementation, this should match the `solrcloud` value of `zimbraEventLoggingBackends`.

#### Testing
The test class _TestSolrCloudEventStore_ contains tests for indexing events with _SolrEventCallback_ as well as deleting events by account and by datasource through _SolrEventStore_. There is a version of each test method for the joint and account-level index use case.

#### Update 11/13
 - Added an event logging backend and _EventStore_ implementation for standalone solr. While not all methods will be implemented due to standalone solr not supporting the streaming API,it will at least allow simple event-driven capabilities.


store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -10667,13 +10674,21 @@ public String getRecipient() {
...
+
         public void setDataSourceId(String dsId) {
             this.dsId = dsId;
         }
 
         public void setRecipient(String recipient) {
             this.recipient = recipient;
         }
+
+        public void setTimestamp(Long timestamp) {
```
grishick: is MessageCallbackContext:: setDataSourceId still being used?

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -10649,6 +10655,7 @@ public void setSavedSearchPromptStatus(OperationContext octxt, String searchStri
     public static class MessageCallbackContext {
```
grishick: Why is this class inside Mailbox.java? If there is no need to share class members, I would prefer to move int to it's own file, because Mailbox.java is enormous as it is.

ilyusha: Yeah good point. I'll move it into its own file.

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
ilyusha: As of this PR, yes. The next pull request will actually remove this in lieu of storing datasource IDs in message metadata.

ilyusha: I actually missed setting the datasource ID in the MailItemImport class.

PR #423, opened 2017-11-08
Ignore failing filter tests

grishick:
These tests fail intermittently on TravisCI, in our dockerized dev environment and on OSX.


grishick:
I updated this PR with approach suggested by @rupalid

PR #425, opened 2017-11-13
zcs-3585,3558,3559:Unit test DeleteHeaderTest failed

ykomiyam:
[Problem]
Unit test com.zimbra.cs.filter.DeleteHeaderTest failed.

[Root cause]
The unit test uses the MockBlob object to keep not only
the blob data, like other real-world Blob object, but also
its message data itself (the member parameter "content").
The MockBlob didn't copy the "content", so the old contents
remained.

[Fix]
Added the Blob.copy() method to perform a deep copy.
The child class of the MockBlob.copy() will copy the data
of the "content" as well.

[Unit test]
ant test-all --> PASS

### DIFF
--- a/build-common.xml
+++ b/build-common.xml
@@ -320,7 +320,6 @@
                     <exclude name="**/cs/filter/HeaderTest.java"/>
                     <exclude name="**/cs/filter/EnvelopeTest.java"/>
                     <exclude name="**/cs/mailbox/ContactAutoCompleteTest.java"/>
-                    <exclude name="**/cs/service/ExternalUserProvServletTest.java"/>
                 </fileset>
             </batchtest>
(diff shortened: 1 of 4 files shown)


build-common.xml
```diff
@@ -320,6 +320,7 @@
                     <exclude name="**/cs/filter/HeaderTest.java"/>
                     <exclude name="**/cs/filter/EnvelopeTest.java"/>
                     <exclude name="**/cs/mailbox/ContactAutoCompleteTest.java"/>
+                    <exclude name="**/cs/service/ExternalUserProvServletTest.java"/>
```
rupalid: Yasuko, can you please remove this.  I am sending the PR to fix the test

PR #428, opened 2017-11-15
ZIMBRA-112: SEEN/READ/REPLIED Events

ilyusha:
This PR introduces three new message event types: `SEEN`, `READ` and `REPLIED`. The purpose of these events is to enable calculation of useful contact-level metrics such as the email open rate and average reply time.

#### EventFlag mechanism
The event state is stored in message metadata using the new `EventFlag` enum. A new message is instantiated as `not_seen`; the new method `Message:advanceEventFlag` is used to advance the event flag through the other states. The events must be triggered in order: a `SEEN` event can only be triggered for a message that has an internal event flag of `not_seen`; a `READ` event can only be triggered for a `seen` event, etc. This event flag metadata field also insures that only one event of each category is logged per message.

#### Other New Metadata Fields
In addition to the EventFlag metadata field, several others have been added:
 - The boolean `sentByMe` is `true` for outgoing messages and false otherwise. This is used to ensure that these three events are only triggered for incoming messages.
 - If the message is created from a datasource, the datasource ID is now stored in metadata so that it can be added to the logged event.

#### Event triggers
The `SEEN` event is triggered in one of two ways:
 - In the `SearchRequest` SOAP handler, with the event being potentially triggered for each message returned to the client. This covers the case where the user loads the UI or selects a folder to view.
 - In `SoapSession`, when a created item is returned to the client as part of notification headers. While it is not guaranteed that this will result in the user actually _seeing_ the item at that time, it is the best approximation we have for the time being.

The `READ` and `REPLIED` events are easier. `READ` is triggered the first time that the `UNREAD` message flag is unset; `REPLIED` is triggered the first time that the `REPLIED` message flag is set.

#### UPDATE 11/20
In order to trigger SEEN events from the IMAP FETCH command, I added a new `MailboxStore::markMsgSeen` interface method. The `Mailbox` implementation simply calls `advanceEventFlag` on the specified message. The `ZMailbox` implementation invokes `MsgActionRequest` with the new "seen" action. `ItemActionHelper` has been updated to handle the new operation.
`ImapHandler::fetch` calls this new method after fetching the message from the MailboxStore.

It should be noted that if FETCH_MARK_READ is true, this will result in `SEEN` and `READ` events being logged back-to-back, which may skew event-driven data. To compensate, queries that calculate differences between `SEEN` and `READ` events may need to be made to ignore time deltas below a certain threshold.

### DIFF
--- a/store/src/java/com/zimbra/cs/event/Event.java
+++ b/store/src/java/com/zimbra/cs/event/Event.java
@@ -1,13 +1,20 @@
 package com.zimbra.cs.event;
 
 import javax.mail.Address;
+import javax.mail.MessagingException;
+import javax.mail.internet.MimeMessage;
 
 import com.google.common.base.Objects;
 import com.google.common.base.Objects.ToStringHelper;
 import com.google.common.base.Strings;
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.util.ZimbraLog;
+import com.zimbra.cs.mailbox.Message;
 import com.zimbra.cs.mime.ParsedAddress;
+import com.zimbra.cs.mime.ParsedMessage;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -30,6 +37,7 @@ public class Event {
         RECEIVED(UniqueOn.MSG_AND_SENDER),
         READ(UniqueOn.MESSAGE),
         SEEN(UniqueOn.MESSAGE),
+        REPLIED(UniqueOn.MESSAGE),
         DELETE_DATASOURCE(UniqueOn.DATASOURCE, true),
         DELETE_ACCOUNT(UniqueOn.ACCOUNT, true);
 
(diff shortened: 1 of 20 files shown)


ilyusha:
Not sure why _travic-ci/pr_ was failing on _EventLoggerTest_; it passes for me. _travis-ci/push_ failed on _SolrEventLoggerTest_, which uncovered an issue with that test that I resolved with the latest commit.

store/src/java/com/zimbra/cs/event/Event.java
```diff
@@ -157,19 +165,38 @@ public static Event generateSentEvent(String accountId, int messageId, String se
...
-            sentEvents.add(generateSentEvent(accountId, messageId, sender.toString(), address.toString(), dsId, timestamp));
+    public static List<Event> generateSentEvents(Message msg, Long timestamp) {
+        try {
+            String acctId = msg.getAccountId();
+            int msgId = msg.getId();
+            String dsId = msg.getDataSourceId();
+            ParsedMessage pm = msg.getParsedMessage();
+            MimeMessage mm = pm.getMimeMessage();
+            Address[] recipients = mm.getAllRecipients();
+            Address sender = mm.getFrom()[0];
```
fatkudu: Need to handle mm.getFrom() returning null:

From https://docs.oracle.com/javaee/7/api/javax/mail/internet/MimeMessage.html#getFrom--

    public Address[] getFrom()
                  throws MessagingException

    Returns the value of the RFC 822 "From" header fields. If this header field is absent, the "Sender" header field is used. If the "Sender" header field is also absent, null is returned.

Not sure about getAllRecipients() return - probably best to handle null return for that too.
May want to include test data to exercise this.

store/src/java/com/zimbra/cs/mailbox/Message.java
```diff
@@ -1290,6 +1301,47 @@ private void getInfoForAssociatedCalendarItem(Account acct, Invite cur, String m
...
+    public EventFlag getEventFlag() {
+        return eventFlag;
+    }
+
+    void setEventFlag(EventFlag eventFlag) {
+        this.eventFlag = eventFlag;
+        try {
+            saveMetadata();
+        } catch (ServiceException e) {
+            ZimbraLog.event.warn("unable to save metadata; duplicate events may be generated");
```
fatkudu: Worth including the exception?  Could make it dependant on say event.isDebugEnabled() if want to keep noise level down.

store/src/java/com/zimbra/cs/mailbox/Message.java
```diff
@@ -1290,6 +1301,47 @@ private void getInfoForAssociatedCalendarItem(Account acct, Invite cur, String m
...
+     * step in the progression, log the appropriate event and update to the current flag
+     */
+    public void advanceEventFlag(EventFlag nextEventFlag) {
+        if (!isSentByMe() && eventFlag.canAdvanceTo(nextEventFlag) && nextEventFlag != EventFlag.not_seen) {
+            Event event = Event.generateMsgEvent(this, nextEventFlag.getEventType());
+            EventLogger.getEventLogger().log(event);
+            try {
+                getMailbox().setMessageEventFlag(this, nextEventFlag);
+            } catch (ServiceException e) {
+                ZimbraLog.event.error("error advancing event flag of msg %d to %s", getId(), nextEventFlag.toString());
```
fatkudu: Log exception?

store/src/java/com/zimbra/cs/event/Event.java
ilyusha: Ah yes, good point.

ilyusha: Actually, probably event better to use `ParsedMessage.getSenderEmail()`

ilyusha:
I did not test commit 3f8db9c enough - I will have to redesign how IMAP FETCH triggers seen events.

grishick:
are you keeping feature/solr rebased on develop?

ilyusha:
@grishick I last rebased it on develop at the start of this sprint

tmclane:
+1

PR #435, opened 2017-11-22
ZCS-3676:Modify AuthRequest Implementation for JWT

shrikant-prasad:
1. added code to generate random string salt, and use this salt for signing jwt in combination with secret key from ldap. salt value is stored in ZM_JWT cookie value and this cookie is set in response of auth request.
2. added JTI claim and token validity value claim in jwt. removed auth token stored earlier in the claim.
3. removed ldap attributes for lifetime and secret key added earlier for JWT. re-used the auth token attributes.
4. added JWTCache to store Salt and expiry time of JWT with JTI as key.

Fixed some junit tests. added ignore to tests failing, they will be fixed when jwt validation logic is updated to use the salt.

ZM_JWT cookie max size validation will be added in ZCS-3689.


store/src/java/com/zimbra/cs/account/JWTCache.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.Map;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import com.zimbra.common.localconfig.LC;
+import com.zimbra.common.util.MapUtil;
+
+public class JWTCache {
+
+    private static final Map<String, ZimbraJWT> JWT_CACHE = MapUtil.newLruMap(LC.zimbra_authtoken_cache_size.intValue());
```
fatkudu: Could you use CacheBuilder - see https://github.com/google/guava/wiki/CachesExplained or even ConcurrentHashMap?  Suspect they have better concurrency handling (less locking)

store/src/java/com/zimbra/cs/service/ZimbraAuthProvider.java
```diff
@@ -125,8 +126,8 @@ protected AuthToken authToken(Element soapCtxt, Map engineCtxt, TokenType tokenT
             if (!StringUtil.isNullOrEmpty(jwt)) {
                 try {
                     Claims body = Auth.validateJWT(jwt);
-                    String encodedAuthToken = (String) body.get(Constants.AUTH_DATA_CLAIM);
-                    at = genAuthToken(encodedAuthToken);
+                    Account acct = Provisioning.getInstance().getAccountById(body.getSubject());
```
fatkudu: How will this code work if acct is null at this point (perhaps the account has just been deleted?)

store/src/java/com/zimbra/cs/account/JWTCache.java
shrikant-prasad: used the guava library cache.

store/src/java/com/zimbra/cs/service/ZimbraAuthProvider.java
shrikant-prasad: added null check for account.

store/src/java/com/zimbra/cs/account/ZimbraAuthToken.java
```diff
@@ -499,25 +500,28 @@ public void deRegister() throws AuthTokenException {
...
-        }
-
         if (isJWT()) {
             try {
                 Account acct = Provisioning.getInstance().get(AccountBy.id, accountId);
                 ZimbraLog.account.debug("auth: generating jwt token for account id: %s", accountId);
-                AuthTokenKey authkey = getCurrentKey();
-                Key key = new SecretKeySpec(authkey.getKey(), SignatureAlgorithm.HS512.getJcaName());
+                RandomStringGenerator generator = new RandomStringGenerator.Builder().withinRange('a', 'z').build();
+                salt = generator.generate(20);
```
rupalid: Do you want to seed the RandomGenerator with something like username or accountId to ensure that the salt is unique. Also can we extract 20 into a constant easier to read the code.

store/src/java/com/zimbra/cs/account/ZimbraAuthToken.java
```diff
@@ -33,8 +34,12 @@
 import org.apache.commons.httpclient.HttpMethod;
 import org.apache.commons.httpclient.HttpState;
 import org.apache.commons.httpclient.cookie.CookiePolicy;
+import org.apache.commons.rng.UniformRandomProvider;
```
Prashantsurana: Can you please confirm there is not any vulnerability reported against version of libraries we are using here?

PR #436, opened 2017-11-22
ZIMBRA-95: Related Contacts Functionality

ilyusha:
This PR introduces "Related Contacts" functionality to ZCS.

### Concepts
This implementation of contact affinity is based on co-occurrence, derived from event data stored in solr and calculated using the Solrcloud Streaming API. This has the benefit of not requiring a separate datastore or dedicated indexes, as the data already exists.

The basic premise is that contacts that co-occur on emails are related. There are various considerations (see "Affinity Scope" below), but the high-level algorithm is this:
1) Given one or more target contacts, find IDs of messages associated with these contacts
2) Find other contacts that were included on these messages, optionally constrained to the requested field (to/cc/bcc)
3) Sort by occurrence count

#### Streaming API
The steps above map neatly to Solr streaming queries.
- IDs of messages associated with the target contacts are fetched using the `search` stream.
- `merge` and `intersect` streams are used to combine the results of multiple `search` streams in cases when there is more than one target contact.
- The `gatherNodes` stream walks from these message IDs to other contacts included on these messages, excluding the targets themselves. Multiple instances of co-occurrence are aggregated in a `count` field.
- The `top` stream decorator lets us sort the results by count, outputting the top related contacts
- The custom `ContactDedupeTupleStream` stream lets us de-duplicate results on email addresses, for cases when the event address data is denormalized. For example, tuples for "ilya.raykin@synacor.com" and "Ilya Raykin <ilya.raykin@synacor.com>" would be merged.

For more information, see https://lucene.apache.org/solr/guide/6_6/streaming-expressions.html

### Affinity Scope 
While the concept of co-occurrence affinity is simple, the devil is in the details.  "Affinity Scope" is introduced to manage the context within which a contact may be related to one or more targets.

#### Incoming vs Outgoing affinity
There are two high-level ways to calculate co-occurrence affinity. 
The first is via outgoing messages. If I often email contacts _A_ and _B_, then _A_ and _B_ are related. This type of affinity can be calculated from the existing `SENT` event data, since these events contains message ID and recipient information.
The second is via incoming messages. The idea is that if I receive messages from _C_ that also include _A_ and _B_ as recipients, then _A_ and _B_ are related. This allows us to make recommendations for contacts that the user has never directly emailed, although the affinity may not be as strong as for outgoing messages.
The downside of incoming message affinity is that it requires a new event type, since we need a separate event for each recipient other than the user. The new `AFFINITY` event is introduced for this purpose. Since this increases the event index size, `AFFINITY` event generation can be turned off, which disables incoming message affinity calculations.

#### Exact Field Matching
The Related Contacts API allows the client to specify which field each target contact is in.
This would be used when composing an email, where each target contact is in one of the _to_, _cc_, or _bcc_ fields. It is then possible to constrain the initial search to only those messages where the contacts occur in the same fields. For example, if the user emails _A_ and cc's _B_ and _C_, this allows us to recommend _C_ _only if_ the user begins to compose and email with _A_ and _B_ in the "to" and "cc" field, respectively.
An affinity scope in which this is turned off would be agnostic of which fields the targets occur in. In the example above, _C_ would be related to _A_ and _B_ even if _A_ and _B_ are both in the "to" field.

#### Union vs Intersection
When calculating outgoing co-occurrence for multiple targets, we can either traverse from messages that _all_ of the targets appear in, or only _some_. The latter provides more results, and is likely to prioritize results which co-occur with more targets, but may have false positives. This is useful when composing an email to a new contact that also includes a group of related contacts.
For example a user often emails _A_, _B_, _C_, but never _X_. If he now writes an email to _X_ that includes _A_ and _B_, _C_ can still be recommended.

#### Incoming Affinity Through Target
The last option pertains only to incoming message affinity. Assume _A_ emails the user with _B_ and _C_ also addressed. If the user composes an email to _A_ and _B_, then _C_ is related _through_ _A_ and would be recommended provided that incoming affinity is enabled. However, if _B_ is the _only_ target, then we have the option of whether to still recommend _C_. This is the loosest form of co-occurrence: incoming message co-occurrence through a sender who is not one of the targets.

#### Currently Defined Scopes
With all that said, here are the affinity scopes that are currently defined, in order of decreasing precedence:

##### OUTGOING_EXACT_MATCH (scope 0)
This scope matches contacts that the user emails along with ALL of the targets, with the targets occurring in the same fields. For example, if the user emails _A_, and cc's _B_ and _C_, then this scope would treat _C_ as a related contact for the targets _to:A_, _cc:B_

##### OUTGOING_EXACT_MATCH_ANY_FIELD (scope 1)
This scope is like the one above, except it is agnostic of which fields the targets appear on in the source messages. With the above example, this scope would treat _C_ as a related contact for the targets _to:A_, _to:B_

##### OUTGOING_BROAD_MATCH (scope 2)
This scope loosens the restriction to match _any_ of the targets. With the above example, it would treat _C_ as a related contact for the targets _to:X_, _to:A_.

##### OUTGOING_BROAD_MATCH_ANY_FIELD (scope 3)
Like above, but agnostic of which fields the targets appear in. _C_ would be a related contact for the targets _to:X_, _cc:A_.

##### INCOMING_FROM_TARGET (scope 4)
This is the first of the two incoming message scopes. If the user receives emails from _A_ that also address _B_ and _C_, then this scope would consider _C_ related to the targets _A_ and _B_. Note that incoming message affinity does not currently take into account the target fields. Also, notice that at least two targets are necessary for this scope, as each target is treated as a potential sender with the other targets as recipients.

##### INCOMING_FROM_ANY_SENDER (scope 5)
Expands on the previous scope to allow for incoming affinity from ANY sender, not just a target. With the previous example, this scope would treat _C_ as related to the target _B_ simply because an external contact _A_ has emailed both of them.

#### Scope Expansion
The API does not actually let the client define the scope. Instead, the client simply requests a number of related contacts to return. The server begins at the narrow-most scope (OUTGOING_EXACT_MATCH) and runs the query. If the number of results at at scope is less than the requested number, the scope is expanded, with the previous results passed in as filters to avoid duplicates. This is repeated until the requested number of contacts is reached or exceeded. The scope at which each result is found is returned in the response as an integer, to allow the client to potentially delineate between different "levels" of affinity.

### Caching
We want to avoid re-running the same contact affinity calculation every time, since it is a non-trivial operation. To this end, a caching layer is included. Results for the same set of targets and for the same target field are cached for a specified time. The cache values include related contacts from all affinity scopes up to and including the scope that found sufficient results, which may exceed the initial number of requested results; the result set is trimmed to the requested number after a cache hit. This allows us to use the cache for a subsequent request that has a higher limit. If there is a cache hit but the requested number exceeds the number of cached related contacts, the query is re-run.
The cache implementation provided with this PR is a simple in-memory cache using Guava. It may be worthwhile to implement a persistent cache that can survive mailbox restarts, but I deemed that outside the scope of this ticket.

### API
The `GetRelatedContactsRequest`takes one or more `targetContact` element, which contains the email of the contact and an optional attribute specifying the field (to/cc/bcc) that contact is in. The request also has an optional `requestedField` attribute that specifies which field the related contacts are requested for. It also takes an optional `limit` attribute that specifies how many related contacts should be returned.

`GetRelatedContactsResponse` contains a list of `relatedContact` elements. Each element contains the email address of the related contact, as well as the personal part (if available) and affinity scope within which the contact was first found. The scope is an integer; the lower the number the more narrow the scope and the closer the relation is between the contact and the targets.

When used to recommend contacts while composing an email, the `requestedField` request attribute and  `field` attribute of target contacts should be set, representing the current set of recipients and the field where the user is typing. When used to find general related contacts, these fields can be left blank.

####  zmmailbox support
The `zmmailbox` CLI supports fetching related contacts. The command is `getRelatedContacts` or `grc`. One or more targets can be specified, optionally prefixed with a colon-separated field. The requested affinity field and limit can be specified with the `field:` and `limit:` options.

### LDAP attributes
Several new LDAP attributes have been added for this feature.
- `zimbraFeatureRelatedContactsEnabled`: whether this feature is enabled for an account/COS
- `zimbraRelatedContactsMinCooccurrenceCount`: the minimum number of co-occurrences that two contacts need in order to be considered related. Defaults to 1.
- `zimbraRelatedContactsMaxAge`: the time window within which co-occurrence is calculated; affinity outside of this range is "forgotten". Defaults to 1 year.
- `zimbraAffinityEventLoggingEnabled`: whether the server should log `AFFINITY` events that allow for incoming message affinity to be calculated.

### Examples
The following two examples are based on an import of my Synacor mailbox into my dev environment.

The first example answers the question "who should I CC on an email where Greg and Gordon are on the 'to' field?"
```
mbox test> grc to:greg.solovyev to:gordon.tillman field:cc
Related Contacts
-----------------------------------------------------
[scope=0] gren.elliot@synacor.com (Gren Elliot)
[scope=0] matthew.fonck@synacor.com (matthew fonck)
[scope=0] travis.mclane@synacor.com (Travis Mclane)
[scope=1] chinmay.pathak@synacor.com (chinmay pathak)
[scope=1] rohan.ambasta@synacor.com (Rohan Ambasta)
[scope=1] tmclane@gmail.com (Travis McLane)
[scope=1] sneha.patil@synacor.com (Sneha Patil)
[scope=1] rupali.desai@synacor.com (Rupali Desai)
[scope=1] sandesh.almeida@synacor.com (Sandesh Almeida)
[scope=1] prashant.surana@synacor.com (Prashant Surana)
```

And here is a more general query for "which contacts are related to Varun?" (my email address shows up because this query is run from a different account; normally it would be filtered out).
```
mbox test> grc varun.risbud                 
Related Contacts
---------------------------------------------------------
[scope=0] michael.medellin@synacor.com (michael medellin)
[scope=0] JMead@synacor.com (Jon Mead)
[scope=0] BWang@synacor.com (Bo Wang)
[scope=0] matthew.fonck@synacor.com (matthew fonck)
[scope=0] michaeltoutonghi@gmail.com (michaeltoutonghi)
[scope=0] john.eastman@synacor.com (John Eastman)
[scope=5] ilya.raykin@synacor.com (ilya raykin)
[scope=5] erynn.petersen@synacor.com (Erynn Petersen)
[scope=5] davide.baldo@zextras.com (Davide Baldo)
[scope=5] sdavi@synacor.com (Steve Davi)
```

#### UPDATE 11/28
As per Gren's recommendation, the `RelatedContact` inner class subclasses the `RelatedContactResult` SOAP class. This removes the need for the `GetRelatedContacts` SOAP handler to convert between the two classes.

---
This change is [Reviewable](https://reviewable.io/reviews/zimbra/zm-mailbox/436)

### DIFF
--- a/soap/src/java/com/zimbra/soap/JaxbUtil.java
+++ b/soap/src/java/com/zimbra/soap/JaxbUtil.java
@@ -898,6 +898,8 @@ public final class JaxbUtil {
             com.zimbra.soap.mail.message.GetPermissionResponse.class,
             com.zimbra.soap.mail.message.GetRecurRequest.class,
             com.zimbra.soap.mail.message.GetRecurResponse.class,
+            com.zimbra.soap.mail.message.GetRelatedContactsRequest.class,
+            com.zimbra.soap.mail.message.GetRelatedContactsResponse.class,
             com.zimbra.soap.mail.message.GetSearchFolderRequest.class,
             com.zimbra.soap.mail.message.GetSearchFolderResponse.class,
             com.zimbra.soap.mail.message.GetSearchHistoryRequest.class,
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9713,7 +9713,7 @@ TODO: delete them permanently from here
     <desc>Time that mailstore will wait for space to free up in the shared indexing queue. Increase this value if you are reindexing several large mailboxes simultaneously.</desc>
 </attr>
 
-<attr id="3038" name="zimbraSolrMaxShardsPerNode" type="integer" cardinality="single" optionalIn="globalConfig,server,alwaysOnCluster" flags="serverInherited,serverPreferAlwaysOn" since="8.8.6">
+<attr id="3038" name="zimbraSolrMaxShardsPerNode" type="integer" min="1" cardinality="single" optionalIn="globalConfig,server,alwaysOnCluster" flags="serverInherited,serverPreferAlwaysOn" since="8.8.6">
   <globalConfigValue>1</globalConfigValue>
   <desc>Limit for how many replicas of the same SOLR Collection are allowed to be created on each node.</desc>
 </attr>
@@ -9729,12 +9729,12 @@ TODO: delete them permanently from here
      </desc>
 </attr>
 
-<attr id="3040" name="zimbraEventLoggingNumThreads" type="integer" cardinality="single" optionalIn="globalConfig,server,alwaysOnCluster" flags="serverInherited,serverPreferAlwaysOn" callback="EventLoggerCallback" since="8.8.6">
+<attr id="3040" name="zimbraEventLoggingNumThreads" type="integer" min="1" cardinality="single" optionalIn="globalConfig,server,alwaysOnCluster" flags="serverInherited,serverPreferAlwaysOn" callback="EventLoggerCallback" since="8.8.6"> 
   <globalConfigValue>10</globalConfigValue>
   <desc>Number of consumer threads used to process the event logging queue</desc>
 </attr>
 
-<attr id="3041" name="zimbraEventBatchMaxSize" type="integer" cardinality="single" optionalIn="globalConfig,server,alwaysOnCluster" flags="serverInherited,serverPreferAlwaysOn" callback="EventLoggerCallback" since="8.8.6">
+<attr id="3041" name="zimbraEventBatchMaxSize" type="integer" min="1" cardinality="single" optionalIn="globalConfig,server,alwaysOnCluster" flags="serverInherited,serverPreferAlwaysOn" callback="EventLoggerCallback" since="8.8.6">
   <globalConfigValue>100</globalConfigValue>
   <desc>Event batch size used by batching EventLogHandlers such as "solrcloud". When this limit is reached, the batch is flushed.</desc>
 </attr>
@@ -9758,4 +9758,23 @@ TODO: delete them permanently from here
   <desc>URL of the event storage backend to be queried for event analytics</desc>
 </attr>
 
+<attr id="3046" name="zimbraContactAffinityEventLoggingEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInherited" callback="EventLoggerCallback" since="8.8.6">
+  <desc>If TRUE, incoming emails will cause an AFFINITY event to be logged for each recipient. This allows for more accurate "related contacts" results, at the cost of higher index usage</desc>
+  <defaultCOSValue>TRUE</defaultCOSValue>
+</attr>
+
+<attr id="3047" name="zimbraRelatedContactsMaxAge" type="duration" cardinality="single" optionalIn="account,cos" flags="accountInherited" since="8.8.6">
+  <desc>The time window for which related contacts are calculated; emails older than this will not affect the contact affinity calculation</desc>
+  <defaultCOSValue>365d</defaultCOSValue>
+</attr>
+
+<attr id="3048" name="zimbraRelatedContactsMinConcurrenceCount" type="integer" min="1" cardinality="single" optionalIn="account,cos" flags="accountInherited" since="8.8.6">
+  <desc>The number of times two contacts have to co-occur on an email in order to be considered related</desc>
+  <defaultCOSValue>1</defaultCOSValue>
+</attr>
+
+<attr id="3049" name="zimbraFeatureRelatedContactsEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInherited" since="8.8.6">
+  <desc>Whether the Related Contacts feature is enabled on the account/COS</desc>
+  <defaultCOSValue>TRUE</defaultCOSValue>
+</attr>
 </attrs>
--- a/store/src/java/com/zimbra/cs/contacts/ContactAffinityQuery.java
+++ b/store/src/java/com/zimbra/cs/contacts/ContactAffinityQuery.java
@@ -1,0 +1,397 @@
+package com.zimbra.cs.contacts;
+
+import java.io.IOException;
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.apache.lucene.index.Term;
+import org.apache.lucene.search.BooleanClause.Occur;
+import org.apache.lucene.search.BooleanQuery;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.TermQuery;
+import org.apache.lucene.search.TermRangeQuery;
+import org.apache.lucene.util.BytesRef;
+import org.apache.solr.client.solrj.io.ModelCache;
+import org.apache.solr.client.solrj.io.SolrClientCache;
+import org.apache.solr.client.solrj.io.Tuple;
+import org.apache.solr.client.solrj.io.comp.ComparatorOrder;
+import org.apache.solr.client.solrj.io.comp.FieldComparator;
+import org.apache.solr.client.solrj.io.comp.StreamComparator;
+import org.apache.solr.client.solrj.io.eq.FieldEqualitor;
+import org.apache.solr.client.solrj.io.graph.GatherNodesStream;
+import org.apache.solr.client.solrj.io.graph.Traversal.Scatter;
+import org.apache.solr.client.solrj.io.stream.CloudSolrStream;
+import org.apache.solr.client.solrj.io.stream.IntersectStream;
+import org.apache.solr.client.solrj.io.stream.MergeStream;
+import org.apache.solr.client.solrj.io.stream.RankStream;
+import org.apache.solr.client.solrj.io.stream.StreamContext;
+import org.apache.solr.client.solrj.io.stream.TupleStream;
+import org.apache.solr.client.solrj.io.stream.metrics.CountMetric;
+import org.apache.solr.client.solrj.io.stream.metrics.Metric;
+import org.apache.solr.client.solrj.util.ClientUtils;
+import org.apache.solr.common.params.MapSolrParams;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
+import com.google.common.collect.Sets;
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.cs.contacts.AffinityScope.MatchType;
+import com.zimbra.cs.contacts.RelatedContactsParams.AffinityTarget;
+import com.zimbra.cs.contacts.RelatedContactsParams.AffinityType;
+import com.zimbra.cs.contacts.RelatedContactsResults.RelatedContact;
+import com.zimbra.cs.event.Event.EventContextField;
+import com.zimbra.cs.event.Event.EventType;
+import com.zimbra.cs.event.SolrEventDocument;
+import com.zimbra.cs.index.LuceneFields;
+import com.zimbra.cs.index.solr.SolrCloudHelper;
+import com.zimbra.cs.mime.ParsedAddress;
+
+/**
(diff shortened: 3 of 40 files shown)


soap/src/java/com/zimbra/soap/JaxbUtil.java
```diff
@@ -1093,7 +1093,55 @@
...
+            //smime
+            com.zimbra.soap.mail.message.SendSecureMsgRequest.class,
+            com.zimbra.soap.mail.message.SendSecureMsgResponse.class,
+            com.zimbra.soap.account.message.GetSmimeCertificateInfoRequest.class,
+            com.zimbra.soap.account.message.GetSmimeCertificateInfoResponse.class,
+            com.zimbra.soap.account.message.SaveSmimeCertificateRequest.class,
+            com.zimbra.soap.account.message.SaveSmimeCertificateResponse.class,
+
+            //IMAP
+            com.zimbra.soap.mail.message.ListIMAPSubscriptionsRequest.class,
```
fatkudu: I recently sorted this list and removed duplicates.  I notice that you have duplicates in here - please remove them.

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9758,4 +9758,23 @@ TODO: delete them permanently from here
   <desc>URL of the event storage backend to be queried for event analytics</desc>
 </attr>
 
+<attr id="3046" name="zimbraAffinityEventLoggingEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInherited" callback="EventLoggerCallback" since="8.8.6">
```
fatkudu: Worth including the word "contact" in the name to make it clear what the affinity is to?

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9758,4 +9758,23 @@ TODO: delete them permanently from here
...
+  <desc>If TRUE, incoming emails will cause an AFFINITY event to be logged for each recipient. This allows for more accurate "related contacts" results, at the cost of higher index usage</desc>
+  <defaultCOSValue>TRUE</defaultCOSValue>
+</attr>
+
+<attr id="3047" name="zimbraRelatedContactsMaxAge" type="duration" cardinality="single" optionalIn="account,cos" flags="accountInherited" since="8.8.6">
+  <desc>The time window for which related contacts are calculated; emails older than this will not affect the contact affinity calculation</desc>
+  <defaultCOSValue>365d</defaultCOSValue>
+</attr>
+
+<attr id="3048" name="zimbraRelatedContactsMinCooccurrenceCount" type="integer" min="1" cardinality="single" optionalIn="account,cos" flags="accountInherited" since="8.8.6">
```
fatkudu: Typo - MonCoocc....  Suggest **zimbraRelatedContactsMinConcurrenceCount**

store/src/java/com/zimbra/cs/contacts/RelatedContactsResults.java
```diff
@@ -0,0 +1,97 @@
+package com.zimbra.cs.contacts;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import com.google.common.base.Joiner;
+import com.google.common.base.Objects;
+
+public class RelatedContactsResults {
```
fatkudu: Looking a bit like very similar classes in store and soap.  Is it possible to use just the SOAP ones or even extend them?

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -10703,7 +10703,15 @@ public void execute(Message msg, MessageCallbackContext ctxt) {
...
                 long timestamp = ctxt.getTimestamp() == null ? System.currentTimeMillis() : ctxt.getTimestamp();
-                EventLogger.getEventLogger().log(Event.generateReceivedEvent(msg, recipient, timestamp));
+                EventLogger logger = EventLogger.getEventLogger();
+                logger.log(Event.generateReceivedEvent(msg, recipient, timestamp));
+                try {
+                    if (getAccount().isAffinityEventLoggingEnabled()) {
+                        logger.log(Event.generateAffinityEvents(msg, recipient, timestamp));
+                    }
+                } catch (ServiceException e) {
+                    ZimbraLog.event.error("unable to determine whether AFFINITY event logging is enabled");
```
fatkudu: Include exception in logging?

soap/src/java/com/zimbra/soap/JaxbUtil.java
ilyusha: Hm not sure what happened here, maybe merge error?

store/conf/attrs/zimbra-attrs.xml
ilyusha: "Cooccurrence" is somewhat of an awkward word to type...

store/src/java/com/zimbra/cs/contacts/RelatedContactsResults.java
ilyusha: Yeah those classes pretty much do the same thing. I generally avoid using SOAP classes in store, but in this case it probably makes sense.

store/conf/attrs/zimbra-attrs.xml
fatkudu: Ah.  Ok, so that was intentional.

store/src/java/com/zimbra/cs/contacts/ContactAffinityQuery.java
```diff
@@ -0,0 +1,388 @@
...
+            throw ServiceException.FAILURE("error reading tuple from contact affinity TupleStream", e);
+        }
+    }
+
+    /**
+     * Compose and run a streaming expression to return related contacts for a given AffinityScope.
+     */
+    @VisibleForTesting
+    public RelatedContactsResults execute(AffinityScope scope) throws ServiceException {
+        try(TupleStream stream = buildAffinityStream(scope)) {
```
varunrisbud: Is the tupleStream being closed? I don't see that it is implementing the AutoClosable interface.

ilyusha: `TupleStream` implements `Closeable`, which extends `AutoCloseable`.

fatkudu:
Review status: 0 of 40 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[store/src/java/com/zimbra/cs/mailbox/Mailbox.java, line 10713 at r1](https://reviewable.io:443/reviews/zimbra/zm-mailbox/436#-L-6V6FT1ZkNIvxrxIdY-r1-10713:-L-6WZcN-IU6lAVC1Mcz:b-smyx3l) ([raw file](https://github.com/zimbra/zm-mailbox/blob/5202146dc14952f983cea2d66e0100888b57cce6/store/src/java/com/zimbra/cs/mailbox/Mailbox.java#L10713)):*
<details><summary><i>Previously, fatkudu (Gren Elliot) wrote…</i></summary><blockquote>

Include exception in logging?
</blockquote></details>

LGTM now

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zimbra/zm-mailbox/436)*

fatkudu:
Reviewed 28 of 40 files at r1, 12 of 12 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zimbra/zm-mailbox/436)*

fatkudu:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[store/conf/attrs/zimbra-attrs.xml, line 9771 at r1](https://reviewable.io:443/reviews/zimbra/zm-mailbox/436#-L-6V6FT1ZkNIvxrxId9-r1-9771:-L-6XgEOyg4Wisfx2OCK:bb6ywmx) ([raw file](https://github.com/zimbra/zm-mailbox/blob/5202146dc14952f983cea2d66e0100888b57cce6/store/conf/attrs/zimbra-attrs.xml#L9771)):*
<details><summary><i>Previously, fatkudu (Gren Elliot) wrote…</i></summary><blockquote>

Ah.  Ok, so that was intentional.
</blockquote></details>

LGTM

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zimbra/zm-mailbox/436)*

PR #439, opened 2017-11-27
ZCS-3557 ensure abbreviateAll is utilized

tmclane:
In the case of the code requesting abbreviation AND the count of the
folder not matching the length of the set; the id list WAS NOT being abbreviated.
This was seen in the field where a user was triggering a search on the server
sending a count of 10k UIDs (well exceeding the maximum of 600) to be included in the SQL query.
This was causing an excessively long computation in the MySQL subsystem.

This code is called "before" the code which tests whether or not the number of UIDs being considered is larger than the LARGEST_FOLDER_BATCH value. The 'search' string was then utilized even in that case which caused all of the entries to be included instead of using the abbreviation.


grishick:
@tmclane can you come up with a test that confirms bad (old) behavior and new (good) behavior?

store/src/java/com/zimbra/cs/imap/ImapSearch.java
```diff
@@ -42,15 +42,15 @@ static String sequenceAsSearchTerm(ImapFolder i4folder, TreeSet<ImapMessage> i4s
         i4set.remove(null);
         if (i4set.isEmpty()) {
             return "item:none";
-        } else if (abbreviateAll && isAllMessages(i4folder, i4set)) {
+        } else if (abbreviateAll) {
```
fatkudu: This looks wrong to me.  It looks like this line is testing whether or not all messages should be returned and if they are, whether an abbreviated search should be used.  This change would cause all messages to be returned, even when only some have been requested.  As Greg said, probably need tests for the problem being addressed but also tests for searching for subsets.

tmclane: The issue is that the test for length isn't used here so if the i4set contains 10000 items they will all be included in the entire search query without this change.

Perhaps we need to rethink all of this?

tmclane:
@grishick Yes I am working on them now. Was just wanting to get some eyes on this sooner than later.

store/src/java/com/zimbra/cs/imap/ImapSearch.java
tmclane: i.e.  abbreviateAll == true && !isAllMesssages()

fatkudu: Yes, I think a rethink is needed for the case when isAllMessages() returns false (or even when abbreviateAll is false)  I think you said there is paging code but that gets involved too late.  I suspect the fix will require making the paging effective.

tmclane:
Closing this since an alternate approach is desired.

PR #441, opened 2017-11-30
ZCS-2655:Recurring all day yearly instance shifts 1 day ahead when ac…

sneha-patil-synacor:
ZCS-2655: Recurring all day yearly instance shifts 1 day ahead when accepted from ZWC

Issue:
When the attendee sends acceptance for all day single instance, it sends the timezone along in instance invite. This modifies the start date of instance on organizer side.

Fix:
Ignore the timezone when modifying the organizer instance. All day invite shouldn't have the timezone in start/end date. When accepted from mac calendar app, the acceptance doesn't contain the timezone.

Added Junit test.
ant test-all passed


store/src/java-test/com/zimbra/cs/mailbox/calendar/AllDayAcceptInstanceTest.java
```diff
@@ -0,0 +1,179 @@
...
+import com.zimbra.cs.mailbox.calendar.Recurrence.IRecurrence;
+import com.zimbra.cs.mailbox.calendar.Recurrence.RecurrenceRule;
+import com.zimbra.cs.mailbox.calendar.Recurrence.SimpleRepeatingRule;
+
+public class AllDayAcceptInstanceTest {
+    @Before
+    public void setUp() throws Exception {
+
+        MailboxTestUtil.initServer();
+        MailboxTestUtil.clearData();
```
rupalid: Not needed since we are clearing in tearDown

store/src/java-test/com/zimbra/cs/mailbox/calendar/AllDayAcceptInstanceTest.java
```diff
@@ -0,0 +1,179 @@
...
+        inviteReply.setIsAllDayEvent(true);
+        // test2 accepts first invite instance and sends acceptance notification to organizer
+        mbox1.addInvite(null, inviteReply, calendarFolder.getId());
+        List<String> uids = new ArrayList<String>();
+        uids.add(uid);
+        Map<String, CalendarItem> calItems = mbox1.getCalendarItemsByUid(null, uids);
+        CalendarItem calItem = calItems.get(uid);
+        Invite exception = calItem.getInvite(rid);
+        // The exception created in organizers series should not have the timezone in start/end time as sent in acceptance notification
+        Assert.assertEquals("VALUE=DATE:20170701", exception.getStartTime().toString());
```
rupalid: Set a message for failure scenario

fatkudu:
Was the reply modifying the ORGANIZER's event times?  If so, that is wrong and scary.  It would be easy to craft a reply that is 3 days off to make the ORGANIZER's recurrence wrong.  I'd be inclined to say that if a reply doesn't exactly match the recurrence rule of the ORGANIZER or exactly match an instance, we should be rejecting/ignoring it.  As such, sounds like this is a ZWC bug and should be fixed there rather than here.

sneha-patil-synacor:
@fatkudu 
When an instance acceptance is received from attendee, it creates an exception in series on organizers calendar as only that particular instance is accepted by the attendee. Now, this exceptional invite contains start and end time which is send in invite associated with acceptance notification from attendee. There is a check which verifies that this exception falls into nearby instances of the calendar item. There is a method in CalendarItem class called 'instancesNear' - it finds instances within 24 hours either side. If there are no nearby instances found, it ignores that notification and does not update the organizers calendar.

store/src/java-test/com/zimbra/cs/mailbox/calendar/AllDayAcceptInstanceTest.java
sneha-patil-synacor: Done

store/src/java-test/com/zimbra/cs/mailbox/calendar/AllDayAcceptInstanceTest.java
sneha-patil-synacor: Done

fatkudu:
@sneha-patil-synacor thanks for the explanation.  I'd forgotten about that case.  I would say that if the target time doesn't exactly match the time of the nearest instance for the main series, then the exception should be created using the timing information for the nearest instance for the main series, rather than based on what is in the reply - I suspect the only thing that should come from the reply is the PARTSTAT setting of the attendee?

sneha-patil-synacor:
@fatkudu 
I agree with that. I have created a new PR with the suggested fix. Please check. https://github.com/Zimbra/zm-mailbox/pull/447

PR #443, opened 2017-12-01
ZCS-3785 move setup call after commandline processing

tmclane:
The 'setup' call initializes all the underlying subsystems which cause the EHCache lock subsystem to hang.

There is no need to execute this before processing the command-line arguments.
Moving the location of the 'setup' call to after the processing of the command-line arguments avoids a thread-lock which was seen when running the `zmplayredo -h` command.


tmclane:
Thanks for the headsup on the needed rebase.

IMO: Rebasing is always a pre-requisite before merging.

PR #445, opened 2017-12-04
Bugfix/zimbra 94 contact frequency graph

varunrisbud:
This PR provides the **Contact Analytics** needed by the machine learning module to classify the emails. It also provides data to display **Contact Frequency Graph** for a contact to the front-end. Two types of analytics are implemented in this PR.

#### Contact Analytics ####
1. **Contact Frequency Count**: It provides the count of the number of emails sent or received to/from a contact in a given time range. It provides four time-ranges:
  - `LAST_DAY(NOW-24hours)`
  - `LAST_WEEK(NOW-7DAYS)`
  - `LAST_MONTH(NOW-1MONTH)`
  - `FOREVER(MATCHES ALL EMAILS)`

You can also specify if you want the **SENT email count** or **RECEIVED email count** or **Both (COMBINED)**. 

2. **Contact Frequency Graph**: It provides the count of the number of emails sent and received to/from a contact for 3 time-ranges grouped by aggregation unit associated with the time range.
  - `CURENT_MONTH` - From the 1st day of the current month to the current day. Grouped by `DAY`.
  - `LAST_SIX_MONTHS` - Zimbra calendar provides the functionality to configure the first day of the week (`zimbraPrefCalendarFirstDayOfWeek`). As per that config first day of the current week is calculated. From the first day of the current week, 6 months are subtracted to get the start date. We group the contact frequency by `WEEK`. 
  - `CURRENT_YEAR` - From the 1st day of the current year to the current day. Grouped by `MONTH`.

#### New Classes ####
A new package `com.zimbra.cs.event.analytics.contact` contains following classes:
 - `ContactAnalytics` - Class that provides methods to fetch the different contact analytics from the event store. 
 - `ContactFrequencyGraphDataPoint` - Represent a data point of the graph which has `label` and `value`.

Classes added to `com.zimbra.soap.mail.message`:
 - `GetContactFrequencyRequest` - Represents the SOAP request.
 - `GetContactFrequencyResponse` - Represents the SOAP response.

Classes added to `com.zimbra.soap.mail.type`
 - `ContactFrequencyData` - Represents the format in which contact frequency graph data will be returned.
 - `ContactFrequencyDataPoint` - Represent a data point for the graph which has `label` and `value`. Same as `ContactFrequencyGraphDataPoint`.

Classes added to `com.zimbra.cs.service.mail`:
 - `GetContactFrequncy` - Request handler to get the contact frequncy graph for a contact. It provides three time ranges `d (CURRENT_MONTH), w (LAST_SIX_MONTHS), m (CURRENT_YEAR)`.

### DIFF
--- a/soap/src/java/com/zimbra/soap/mail/message/GetContactFrequencyResponse.java
+++ b/soap/src/java/com/zimbra/soap/mail/message/GetContactFrequencyResponse.java
@@ -1,0 +1,22 @@
+package com.zimbra.soap.mail.message;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+
+import com.zimbra.common.soap.MailConstants;
+import com.zimbra.soap.mail.type.ContactFrequencyData;
+
+@XmlRootElement(name=MailConstants.E_GET_CONTACT_FREQUENCY_RESPONSE)
+public class GetContactFrequencyResponse {
+
+    public GetContactFrequencyResponse() {}
+
+    @XmlElement(name=MailConstants.E_CONTACT_FREQUENCY_DATA, type=ContactFrequencyData.class)
+    private List<ContactFrequencyData> frequencyGraphs = new ArrayList<>();
+
+    public List<ContactFrequencyData> getFrequencyGraphs() { return frequencyGraphs; }
+    public void addFrequencyGraph(ContactFrequencyData data) { this.frequencyGraphs.add(data); }
+}
--- a/soap/src/java/com/zimbra/soap/mail/type/ContactFrequencyData.java
+++ b/soap/src/java/com/zimbra/soap/mail/type/ContactFrequencyData.java
@@ -1,0 +1,33 @@
+package com.zimbra.soap.mail.type;
+
+import java.util.List;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlElement;
+
+import com.zimbra.common.soap.MailConstants;
+
+@XmlAccessorType(XmlAccessType.NONE)
+public class ContactFrequencyData {
+
+    public ContactFrequencyData() {}
+
+    public ContactFrequencyData(String frequencyBy, List<ContactFrequencyDataPoint> dataPoints) {
+        setFrequencyBy(frequencyBy);
+        setDataPoints(dataPoints);
+    }
+
+    @XmlAttribute(name=MailConstants.A_CONTACT_FREQUENCY_BY, required=true)
+    private String frequencyBy;
+
+    @XmlElement(name=MailConstants.E_CONTACT_FREQUENCY_DATA_POINT, type=ContactFrequencyDataPoint.class)
+    private List<ContactFrequencyDataPoint> dataPoints;
+
+    public String getFrequencyBy() { return frequencyBy; }
+    public void setFrequencyBy(String frequencyBy) { this.frequencyBy = frequencyBy; }
+
+    public List<ContactFrequencyDataPoint> getDataPoints() { return dataPoints; }
+    public void setDataPoints(List<ContactFrequencyDataPoint> dataPoints) { this.dataPoints = dataPoints; }
+}
--- a/store/src/java/com/zimbra/cs/event/SolrEventStore.java
+++ b/store/src/java/com/zimbra/cs/event/SolrEventStore.java
@@ -1,15 +1,33 @@
 package com.zimbra.cs.event;
 
 import java.io.IOException;
+import java.sql.Timestamp;
+import java.time.LocalDateTime;
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.TemporalAdjusters;
+import java.time.temporal.WeekFields;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Locale;
 
+import org.apache.commons.lang.NotImplementedException;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.search.BooleanClause.Occur;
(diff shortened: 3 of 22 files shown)


tmclane:
Would a 'LAST_30_DAYS' not be more relevant than CURRENT_MONTH?

store/src/java/com/zimbra/cs/event/SolrEventStore.java
```diff
@@ -73,13 +244,13 @@ public Factory() throws ServiceException {
...
-            default:
-                locator = new JointCollectionLocator(SolrConstants.EVENT_CORE_NAME_OR_PREFIX);
-                break;
+                case account:
+                    locator = new AccountCollectionLocator(SolrConstants.EVENT_CORE_NAME_OR_PREFIX);
+                    break;
+                case combined:
+                default:
+                    locator = new JointCollectionLocator(SolrConstants.EVENT_CORE_NAME_OR_PREFIX);
+                    break;
```
tmclane: Is this tabs and whitespace?

ilyusha:
@tmclane It is a feature requirement to provide contact frequency metrics starting at the beginning of the current month.

tmclane:
@ilyusha I understand that.  I was just thinking that if it's day 4 of the month for example I might want to see the last 30 days worth not just to 12/1 for example.

Or does it actually mean the last 30 days?

store/src/java/com/zimbra/cs/event/SolrEventStore.java
```diff
@@ -58,6 +77,158 @@ protected void deleteEventsByDataSource(String dataSourceId) throws ServiceExcep
...
+    private String getContactFrequencyQueryStartDate(ContactAnalytics.ContactFrequencyTimeRange timeRange) throws ServiceException {
+        switch (timeRange) {
+            case LAST_DAY:
+                return "NOW-1DAY";
+            case LAST_WEEK:
+                return "NOW-7DAY";
+            case LAST_MONTH:
+                return "NOW-1MONTH";
+            default:
+                throw ServiceException.FAILURE("Time range not supported " + timeRange, new NotImplementedException());
```
ilyusha: You can probably use `ServiceException.INVALID_REQUEST` here instead of NotImplementedException.

varunrisbud:
@tmclane As per the feature request, if its 4th day of the month then data for just 4 days will be returned. We can change it if the product team wants it to be last 30 days instead.

tmclane:
@varunrisbud Ok. Fair enough. Perhaps a new constant would be appropriate (don't add it now please - not trying to push feature creep).

varunrisbud:
@tmclane I agree with you that Last_30_Days makes more sense. But then this whole feature to show contact frequency graph to the user does not make sense to me. I cant think of a use case for which this graph will be useful for a normal user. But that's what the client wants.

varunrisbud:
@tmclane - I have fixed the switch case indentation issue. Turns out IntelliJ has different switch case indentation format than the majority of other IDEs.

@ilyusha - Changed the exceptions to `ServiceException.INVALID_REQUEST` as suggested.

soap/src/java/com/zimbra/soap/mail/message/GetContactFrequencyResponse.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.zimbra.common.soap.MailConstants;
+import com.zimbra.soap.mail.type.ContactFrequencyData;
+
+@XmlRootElement(name=MailConstants.E_GET_CONTACT_FREQUENCY_RESPONSE)
+public class GetContactFrequencyResponse {
+
+    public GetContactFrequencyResponse() {}
+
+    @XmlElements({
```
fatkudu: If there is only one class, don't need `@XmlElements`.  Suspect this makes the WSDL un-necessarily complicated - you probably end up with list of elements which can be a choice from the set [<data>].  If you just use `@XmlElement` with a `List` then the WSDL will allow a list of "<data>" elements.

soap/src/java/com/zimbra/soap/mail/type/ContactFrequencyData.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public ContactFrequencyData(String frequencyBy, List<ContactFrequencyDataPoint> dataPoints) {
+        setFrequencyBy(frequencyBy);
+        setDataPoints(dataPoints);
+    }
+
+    @XmlAttribute(name=MailConstants.A_CONTACT_FREQUENCY_BY, required=true)
+    private String frequencyBy;
+
+    @XmlElements({
```
fatkudu: Again, no need for the `@XmlElements` wrapper.

store/src/java/com/zimbra/cs/event/SolrEventStore.java
```diff
@@ -58,6 +77,158 @@ protected void deleteEventsByDataSource(String dataSourceId) throws ServiceExcep
...
+                graphDataPoints.add(new ContactFrequencyGraphDataPoint(rangeResult.getValue(), rangeResult.getCount()));
+            }
+        }
+        return graphDataPoints;
+    }
+
+    private String getAggregationBucketForContactFrequencyGraphTimeRange(ContactAnalytics.ContactFrequencyGraphTimeRange timeRange) throws ServiceException {
+        switch (timeRange) {
+        case CURRENT_MONTH:
+            return "+1DAY/DAY";
```
fatkudu: Don't quite understand what this is.  Perhaps a comment to explain the format and why is has been chosen to be related to the current month?  Maybe similar comments for the other cases too.

soap/src/java/com/zimbra/soap/mail/message/GetContactFrequencyResponse.java
ilyusha: Ah OK, thanks!

store/src/java/com/zimbra/cs/event/SolrEventStore.java
varunrisbud: "+1DAY/DAY" represent the "Gap/Rounding" needed for Solr facet query. I will add comments to the code.

PR #446, opened 2017-12-04
ZIMBRA-98: Sort By Relevance

ilyusha:
This PR Introduces the ability to sort search results by relevance.

Relevance sort is unique in that the sort key is not intrinsic to messages, like name or date, but is dependent on the query. Prior to this change, all sort keys were stored in both the index and the database (with the exception of recipient sort, where the sort key is _only_ stored in the db, making this sort incompatible with index-bound clauses). This allowed both index-first and db-first query plans to specify the sort order of the initial operation. Furthermore, since all query operations have a db component, the db search result object `DbSearch.Result` could always be instantiated with a sort value taken directly from the database.

Since this assumption no longer holds, the following restrictions apply to relevance sort:
* Relevance sort can only be specified for searches where _all_ clauses have a text component. This is because relevance ordering is undefined on db-only clauses. Note that this means that union queries `foo or in:sent` cannot be sorted by relevance, while intersection queries `foo and in:sent` can, since the latter compiles to a single db operation with a nested index operation.
* Queries with relevance sort _must_ be executed index-first

That said, the following changes are introduced to make relevance sort possible:
* The `DbSearch` class can be instantiated with a `LuceneResultsChunk` instance. This must be done when sorting by relevance, as it lets us sort the db results into the order provided by the index.
* If relevance sort is specified,`DbSearch::searchInternal` explicitly sorts the results after the sql query runs, using the `LuceneResultsChunk` mentioned above.
* The `DbSearch.Result` class accepts sort values of type `float`, to accommodate the relevance score value.

#### Solr query changes
The solr dismax query has been updated for better relevance matching:
* The `pf` parameter has been added to prioritize documents where the provided terms match as a phrase. This prioritizes documents where a query for `+foo +bar` contains `foo` and `bar` next to each other.
* The `tie` parameter has been added with a value of 0.1. This causes lower-scoring fields to still contribute to the total relevance score, acting as a tiebreaker (0.1 is the value recommended by solr documentation).

#### Other changes
Several other changes have been made as part of this PR:
* Solr query local params are now built programmatically using the helper `LocalParams` class, instead of being constructed from strings.
* The server no longer fetches index results one-by-one when building `ZimbraHit` objects, eliminating a large bottleneck. `SolrIndexBase` was already returning the full set of requested fields as part of the initial query; it was only matter of updating `LuceneQueryOperation` to make use of this data.

#### UPDATE 12/15/2017
* If a relevance sort is requested but the search query does not support it, it defaults to `dateDesc` instead.
* `SearchResponse` includes new optional boolean flag `relevanceSortSupported` that is set to `false` for cases when not all search clause can be sorted by relevance. This is so that the UI knows whether to display the "relevance" sort option.
* New LDAP attribute `zimbraDefaultSortByRelevance` controls whether sort defaults to `relevanceDesc` if no sort value is specified.


store/src/java/com/zimbra/cs/index/DBQueryOperation.java
```diff
@@ -787,10 +791,26 @@ private void luceneFirstGetNextChunk(SortBy sort) throws ServiceException {
             } else {
                 // must not ask for offset,limit here b/c of indexId constraints!,
                 dbSearch(dbHits, sort, -1, -1);
+                //if we are sorting by relevance, we have to re-sort the DB results here
+//                if (sort.getKey() == SortBy.Key.RELEVANCE) {
+//                    sortDbHitsByLuceneOrder();
+//                }
             }
```
tmclane: Would prefer not to see commented out code checked in.

store/src/java/com/zimbra/cs/index/DBQueryOperation.java
```diff
@@ -787,10 +791,26 @@ private void luceneFirstGetNextChunk(SortBy sort) throws ServiceException {
...
+//        Map<Integer, DbSearch.Result> hitsMap = new HashMap<>();
+//        for (DbSearch.Result hit: dbHits) {
+//            hitsMap.put(hit.getIndexId(), hit);
+//        }
+//        List<DbSearch.Result> sorted = new ArrayList<>(dbHits.size());
+//        for (Integer indexId: luceneChunk.getIndexIds()) {
+//            sorted.add(hitsMap.get(indexId));
+//        }
+//        dbHits = sorted;
+//    }
```
tmclane: Would prefer not to see commented out code checked in.

store/src/java/com/zimbra/cs/index/DBQueryOperation.java
ilyusha: Oh, good catch. I meant to delete that.

store/src/java/com/zimbra/qa/unittest/TestSearchSortByRelevance.java
```diff
@@ -0,0 +1,138 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2014, 2016 Synacor, Inc.
```
gordyt: Should we update the copyright date?

ilyusha: I think we're going to have to update the copyright all over the codebase :)

Victor-D:
In 2018 :D

store/src/java/com/zimbra/qa/unittest/TestSearchSortByRelevance.java
Victor-D: In 2018 :D

PR #448, opened 2017-12-05
Bug 80485: Introduce attribute to disable local auth fallback for administrators

mss:
Zimbra has a surprising feature: On a domain which is configured for LDAP authentication and where `zimbraAuthFallbackToLocal` is false, a user who is either an admin or a domain admin still login with the local password. As I commented on [bug 80485](https://bugzilla.zimbra.com/show_bug.cgi?id=80485) do I consider this a security issue since when you auto provision your accounts via scripting you have to specify a password for `zmprov createAcount` and since it is never used, you might just choose some trivial dummy value. And suddenly you've got an unexpected backdoor to your system.

These changes introduce an extra attribute `zimbraAdminAuthFallbackToLocal` (domain and global config) which still defaults to true for two reasons:
* To keep backwards compatibility
* To make sure an admin can still access the admin console in case LDAP doesn't work since the existing attribute `zimbraAdminConsoleLDAPAuthEnabled` is ignored.

To reproduce this issue and verify the fix:

1. Create a test domain, say `example.com`.
2. Create an admin account with a password like `test1234`.
3. Make sure you can use this account to log in to both the Web UI and Admin UI.
4. Configure the domain for a broken AD authentication: `zmprov md example.com zimbraAuthMech ad zimbraAuthLdapBindDn "%u@test.local" zimbraAuthLdapURL "ldap://localhost:1389"`
5. Just to be sure call `zmprov fc -a all`.
6. Try to login with the local password. This should still work. for both the Web UI and Admin UI.
7. Change the attribute to false: `zmprov md example.com zimbraAuthFallbackToLocal FALSE zimbraAdminAuthFallbackToLocal FALSE`.
8. Just to be sure call `zmprov fc -a all`.
9. Try to log in. This should now fail for both the Web UI and Admin UI.

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9638,6 +9638,10 @@ TODO: delete them permanently from here
 <attr id="3023" name="zimbraLDAPSchemaVersion" type="string" cardinality="single" requiredIn="globalConfig" since="8.8.8">
   <desc>LDAP schema version for the system.</desc>
   <globalConfigValue>1518163473</globalConfigValue>
+
+<attr id="3024" name="zimbraAdminAuthFallbackToLocal" type="boolean" cardinality="single" optionalIn="globalConfig,domain" since="8.8.8">
+  <desc>Fall back to local auth if external mech fails for an administrator account</desc>
+  <globalConfigValue>TRUE</globalConfigValue>
 </attr>
 
 </attrs>
--- a/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
+++ b/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
@@ -2936,6 +2936,15 @@ public class ZAttrProvisioning {
     public static final String A_zimbraAdminAccessControlMech = "zimbraAdminAccessControlMech";
 
     /**
+     * Fall back to local auth if external mech fails for an administrator
+     * account
+     *
+     * @since ZCS 8.8.6
+     */
+    @ZAttr(id=3022)
+    public static final String A_zimbraAdminAuthFallbackToLocal = "zimbraAdminAuthFallbackToLocal";
+
+    /**
      * lifetime of newly created admin auth tokens. Must be in valid duration
      * format: {digits}{time-unit}. digits: 0-9, time-unit: [hmsd]|ms. h -
      * hours, m - minutes, s - seconds, d - days, ms - milliseconds. If time
--- a/store/src/java/com/zimbra/cs/account/ZAttrConfig.java
+++ b/store/src/java/com/zimbra/cs/account/ZAttrConfig.java
@@ -1207,6 +1207,83 @@ public abstract class ZAttrConfig extends Entry {
     }
 
     /**
+     * Fall back to local auth if external mech fails for an administrator
+     * account
+     *
+     * @return zimbraAdminAuthFallbackToLocal, or true if unset
+     *
+     * @since ZCS 8.8.6
+     */
+    @ZAttr(id=3022)
+    public boolean isAdminAuthFallbackToLocal() {
+        return getBooleanAttr(Provisioning.A_zimbraAdminAuthFallbackToLocal, true, true);
+    }
+
+    /**
+     * Fall back to local auth if external mech fails for an administrator
+     * account
+     *
+     * @param zimbraAdminAuthFallbackToLocal new value
+     * @throws com.zimbra.common.service.ServiceException if error during update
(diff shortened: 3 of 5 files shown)


mss:
Hmm… something went wrong when I rebased this branch. I'll push again.

mss:
I fixed it up. Did somebody force-push to `develop` before?

There was an attribute with the id 3021 (something about profile pictures) which is gone now. I didn't renumber my attribute though, it is still 3022. Should I change that?

gordyt:
@mss I did a force push the other day (by request) to remove two commits from `develop`.  Not sure if you need to renumber.  @rupalid would you comment on that?

mss:
I noticed something I don't like about that part of the code; I'll do a refactoring later on and will update the branch. No need to review before I do that :smile:

barrydegraaff:
One bottle of beer for Malte, this is a really important fix! Thanks man!

barrydegraaff:
Also can we change the default behavior for new installs or when users enable LDAP auth from now on, that zimbraAuthFallbackToLocal is set to FALSE by default. That would be even better!

mss:
I had another look and the way I wanted to change the code won't work. So the code is ready for review.

@rupalid Do I have to/should I renumber my attribute?

@barrydegraaff thanks; my code originally set the value of `zimbraAdminAuthFallbackToLocal` to false but that can have the effect of locking people out of the Admin UI. There should probably a special case (and yet another attribute) to take care of that but that would probably be a bigger change. Maybe for an upcoming version this default could be changed.

mss:
Ah, wait. I should be able to get the information if this is a login to the Admin UI from the `authCtxt` like it is done in `AuthMechanism.newInstance`:
```java
Boolean asAdmin = context == null ? null : (Boolean) context.get(AuthContext.AC_AS_ADMIN);
```

The semantic of the `zimbraAdminAuthFallbackToLocal` attribute I coded here is different from what I orifinally suggested on [bug 80485](https://bugzilla.zimbra.com/show_bug.cgi?id=80485#c9):

> Falling back to local auth often makes sense for the Admin UI on port 7071 (which should be behind a strict firewall anyway) so you're still able to access it when your LDAP server is down.  But the mail part should not fall back to local.
> 
> I think Adam's suggestion is a good one:  make the Web UI (and IMAP etc.) login look at `zimbraAuthFallbackToLocal`, not for both admin users and mere mortals.  Introduce an extra attribute `zimbraAdminAuthFallbackToLocal` which allows the login to the Admin UI use the fallback.

I'd prefer the one I suggested there, ie. never treat admin accounts differently to normal accounts for non-administrative logins, which would be a change in behaviour though. Behaviour I hope nobody relies on. What do you people think?

barrydegraaff:
I would like to be able to use the Admin UI with fallback to local auth if LDAP goes down, 
as that makes sense in case one needs to troubleshoot, and the admin UI should not be exposed 
to public. But I consider that to a `nice to have`. If the Admin UI does not fall back, one
can  still use the CLI.


I do not want the use of a local set password in case the LDAP is configured and the fallback
is set to false, so in short, I agree with your idea Malte. I also thing that the setting for
fallback should be set to false automatically if LDAP is configured, as that makes the most
sense to me, and prevents unwanted surprises, it can be a really big deal if account get 
compromised because of this bug.

Barry

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9624,4 +9624,9 @@ TODO: delete them permanently from here
   </desc>
 </attr>
 
+<attr id="3022" name="zimbraAdminAuthFallbackToLocal" type="boolean" cardinality="single" optionalIn="globalConfig,domain" since="8.8.6">
```
grishick: I think attributed description has to be more specific to describe the logic of the code being added to LdapProvisioning.java, which is that if zimbraAuthFallbackToLocal is set to TRUE, then any account (admin or non admin) will fall back to local authentication, but if zimbraAuthFallbackToLocal is set to FALSE, then it is possible to allow only admin accounts to fall back to local authentication by setting this new attribute to TRUE, which is also the default behavior. If the desired outcome is to not allow falling back to local authentication for any accounts, then it is necessary to set both attributes to FALSE.

barrydegraaff:
**Hello All, can we please fix this one, as it is a freaking security issue.** Thanks!

barrydegraaff:
Linked to ZCS-8285

barrydegraaff:
ZBUG-2859

mss:
@barrydegraaff does that mean that this will finally be fixed in the foreseeable future?

PR #449, opened 2017-12-06
Bugfix/zcs 3557

grishick:
Factor out the method that generates zimbra search query, fix it to never exceed max folder batch size and add a unit test for the new method. I wrote the test before fixing the method and ran the test to make sure it was actually producing ridiculously long queries.


store/src/java-test/com/zimbra/cs/imap/ImapFolderTest.java
```diff
@@ -61,4 +61,47 @@ public void testGetSubsequence() throws Exception {
...
+        for(int i = 1; i < (ImapFolder.LARGEST_FOLDER_BATCH * 3); i++) {
+            i4folder.cache(new ImapMessage(i, Type.of((byte) 5), i, 0, null), true);
+        }
+        String searchStr = i4folder.getSearchQuery(seqSearch);
+        Assert.assertEquals("wrong search query when sequence is 1:*", "in:\"INBOX\" item:all", searchStr);
+
+        //2X max batch size
+        seqSearch = new SequenceSearch("S", String.format("1:%d", ImapFolder.LARGEST_FOLDER_BATCH * 2), true);
+        searchStr = i4folder.getSearchQuery(seqSearch);
+        Assert.assertEquals("wrong search query when sequence is twice the size of max batch", "in:\"INBOX\" item:all", searchStr);
```
fatkudu: This looks to me as if the search will return all items in the folder - but we only want the first 1200 items of 1800.
Am I missing something?
If there is something that cleans up the mess later, then perhaps a SOAP IMAP test would prove that is working.  Also, if that is the case, the code could definitely benefit from more comments to explain how it works - and this test should have a comment that the list of all items gets pruned later for the real IMAP case.

store/src/java-test/com/zimbra/cs/imap/ImapFolderTest.java
```diff
@@ -61,4 +61,47 @@ public void testGetSubsequence() throws Exception {
...
+
+        //just over max batch size
+        seqSearch = new SequenceSearch("S", String.format("1:%d", ImapFolder.LARGEST_FOLDER_BATCH+1), true);
+        searchStr = i4folder.getSearchQuery(seqSearch);
+        Assert.assertEquals("wrong search query when sequence is longer than max batch size by one", "in:\"INBOX\" item:all", searchStr);
```
fatkudu: Same applies here as for previous comment.

store/src/java-test/com/zimbra/cs/imap/ImapFolderTest.java
```diff
@@ -61,4 +61,47 @@ public void testGetSubsequence() throws Exception {
...
...3,294,295,296,297,298,299,300,301,302,303,304,305,306,307,308,309,310,311,312,313,314,315,316,317,318,319,320,321,322,323,324,325,326,327,328,329,330,331,332,333,334,335,336,337,338,339,340,341,342,343,344,345,346,347,348,349,350,351,352,353,354,355,356,357,358,359,360,361,362,363,364,365,366,367,368,369,370,371,372,373,374,375,376,377,378,379,380,381,382,383,384,385,386,387,388,389,390,391,392,393,394,395,396,397,398,399,400,401,402,403,404,405,406,407,408,409,410,411,412,413,414,415,416,417,418,419,420,421,422,423,424,425,426,427,428,429,430,431,432,433,434,435,436,437,438,439,440,441,442,443,444,445,446,447,448,449,450,451,452,453,454,455,456,457,458,459,460,461,462,463,464,465,466,467,468,469,470,471,472,473,474,475,476,477,478,479,480,481,482,483,484,485,486,487,488,489,490,491,492,493,494,495,496,497,498,499,500,501,502,503,504,505,506,507,508,509,510,511,512,513,514,515,516,517,518,519,520,521,522,523,524,525,526,527,528,529,530,531,532,533,534,535,536,537,538,539,540,541,542,543,544,545,546,547,548,549,550,551,552,553,554,555,556,557,558,559,560,561,562,563,564,565,566,567,568,569,570,571,572,573,574,575,576,577,578,579,580,581,582,583,584,585,586,587,588,589}", searchStr);
```
fatkudu: Perhaps a comment here that the list is 10 items shorter because item 589 has IMAP ID=LARGEST_FOLDER_BATCH-1.  Took me a while to spot that.

store/src/java/com/zimbra/cs/imap/ImapSearch.java
```diff
@@ -35,7 +35,7 @@
 
     static boolean isAllMessages(ImapFolder i4folder, Set<ImapMessage> i4set) {
         int size = i4set.size() - (i4set.contains(null) ? 1 : 0);
-        return size == i4folder.getSize();
+        return size > ImapFolder.LARGEST_FOLDER_BATCH || size == i4folder.getSize();
```
fatkudu: This goes back to what I said about Travis's original proposed fix.  My assumption would be that `isAllMessages` means `i4set` contains all messages.  This change makes it "contains all messages or is larger than the batch size".  With this change, certainly need to rename the method to something more accurate.  My worry is that it actually introduces a bug.  See comments against unit test on how a SOAP IMAP test might prove one way or the other.

grishick: @fatkudu you are correct. This change modifies how many records are returned by UID SEARCH with a range that covers a subset of messages. Remote IMAP returns 791 messages for 'UID SEARCH 10:800 UNDELETED' before this change and 1000 messages after this change regardless of whether the selected folder is virtual or regular.

grishick:
I filed 3 user stories that we need to complete to fix the underlying problems. Closing this PR since it is not complete.

PR #450, opened 2017-12-06
Lock Abstraction WIP

spoon16:
This is the PR of the active work we are doing on the mailbox lock abstraction.

Please do not merge.

### DIFF
--- a/store/src/java/com/zimbra/cs/mailbox/Mailbox.java
+++ b/store/src/java/com/zimbra/cs/mailbox/Mailbox.java
@@ -5773,7 +5773,8 @@ public class Mailbox implements MailboxStore {
         final MailboxLock l = lockFactory.writeLock();
         try {
             ZimbraLog.mailbox.info("addMessage tries to get lock");
-            l.lock();
+			// "addMessageInternal" method calls MailboxTRansaction and inside of it, is called lock(), if this line is not commented we should have 2 locks and that is incorrect
+            //l.lock(); 
             ZimbraLog.mailbox.info("addMessage  gets lock");
             try {
                 Message message = addMessageInternal(l, octxt, pm, folderId, noICal, flags, tags, conversationId,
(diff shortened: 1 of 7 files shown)


store/src/java/com/zimbra/cs/mailbox/MailboxManager.java
```diff
@@ -920,6 +921,7 @@ private synchronized Mailbox createMailboxInternal(OperationContext octxt, Accou
             try {
                 if (t != null) {
                     mboxTransaction.close();
+                    lock.close();
```
spoon16: ⭐️

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -5773,7 +5773,8 @@ private Message addMessage(OperationContext octxt, ParsedMessage pm, int folderI
         final MailboxLock l = lockFactory.writeLock();
         try {
             ZimbraLog.mailbox.info("addMessage tries to get lock");
-            l.lock();
+			// "addMessageInternal" method calls MailboxTRansaction and inside of it, is called lock(), if this line is not commented we should have 2 locks and that is incorrect
+            //l.lock();
```
spoon16: @Raffaell0 I think it is ok to call `l.lock()` multiple times. If the current thread holds the lock than calling `.lock` should be a noop.

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -5810,7 +5811,8 @@ private Message addMessageInternal(MailboxLock l, OperationContext octxt, Parsed
             int flags, String[] tags, int conversationId, String rcptEmail, Message.DraftInfo dinfo,
             CustomMetadata customData, DeliveryContext dctxt, StagedBlob staged, MessageCallbackContext callbackContext, String dsId)
     throws IOException, ServiceException {
-        assert l.isWriteLockedByCurrentThread();
+		// at this point isn't possible to know the kind of the lock (reader/writer), we know it after call "lock()" method
+        //assert l.isWriteLockedByCurrentThread();
```
spoon16: This is asserting that `lock` has been called on `l`. `l` knows that it is a write lock because we used `lockFactory.writeLock` to create it.

store/src/java/com/zimbra/cs/mailbox/MailboxManager.java
```diff
@@ -920,6 +921,7 @@ private synchronized Mailbox createMailboxInternal(OperationContext octxt, Accou
             try {
                 if (t != null) {
```
spoon16: What is `t` here?

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
Raffaell0: doesn't matter if you put lockFactory.writeLock, at this point there isn't lock yet, just after call lock() you are able to check if the lock is a writer

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
Raffaell0: multiple instances are created if you call lock()

PR #451, opened 2017-12-07
ZCS-3809 Adding new libs required by JWT

rupalid:
In JWT we are using newer version of jackson libraries. So updating it in soap and client projects to the latest version.

### DIFF
--- a/soap/ivy-soapdoclet.xml
+++ b/soap/ivy-soapdoclet.xml
@@ -6,9 +6,9 @@
  <dependencies>
   <dependency org="log4j" name="log4j" rev="1.2.16" />
   <dependency org="dom4j" name="dom4j" rev="1.5.2" />
-  <dependency org="org.codehaus.jackson" name="jackson-core-asl" rev="1.9.2" />
-  <dependency org="org.codehaus.jackson" name="jackson-mapper-asl" rev="1.9.2" />
-  <dependency org="org.codehaus.jackson" name="jackson-xc" rev="1.9.2" />
+<dependency org="com.fasterxml.jackson.core" name="jackson-core" rev="2.9.2"/>
+  <dependency org="com.fasterxml.jackson.core" name="jackson-databind" rev="2.9.2"/>
+  <dependency org="com.fasterxml.jackson.module" name="jackson-module-jaxb-annotations" rev="2.9.2"/>
   <dependency org="com.google.guava" name="guava" rev="13.0.1"/>
   <dependency org="commons-lang" name="commons-lang" rev="2.6"/>
   <dependency org="org.freemarker" name="freemarker" rev="2.3.19"/>
--- a/soap/ivy.xml
+++ b/soap/ivy.xml
@@ -17,9 +17,9 @@
   <dependency org="commons-cli" name="commons-cli" rev="1.2" />
   <dependency org="org.freemarker" name="freemarker" rev="2.3.23"/>
   <dependency org="log4j" name="log4j" rev="1.2.16" />
-  <dependency org="org.codehaus.jackson" name="jackson-core-asl" rev="1.9.2" />
-  <dependency org="org.codehaus.jackson" name="jackson-mapper-asl" rev="1.9.2" />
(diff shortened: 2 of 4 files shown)


fatkudu:
This doesn't build for me (or for Travis CI):
compile:
    [javac] Compiling 1916 source files to /home/zimbra/fatkudu/ZCS/zm-mailbox/soap/build/classes
    [javac] /home/zimbra/fatkudu/ZCS/zm-mailbox/soap/src/java/com/zimbra/soap/type/ZmBoolean.java:24: error: package org.codehaus.jackson.map.annotate does not exist
    [javac] import org.codehaus.jackson.map.annotate.JsonSerialize;
    [javac]                                         ^
    [javac] /home/zimbra/fatkudu/ZCS/zm-mailbox/soap/src/java/com/zimbra/soap/type/ZmBoolean.java:56: error: cannot find symbol
    [javac] @JsonSerialize(using=ZmBooleanSerializer.class)
...

fatkudu:
A while back (years!) someone else and then me had a quick look at moving to a more up to date Jackson version for zm-soap and I think there were a few non-trivial issues.

fatkudu:
Found this note in a dir with some experimental changes in P4 to try and move things along.  Afraid I didn't write notes on what the problems were:

> http://wiki.fasterxml.com/JacksonUpgradeFrom19To20
> 
> https://github.com/spring-projects/spring-framework/wiki/Migrating-from-earlier-versions-of-the-spring-framework
> Has this to say about Jackson:
> Jackson 1.8.6 (note: minimum 2.1 as of Spring Framework 4.1, with Jackson 2.2.2 or later recommended)
> 
> 
> https://github.com/FasterXML/jackson-docs
> Says this:
> On Jackson versioning
> 
> Note that there are two major Jackson versions: 1.x (1.0 - 1.9) and 2.x (2.0 - 2.3). These versions can co-exist as they are located in different Java package and use different jar naming and Maven group/artifact ids. But this means that you have to make sure that all components in use have matching major versions: specifically, Jackson 2.x code does NOT understand or support Jackson 1.x annotations, or vice versa.
> 
> Minor versions (like 2.1 and 2.2) are backwards compatible with respect to public API: old code should work without recompilation, if (but only if) it relies on external API; public methods, annotations. When overriding internal functionality, we try hard to maintain backwards compatibility between adjacent minor versions; need for changes is indicated by deprecating internal methods. Recompilation is thus recommended when extending by sub-classing, for example.
> 
> https://github.com/FasterXML/jackson
>   This is a tiny umbrella project which links to the real meat - see "Core modules"
>      https://github.com/FasterXML/jackson-core
>          For non-Maven use cases, you download jars from Central Maven repository
>              http://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-core/
>              http://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-core/2.2.4/jackson-core-2.2.4.jar
> 
>              and:
>              http://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-annotations/2.2.0/jackson-annotations-2.2.0.jar
>              http://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-databind/2.2.0/jackson-databind-2.2.0.jar
>              http://repo1.maven.org/maven2/com/fasterxml/jackson/module/jackson-module-jaxb-annotations/2.2.0/jackson-module-jaxb-annotations-2.2.0.jar

I've got a tar archive of some source files from P4 if you're interested but I don't know what state they were in or what wasn't working.

rupalid:
@fatkudu  I am working on the compilation issues. For some reason publish-local-all was successful and hence I sent the pull request.
Meanwhile does your second comment mean that we should try to maintain 2 versions. I have resolved many of the compilation errors but a few remain.

fatkudu:
@rupalid Full testing would be needed - all SOAP tests and unit tests in soap and store.  Plus also make sure that `cd soap && ant package-docs` works and produces reasonable documentation.

On whether shipping 2 versions is acceptable, I initially thought that was fine but from something @plobbes said elsewhere, we would need to be sure there aren't known security problems in the old library.

PR #454, opened 2017-12-08
ZCS-3258:Modify existing JWT validation implementation

shrikant-prasad:
added code to find salt from ZM_JWT cookie or jwtSalt element in soap context.
jwt validation logic modified to include salt in the key which is used for validation of jwt.
refactored jwt related code and moved all methods into JWTUtil class.
added junit tests.

-- | -- | -- | -- | -- | --
1120 | 0 | 0 | 10 | 100.00% | 244.402


store/src/java/com/zimbra/cs/service/util/JWTUtil.java
```diff
@@ -0,0 +1,182 @@
+/*
```
rupalid: Please add detailed comments for each of the public methods.
Also add logging statements, which might be useful while debugging.

Add unit test for the public utility methods

store/src/java/com/zimbra/cs/service/util/JWTUtil.java
```diff
@@ -0,0 +1,182 @@
...
+                throw AccountServiceException.NO_SUCH_ACCOUNT(claims.getSubject());
+            }
+            int acctValidityValue = acct.getAuthTokenValidityValue();
+            int tokenValidityValue = (Integer) claims.get(Constants.TOKEN_VALIDITY_VALUE_CLAIM);
+            if (acctValidityValue != tokenValidityValue){
+                throw AuthFailedServiceException.AUTH_FAILED("Token not valid");
+            }
+        } catch(ExpiredJwtException eje) {
+            throw ServiceException.AUTH_EXPIRED(eje.getMessage());
+        } catch(SignatureException se) {
```
rupalid: Should we log the stack trace at debug level.

store/src/java/com/zimbra/cs/service/util/JWTUtil.java
shrikant-prasad: done.

PR #455, opened 2017-12-08
Bugfix/zimbra 127 imp email analytics

varunrisbud:
This PR adds 4 email analytics which are needed by the machine learning module to classify emails. This PR is built on top of work done in bugfix/ZIMBRA-94-contactFrequencyGraph. No new classes are added.

#### Contact Analytics ####
- `Percentage Open Emails` - Provides the percentage of emails read to received from a contact.

- `Average Time to Open Emails for a Contact` - Provides the mean of the delta of time when email was read to the time when email was received. This is calculated using the Solr streaming queries. It creates an inner join of the stream of the read events to the stream of received events on message id. It calculates the difference between the timestamp of the read event to the timestamp of the received event. If the delta is less then 500ms then that event pair is ignored. Then a mean delta is calculated.

- `Average Time to Open Emails for an Account` - Same as `Average Time to Open Emails for a Contact`, except for getting read and received events for a specific contact, all the read and received events for an account(all contacts) are used.

- `Ratio of Avg Time to Open Emails for a Contact to Avg Time to Open Emails for an Account` - Provides a ratio of above two analytics. Smaller the ratio more importance can be assigned to the emails from that contact. This is assuming a User will read an email from important contact immediately after it is received. 

- `Percentage Replied Emails for a Contact` - Provides the percentage of emails replied to emails received from a contact.

### DIFF
--- a/store/src/java/com/zimbra/cs/event/SolrEventStore.java
+++ b/store/src/java/com/zimbra/cs/event/SolrEventStore.java
@@ -6,12 +6,9 @@ import java.time.LocalDateTime;
 import java.time.temporal.ChronoUnit;
 import java.time.temporal.TemporalAdjusters;
 import java.time.temporal.WeekFields;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
+import java.util.*;
 
+import com.google.common.base.Joiner;
 import com.google.common.collect.Maps;
 
 import org.apache.commons.lang.NotImplementedException;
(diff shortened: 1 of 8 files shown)


store/src/java/com/zimbra/cs/event/SolrEventStore.java
```diff
@@ -234,6 +237,150 @@ private String getAccountFilter(String accountId) {
...
+            int count = 0;
+            try (JoinStream joinStream = new InnerJoinStream(seenEventSelectStream, readEventSelectStream, equalitor)) {
+                joinStream.setStreamContext(getStreamContext());
+                joinStream.open();
+                Tuple tuple = joinStream.read();
+                while (!tuple.EOF) {
+                    Date seenDate = tuple.getDate(seenEventTimestampFieldName);
+                    Date readDate = tuple.getDate(readEventTimestampFieldName);
+                    double delta = readDate.getTime() - seenDate.getTime();
+                    if(delta > 500) {
```
tmclane: nit: space between if and (

store/src/java/com/zimbra/qa/unittest/SolrEventStoreTestBase.java
```diff
@@ -308,6 +309,94 @@ protected void testContactFrequencyGraph(ContactAnalytics.ContactFrequencyGraphT
         }
     }
 
+    protected void testPercentageOpenedEmails(SolrEventCallback eventCallback, String collectionName, SolrEventStore eventStore) throws Exception {
+        List<Timestamp> timestamps = getTimestampsForContactFrequencyCountTest(ContactAnalytics.ContactFrequencyTimeRange.LAST_WEEK);
+        List<Event> events = new ArrayList<>(timestamps.size());
+        //generate 15 received events
+        for (int i = 0; i < timestamps.size(); i++) {
```
tmclane: Curious what makes this exactly 15 events?

tmclane: Is getTimestampsForContactFrequencyCountTest always guaranteed to return 15?

PR #458, opened 2017-12-11
Bugfix/zcs 2184

tmclane:
Create nicer more understandable output message for the case that someone passes in an invalid attribute.

    zmmigrateattrs  ssdb:127.0.0.1:8888 -r Foo
    [] FATAL: Migration can't proceed: Invalid attribute specified: Foo


tmclane:
Note: This needs to be rebased to the end of the bugfix/ZCS-2187 branch to compile.  Separated for ease of review.

store/src/java/com/zimbra/cs/ephemeral/migrate/AttributeMigration.java
```diff
@@ -100,7 +101,8 @@ public AttributeMigration(Collection<String> attrsToMigrate, Integer numThreads)
      * @param numThreads - number of threads to use during migration. If null, the migration happens synchronously.
```
grishick: nit: add destUrl to javadoc comment.

PR #459, opened 2017-12-11
 ZCS-2191 handle ctrl-c gracefully

tmclane:
Update zmmigrateattrs to utilize a 'Shutdown-hook' such that when the JVM is exiting it can decide if it was successful or failed in the conversion process.

This enables the utility to output the appropriate message before fully exiting.

### DIFF
--- a/store/src/java/com/zimbra/cs/ephemeral/migrate/AttributeMigration.java
+++ b/store/src/java/com/zimbra/cs/ephemeral/migrate/AttributeMigration.java
@@ -69,6 +69,7 @@ import com.zimbra.soap.admin.type.CacheEntryType;
  *
  */
 public class AttributeMigration {
+    private ShutdownHook hook;
     private EntrySource source;
     private final Collection<String> attrsToMigrate;
     private int numThreads;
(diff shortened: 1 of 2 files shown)


tmclane:
It does seem to work as is. I added a Thread.sleep(4000) call during the processing so I could see if it output the message or not. Either way it appears the CSV is 'closed' although perhaps it's missing a message or two? Hard to know I feel like. 
I'll look into calling the `closeReports` in the hook as well.

tmclane:
Note: I also looked into adding a signal handler to intercept the `Ctrl-C` directly although this was not successful so far.  Figured I could add some cleanup code there if I could intercept it.

fatkudu:
@tmclane the ticket/bug is rather unhelpful in explaining exactly what the problem is but I assume it is missing entries in the .csv files rather than a complaint about what gets written to the terminal session.

tmclane:
@fatkudu if you `Ctrl-C` a program how do you know what is missing? h

This is what is output `after` this change when you push `Ctrl-C`:

    [] INFO: no ephemeral data to migrate for account 'admin@zcs-dev.test'
    [] INFO: no ephemeral data to migrate for account 'galsync.qpfofgulvo@zcs-dev.test'
    [] INFO: See PARTIAL report (migration abandoned) : '/opt/zimbra/data/tmp/zmmigrateattrs-2017-12-11_22-01-33.csv'

tmclane:
I feel like this satisfies the intent of the ticket personally. Although I can look into attempting to get the additional call made in the shutdown hook.
It would seem that more reorganization would be required to accomplish that since there would no longer be any reason to call it the way it is called currently.  The shutdown hook is always called after all.

fatkudu:
@tmclane that actually looks kind of cool.

PR #460, opened 2017-12-12
ZCS-3456:Attendee's status isn't updated

ykomiyam:
[Bug]
Organizer see the attendee's status as Tentative instead of Accepted
even when the attendee proposed a new time in invite and the organizer
accepted it.

[Root cause]
The the CounterAppointmentRequest proposed by the attendee contains
the attendee info ("PROPERTY:ATTENDEE(ATTENDEE)"), but it does not
contain the data of RSVP in it; that should be acceptable, and expected
behaviour because such an information has not been updated on the attendee
side.  On the other hand, when the organizer received the counter
appointment (via email), if there is no RSVP data in the attendee info,
the organizer turned the RSVP off for to the corresponding appointment
object.  Therefore, if the attendee later sent the acceptance, the
organizer ignored it even the acceptance email asked to update the
RSVP data on the organizer ("updateOrganizer":"TRUE").

[Fix]
Changed the default value of the RSVP data to true.

### DIFF
--- a/store/src/java-test/com/zimbra/cs/mailbox/calendar/Calendar_NewMeetingRequest_RsvpTrue.txt
+++ b/store/src/java-test/com/zimbra/cs/mailbox/calendar/Calendar_NewMeetingRequest_RsvpTrue.txt
@@ -1,0 +1,96 @@
+To: attendee1 <attendee1@zimbra.com>
+Message-ID: <2103105317.2993.1513012884769.JavaMail.zimbra@synacorjapan.com>
+Subject: Party
+MIME-Version: 1.0
+Content-Type: multipart/alternative;
+	boundary="=_2eb7dab5-3168-43b0-ba14-80504c7e08da"
+
+--=_2eb7dab5-3168-43b0-ba14-80504c7e08da
+Content-Type: text/plain; charset=utf-8
+Content-Transfer-Encoding: 7bit
+
+The following is a new meeting request:
+
+Subject: Party
+Organizer: org@synacorjapan.com
+
+Time: Tuesday, December 12, 2017, 10:00:00 AM - 10:30:00 AM GMT +09:00 Japan
+
+Invitees: attendee1@zimbra.com
+
+
+*~*~*~*~*~*~*~*~*~*
+
+
+
+--=_2eb7dab5-3168-43b0-ba14-80504c7e08da
+Content-Type: text/html; charset=utf-8
+Content-Transfer-Encoding: 7bit
+
+<html><body id='htmlmode'><h3>The following is a new meeting request:</h3>
+
+<p>
+<table border='0'>
+<tr><th align=left>Subject:</th><td>Party </td></tr>
+<tr><th align=left>Organizer:</th><td>org@synacorjapan.com </td></tr>
+</table>
+<p>
+<table border='0'>
+<tr><th align=left>Time:</th><td>Tuesday, December 12, 2017, 10:00:00 AM - 10:30:00 AM GMT +09:00 Japan
+ </td></tr></table>
+<p>
+<table border='0'>
+<tr><th align=left>Invitees:</th><td>attendee1@zimbra.com </td></tr>
+</table>
+<div>*~*~*~*~*~*~*~*~*~*</div><br></body></html>
+--=_2eb7dab5-3168-43b0-ba14-80504c7e08da
+Content-Type: text/calendar; charset=utf-8; method=REQUEST; name=meeting.ics
+Content-Transfer-Encoding: 7bit
+
+BEGIN:VCALENDAR
+PRODID:Zimbra-Calendar-Provider
+VERSION:2.0
+METHOD:REQUEST
+BEGIN:VTIMEZONE
+TZID:Asia/Tokyo
+BEGIN:STANDARD
+DTSTART:16010101T000000
+TZOFFSETTO:+0900
+TZOFFSETFROM:+0900
+END:STANDARD
+END:VTIMEZONE
+BEGIN:VEVENT
+UID:8282cdfc-a5a6-4ed4-84e2-be12c2bc4af8
+SUMMARY:Party
+ATTENDEE;CN=attendee1;ROLE=REQ-PARTICIPANT;PARTSTAT=NEEDS-ACTION;RSVP=TRUE:mailto
+ :attendee1@zimbra.com
+ORGANIZER:mailto:org@synacorjapan.com
+DTSTART;TZID="Asia/Tokyo":20171212T100000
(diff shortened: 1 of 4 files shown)


store/src/java-test/com/zimbra/cs/mailbox/calendar/Calendar_NewMeetingRequest_RsvpTrue.txt
```diff
@@ -0,0 +1,96 @@
...
+DTSTART:16010101T000000
+TZOFFSETTO:+0900
+TZOFFSETFROM:+0900
+END:STANDARD
+END:VTIMEZONE
+BEGIN:VEVENT
+UID:8282cdfc-a5a6-4ed4-84e2-be12c2bc4af8
+SUMMARY:Party
+ATTENDEE;CN=attendee1;ROLE=REQ-PARTICIPANT;PARTSTAT=NEEDS-ACTION;RSVP=TRUE:mailto
+ :attendee1@zimbra.com
```
jiteshsojitra: I would suggest not to use @zimbra.com for test accounts for public code.

store/src/java-test/com/zimbra/cs/mailbox/calendar/ZAttendeeTest.java
```diff
@@ -16,16 +16,44 @@
...
+import com.zimbra.cs.mailbox.MailboxTestUtil;
+import com.zimbra.cs.mime.ParsedMessage;
+import com.zimbra.cs.mime.ParsedMessage.CalendarPartInfo;
 
 public class ZAttendeeTest {
 
+    @BeforeClass
+    public static void init() throws Exception {
+        MailboxTestUtil.initServer();
+        Provisioning prov = Provisioning.getInstance();
```
rupalid: Move the account creation to setUp

store/src/java-test/com/zimbra/cs/mailbox/calendar/ZAttendeeTest.java
```diff
@@ -16,16 +16,44 @@
...
 import com.zimbra.common.calendar.ZCalendar.ZParameter;
 import com.zimbra.common.calendar.ZCalendar.ZProperty;
 import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.util.ByteUtil;
+import com.zimbra.cs.account.Provisioning;
+import com.zimbra.cs.mailbox.MailboxTestUtil;
+import com.zimbra.cs.mime.ParsedMessage;
+import com.zimbra.cs.mime.ParsedMessage.CalendarPartInfo;
 
 public class ZAttendeeTest {
```
rupalid: Can you add the following, so we can find which test fails in Circle CI
 @Rule public TestName testName = new TestName();
    @Rule public MethodRule watchman = new ZTestWatchman();

store/src/java-test/com/zimbra/cs/mailbox/calendar/ZAttendeeTest.java
```diff
@@ -16,16 +16,44 @@
...
+    public static void init() throws Exception {
+        MailboxTestUtil.initServer();
+        Provisioning prov = Provisioning.getInstance();
+
+        prov.createAccount("test@zimbra.com", "secret", Maps.<String, Object>newHashMap());
+    }
+
+    @Before
+    public void setUp() throws Exception {
+        MailboxTestUtil.clearData();
```
rupalid: Please also add the following , so we can know which test fails in CircleCI

System.out.println(testName.getMethodName());

store/src/java-test/com/zimbra/cs/mailbox/calendar/Calendar_NewMeetingRequest_RsvpTrue.txt
ykomiyam: Fixed.  Use "@testdomain.com"

store/src/java-test/com/zimbra/cs/mailbox/calendar/ZAttendeeTest.java
ykomiyam: Fixed.

store/src/java-test/com/zimbra/cs/mailbox/calendar/ZAttendeeTest.java
ykomiyam: Fixed.

store/src/java-test/com/zimbra/cs/mailbox/calendar/ZAttendeeTest.java
ykomiyam: Fixed.

store/src/java/com/zimbra/cs/mailbox/calendar/ZAttendee.java
```diff
@@ -142,10 +142,10 @@ public ZAttendee(ZProperty prop) {
         setRole(prop.paramVal(ICalTok.ROLE, null));
         setPartStat(prop.paramVal(ICalTok.PARTSTAT, null));
 
-        String rsvpStr = prop.paramVal(ICalTok.RSVP, "FALSE");
-        boolean rsvp = false;
-        if (rsvpStr.equalsIgnoreCase("TRUE"))
-            rsvp = true;
+        String rsvpStr = prop.paramVal(ICalTok.RSVP, "TRUE");
+        boolean rsvp = true;
```
chinmay15: this is setting rsvp to true by default. did you test by sending appointment with rsvp=false?

store/src/java/com/zimbra/cs/mailbox/calendar/ZAttendee.java
```diff
@@ -142,10 +142,10 @@ public ZAttendee(ZProperty prop) {
         setRole(prop.paramVal(ICalTok.ROLE, null));
         setPartStat(prop.paramVal(ICalTok.PARTSTAT, null));
 
-        String rsvpStr = prop.paramVal(ICalTok.RSVP, "FALSE");
-        boolean rsvp = false;
-        if (rsvpStr.equalsIgnoreCase("TRUE"))
-            rsvp = true;
+        String rsvpStr = prop.paramVal(ICalTok.RSVP, "TRUE");
```
fatkudu: See https://tools.ietf.org/html/rfc2445#section-4.2.17 - the default value of RSVP is FALSE.
I think this change is too intrusive and may lead to bugs.  Presumably, the attendee information from the REPLY is at some point being applied to the ORGANIZER's calendar entry.  The code that does that change needs to consider what should be taken from the reply and what should not.

ykomiyam:
Retract the pull request as per Gren's comment

PR #462, opened 2017-12-14
Bugfix/zcs 3809

rupalid:
ZCS-3809

Added the new libraries requiredd for JWT.

Currently have 2 version of jackson libraries. This is temporary.  Will open a new bug to have a single version of jackson library, resolve compilation errors and unit test failures.

This fix is only so that we can create a fix from JWT branch


shrikant-prasad:
need to raise new PR after taking latest feature/jwt branch code as many changes not specific to this fix have come up in this PR.

PR #463, opened 2017-12-14
ZCS-3258: add null check for empty soap context

shrikant-prasad:
add null check for empty soap context


shrikant-prasad:
@ykomiyam handled that code too.

PR #469, opened 2017-12-22
ZCS-2655:Updating unit test to include different timezones

sneha-patil-synacor:
In extent to https://github.com/Zimbra/zm-mailbox/pull/466

ant test-all passed

### DIFF
--- a/store/src/java-test/com/zimbra/cs/mailbox/calendar/AllDayAcceptInstanceTest.java
+++ b/store/src/java-test/com/zimbra/cs/mailbox/calendar/AllDayAcceptInstanceTest.java
@@ -1,5 +1,6 @@
 package com.zimbra.cs.mailbox.calendar;
 
+import java.io.File;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


store/src/java-test/com/zimbra/cs/mailbox/calendar/AllDayAcceptInstanceTest.java
```diff
@@ -93,13 +109,125 @@ public MailSender getMailSender() {
...
+        System.out.println(testName.getMethodName());
+        MailboxTestUtil.clearData();
+    }
+
+    /*
+     * Test with organizer timezone GMT+
+     * In this case GMT+5:30 Asia/Kolkata
+     */
+    @Test
+    public void testZCS2655PositiveTimeZone() throws Exception {
```
fatkudu: I assume large chunks of this test are the same as the next one.  Would be good to share code as much as possible to ensure they stay in sync and make it clear exactly what the differences are.

PR #471, opened 2017-12-27
ZCS-3828:ZWC affected by Mailsploit

sneha-patil-synacor:
Stripping these categories altogether: 
"Other, Control", "Other, Private Use", "Other, Unassigned", "Other, Format" and "Other, Surrogate"
Replace all "Separator, *" characters with a simple space (U+0020)

Change the default value of zimbraPrefShortEmailAddress to FALSE

Will be updating unit test

### DIFF
--- a/common/src/java/com/zimbra/common/util/StringUtil.java
+++ b/common/src/java/com/zimbra/common/util/StringUtil.java
@@ -28,7 +28,6 @@ import java.io.InputStreamReader;
 import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


common/src/java/com/zimbra/common/mime/InternetAddress.java
```diff
@@ -497,10 +497,40 @@ private void parse(byte[] content, int start, int length, boolean angle) {
...
 
+    private static String sanitizeString(String value) {
+        if (value != null) {
+            StringBuilder newString = new StringBuilder(value.length());
+            for (int offset = 0; offset < value.length();) {
+                int codePoint = value.codePointAt(offset);
+                offset += Character.charCount(codePoint);
+                // Replace invisible control characters and unused code points
+                switch (Character.getType(codePoint)) {
+                case Character.CONTROL: // \p{Cc}
```
plobbes: nit: per internallly doc'd coding standard (Zimbra_Server_Code_Style in engineering wiki) we should indent case one level from the switch statement.

plobbes: replying to @fatkudu - it might be worth while to use a more generic sanitizer although I'll add these caveats:
1. we'd probably need a custom 'policy' as the html sanitizer doesn't seem to ship by default with a sanitizer that would be generic for rfc822 data - we're not dealing with an HTML encoding issue here, but a failure to handle data robustly when displaying the email addresses (and/or when replying to the given address)
2. our sanitization is removing data that could be (really is) valid data per the spec, but just because the spec allows for it, we probably still don't want to allow it due to the potential for abuse like we're trying to address here

sneha-patil-synacor:
@plobbes 
Could you please help understand why do we need install/upgrade step. With this change, we sanitize the address header in the response  to the soap request. We are not changing actual mime that is present on server.

plobbes:
Sorry for the delay.  The idea behind the need for an "upgrade" step is that any time we change defaults, there's typically a corresponding upgrade step that we would perform to have any sites that haven't customized their environment (for this specific attribute/setting) to get the new default via the upgrade step. Even though we're doing steps to make the old default more tolerant of this, if we change the default, then I would expect us to do the same type of upgrade step as usual. Of course, it's up to the PM to confirm that this is their desire/expectation as well.

PR #473, opened 2017-12-28
Bugfix/soap tests

tmclane:
Putting my fixes for each individual soap test from ZCS-1101

### DIFF
--- a/store/src/java/com/zimbra/qa/unittest/SharedImapNotificationTests.java
+++ b/store/src/java/com/zimbra/qa/unittest/SharedImapNotificationTests.java
@@ -358,7 +358,7 @@ public abstract class SharedImapNotificationTests extends ImapTestBase {
                 try {
                     connection.fetch("1:*", "(ENVELOPE BODY)");
                     return "should not be able to connect; connection should be closed";
-                } catch (CommandFailedException e) {}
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/qa/unittest/SharedImapNotificationTests.java
```diff
@@ -359,6 +359,7 @@ protected String checkResult() throws Exception {
                     connection.fetch("1:*", "(ENVELOPE BODY)");
                     return "should not be able to connect; connection should be closed";
                 } catch (CommandFailedException e) {}
+                  catch (java.net.SocketException e) {}
```
tmclane: Wondering if it would be better to instead make CommandFailedException inherit from SocketException somewhere in the chain?

fatkudu: @tmclane Personally, don't think so, the way my mind thinks, a command could fail for a number of reasons unrelated to sockets.
I'd suggest using the idiom:
`catch (CommandFailedException | java.net.SocketException e) {}`

tmclane:
Updated with @fatkudu 's suggestion.

PR #475, opened 2018-01-02
Bugfix/zimbra 128 contact freq graph user time zone fix

varunrisbud:
As per discussion with the product team, the contact frequency graph should be displayed as per user's current time zone. This affects the calculation of start date of a time range. User's current time zone will be provided in the soap request as 'offsetInMinutes'. If the timezone (offsetInMinutes) is not provided then the start date of the time range will be calculated as per timezone set by the user in preferences.

### DIFF
--- a/store/src/java/com/zimbra/cs/event/SolrEventStore.java
+++ b/store/src/java/com/zimbra/cs/event/SolrEventStore.java
@@ -195,8 +195,7 @@ public abstract class SolrEventStore extends EventStore {
 
     private ZonedDateTime getStartDateForCurrentMonth(ZoneId userZoneId) {
         ZonedDateTime userTZNow = ZonedDateTime.now(userZoneId);
-        ZonedDateTime userTZFirstDayOfMonth = userTZNow.with(TemporalAdjusters.firstDayOfMonth()).truncatedTo(ChronoUnit.DAYS);
-        return userTZFirstDayOfMonth;
+        return userTZNow.with(TemporalAdjusters.firstDayOfMonth()).truncatedTo(ChronoUnit.DAYS);
     }
 
     private ZonedDateTime getStartDateForLastSixMonths(ZoneId userZoneId) throws ServiceException {
@@ -211,16 +210,14 @@ public abstract class SolrEventStore extends EventStore {
 
     private ZonedDateTime getStartDateForCurrentYear(ZoneId userZoneId) {
         ZonedDateTime userTZNow = ZonedDateTime.now(userZoneId);
-        ZonedDateTime userTZFirstDayOfCurrentYear = userTZNow.with(TemporalAdjusters.firstDayOfYear()).truncatedTo(ChronoUnit.DAYS);
-        return userTZFirstDayOfCurrentYear;
+        return userTZNow.with(TemporalAdjusters.firstDayOfYear()).truncatedTo(ChronoUnit.DAYS);
     }
 
     private String getSolrTimeZone(Integer offsetInMinutes) {
         String sign = offsetInMinutes >= 0 ? "+" : "-";
         int hours = Math.abs(offsetInMinutes) / 60;
         int minutes = Math.abs(offsetInMinutes) % 60;
-        String solrTimeZone = String.format("GMT%s%02d%02d", sign, hours, minutes); //e.g. format is GMT-0500 for Eastern time zone.
-        return solrTimeZone;
+        return String.format("GMT%s%02d%02d", sign, hours, minutes); //e.g. format is GMT-0500 for Eastern time zone.
     }
 
     private BooleanQuery getQueryToSearchContactAsSenderOrReceiver(String contact) {
--- a/store/src/java/com/zimbra/cs/event/SolrEventStore.java
+++ b/store/src/java/com/zimbra/cs/event/SolrEventStore.java
@@ -219,10 +219,8 @@ public abstract class SolrEventStore extends EventStore {
         String sign = offsetInMinutes >= 0 ? "+" : "-";
         int hours = Math.abs(offsetInMinutes) / 60;
         int minutes = Math.abs(offsetInMinutes) % 60;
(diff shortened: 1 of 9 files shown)


store/src/java/com/zimbra/cs/event/SolrEventStore.java
```diff
@@ -219,10 +219,8 @@ private String getSolrTimeZone(Integer offsetInMinutes) {
         String sign = offsetInMinutes >= 0 ? "+" : "-";
         int hours = Math.abs(offsetInMinutes) / 60;
         int minutes = Math.abs(offsetInMinutes) % 60;
-        String timeFormat = "%02d";
-        StringBuilder solrTimeZone = new StringBuilder();
-        solrTimeZone.append("GMT").append(sign).append(String.format(timeFormat, hours)).append(String.format(timeFormat, minutes));
-        return solrTimeZone.toString();
+        String solrTimeZone = String.format("GMT%s%02d%02d", sign, hours, minutes); //e.g. format is GMT-0500 for Eastern time zone.
```
gordyt: Minor nit - could just `return String.format...`.

store/src/java/com/zimbra/cs/event/SolrEventStore.java
```diff
@@ -174,39 +180,47 @@ private String getAggregationBucketForContactFrequencyGraphTimeRange(ContactAnal
...
     }
 
-    private LocalDateTime getStartDateForCurrentMonth() {
-        LocalDateTime now = LocalDateTime.now();
-        LocalDateTime firstDayOfMonth = now.with(TemporalAdjusters.firstDayOfMonth()).truncatedTo(ChronoUnit.DAYS);
-        return firstDayOfMonth;
+    private ZonedDateTime getStartDateForCurrentMonth(ZoneId userZoneId) {
+        ZonedDateTime userTZNow = ZonedDateTime.now(userZoneId);
+        ZonedDateTime userTZFirstDayOfMonth = userTZNow.with(TemporalAdjusters.firstDayOfMonth()).truncatedTo(ChronoUnit.DAYS);
+        return userTZFirstDayOfMonth;
```
tmclane: nitpick: Any reason you can't just return the `userTZNow.with(TemporalAdjusters.firstDayOfMonth()).truncatedTo(ChronoUnit.DAYS);` value directly?

store/src/java/com/zimbra/cs/event/SolrEventStore.java
```diff
@@ -174,39 +180,47 @@ private String getAggregationBucketForContactFrequencyGraphTimeRange(ContactAnal
...
-        return firstDayOfWeek6MonthsBack;
+        ZonedDateTime userTZFirstDayOfCurrentWeekAsConfigured = ZonedDateTime.now(userZoneId).with(WeekFields.of(Locale.US).dayOfWeek(), firstDayOfWeekAdjusted).truncatedTo(ChronoUnit.DAYS);
+        ZonedDateTime userTZFirstDayOfWeek6MonthsBack = userTZFirstDayOfCurrentWeekAsConfigured.minusMonths(6).with(WeekFields.of(Locale.US).dayOfWeek(), firstDayOfWeekAdjusted);
+        return userTZFirstDayOfWeek6MonthsBack;
+    }
+
+    private ZonedDateTime getStartDateForCurrentYear(ZoneId userZoneId) {
+        ZonedDateTime userTZNow = ZonedDateTime.now(userZoneId);
+        ZonedDateTime userTZFirstDayOfCurrentYear = userTZNow.with(TemporalAdjusters.firstDayOfYear()).truncatedTo(ChronoUnit.DAYS);
+        return userTZFirstDayOfCurrentYear;
```
tmclane: same here.

store/src/java/com/zimbra/cs/service/mail/GetContactFrequency.java
```diff
@@ -35,21 +37,27 @@ public Element handle(Element request, Map<String, Object> context)
         Account acct = getRequestedAccount(zsc);
         GetContactFrequencyRequest req = zsc.elementToJaxb(request);
         String contactEmail = req.getEmail();
-        String freqByCSV = req.getFrequencyBy();
+        String freqByOptions = req.getFrequencyBy();
```
fatkudu: What happens if this (or offsetInMinutes) is null?

store/src/java/com/zimbra/qa/unittest/SolrEventStoreTestBase.java
```diff
@@ -286,7 +288,7 @@ void testContactFrequencyCount(ContactAnalytics.ContactFrequencyTimeRange timeRa
     }
 
     protected void testContactFrequencyGraph(ContactAnalytics.ContactFrequencyGraphTimeRange timeRange, SolrEventCallback eventCallback, String collectionName, SolrEventStore eventStore) throws Exception {
-        List<Timestamp> timestamps = getTimestampForContactFrequencyGraphTest(timeRange);
+        List<Timestamp> timestamps = getTimestampForContactFrequencyGraphTest(timeRange, ZoneId.of("GMT-0800")); //LA time zone
```
fatkudu: I assume "GMT+0000" is a valid ZoneId and it doesn't collapse to "GMT"?  Might be worth a test to confirm.

varunrisbud: Ran test for GMT+0000 timezone. Seems to be working. Thanks for pointing this out. I didn't even know that "GMT+0000" was a valid timezone Id.

store/src/java/com/zimbra/cs/service/mail/GetContactFrequency.java
varunrisbud: If the offset is null we calculated the offset from UTC to the preferred timezone set by the user. This is handled on line 56 and in ContactAnalytics class. The 'FrequencyBy' is a required option, so I think an exception would be raised.

PR #476, opened 2018-01-03
ZIMBRA-131: SmartFolder functionality

ilyusha:
Upcoming NextGen work requires the server to use machine learning to classify incoming emails into one or more configurable "smart folders", such as `important`, `social`, or `finance`.  This makes a smart folder very similar to tag, with the difference being that the tagging is performed by the server instead of by the client.

#### Implementation as Tags
`SmartFolder` subclasses `Tag` to enable us to take advantage of the existing tagging infrastructure. To avoid potential name collisions between smart folders and user-created tags, the internal tag corresponding to a smart folder has its name prefixed with two backslashes `\\`. This places smart folders into the namespace carved out for Flags, which begin with a single backslash. The method `SmartFolder.getSmartFolderName()` returns the "external" name which does not include the leading backslashes. This makes it possible for there to be a tag and smart folder with the same user-facing name.

While smart folder names are stored alongside tags in the database, the `UnderlyingData::setTags` method separates the input into two distinct `tags` and `smartFolders` arrays. This ensures that smart folders can only be reached by the new `MailItem::getSmartFolders` method, without polluting tag-related calls.

#### SmartFolder management
`Mailbox.java` has a new method `syncSmartFolders`that takes a `SmartFolderProvider` instance. The provider exposes a set of strings representing the smart folders that should be present on account. These names are compared to the existing smart folders, which are then created/deleted as necessary.
It should be noted that this PR does not provide any concrete implementations of `SmartFolderProvider`, but a future PR will include an implementation that gets these values from LDAP. That PR will also include actually calling `syncSmartFolders` at the appropriate time.

#### SOAP endpoint
As smart folders are currently read-only, the only new  API endpoint is `GetSmartFoldersRequest`, lets the client fetch a list of all smart folders on the account. 

#### Notifications
Notifications about changes to smart folders are returned in SOAP response context headers. These  notifications are a subset of tag notifications, without support for renaming or changing color.

#### ZClient updates
`ZMailbox` has a new method `getSmartFolders()` that invokes `GetSmartFoldersRequest`. Smart folders are cached with the same mechanism used for caching folders and tags. Response header notifications update this cache accordingly. The method `getSmartFolderById` retrieves smart folders from this cache.

#### Tests
The `SmartFolderTest` unit test class includes tests for creating/deleting smart folders, tagging items with smart folders, creating a smart folder with the same names as a tag, and testing the `Mailbox::syncSmartFolders` method.
The `TestSmartFolders` SOAP test class has tests for `GetSmartFolderRequest` and for response header notifications.

#### UPDATE 1/8/2018
* `ZTag` and `ZSmartFolder` now subclass new `ZTagInfo` class to reduce code duplication
* New method`ToXML::encodeTagBase` encodes fields common to both tags and smart folders
* Disallows empty smart folder names
* Several changes recommended by Codacy.
* Rebased on new head of `feature/solr`

### DIFF
--- a/client/src/java/com/zimbra/client/ZMailbox.java
+++ b/client/src/java/com/zimbra/client/ZMailbox.java
@@ -76,6 +76,7 @@ import com.zimbra.client.event.ZCreateFolderEvent;
 import com.zimbra.client.event.ZCreateMessageEvent;
 import com.zimbra.client.event.ZCreateMountpointEvent;
 import com.zimbra.client.event.ZCreateSearchFolderEvent;
+import com.zimbra.client.event.ZCreateSmartFolderEvent;
 import com.zimbra.client.event.ZCreateTagEvent;
 import com.zimbra.client.event.ZCreateTaskEvent;
 import com.zimbra.client.event.ZDeleteEvent;
@@ -89,6 +90,7 @@ import com.zimbra.client.event.ZModifyMailboxEvent;
 import com.zimbra.client.event.ZModifyMessageEvent;
 import com.zimbra.client.event.ZModifyMountpointEvent;
 import com.zimbra.client.event.ZModifySearchFolderEvent;
+import com.zimbra.client.event.ZModifySmartFolderEvent;
 import com.zimbra.client.event.ZModifyTagEvent;
 import com.zimbra.client.event.ZModifyTaskEvent;
 import com.zimbra.client.event.ZModifyVoiceMailItemEvent;
@@ -191,6 +193,8 @@ import com.zimbra.soap.mail.message.GetRelatedContactsRequest;
 import com.zimbra.soap.mail.message.GetRelatedContactsResponse;
 import com.zimbra.soap.mail.message.GetSearchHistoryRequest;
 import com.zimbra.soap.mail.message.GetSearchHistoryResponse;
+import com.zimbra.soap.mail.message.GetSmartFoldersRequest;
+import com.zimbra.soap.mail.message.GetSmartFoldersResponse;
 import com.zimbra.soap.mail.message.IMAPCopyRequest;
 import com.zimbra.soap.mail.message.IMAPCopyResponse;
 import com.zimbra.soap.mail.message.ImportContactsRequest;
@@ -230,6 +234,7 @@ import com.zimbra.soap.mail.type.NewContactAttr;
 import com.zimbra.soap.mail.type.NewContactGroupMember;
 import com.zimbra.soap.mail.type.NewSearchFolderSpec;
 import com.zimbra.soap.mail.type.RelatedContactsTarget;
+import com.zimbra.soap.mail.type.TagInfo;
 import com.zimbra.soap.mail.type.TestDataSource;
 import com.zimbra.soap.type.AccountSelector;
 import com.zimbra.soap.type.AccountWithModifications;
@@ -284,6 +289,7 @@ public class ZMailbox implements ToZJSONObject, MailboxStore {
     private SoapHttpTransport mTransport;
     private SessionPreference mNotifyPreference;
     private Map<String, ZTag> mNameToTag;
+    private Map<String, ZSmartFolder> mNameToSmartFolder;
     private ItemCache mItemCache;
     private ZGetInfoResult mGetInfoResult;
     private ZFolder mUserRoot;
@@ -1051,6 +1057,8 @@ public class ZMailbox implements ToZJSONObject, MailboxStore {
                 event = new ZModifyAppointmentEvent(e);
             } else if (e.getName().equals(MailConstants.E_TASK)) {
                 event = new ZModifyTaskEvent(e);
+            } else if (e.getName().equals(MailConstants.E_SMART_FOLDER)) {
+                event = new ZModifySmartFolderEvent(e);
             }
             if (event != null) {
                 handleEvent(event);
@@ -1119,6 +1127,9 @@ public class ZMailbox implements ToZJSONObject, MailboxStore {
             } else if (e.getName().equals(MailConstants.E_TAG)) {
                 event = new ZCreateTagEvent(new ZTag(e, this));
                 addTag(((ZCreateTagEvent)event).getTag());
+            } else if (e.getName().equals(MailConstants.E_SMART_FOLDER)) {
+                event = new ZCreateSmartFolderEvent(new ZSmartFolder(e, this));
+                addSmartFolder(((ZCreateSmartFolderEvent)event).getSmartFolder());
             }
             if (event != null) {
                 if (events == null) {
@@ -1194,6 +1205,7 @@ public class ZMailbox implements ToZJSONObject, MailboxStore {
         public synchronized void handleRefresh(ZRefreshEvent event, ZMailbox mailbox) {
             ZFolder root = event.getUserRoot();
             List<ZTag> tags = event.getTags();
+            List<ZSmartFolder> smartFolders = event.getSmartFolders();
 
             mItemCache.clear();
             mMessageCache.clear();
@@ -1212,6 +1224,16 @@ public class ZMailbox implements ToZJSONObject, MailboxStore {
                 }
                 for (ZTag tag : tags) {
                     addTag(tag);
+                }
+            }
+            if (smartFolders != null) {
+                if (mNameToSmartFolder == null) {
+                    mNameToSmartFolder = new HashMap<String, ZSmartFolder>();
+                } else {
+                    mNameToSmartFolder.clear();
+                }
+                for (ZSmartFolder smartFolder: smartFolders) {
+                    addSmartFolder(smartFolder);
                 }
             }
         }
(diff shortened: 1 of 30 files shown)


client/src/java/com/zimbra/client/ZMailbox.java
```diff
@@ -1262,6 +1284,12 @@ public synchronized void handleModify(ZModifyEvent event, ZMailbox mailbox) thro
                 if (contact != null) {
                     contact.modifyNotification(mce);
                 }
+            } else if (event instanceof ZModifySmartFolderEvent) {
+                ZModifySmartFolderEvent msfe = (ZModifySmartFolderEvent) event;
+                ZSmartFolder smartFolder = getSmartFolderById(msfe.getId());
+                if (smartFolder != null) {
```
fatkudu: Worth a warning if smartFolder is null?

client/src/java/com/zimbra/client/ZSmartFolder.java
```diff
@@ -0,0 +1,76 @@
...
+import com.zimbra.common.soap.Element;
+import com.zimbra.common.soap.MailConstants;
+import com.zimbra.soap.mail.type.RetentionPolicy;
+import com.zimbra.soap.mail.type.TagInfo;
+
+public class ZSmartFolder implements Comparable<ZSmartFolder>, ZItem, ToZJSONObject {
+
+    private String id;
+    private String name;
+    private int unreadCount;
```
fatkudu: I notice that ZTag also has a `mColor`.  Is there any mileage in extending `ZTag`?  Hmm, or perhaps that makes the client mirror the underlying server implementation rather than reflecting the client view.  Either way, I could imagine a re-org with both extending another class being of value.

client/src/java/com/zimbra/client/event/ZCreateSmartFolderEvent.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2006, 2007, 2009, 2010, 2011, 2013, 2014, 2016 Synacor, Inc.
```
fatkudu: Worth just using 2018 in copyright notices for new code.

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -2895,7 +2900,7 @@ private PendingLocalModifications snapshotModifications(PendingLocalModification
                     }
                     snapshot.recordModified(folder, chg.why, (MailItem) chg.preModifyObj);
                 } else if (item instanceof Tag) {
-                    if (((Tag) item).isListed()) {
+                    if (((Tag) item).isListed() || item instanceof SmartFolder) {
```
fatkudu: Is it worth having a `listedTagOrSmartFolder(item)` method that does this test as seems to be used a few times.

store/src/java/com/zimbra/cs/redolog/op/CreateSmartFolder.java
```diff
@@ -0,0 +1,70 @@
...
+    public CreateSmartFolder() {
+        super(MailboxOperation.CreateSmartFolder);
+        mId = UNKNOWN_ID;
+    }
+
+    public CreateSmartFolder(int mailboxId, String name) {
+        this();
+        setMailboxId(mailboxId);
+        mId = UNKNOWN_ID;
+        mName = name != null ? name : "";
```
fatkudu: Are we OK with having an empty name?  Inclined to think should throw an Exception.

store/src/java/com/zimbra/cs/redolog/op/CreateSmartFolder.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override public void redo() throws Exception {
+        int mboxId = getMailboxId();
+        Mailbox mbox = MailboxManager.getInstance().getMailboxById(mboxId);
+
+        try {
+            mbox.createSmartFolder(getOperationContext(), mName);
+        } catch (MailServiceException e) {
+            String code = e.getCode();
```
fatkudu: Yuck.  Is it possible to refactor the code so that we have a subclass AlreadyExistsMailServiceException and change all places where ALREADY_EXISTS is used to create a MailServiceException to create that instead?  Then we can just catch that here.  We shouldn't really be needing to introspect an exception to find out what type of exception it is...  (I guess this is all a question of how much effort is needed to improve the pre-existing code)

store/src/java/com/zimbra/cs/service/mail/ToXML.java
```diff
@@ -1010,6 +1013,44 @@ public static Element encodeTag(Element parent, ItemIdFormatter ifmt, OperationC
         return encodeTag(parent, ifmt, octxt, tag, NOTIFY_FIELDS);
     }
 
+    public static Element encodeSmartFolder(Element parent, ItemIdFormatter ifmt, OperationContext octxt, SmartFolder smartFolder, int fields)
+    throws ServiceException {
+        Element el = parent.addNonUniqueElement(MailConstants.E_SMART_FOLDER);
+        el.addAttribute(MailConstants.A_ID, ifmt.formatItemId(smartFolder));
```
fatkudu: Can some of this code be shared with Tag?

client/src/java/com/zimbra/client/ZMailbox.java
ilyusha: I modeled this after how ZMailbox handles other modification notifications, which don't do anything if the underlying object doesn't exist. Perhaps it's possible for a modify notification to come through for which the item is not cached on the client?

client/src/java/com/zimbra/client/ZSmartFolder.java
ilyusha: I tried to avoid having the client be aware that SmartFolders are just Tags, since that's an implementation detail. That said, it might be worthwhile to have some shared code there, perhaps with a `ZTagBase` or some such.

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
ilyusha: Ah yeah that's a good idea.

store/src/java/com/zimbra/cs/redolog/op/CreateSmartFolder.java
ilyusha: Agreed, good point.

store/src/java/com/zimbra/cs/redolog/op/CreateSmartFolder.java
ilyusha: Yeah... I'll see how often this is done in the codebase and make that change if it's not too much of a pain.

store/src/java/com/zimbra/cs/service/mail/ToXML.java
ilyusha: Probably, I just tried to avoid making too many changes to the mess that is the `ToXML` class :)

store/src/java/com/zimbra/cs/redolog/op/CreateSmartFolder.java
ilyusha: Handling this in the `Mailbox::createSmartFolder` call, since a RedoOp with an empty name shouldn't be created in the first place

client/src/java/com/zimbra/client/event/ZModifySmartFolderEvent.java
```diff
@@ -0,0 +1,91 @@
...
+            throw new JSONException(se);
+        }
+    }
+
+    @Override
+    public String toString() {
+        try {
+            return String.format("[ZModifySmartFolderEvent %s]", getId());
+        } catch (ServiceException e) {
+            throw new RuntimeException(e);
```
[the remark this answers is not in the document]
ilyusha: This is consistent with the behavior of other `ZModify*` classes

soap/src/java/com/zimbra/soap/mail/message/GetSmartFoldersResponse.java
```diff
@@ -0,0 +1,48 @@
...
+
+@XmlRootElement(name=MailConstants.E_GET_SMART_FOLDERS_RESPONSE)
+public class GetSmartFoldersResponse {
+    /**
+     * @zm-api-field-description Information about smart folders
+     */
+    @XmlElement(name=MailConstants.E_SMART_FOLDER, required=false)
+    private List<TagInfo> smartFolders = Lists.newArrayList();
+
+    public GetSmartFoldersResponse() {
```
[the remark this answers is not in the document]
ilyusha: No-argument constructors are required by JAXB

store/src/java/com/zimbra/cs/mailbox/SmartFolder.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private void init() {
+        if (mData.type != Type.SMARTFOLDER.toByte()) {
+            throw new IllegalArgumentException();
+        }
+    }
+
+
+    static SmartFolder create(Mailbox mbox, int id, String requestedName) throws ServiceException {
```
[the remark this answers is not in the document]
ilyusha: This behavior is consistent with other `MailItem` subclasses.

PR #478, opened 2018-01-09
ZCS-3890:fix issue when request is proxied to another server

shrikant-prasad:
removed JWT related changes from ZimbraAuthToken. Added ZimbraJWToken class which extends ZimbraAuthToken for now. (created ZCS-3971 to refactor ZimbraAuthToken to remove this dependency).
As far as request proxy is concerned, when request flow using jwt is proxied to another server, jwt and it's salt will be added in the envelope context instead of auth token.

### DIFF
--- a/store/src/java/com/zimbra/cs/account/ZimbraAuthToken.java
+++ b/store/src/java/com/zimbra/cs/account/ZimbraAuthToken.java
@@ -30,12 +30,8 @@ import org.apache.commons.httpclient.HttpClient;
 import org.apache.commons.httpclient.HttpMethod;
 import org.apache.commons.httpclient.HttpState;
 import org.apache.commons.httpclient.cookie.CookiePolicy;
-import org.apache.commons.rng.UniformRandomProvider;
-import org.apache.commons.rng.simple.RandomSource;
-import org.apache.commons.text.RandomStringGenerator;
 
 import com.google.common.base.Objects;
-import com.google.common.primitives.Bytes;
 import com.zimbra.common.account.Key.AccountBy;
 import com.zimbra.common.auth.ZAuthToken;
 import com.zimbra.common.localconfig.LC;
@@ -47,9 +43,7 @@ import com.zimbra.common.util.BlobMetaData;
 import com.zimbra.common.util.Log;
 import com.zimbra.common.util.LogFactory;
 import com.zimbra.common.util.MapUtil;
-import com.zimbra.common.util.StringUtil;
 import com.zimbra.common.util.ZimbraCookie;
-import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.cs.account.auth.AuthMechanism.AuthMech;
 import com.zimbra.cs.ephemeral.EphemeralInput;
 import com.zimbra.cs.ephemeral.EphemeralInput.AbsoluteExpiration;
@@ -58,7 +52,6 @@ import com.zimbra.cs.ephemeral.EphemeralKey;
 import com.zimbra.cs.ephemeral.EphemeralLocation;
 import com.zimbra.cs.ephemeral.EphemeralStore;
 import com.zimbra.cs.ephemeral.LdapEntryLocation;
-import com.zimbra.cs.service.util.JWTUtil;
 
 /**
  * @since May 30, 2004
@@ -73,7 +66,6 @@ public class ZimbraAuthToken extends AuthToken implements Cloneable {
     private static final String C_DOMAIN = "domain";
     private static final String C_DLGADMIN = "dlgadmin";
     private static final String C_TYPE = "type";
-    private static final String C_TOKEN_TYPE = "tokenType";
     private static final String C_TYPE_ZIMBRA_USER = "zimbra";
     private static final String C_TYPE_EXTERNAL_USER = "external";
     private static final String C_TYPE_ZMG_APP = "zmgapp";
@@ -89,14 +81,12 @@ public class ZimbraAuthToken extends AuthToken implements Cloneable {
     private static final String C_CSRF = "csrf";
     private static final Map<String, ZimbraAuthToken> CACHE = MapUtil.newLruMap(LC.zimbra_authtoken_cache_size.intValue());
     private static final Log LOG = LogFactory.getLog(AuthToken.class);
-    private static final int  SALT_LENGTH = 20;
 
-    private String accountId;
+    protected String accountId;
     private String adminAccountId;
     private int validityValue = -1;
-    private long expires;
-    private long issuedAt;
-    private String encoded;
+    protected long expires;
+    protected String encoded;
     private boolean isAdmin;
     private boolean isDomainAdmin;
     private boolean isDelegatedAdmin;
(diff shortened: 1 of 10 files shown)


store/src/java/com/zimbra/cs/account/ZimbraAuthToken.java
```diff
@@ -235,16 +236,16 @@ protected ZimbraAuthToken(String encoded) throws AuthTokenException {
 
             String tid = (String)map.get(C_TOKEN_ID);
             if(tid !=null) {
-            	try {
-            		tokenID = Integer.parseInt(tid);
-            	} catch (NumberFormatException e) {
-            		tokenID = -1;
-                }
+            try {
```
Prashantsurana: Fix indentations.

shrikant-prasad:
IMO, there is no need of ZimbraJWToken as it doesn't have anything else other than jwtId to add on top of ZimbraAuthToken.

fatkudu:
@Prashantsurana. I do kind of agree as philosophically, this is a different type of authentication.  We could perhaps use a base class which happens to have anything that is shared between `ZimbraAuthToken` and `ZimbraJWToken`

shrikant-prasad:
@Prashantsurana @fatkudu : JWT is different type of authentication but once the authentication is done, internally we are generating ZimbraAuthToken only to minimise the code changes(as the workflow is based on operation context which makes use of ZimbraAuthToken). If we decide to add new class ZimbraJWToken(which would basically extend ZimbraAuthToken as it shares everything with it, and just add one jwtId new field),  changes needs to be done at several places to make all existing flows work.

fatkudu:
@shrikant-prasad @Prashantsurana - presumably if ZimbraJWToken extended ZimbraAuthToken only minimal changes would be required as ZimbraJWToken would be a ZimbraAuthToken?

shrikant-prasad:
the problem occurs when auth token is re-created from it's encoded string. currently an object of ZimbraAuthToken is created from the string. if we add ZimbraJWToken, now will have to first decode the string to find type of token, then accordingly create instance of either ZimbraAuthToken or ZimbraJWToken.

Prashantsurana:
Mixing up ZimbraAuthToken with JWT difficult to maintain. 
In SoapContext  you can always pass JWT instead of encoded authtoken.
Current  code is difficult to maintain. 


code: 
```
            String tokenTypeCode = (String)map.get(C_TOKEN_TYPE);
            if (tokenTypeCode != null) {
                tokenType = TokenType.fromCode(tokenTypeCode);
            }
```

Here possibility of NPE for previous AuthToken.  Similar issues can happen if we use conditional logic. 

While proxying just pass JWT add support for JWT in Context. 
ZimbraJWToken can inherit from AuthToken as it completely different token. 
There is no need to decode the JWT and create ZimbraAuthToken. This will keep dependency on ZimbraAuthToken forever.

store/src/java/com/zimbra/cs/service/account/EndSession.java
```diff
@@ -56,13 +56,13 @@ public Element handle(Element request, Map<String, Object> context)
             try {
                 AuthToken at = zsc.getAuthToken();
                 if (at.isJWT()) {
```
grishick: I would rather encapsulate the logic that removes token from session in the class that implements the token instead of having if/else branches inside EndSession.java.

grishick:
@shrikant-prasad is there a need to register JWT token in LDAP/SSDB the same way we are registering Zimbra auth tokens? The need to register zimbra auth token comes from the danger of attacker stealing a token and reusing it. Is the same kind of attack possible with JWT?

shrikant-prasad:
@grishick there is no need to register JWT in ldap/ssdb. 
only on logout, jwt id is added in a blacklist stored in ldap/ssdb.

grishick:
@shrikant-prasad with current implementation JWT tokens seem to be registered and deregistered in LDAP/SSDB. Could you point me to the implementation of the blacklist?

shrikant-prasad:
@grishick  jwt is not registered in ldap. only logged out jwt id is added in ldap.
 in derigster method :  acct.addInvalidJWTokens(jwtId, server_version, expiration);

store/src/java/com/zimbra/cs/service/account/EndSession.java
shrikant-prasad: Inside the auth token class, there is no reference of the context which holds the http req and response, hence it won't be possible to set the cookie inside the auth token class. Thats why this logic is here in EndSession.

grishick:
@shrikant-prasad thank you for the answers. One more thing, since you are adding an ephemeral attribute, it should be added to zmmigrateattrs

shrikant-prasad:
@Prashantsurana @fatkudu @grishick removed JWT related changes from ZimbraAuthToken. Added ZimbraJWToken class which extends ZimbraAuthToken for now. (created ZCS-3971 to refactor ZimbraAuthToken to remove this dependency).
As far as request proxy is concerned, when request flow using jwt is proxied to another server, jwt and it's salt will be added in the envelope context instead of auth token.

store/src/java/com/zimbra/cs/account/ZimbraJWToken.java
```diff
@@ -0,0 +1,145 @@
...
+import com.zimbra.common.util.StringUtil;
+import com.zimbra.common.util.ZimbraLog;
+import com.zimbra.cs.account.auth.AuthMechanism.AuthMech;
+import com.zimbra.cs.ephemeral.EphemeralInput.AbsoluteExpiration;
+import com.zimbra.cs.ephemeral.EphemeralInput.Expiration;
+import com.zimbra.cs.service.util.JWTUtil;
+
+import io.jsonwebtoken.Claims;
+
+public class ZimbraJWToken extends ZimbraAuthToken {
```
Prashantsurana: Better to avoid extending from ZimbraAuthToken  there are several public method in ZimbraAuthToken are not applicable ZimbraJWToken. 

As we have discussed, we need proper inheritance in place with cleaner approach.

store/src/java/com/zimbra/cs/account/ZimbraJWToken.java
```diff
@@ -0,0 +1,145 @@
...
+
+    public static AuthToken getAuthToken(String jwt, String salt) throws AuthTokenException {
+        AuthToken at = null;
+        try {
+            Claims body = JWTUtil.validateJWT(jwt, salt);
+            Account acct = Provisioning.getInstance().getAccountById(body.getSubject());
+            if (acct == null ) {
+                throw AccountServiceException.NO_SUCH_ACCOUNT(body.getSubject());
+            }
+            at = new ZimbraJWToken(acct, body.getExpiration().getTime(), body.getId(), jwt);
```
fatkudu: Can body.getExpiration() return null?  If so, it needs handling.

store/src/java/com/zimbra/soap/ProxyTarget.java
```diff
@@ -146,17 +146,19 @@ public Element dispatch(Element request, ZimbraSoapContext zsc) throws ServiceEx
                 transport.setTimeout((int) Math.min(mTimeout, Integer.MAX_VALUE));
 
             transport.setResponseProtocol(zsc.getResponseProtocol());
-            AuthToken authToken = AuthToken.getCsrfUnsecuredAuthToken(zsc.getAuthToken());
-            if (authToken != null && !StringUtil.isNullOrEmpty(authToken.getProxyAuthToken())) {
-                transport.setAuthToken(authToken.getProxyAuthToken());
+            if (!zsc.getAuthToken().isJWT()) {
```
fatkudu: The previous code implied that perhaps `zcs.getAuthToken()` could return null.  Needs handling if that is possible.

PR #479, opened 2018-01-09
ZIMBRA-129: In-Memory Event Metrics

ilyusha:
Adding machine learning to ZCS requires the ability to generate features to be used as inputs to classifiers. One important category of features are those calculated from event data to drive "important email" classification, such as sender contact frequency, read rate, or reply rate. Since these features potentially need to be generated for each incoming message, the mailbox server needs to be able to efficiently calculate this data without issuing a Solr query every time, which may be prohibitively expensive.

To enable event-driven ML features described above, This PR introduces the concept of  an `EventMetric`, which is an in-memory representation of some calculation derived from event data, which can be incrementally updated by the event stream. 

This code relies heavily on generics. This is done in order to provide as much flexibility as possible in handling different metric types, and to allow future development of other calculated metrics.

#### Incremental Metrics
A low-level `IncrementableMetric` interface represents a wrapper for some internal state that can expose some value. It can be queried for this value, and updated with an `Increment`, which updates the internal state accordingly. Since metric calculations differ in what they return and how they are updated, `IncrementMetric` is parameterized by both `T`, the return type, and `I`, the `Increment` type.

Two `IncrementableMetric` implementations are used in this PR:
* `ValueMetric` tracks a value represented by an `Integer`, with `IntIncrement` used to simply add the increment to the current value
* `RatioMetric` is used when the underlying calculation is a fraction. The return type is a `Double`; the update process adds the current numerator and denominator to the ones passed in by the provided `RatioIncrement`.

#### EventMetric
An `EventMetric` is the primary focus of this PR. It contains an underlying state stored in an `IncrementableMetric` field, and has the ability to increment this state with an `Increment` derived from a list of events.
`EventMetric` is parameterized by three types:
* `M` is the type of the underlying `IncrementableMetric`
* `T` is the return type of the `getValue()` method
* `I` is the type of `Increment` the underlying metric is incremented by

Three subclasses of `EventMetric` are implemented:
* `ContactFrequencyMetric` that tracks the number of sent/received/total interactions with the a contact over one of several time horizons 
* `EventRatioMetric` that calculates the ratio between two message event types (seen, read, replied), either globally or for a given contact
* `TimeDeltaMetric` that calculates the average time delta between two message event types, either globally or for a given contact.

##### EventMetric Initialization
The initial `IncrementableMetric` value underlying an `EventMetric` is fetched using a `MetricInitializer` class.  In production, an `EventStoreInitializer` is used, which executes the appropriate `EventStore` calls to get the initial data. For unit tests, a dummy initializer is provided. 
This initializer has three methods:
* `getInitialData()`returns an `IncrementableMetric` of type `M`
* `getTimestamp()` returns the millisecond timestamp of when the initialization happened
* `getMetricLifetime()` returns a lifetime, in milliseconds, of how long the initialized value is valid for (see "EventMetric Expiration" below)

#### EventMetric Parameters
An `EventMetric` has parameters passed to it. For example, `ContactFrequencyMetric` requires a `ContactFrequencyTimeRange` and `ContactFrequencyEventType` parameters. Likewise, both `EventRatioMetric` and `TimeDeltaMetric` are defined by two `EventType` values ( numerator/denominator and from/to, respectively). Additionally, All of the concrete `EventMetric` subclasses take a String "contact" parameter. This is required by `ContactFrequencyMetric`, but optional for the other two; this allows us to calculate a ratio or time delta metric for the sender of a message relative to the global account value.

#### EventMetric Expiration
A metric like "contact frequency" may have a timeframe associated with it (for example, "sent frequency for the last 24 hours"). For time-based metrics, we don't want to increment them indefinitely, as we will exceed the "window" the metric represents. In fact, we cannot have a truly accurate time-based metric without keeping a record of the timestamps of each data point that contributes to the total - and while this is theoretically possible (we could implement an `IncrementableMetric` that does this), it seems unnecessary.

Instead, `MetricInitializer` defines an optional millisecond lifetime value for the data it returns. If the lifetime is positive, then when sufficient time passes, the `EventMetric.getValue()` method will re-fetch a fresh value from the initializer instead of using the internal `IncrementableMetric` value, "resetting" it to the ground truth.

Specifically, the lifetimes are currently configured as follows:
* "last day" contact frequency has a lifetime of 12 hours
* "last week" contact frequency has a lifetime of 1 day
* "last month" contact frequency has a lifetime of 7 days

It should be noted that this may result in these time-bounded metrics going "outside the bounds" of their time window somewhat. However, given what these metrics are used for, this should not pose any real problems.

#### Updating EventMetrics
EventMetrics are updated using the `increment` method, which takes a list of `Event` objects. It is up to the implementation to filter this event list accordingly construct an `Increment` of parameterized type `I`, which is then used to update the internal `IncrementableMetric`. 
For example, `EventRatioMetric` calculates how many events in the provided list match the numerator and denominator conditions, and create a `RatioIncrement` that updates the `RatioMetric` value.

#### EventMetric management
The `EventMetricManager` is a singleton class used to manage event metrics. It contains a registry of Factory classes used to instantiate `EventMetric` instances; these are mapped from `MetricType` enums.
The `EventMetricManager::getMetrics` method takes an account ID and returns a cached `AccountEventMetrics` instance. This, in turn, provides access to all the currently loaded metrics for a given account. A metric can then be retrieved by a `MetricKey`, which is parameterized on the same three type parameters as `EventMetric`. This is done so that the key contains all the information necessary to instantiate the requested metric instance with the appropriate Factory if it does not exist in the cache.
The `AccountEventMetrics::incrementAll` method is used to increment all metrics currently loaded for the account with the given event list.

#### Updating Log Handler
A special `EventLogHandler` instance is used to update event metrics currently loaded into memory. This log handler is created by default alongside the other user-configured ones. It uses the existing `BatchingEventLogger`, with a `BatchedEventCallback` that calls `AccountEventMetrics::incrementAll` appropriately. Note that this means that an EventMetric may not always represent the most up-to-date view of the data, as event batching introduces a lag. As with the metric lifetime issue described above, this should be OK.

#### Changes to EventStore interface
A slight change has been made to the `EventStore` interface to allow for calculating arbitrary ratios and time deltas. Previously, the interface had methods for specific calculations such as `getPercentageRepliedEmails` and `getAvgTimeToOpenEmail`. These have been replaced with more general methods `getEventRatio` and `getEventTimeDeta`. `SolrEventStore` doesn't change in any meaningful way, as it already had private methods that supported these calculations; these methods have simply been made public. Additionally, the return type of these methods has been changed from `Double` to `RatioMetric` so that they can be used by EventMetric initializers.

#### Testing
`EventMetricTest` is a unit test class containing test methods for each implementation of `IncrementableMetric` as well as every `EventMetric` type.

#### UPDATE 1/10
Removed `MetricInitializer.getTimestamp() method, small fixes to unit tests

#### UPDATE 1/11
Rebased on top of new `feature/solr` branch head

### DIFF
--- a/store/src/java-test/com/zimbra/cs/event/analytics/EventMetricTest.java
+++ b/store/src/java-test/com/zimbra/cs/event/analytics/EventMetricTest.java
@@ -1,0 +1,263 @@
+package com.zimbra.cs.event.analytics;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.util.List;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.cs.event.Event;
+import com.zimbra.cs.event.Event.EventType;
+import com.zimbra.cs.event.analytics.AccountEventMetrics.MetricKey;
+import com.zimbra.cs.event.analytics.ContactFrequencyMetric.ContactFrequencyParams;
+import com.zimbra.cs.event.analytics.EventDifferenceMetric.EventDifferenceParams;
+import com.zimbra.cs.event.analytics.EventMetric.MetricInitializer;
+import com.zimbra.cs.event.analytics.EventMetric.MetricType;
+import com.zimbra.cs.event.analytics.IncrementableMetric.Increment;
+import com.zimbra.cs.event.analytics.RatioMetric.RatioIncrement;
+import com.zimbra.cs.event.analytics.ValueMetric.IntIncrement;
+import com.zimbra.cs.event.analytics.contact.ContactAnalytics.ContactFrequencyEventType;
+import com.zimbra.cs.event.analytics.contact.ContactAnalytics.ContactFrequencyTimeRange;
+import com.zimbra.cs.event.logger.BatchingEventLogger;
+import com.zimbra.cs.event.logger.EventMetricCallback;
+
+public class EventMetricTest {
+
+    private static final String CONTACT_1 = "test1@zimbra.com";
+    private static final String CONTACT_2 = "test2@zimbra.com";
+    private static final String ACCT_ID = "accountId";
+    private BatchingEventLogger logger;
+    private AccountEventMetrics metrics;
+
+    @Before
+    public void setUp() throws Exception {
+        logger = new BatchingEventLogger(10, 0, new EventMetricCallback());
+        metrics = EventMetricManager.getInstance().getMetrics(ACCT_ID);
+    }
+
+    @Test
+    public void testValueMetric() throws Exception {
+
+        MetricInitializer<ValueMetric, Integer, IntIncrement> initializer = new MetricInitializer<ValueMetric, Integer, IntIncrement>() {
+
+            @Override
+            public ValueMetric getInitialData() {
+                return new ValueMetric(0);
+            }
+
+            @Override
+            public long getMetricLifetime() {
+                return 0;
+            }
+        };
+
+        EventMetric<ValueMetric, Integer, IntIncrement> metric = new EventMetric<ValueMetric, Integer, IntIncrement>("testAccountID", null, initializer) {
+
+            @Override
+            protected IntIncrement getIncrement(List<Event> events) throws ServiceException {
+                return new IntIncrement(1);
+            }
+        };
+
+        assertEquals("initial value should be 0", (Integer)0, metric.getValue());
+        metric.increment(null);
+        metric.increment(null);
+        assertEquals("value after two increments should be 2", (Integer)2, metric.getValue());
+    }
+
+    @Test
+    public void testRatioMetric() throws Exception {
+
+        MetricInitializer<RatioMetric, Double, RatioIncrement> initializer = new MetricInitializer<RatioMetric, Double, RatioIncrement>() {
+
+            @Override
+            public RatioMetric getInitialData() {
+                return new RatioMetric(0d, 1);
+            }
+
+            @Override
+            public long getMetricLifetime() {
+                return 0;
+            }
+        };
+
+        EventMetric<RatioMetric, Double, RatioIncrement> metric = new EventMetric<RatioMetric, Double, RatioIncrement>("testAccountID", null, initializer) {
+
+            @Override
+            protected RatioIncrement getIncrement(List<Event> events) throws ServiceException {
+                return new RatioIncrement(1d, 2);
+            }
+        };
+
+        assertEquals("initial value should be 0", new Double(0), metric.getValue());
+        metric.increment(null);
+        assertEquals("new value should be 1/3", new Double(1d/3), metric.getValue());
+        metric.increment(null);
+        assertEquals("new value should be 2/5", new Double(2d/5), metric.getValue());
+    }
+
+    private void logContactFrequencyEvents(String contactEmail) {
+        logger.log(Event.generateEvent(ACCT_ID, 1, contactEmail, "me@zimbra.com", EventType.RECEIVED, null, null, System.currentTimeMillis()));
+        logger.log(Event.generateSentEvent(ACCT_ID, 2, "me@zimbra.com", contactEmail, null, null, System.currentTimeMillis()));
+        logger.log(Event.generateSentEvent(ACCT_ID, 3, "me@zimbra.com", "other@zimbra.com", null, null, System.currentTimeMillis()));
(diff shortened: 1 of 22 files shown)


store/src/java-test/com/zimbra/cs/event/analytics/EventMetricTest.java
```diff
@@ -0,0 +1,270 @@
...
+        assertTrue("initializer.getInitialValue() should have been triggered", initializedFlag.get() == true);
+        assertEquals("initial contact1 TTO should be 0", new Double(0), contactTTO.getValue());
+        assertEquals("initial global TTO should be 0", new Double(0), globalTTO.getValue());
+
+        long timestamp = System.currentTimeMillis();
+        //1 second difference for contact1
+        logMsgEvents(CONTACT_1, 1, 1, EventType.SEEN, timestamp-1000);
+        logMsgEvents(CONTACT_1, 1, 1, EventType.READ, timestamp);
+
+        //2 second difference for contact2
```
grishick: why do you need to log events for CONTACT_2 if you are not checking them in this test?

store/src/java-test/com/zimbra/cs/event/analytics/EventMetricTest.java
```diff
@@ -0,0 +1,270 @@
...
+        logMsgEvents(CONTACT_2, 2, 1, EventType.SEEN, timestamp-2000);
+        logMsgEvents(CONTACT_2, 2, 1, EventType.READ, timestamp);
+
+        assertEquals("new contact1 TTO should be 1 second", new Double(1), contactTTO.getValue());
+        assertEquals("initial global TTO should be 1.5 seconds", new Double(1.5), globalTTO.getValue());
```
grishick: what makes initial global TTO to change here after it was set to 0 by the initializer?

store/src/java/com/zimbra/cs/event/analytics/EventMetric.java
```diff
@@ -0,0 +1,139 @@
...
+
+
+    /**
+     * Helper interface used to initialize EventMetric
+     */
+    public static abstract class MetricInitializer<M extends IncrementableMetric<T, I>, T, I extends Increment> {
+
+        public abstract M getInitialData() throws ServiceException;
+
+        public long getTimestamp() {
```
grishick: I am not sure about the purpose of encapsulating System.getCurrentTimeMillis() in MetricInitializer.

store/src/java-test/com/zimbra/cs/event/analytics/EventMetricTest.java
```diff
@@ -0,0 +1,270 @@
...
+        }
+        //log an event for a different account id to make it doesn't affect the metrics
+        logger.log(Event.generateEvent("other", startingMsgId, contactEmail, "me@zimbra.com", eventType, null, null, timestamp));
+        logger.sendAllBatched();
+    }
+
+    @Test
+    public void testContactFrequency() throws Exception {
+
+        AtomicBoolean initialized = new AtomicBoolean(false);
```
grishick: why do you need to use an AtomicBoolean and not simple boolean here and in similar places?

store/src/java-test/com/zimbra/cs/event/analytics/EventMetricTest.java
ilyusha: Logging events for CONTACT_2 changes the global time-to-open to 1.5 seconds.
Although I just realized there's a copy/paste error in the error message below this.

store/src/java-test/com/zimbra/cs/event/analytics/EventMetricTest.java
ilyusha: This is a copy/paste error, it should say "new global TTO". Logging a TTO of 1  second for contact1 and 2 seconds for contact2 makes the global TTO 1.5.

store/src/java/com/zimbra/cs/event/analytics/EventMetric.java
ilyusha: Yeah you're right, I think I originally wanted to do more with this but it didn't pan out. I'll get rid of this method.

store/src/java-test/com/zimbra/cs/event/analytics/EventMetricTest.java
ilyusha: I originally had dummy initializers defined as anonymous classes, with the getInitialData() calls setting a flag outside the class scope; I had to use an AtomicBoolean for this. When I refactored that to an inner class, I kept the AtomicBoolean even though it's unnecessary. I'll change this.

PR #480, opened 2018-01-10
ZCS-3888:Better SOAP BatchRequest log

ykomiyam:
Although individual SOAP requests are logged, the system administrator
can't see which SOAP requests are submitted along with the BatchRequest.
It is useful for the user support purpose to get the list of the real
"handleable" requests in the BatchRequest.

Changed the existing log text to display the list of the SOAP requests
under the BatchRequest.

### DIFF
--- a/store/src/java/com/zimbra/soap/SoapEngine.java
+++ b/store/src/java/com/zimbra/soap/SoapEngine.java
@@ -1,7 +1,7 @@
 /*
  * ***** BEGIN LICENSE BLOCK *****
  * Zimbra Collaboration Suite Server
- * Copyright (C) 2004, 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016 Synacor, Inc.
+ * Copyright (C) 2004, 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018 Synacor, Inc.
  *
  * This program is free software: you can redistribute it and/or modify it under
  * the terms of the GNU General Public License as published by the Free Software Foundation,
@@ -339,8 +339,15 @@ public class SoapEngine {
             if (!doc.getName().equals("BatchRequest")) {
                 doCsrfCheck =  false;
             } else {
-                LOG.info("Only BatchRequest does not have a handler mapped to it. Request: %s, does not have a "
-                    + "handler, log for future handling.", path);
+                StringBuilder sb = new StringBuilder();
+                for (Element req : doc.listElements()) {
+                    if (sb.length() > 0) {
+                        sb.append(",");
+                    }
+                    sb.append(req.getName());
+                }
+                LOG.info("BatchRequest [%s] contains %d sub-request(s): %s",
+                        path, doc.listElements().size(), sb.toString());
             }
         } else {
             if (doc.getName().equals("AuthRequest")) {


plobbes:
Can we get a better description of the motivation (problem being fixed) by this change? I see that https://bugzilla.zimbra.com/show_bug.cgi?id=108720) was mentioned internally, but I'm not seeing much detail to support the change.

Before this change, each BatchRequest is logged with thread info (including url) similar to:
```
qtp998351292-45503:http://localhost:8080/service/soap/BatchRequest
```

And via loop starting at line 453:
https://github.com/Zimbra/zm-mailbox/blob/244dfb704069c2e739ac8d3d55e542b19a90e0d6/store/src/java/com/zimbra/soap/SoapEngine.java#L453

We subsequently log each batch "subrequest" with the same thread info and prefix the request with (batch) via line 458:
https://github.com/Zimbra/zm-mailbox/blob/244dfb704069c2e739ac8d3d55e542b19a90e0d6/store/src/java/com/zimbra/soap/SoapEngine.java#L458

With this proposed change we're effectively adding a second loop through doc.listElements to list all the subrequests on a single line, which may have some limited value (looking for that to be explained) I suppose but I'm not sure that the value exceeds the cost of looping through the list a second time within mailboxd. The value may be there, I'm just not seeing it yet. While I have a few possible ideas, I'd really like us to spell out what we see as the benefit.

ykomiyam:
[Requirement]
The customer wants the list of the SOAP request in the BatchRequest (1) in one line, not in separate lines, and (2) at the time when the SOAP request is received.   Many ISP are still using 'grep' to search a log event from mailbox.log;   finding one event in one line is much easier than combining multiple lines.  And the log event you mentioned (SoapEngine.java, Line 458) is logged after the request is processed; which may not be logged at all if an exception is thrown.

[Background]
Originally, this request arose from the compliant from the system  at the large ISP; when they have a call from their end user, they need to know the exact actions which he/she took on the web mail pages.  Ideally they want the information about "which item (button, pulldown menu, etc.) was clicked when", but the Zimbra logs the events only per SOAP API call.  For most of the SOAP calls but BatchRequest, the operator can know the end user's action to the log events easily because each event represents the user's operation respectively.  But, because various end user action cause the BatchRequest, the operator can't even guess which action was taken without the detailed SOAP request inside the BatchRequest.  
On the other hand, mailbox.log has already logged an event "Only BatchRequest does not have a handler mapped to it. Request: <path>, does not have a ..." when the BatchRequest is received.  Therefore, I modified it to add the list of the requests in the BatchRequest.

store/src/java/com/zimbra/soap/SoapEngine.java
```diff
@@ -339,8 +339,15 @@ private Element dispatch(String path, Element envelope, Map<String, Object> cont
...
-                LOG.info("Only BatchRequest does not have a handler mapped to it. Request: %s, does not have a "
-                    + "handler, log for future handling.", path);
+                StringBuilder sb = new StringBuilder();
+                for (Element req : doc.listElements()) {
+                    if (sb.length() > 0) {
+                        sb.append(",");
+                    }
+                    sb.append(req.getName());
+                }
+                LOG.info("BatchRequest [%s] contains %d sub-request(s) [%s]",
```
plobbes: minor suggestion: change
  "BatchRequest [%s] contains %d sub-request(s) [%s]"
to
  "BatchRequest [%s] contains %d sub-request(s): %s"

This is very minor, but it would seem (to me) to be more consistent with how we currently log other items.

PR #481, opened 2018-01-10
ZCS-2518 initialize LicenseExtension in ImapDaemon

tmclane:
Currently this emits a 'WARNING' level message if the extension is not found but continues loading in the case of FOSS and causes the appropriate 'ALERT' to be sent to an IMAP client connecting to the IMAPD when the license is bad / expired.

The rules for the alert appearance are the same as the internal imap server since it is using the same code and mechanism.

IMAPD

    telnet localhost 8143
    Trying ::1...
    Connected to localhost.
    Escape character is '^]'.
    * OK zcs-dev-testbed.dev Zimbra IMAP4rev1 server ready
    * OK [ALERT] Your mail server's license is missing, invalid or expired.

Mailbox / IMAP

    telnet localhost 7143
    Trying ::1...
    Connected to localhost.
    Escape character is '^]'.
    * OK zcs-dev-testbed.dev Zimbra IMAP4rev1 server ready
    * OK [ALERT] Your mail server's license is missing, invalid or expired.

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/ImapDaemon.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapDaemon.java
@@ -22,6 +22,7 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.util.Properties;
 
+import com.zimbra.cs.extension.ExtensionUtil;
 import org.apache.log4j.PropertyConfigurator;
 
 import com.zimbra.common.calendar.WellKnownTimeZones;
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/imap/ImapDaemon.java
```diff
@@ -119,6 +120,8 @@ public static void main(String[] args) {
                     errorExit("ImapDaemon: imapd service was unable to intialize timezones EXITING.");
                 }
 
+                ExtensionUtil.init("com.zimbra.cs.network.license.LicenseExtension");
```
fatkudu: Would be good to have a comment saying that this loads the LicenseExtension if it is present and that this is used to allow users to be notified of licensing issues rather than just getting unexpected behavior.

Talking of which, what happens on FOSS systems?  The comment should also re-assure that this
What testing has been done? with this change, do users get a warning if their license has expired?

tmclane: Yes the alert is now returned when you open a connection to the server.
In the case of FOSS it will return a 'WARN' level log message in the imapd.log file but has no other impact.

I'll add the comment.

PR #483, opened 2018-01-11
ZIMBRA-130: Feature Generation for Machine Learning

ilyusha:
This PR builds on the event metric work from #ZIMBRA-129 to introduce classifier feature management. This is a direct prerequisite to adding the concept of "Classifiers" to ZCS.

In machine learning, a _feature_ is some measurable property of the object to be classified. A set of features is the input into a classifier at both the training and classification steps.

For email classification, we need to be able to generate a wide variety of features. Some are event-driven, representing some metric related to the sender of the email. Others are properties of the message itself - how many recipients it has, whether it's part of a conversation, etc.
It should be noted that while content-based features such as subject and body words will certainly be used for classification, the task of parsing text to extract these tokens is left to the machine learning backend. An upcoming PR for classifier management will address this.

The key components in this PR are as follows:

* `Classifiable` is a new marker interface used throughout the ML system. The only class currently implementing this is `Message`.
* A `Feature` is a simple abstract class for some computed value. It is parameterized by `T`, which is the return type of the `getFeatureValue()` method. Subclasses are described below.
* A `FeatureFactory` builds a feature for some `Classifiable` item. It is parameterized by `T` as above, as well as by `C`, which implements `Classifiable`. Subclasses are described below.
* A `FeatureSpec` lets us define a single feature in a declarative way, without directly referencing factories. Its utility will be apparently in future PRs, where we will need to easily define classifier features outside of java code.
* A `FeatureSet` is a combination of one or more FeatureSpecs representing all features used by a classifier.
* `ComputedFeatures` is a wrapper for a list of `Feature` instances created for a given `Classifiable` item.
* `FeatureParam` represents a single key/value pair used as a parameter to a feature factory. It is parameterized by the underlying value type.  The `ParamKey` enum is used as the key; the enum includes a short string representation used for serializing the parameter.
* `FeatureParams` represents a collection of `FeatureParam` instances, used for `FeatureFactory` configuration. More details below.

#### Feature Implementations
The following `Feature` subclasses are provided:
* `PrimitiveFeature` simply exposes some value passed in through the constructor.
* `MetricFeature` is a wrapper for some underlying `EventMetric`.
* `RatioFeature` is a decorator that returns the ratio of two other features.

#### Feature Factories
The following feature factories are defined in this PR:
* `ConversationFeatureFactory` returns a Boolean PrimitiveFeature representing whether the message is part of an existing conversation
* `NumRecipientsFeatureFactory` returns an Integer PrimitiveFeature representing how many recipients the message has. This can be the total number of recipients, or limited to the TO or CC fields.
* `RecipientFieldFeatureFactory` returns an Integer PrimitiveFeature representing which field the recipient is on. It returns 0 if the recipient is on the TO field, 1 if CC, 2 if BCC, and 3 otherwise (for example, if the recipient is part of a distribution list).
* `ContactFrequencyFeatureFactory` returns a MetricFeature wrapping a `ContactFrequencyMetric`.
* `EventRatioFeatureFactory` returns a MetricFeature wrapping an appropriate `EventRatioMetric`
* `EventTimeDeltaFeatureFactory` returns a MetricFeature wrapping an appropriate `TimeDeltaMetric`.

#### Feature Parameters
Feature params that can be used to configure the feature factories described above are:
* METRIC_INITIALIZER: sets the `MetricInitializer` used by `MetricFeature` instances
* `FREQUENCY_TYPE`: sets `ContactFrequencyEventType` enum in `ContactFrequencyFeatureFactory`; one of {`SENT`, `RECEIVED`, `COMBINED`}
* `TIME_RANGE`: sets the `ContactFrequencyTimeRange` enum in `ContactFrequencyFeatureFactory`; one of {`LAST_DAY`, `LAST_WEEK`, `LAST_MONTH`, FOREVER`}
* `RECIPIENT_TYPE`: sets the `RecipientCountType` enum in `NumRecipientsFeatureFactory`; one of {`TO`, `CC`, `ALL`}
* `GLOBAL_RATIO`: optional boolean parameter passed into`EventTimeDeltaFeatureFactory`; if `True`, the contact-specific metric value will be divided by the global account value, "normalizing" it.
* `NUMERATOR`: required `EventType` parameter for `EventRatioFeatureFactory`
* `DENOMINATOR`: required `EventType` parameter for `EventRatioFeatureFactory`
* `FROM_EVENT`: required `EventType` parameter for `EventTimeDeltaFeatureFactory`
* `TO_EVENT`: required `EventType` parameter for `EventTimeDeltaFeatureFactory`

#### FeatureSpec mechanism
The `FeatureSpec` class allows features to be defined without explicitly referencing FeatureFactories. In fact, a `FeatureSet`, representing all features to be generated for a message, is built with FeatureSpec instances.
A `FeatureSpec` is a combination of a `KnownFeature` enum and zero or more `FeatureParam` objects. A `KnownFeature` enum value refers to a combination of some feature factory class along with some hard-coded parameters. 
The `FeatureSpec.buildFactory()` method returns an appropriate `FeatureFactory` instance created from the underlying `KnownFeature` and optional parameters.
`FeatureSpecs` can be serialized/deserialized. This will become important in an upcoming PR where the information about registered classifiers needs to be persisted in LDAP. To enable this, as with the `ParamKey` enum, `KnownFeature` values have a short-form string value.

#### Testing
The `FeatureFactoryTest` suite has tests for each of the `FeatureFactory` implementations, as well as for `FeatureSpec` encoding/decoding and building `ComputedFeatures` results from a given `FeatureSet`.

### DIFF
--- a/common/src/java/com/zimbra/common/util/ZimbraLog.java
+++ b/common/src/java/com/zimbra/common/util/ZimbraLog.java
@@ -478,6 +478,11 @@ public final class ZimbraLog {
     public static final Log eventlog = LogFactory.getLog("zimbra.eventlog");
 
     /**
+     * the "zimbra.ml" logger. For machine learning-related events
+     */
+    public static final Log ml = LogFactory.getLog("zimbra.ml");
+
+    /**
      * Maps the log category name to its description.
      */
     public static final Map<String, String> CATEGORY_DESCRIPTIONS;
@@ -568,6 +573,7 @@ public final class ZimbraLog {
         descriptions.put(contactbackup.getCategory(), "Contact Backup and restore");
         descriptions.put(event.getCategory(), "Event log operations");
         descriptions.put(eventlog.getCategory(), "Serialized events");
+        descriptions.put(ml.getCategory(), "Machine learning operations");
         CATEGORY_DESCRIPTIONS = Collections.unmodifiableMap(descriptions);
     }
 
--- a/store/src/java-test/com/zimbra/cs/ml/feature/FeatureFactoryTest.java
+++ b/store/src/java-test/com/zimbra/cs/ml/feature/FeatureFactoryTest.java
@@ -1,0 +1,357 @@
+package com.zimbra.cs.ml.feature;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+import javax.mail.internet.MimeMessage;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import com.zimbra.cs.account.MockProvisioning;
+import com.zimbra.cs.account.Provisioning;
+import com.zimbra.cs.event.Event;
+import com.zimbra.cs.event.Event.EventType;
+import com.zimbra.cs.event.analytics.EventMetric.MetricInitializer;
+import com.zimbra.cs.event.analytics.EventMetricManager;
+import com.zimbra.cs.event.analytics.RatioMetric;
+import com.zimbra.cs.event.analytics.RatioMetric.RatioIncrement;
+import com.zimbra.cs.event.analytics.ValueMetric;
+import com.zimbra.cs.event.analytics.ValueMetric.IntIncrement;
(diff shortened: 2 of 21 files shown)


ilyusha:
This PR had an incorrect branch name; moved to https://github.com/Zimbra/zm-mailbox/pull/484

PR #485, opened 2018-01-12
Bugfix/zimbra 135 event logger shutdown fix

varunrisbud:
While testing the code for the Event backfill from the database an issue with the EventLogger::shutdownEventLogger was uncovered. The shutdownEventLogger method was immediately killing the EventNotifier thread. When the intended behavior was 

1. Initiate shutdown. Notify EventNotifier treads that shutdown was initiated. The thread should try to complete execution of the current task and stop. (Don't drain the event queue)
2. If the drain event queue flag is set it should try to drain the event queue and once its empty it should stop.
3. After initiating the shutdown the executor service would wait for 30 seconds to allow the threads to complete execution. (shutdown gracefully)
4. If the threads don't complete the execution it would call shutdownNow and kill all the threads.

To satisfy this behavior I am using two atomic boolean flags and *poison pill* approach.
1. When the shutdownExecutor flag is set the EventNotifier thread exits from the while loop in which it waits to consume events from the event queue. In order to check frequently the state of the shutdownExecutor poll method with a 3 seconds timeout is used to retrieve events from the event queue. Poll with timeout is used instead of take so in case of event queue the code does not get blocks on it till there is some event added to the queue.
2. If we want to shutdown the executor service as well as drain the event queue before the executor service is killed then drainQueueBeforeShutdown flag is set. In order to determine the end of event queue *poison pills* are added to the event queue. So now EventNotifier thread will consume events from the event queue till it encounters the poison pill. 

#### Cons for this approach

1. Events added to the event queue after poison pill will not be consumed by the EventNotifier. We might lose events in this case. But in a production system EventLogger is closed only when the zm-mailbox shutdown is initiated. It that case we do not expect more events to be generated. 
2. As the event queue is a bounded queue in case of the queue being full, we will not be able to add poison pills immediately to the event queue. We will need to wait till space is available in queue. Offer method with 3 seconds timeout is used. In that case executor shutdown won't happen gracefully for all the EventNotifier threads. But will avoid the case of waiting infinitely for space to because available on the queue and shutdown to happen.

### DIFF
--- a/store/src/java/com/zimbra/cs/event/Event.java
+++ b/store/src/java/com/zimbra/cs/event/Event.java
@@ -45,7 +45,9 @@ public class Event {
         DELETE_ACCOUNT(UniqueOn.ACCOUNT, true),
         //auxiliary event used to allow contact affinity to be calculated from
         //incoming messages
-        AFFINITY(UniqueOn.MSG_AND_RECIPIENT);
+        AFFINITY(UniqueOn.MSG_AND_RECIPIENT),
+        //poison pill needed to stop the executor service in EventLogger class.
+        POISON_PILL(UniqueOn.NONE, true);
 
         private boolean internal;
         private UniqueOn uniqueOn;
--- a/store/src/java/com/zimbra/cs/event/Event.java
+++ b/store/src/java/com/zimbra/cs/event/Event.java
@@ -23,8 +23,7 @@ import com.zimbra.cs.mailbox.Message;
 import com.zimbra.cs.mime.ParsedMessage;
 
 public class Event {
-    public static final String MULTI_VALUE_SEPARATOR = ",";
-
+    public static Event POISON_PILL = new Event("POISON_PILL_ACCOUNT_ID", EventType.POISON_PILL, 1L);
     private String accountId;
     private EventType eventType;
     private long timestamp;
--- a/store/src/java/com/zimbra/cs/event/logger/EventLogger.java
+++ b/store/src/java/com/zimbra/cs/event/logger/EventLogger.java
@@ -23,6 +23,7 @@ public class EventLogger {
     private static final LinkedBlockingQueue<Event> eventQueue = new LinkedBlockingQueue<>();
     private static final AtomicBoolean executorServiceRunning = new AtomicBoolean(false);
     private static final AtomicBoolean drainQueueBeforeShutdown = new AtomicBoolean(false);
+    private static final AtomicBoolean shutdownExecutor = new AtomicBoolean(false);
     private ExecutorService executorService;
     private ConfigProvider config;
     private boolean enabled;
@@ -133,6 +134,7 @@ public class EventLogger {
         int numThreads = config.getNumThreads();
         ZimbraLog.event.info("Starting Event Notifier Logger with %s threads; initial event queue size is %s", numThreads, eventQueue.size());
         drainQueueBeforeShutdown.set(false);
+        shutdownExecutor.set(false);
 
         ThreadFactory namedThreadFactory = new ThreadFactoryBuilder().setNameFormat("EventLogger-Worker-Thread-%d").build();
         executorService = Executors.newFixedThreadPool(numThreads, namedThreadFactory);
@@ -154,8 +156,14 @@ public class EventLogger {
         }
 
         ZimbraLog.event.warn("Shutdown called for Event Notifier Executor; initiating shutdown sequence...");
-        executorService.shutdownNow();
+        shutdownExecutor.set(true);
+        executorService.shutdown();
         try {
+            if(drainQueueBeforeShutdown.get()) {
+                for (int i = 0; i < config.getNumThreads(); i++) {
+                    eventQueue.put(new Event("POISON_PILL_ACCOUNT_ID", Event.EventType.POISON_PILL, System.currentTimeMillis()));
+                }
+            }
             executorService.awaitTermination(30, TimeUnit.SECONDS);
         } catch (InterruptedException e) {
             executorService.shutdownNow();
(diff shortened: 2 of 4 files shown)


store/src/java/com/zimbra/cs/event/logger/EventLogger.java
```diff
@@ -197,29 +210,32 @@ private EventNotifier(Map<String, EventLogHandler.Factory> knownFactories, Map<S
...
+                    event = eventQueue.poll(POLL_TIMEOUT, TimeUnit.SECONDS);
+                    if (event != null && event != POISON_PILL) {
+                         notifyEventLogHandlers(event);
+                    }
+                }
+                //drain the event queue till POISON_PILL is encountered.
+                if (drainQueueBeforeShutdown.get() && !(event != null && event != POISON_PILL)) {
+                    ZimbraLog.event.debug("Draining the queue");
+                    event = eventQueue.poll(); //wait till event is added to the queue. In case of empty queue wait for POISON_PILL
+                    while (event != null && event != POISON_PILL) { //exit loop if POISON_PILL is encountered
```
tmclane: Would processing all messages after the POISON_PILL is encountered be such a bad thing?
As soon as the queue empties then it could exit.  This would ensure that nothing is 'lost' even if produced after the poisoning.   I presume the events are not produced at a faster rate than they can be processed.

varunrisbud: @tmclane Actually, thats a good point. In that case we would not need any poison pills. In the drainQueueBeforeShutdown if block, if we get a null back from a poll it will exit. I will test this out. Thanks for pointing this out.

store/src/java/com/zimbra/cs/event/logger/EventLogger.java
```diff
@@ -154,8 +159,16 @@ public void shutdownEventNotifierExecutor() {
...
+        executorService.shutdown();
         try {
+            if(drainQueueBeforeShutdown.get()) {
+                for (int i = 0; i < config.getNumThreads(); i++) {
+                    /* In case of event queue being full we dont want to wait till space is available
+                    In that case some EventNotifier thread may not shutdown gracefully */
+                    eventQueue.offer(POISON_PILL, POISON_PILL_OFFER_TIMEOUT, TimeUnit.SECONDS);
+                }
+            }
             executorService.awaitTermination(30, TimeUnit.SECONDS);
```
fatkudu: I know this isn't new but 30 seconds seems like a long time, would be nice if it was configurable.  Would this affect `zmcontrol stop`?  For instance, I think Zimbra is started and stopped several times during install.  Hope this wouldn't result in each of those taking up to 30 seconds?

store/src/java/com/zimbra/cs/event/logger/EventLogger.java
```diff
@@ -197,29 +210,32 @@ private EventNotifier(Map<String, EventLogHandler.Factory> knownFactories, Map<S
         @Override
         public void run() {
             try {
-                while (!Thread.currentThread().isInterrupted()) {
-                    consume(eventQueue);
+                Event event = null;
+                while (!shutdownExecutor.get()) { //don't process any more events if shutdownExecutor is set to true.
+                    /* poll is used so we can check if shutdownExecutor is set every 3 seconds.
+                    if take was used the code would block on it in case of empty event queue. */
+                    event = eventQueue.poll(POLL_TIMEOUT, TimeUnit.SECONDS);
```
fatkudu: Is there any way to do something like this in a way where it gets stopped when we know that we want to shutdown?  Un-necessary 3 second delays considered harmful.

store/src/java/com/zimbra/cs/event/logger/EventLogger.java
varunrisbud: @fatkudu 30 seconds is the maximum time to wait before killing the EventNotifier threads. This time is needed to flush all the events in the logger to Solr as we are using a batch mechanism to commit events to the Solr.  

This would affect the `zmcontrol stop` process. The maximum time added by EventNotifier would be 30 seconds and minimum would be 3 seconds in case of an empty event queue.(after implementing the code change suggested by Travis). 

With the poison pill approach (current implementation), it could be max 30 sec and min would be time needed for all threads to consume the poison pill (~ 1 sec), again assuming the event queue is empty.

In case the event queue is not empty it could be anywhere between the max and min value.

In case of the `zimbra install` I would assume that event queue will be empty hence min time will be taken.

store/src/java/com/zimbra/cs/event/logger/EventLogger.java
varunrisbud: @fatkudu I could not think of a way to do it without using a poll method.

store/src/java/com/zimbra/cs/event/logger/EventLogger.java
```diff
@@ -197,29 +200,33 @@ private EventNotifier(Map<String, EventLogHandler.Factory> knownFactories, Map<S
...
+                    if take was used the code would block on it in case of empty event queue. */
+                    Event event = eventQueue.poll(POLL_TIMEOUT, TimeUnit.SECONDS);
+                    if (event != null) {
+                         notifyEventLogHandlers(event);
+                    }
+                }
+                //drain the event queue till it is empty.
+                if (drainQueueBeforeShutdown.get()) {
+                    ZimbraLog.event.debug("Draining the queue");
+                    Event event = eventQueue.poll(POLL_TIMEOUT, TimeUnit.SECONDS);
```
fatkudu: Could you interrupt the poll using Thread.interrupt() where shutdownExecutor is currently set? I do thing that if every restart during an install takes 3 seconds more that is a problem.

store/src/java/com/zimbra/cs/event/Event.java
```diff
@@ -23,8 +23,7 @@
 import com.zimbra.cs.mime.ParsedMessage;
 
 public class Event {
-    public static final String MULTI_VALUE_SEPARATOR = ",";
-
+    public static String MULTI_VALUE_SEPARATOR = ",";
```
grishick: what is the reason for removing 'final' keyword here?

store/src/java/com/zimbra/cs/event/logger/EventLogger.java
grishick: While I am less concerned about installs - I think we should optimize the installer to avoid multiple restarts as if we are installing Windows 95 - I am concerned about delaying shutdown by 3 seconds. For scenarios where an admin just wants to shut down ZCS and does not care about events in the queue. Perhaps, you could make the timeout configurable via LC?

PR #486, opened 2018-01-13
ZIMBRA-130: Classifier Support

ilyusha:
This PR introduces classifiers to ZCS.

### The ML Server
The actual machine learning work is handled by the ML server found here: https://github.com/Zimbra/zimbra-ml. 

#### Introduction
This ML server uses The _Intel Neon_ deep learning library to build LSTM and Convolutional neural networks. It allows for registering classifiers for a given specification, training them, and classifying inputs with them. It also supports broader classifier management operations, such as listing all registered classifiers, getting detailed info for a classifier by ID, and deleting an existing classifier. It exposes a GraphQL API with a Tornado webserver.

#### Exclusive vs Overlapping Classes
The ML server supports classifying inputs into both _exclusive_ and _overlapping_ classes. A message can be classified into zero or one exclusive class, and into any number of overlapping classes. The distinction is that exclusive classifications are binary, while overlapping classifications include a probability value; the mailbox server must decide whether the probability is high enough to warrant action. 
An example of exclusive classes is folders such as "work", "social", "shopping", and "finance". An email can belong to no more than one of these.
An example of an overlapping class is "important". An email can be long to this class independently of any other classifications.

### Mailbox Client
The ZCS mailbox server acts the client to this backend. While this PR does not include any GraphQL code needed to actually connect to this server (this will be added later), it has all the abstractions necessary to manage classifiers given an _arbitrary_ backend. In short, the code introduced here lays the groundwork for managing classifiers in conjunction with the ML server described above.

First, I will quickly address two packages that don't have much functionality.

#### The `ml/schema` package
This package contains classes that mirror the GraphQL schema defined in
https://github.com/Zimbra/zimbra-ml/blob/master/schema/schema.py.
Since work on both the zimbra-ml repo and the ZCS ML framework is still ongoing, the correlation between these classes and the GraphQL schema may not be exact. These classes are mainly containers for data to be passed to the API, so there is little to no business logic here.

#### The `ml/query` package
This package contains classes representing queries that can be issued to the ML backend.
The all extend the `MLQuery` base class, and are parameterized by the return type of the `execute()` method. The queries correspond to operations that can be performed on the ML server API.

#### Classifier Management
The `ClassifierManager` class is the primary entry point into the machine learning system on ZCS. It is a singleton class that provides methods to manage classifiers, such as creating, deleting, and listing them.

A `ClassifierRegistry` is a helper class used to persist and fetch information about registered classifiers. It is an abstract class; the primary implementation `LdapClassifierRegistry` stores the information in the multivalued LDAP attribute `zimbraMachineLearningClassifierInfo`. The info is stored as a BEncoded string.

### The `Classifier` Class
The most important class introduced in this PR is `Classifier`.

#### Definition & Creation
The first class involved with classifier creation is `ClassifierSpec`. This is the blueprint for a classifier, from the perspective of the ML server. It contains the following fields:
* An ID (optional; if not provided, the server will generate a random one)
* The number of words the classifier will extract from the email subject
* The number of words the classifier will extract from the email content
* The number of other features that the classifier will use
* The overlapping and exclusive class labels that the classifier will output
* An optional path to the vocabulary file on the ML server (if not provided, a default is used), for when we want to use a pre-computed vocabulary
* The size of the lookup table to be created by the classifier for when we create our own word embeddings
* The number of dimensions to be used for word embedding generation

In short, these fields contain all the information required for the ML server to generate the underlying neural network topology. 

When it comes to numeric features, the backend does not care what they represent - only how many there are. The mailbox, however _does_ need to know what the features are, so that it can generate them from messages. This means that a  definition of a classifier on the mailbox is a combination of a `ClassifierSpec` and a `FeatureSet` (introduced in #ZIMBRA-132).

With this in mind, we can construct an instance of the `ClassifierData` class, which represents all the information necessary to register a classifier, both on the mailbox and on the ML backend. The class contains:
* A `ClassifierSpec` instance representing the information needed by the backend
* A `FeatureSet` instance representing the features that will generated for this classifier
* A string label that the classifier will be referred by (distinct from the internal classifier ID)
* A string description of the classifier (this will be useful in a future PR that introduces  a command-line utility for classifier management)

The `ClassifierManager::registerClassifier` method takes a `ClassifierData` instance and does the following:
1.  Checks that the classifier label is not a duplicate
2. Registers the `ClassifierSpec` on the ML backend using a `CreateClassifierQuery`
3. Instantiates a `Classifier` with the ID returned by the backend
4. Registers this classifier with the `ClassifierRegistry`.

A `Classifier`, then, has the following fields, accessible by getters:
* String ID
* String label
* String description
* `FeatureSet` instance
* A `ClassifierInfo` object, which extends the `ClassifierSpec` used during its creation, adding several metadata fields.

#### Training a classifier
The `train` method takes a `TrainingSpec` object. A `TrainingSpec` contains the following fields:
* A `TrainingData` instance, which wraps a list of `TrainingDocument` instances, which contain the actual feature and label information
* The number of epochs to train over
* The learning rate
* An optional percent training data holdout for testing the classifier
* An optional `TrainingData` instance containing data to test on (either this or percent holdout should be specified)
* An optional boolean "persist" flag. If true, the training set will be persisted on the server, so that multiple training runs can be performed without having to send over the entire set every time.

When training is kicked off, the backend returns an updated `ClassifierInfo` response. If the "persist" flag was set, this response will contain a `TrainingSetInfo` object which has metadata about the persisted training set.

#### Classifying an email
The `Classifier::classify` method is used to get classification results for a `Classifiable` item.  This method generates the input features using its `FeatureSet`, and instantiates the appropriate `AbstractClassificationQuery`.
The classification query returns a `ClassificationResult`, which contains:
* The URL of the underlying text (if provided during the query input)
* The exclusive class label, if available
* An array of `OverlappingClassification` instances, each of which contain the class label and a probability assigned to that class label.

#### MachineLearningBackend
The abstract `MachineLearningBackend` class represents a gateway to a ML server. It accepts `MLQuery` subclasses and invokes callbacks on them; these callbacks have the same return type as the `MLQuery.execute()` method. The backend to be used is specified by the new `zimbraMachineLearningBackendURL` LDAP attribute.

The only implementation of `MachineLearningBackend` in this PR is the `DummyMachineLearningBackend`. This is a mock backend that can be used for testing and development. It has an in-memory "registry" of classifiers, and when "classifying" messages, applies the exclusive and overlapping class labels that are present in email subject. For example, if the the classifier supports exclusive classes `shopping`, `work`, `social` and overlapping classes `important`, an email with the subject "important work message" would be classified as "work" and "important". This lets us test the full classifier system without actually being connected to a functional ML backend.

#### Testing
The `ClassifierManagerTest` class contains unit tests for various aspects of classifier management.

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9785,4 +9785,14 @@ TODO: delete them permanently from here
   <desc>If true, search results will be sorted by relevance if a sortBy value is not provided.</desc>
   <defaultCOSValue>FALSE</defaultCOSValue>
 </attr>
+
+<attr id="3051" name="zimbraMachineLearningClassifierInfo" type="string" cardinality="multi" optionalIn="globalConfig" since="8.8.6">
+  <desc>Serialized info about registered classifiers</desc>
+</attr>
+
+<attr id="3052" name="zimbraMachineLearningBackendURL" type="string" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="8.8.6">
+    <desc>URL for accessing the machine learning service.
+    First part of the URL before the first colon identifies the implementation Factory.
+    Only the "zimbra" prefix is currently supported.</desc>
+</attr>
 </attrs>
--- a/store/src/java/com/zimbra/cs/ml/classifier/ClassifierData.java
+++ b/store/src/java/com/zimbra/cs/ml/classifier/ClassifierData.java
@@ -1,0 +1,57 @@
+package com.zimbra.cs.ml.classifier;
+
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.cs.mailbox.Message;
+import com.zimbra.cs.ml.Classifiable;
+import com.zimbra.cs.ml.classifier.Classifier.ClassifiableType;
+import com.zimbra.cs.ml.feature.FeatureSet;
+import com.zimbra.cs.ml.schema.ClassifierInfo;
+import com.zimbra.cs.ml.schema.ClassifierSpec;
+
+/**
+ * Class encompassing all data needed to register a classifier
+ */
+public class ClassifierData<T extends Classifiable> {
+    private String label;
+    private String description;
+    private FeatureSet<T> featureSet;
+    private ClassifierSpec spec;
+    private ClassifiableType type;
+
+    public ClassifierData(ClassifiableType type, String label, ClassifierSpec spec, FeatureSet<T> featureSet) {
+        this(type, label, spec, featureSet, null);
+    }
+
+    public ClassifierData(ClassifiableType type, String label, ClassifierSpec spec, FeatureSet<T> featureSet, String description) {
+        this.type = type;
+        this.label = label;
+        this.description = description;
+        this.featureSet = featureSet;
+        this.spec = spec;
+    }
+
+    public String getLabel() {
+        return label;
+    }
+
+    public ClassifierSpec getSpec() {
+        return spec;
+    }
+
+    @SuppressWarnings("unchecked")
+    public Classifier<T> create(ClassifierInfo info) throws ServiceException {
+        Classifier<T> classifier;
+        switch (type) {
+        case MESSAGE:
+            classifier = (Classifier<T>) new MessageClassifier(info.getClassifierId(), label, (FeatureSet<Message>) featureSet);
+            break;
+        default:
+            throw ServiceException.FAILURE(String.format("Cannot create Classifier for ClassifiableType %s; only MESSAGE is supported at this time", type), null);
+        }
+        if (description != null) {
+            classifier.setDescription(description);
+        }
+        classifier.setClassifierInfo(info);
+        return classifier;
+    }
+}
--- a/store/src/java/com/zimbra/cs/ml/classifier/ClassifierManager.java
+++ b/store/src/java/com/zimbra/cs/ml/classifier/ClassifierManager.java
@@ -1,0 +1,125 @@
+package com.zimbra.cs.ml.classifier;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.util.ZimbraLog;
+import com.zimbra.cs.ml.Classifiable;
+import com.zimbra.cs.ml.query.CreateClassifierQuery;
+import com.zimbra.cs.ml.query.DeleteClassifierQuery;
+import com.zimbra.cs.ml.query.ListClassifiersQuery;
+import com.zimbra.cs.ml.schema.ClassifierInfo;
+import com.zimbra.cs.ml.schema.ClassifierSpec;
+
+/**
+ * Primary entry point into the classifier system
+ * @author iraykin
+ */
+public class ClassifierManager {
+
+    private ClassifierRegistry registry;
+
+    private static ClassifierManager instance = null;
+
+    private ClassifierManager() throws ServiceException {
+        this(new LdapClassifierRegistry());
+    }
+
+    private ClassifierManager(ClassifierRegistry registry) {
+        this.registry = registry;
+    }
+
+    public void setRegistry(ClassifierRegistry registry) {
+        this.registry  = registry;
+    }
+
+    public static synchronized ClassifierManager getInstance() throws ServiceException {
+        if (instance == null) {
+            instance = new ClassifierManager();
+        }
+        return instance;
+
+    }
+
+    /**
+     * return a map of registered classifiers keyed by the IDs. Used by the classifier registry
+     * to construct full {@link Classifier} instances
+     */
+    public Map<String, ClassifierInfo> getAllClassifierInfo() throws ServiceException {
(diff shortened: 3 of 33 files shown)


store/src/java/com/zimbra/cs/ml/classifier/ClassifierData.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public ClassifierSpec getSpec() {
+        return spec;
+    }
+
+    @SuppressWarnings("unchecked")
+    public Classifier<T> create(ClassifierInfo info) {
+        Classifier<T> classifier;
+        switch (type) {
```
fatkudu: Kind of looks like don't need the switch statement?  Or perhaps need a comment like similar cases lower down.  Would it make more sense to only allow MESSAGE at the moment and anything else is an error?

store/src/java/com/zimbra/cs/ml/classifier/ClassifierManager.java
```diff
@@ -0,0 +1,125 @@
...
+ */
+public class ClassifierManager {
+
+    private ClassifierRegistry registry;
+
+    private static ClassifierManager instance = null;
+
+        private ClassifierManager() throws ServiceException {
+        this(new LdapClassifierRegistry());
+    }
```
fatkudu: indenting looks off somewhere here.

store/src/java/com/zimbra/cs/ml/classifier/ClassifierRegistry.java
```diff
@@ -0,0 +1,197 @@
...
+    private Map<String, Classifier<?>> idMap;
+    private Map<String, Classifier<?>> labelMap;
+    private boolean loaded = false;
+
+    public ClassifierRegistry() {
+        idMap = new HashMap<>();
+        labelMap = new HashMap<>();
+    }
+
+    void loadKnownClassifiers() throws ServiceException {
```
fatkudu: Does this need to be synchronized?  Seem to be a few calls to it.

store/src/java/com/zimbra/cs/ml/classifier/ClassifierRegistry.java
```diff
@@ -0,0 +1,197 @@
...
+        map.put(KEY_DESCRIPTION, classifier.getDescription());
+        return BEncoding.encode(map);
+    }
+
+    protected Classifier<?> decode(String encoded) throws ServiceException {
+        Map<String, Object> map;
+        try {
+            map = BEncoding.decode(encoded);
+        } catch (BEncodingException e) {
+            throw ServiceException.FAILURE("unable to decode classifier", null);
```
fatkudu: Worth logging the value of `encoded`?

store/src/java/com/zimbra/cs/ml/classifier/ClassifierRegistry.java
```diff
@@ -0,0 +1,197 @@
...
+        try {
+            type = ClassifiableType.valueOf(typeStr);
+        } catch (IllegalArgumentException e) {
+            throw ServiceException.FAILURE("invalid ClassifiableType value encountered during decoding classifier: " + typeStr, null);
+        }
+        //this may seem unnecessary now, but it allows us to add classifiers for things other than messages in the future
+        switch (type) {
+        case MESSAGE:
+        default:
+            return decodeMessageClassifier(map);
```
fatkudu: If the value of `type` isn't MESSAGE then I suspect this will make no sense.  Suggest treat anything other than MESSAGE as an error.

store/src/java/com/zimbra/cs/ml/classifier/ClassifierRegistry.java
ilyusha: The encoded values can be pretty long, but it's probably worth logging in error cases

store/src/java/com/zimbra/cs/ml/classifier/ClassifierRegistry.java
ilyusha: yeah - good catch

store/src/java/com/zimbra/cs/ml/classifier/ClassifierRegistry.java
ilyusha: Right, I think I filed that away as a "fix later" and then forgot about it. Thanks for noticing.

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9785,4 +9785,14 @@ TODO: delete them permanently from here
   <desc>If true, search results will be sorted by relevance if a sortBy value is not provided.</desc>
   <defaultCOSValue>FALSE</defaultCOSValue>
 </attr>
+
+<attr id="3051" name="zimbraClassifierInfo" type="string" cardinality="multi" optionalIn="globalConfig" since="8.8.6">
+  <desc>Serialized info about registered classifiers</desc>
```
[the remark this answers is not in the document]
ilyusha: Yeah, it's probably a good idea to preface ML-related attributes with "zimbraMachineLearning"

PR #492, opened 2018-01-17
Bugfix/zcs 2220wait set cleanup

fatkudu:
Various cleanups made whilst investigating ZCS-2220.
Main changes are that SessionMap's countActiveSessions was returning some completely unrelated value and that TestUtil.getAdminSoapTransport was un-necessarily using a session which never gets used (and doesn't get cleaned up until inactivity is spotted) - so stopped that.  Guessing this was the root of why we thought AdminWaitSets were consuming Admin sessions on the mailbox.

### DIFF
--- a/store/src/java/com/zimbra/cs/session/SessionCache.java
+++ b/store/src/java/com/zimbra/cs/session/SessionCache.java
@@ -96,6 +96,17 @@ public final class SessionCache {
         return getSessionMap(Session.Type.SOAP).get(accountId);
     }
 
+    public static int countActiveSessionsForAccount(String accountId, Session.Type type) {
+        SessionMap sessionMap = getSessionMap(type);
+        if (sessionMap == null) {
+            return 0;
+        }
+        int num = sessionMap.countActiveSessions(accountId);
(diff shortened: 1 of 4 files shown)


store/src/java/com/zimbra/cs/session/SessionCache.java
```diff
@@ -42,6 +42,27 @@
  **/
 public final class SessionCache {
 
+    private static final String sRunIdentifier = (System.currentTimeMillis() / 1000) + "." + new java.util.Random().nextInt(100);
```
tmclane: Would a UUID be better? This is probably unique enough in reality.

fatkudu: I'm inclined to leave that alone for now although I take your point :-)

tmclane: It was just a thought. I agree this is probably unique enough. If any sessions double up in exactly the same millisecond theoretically the 'nextInt' will not generate the same exact number although I suppose that possibility exists however small.

I am happy for you to leave it alone as well.

PR #493, opened 2018-01-17
ZIMBRA-133: Classification Tasks

ilyusha:
This PR builds on ZIMBRA-130 to introduce the concept of a _Classification Task_ to the mailbox.

A `ClassificationTask` represents some piece of mailbox functionality that is powered by machine learning; for example, tagging emails with SmartFolders. This task can then be _assigned_ a classifier, at runtime, to power it. Classifier development is an iterative process, requiring experimentation with feature engineering and hyper-parameter tuning; this mechanism allows for classifiers to be "plugging in" without any changes to the mailbox code.

#### Binary vs Multi-Label Tasks
`ClassificationTask` has two subclasses: `BinaryClassificationTask` and` MultilabelClassificationTask`.
The difference between them is that a multi-label task _must_ be handled by exclusive class labels, while a binary task can be handled by either overlapping or exclusive class labels. This is because a binary task can be represented by a single thresholded overlapping class, or by two exclusive classes representing a positive vs negative prediction.

#### Task Registration
A classification task must first be _registered_ with the `ClassifierManager` with a unique name. This allows for tasks to be defined in zimlets or server extensions. When first registered, a task is "unassigned"; that is, no classifier is designated to handle it.

#### Classifier-Task Assignment
A task can be _assigned_ to a classifier using a `ClassificationTaskConfigProvider`. This is an abstract class that manages a mapping from classification task names to `TaskConfig` instances, which are wrappers around a classifier label and an optional probability threshold (used for overlapping class labels).
It is important to note that multiple tasks may be assigned to the same classifier. This is possible due to the classifiers' ability to classify into exclusive/overlapping labels. For example, two binary tasks may be handled by a single classifier as two overlapping labels, or by two classifiers, each using positive/negative exclusive labels.
A classifier can only be assigned to a task if it is _compatible_ with it, which is determined by whether the exclusive/overlapping class labels expected by the task are a subset of the class labels used by the classifier.

The `LdapClassificationTaskConfigProvider` implementation stores task-classifier mappings in the multivalued `zimbraMachineLearningTaskConfig` LDAP attribute.

#### Classification Callbacks
Classification tasks interact with the mailbox using a callback system.
A `ClassificationCallback` represents logic to be invoked depending on class labels returned by `ClassificationResult`. These callbacks are registered during `ClassificationTask` definition.
There are two abstract subclasses:
* An `ExclusiveClassCallback` is instantiated with a set of the possible exclusive class labels; the `handle` method is passed the classified item and the class label. It is up to the method logic to determine what action to take based on this label.
* An `OverlappingClassCallback` is instantiated with a single class label; the `handle` method is invoked with the classified item only if the label probability exceeds the threshold.

#### ClassificationHandler class
A `ClassificationHandler` wraps all callbacks registered for a task. Its `handle` method takes a `Classifiable` item and its corresponding `ClassificationResult` instance and invokes all necessary callbacks.

#### ClassificationExecutionContext class
This is the primary class through which the mailbox should interact with the machine learning system. 
The need for this class arises from the fact that there is not a one-to-one mapping between tasks and classifiers. As described above, a classifier may be assigned to multiple tasks; likewise, some tasks may not have any classifier assignments at all. A `ClassificationExecutionContext` abstracts away the details of task/classifier assignments from the mailbox caller.
A `ClassificationExecutionContext` is constructed using the `ClassifierManager::resolveConfig` method, which takes a `ClassificationTaskConfigProvider` instance. This method iterates over the classifier mappings and assembles a `ClassificationHandler` for each unique classifier. The `execute` method iterates over each classifier and invokes the correct handler.
`ClassificationExecutionContext` instances are cached in `ClassifierManager` using a type-safe heterogeneous map, keyed off the `Classifiable` type that they handle. This avoids having to resolve the classifiers for every incoming message. The `getExecutionContext` method handles fetching an existing context from this cache and resolving it if necessary. The `clearExecutionContextCache` method should be invoked when task assignments are changed.

#### Invocation from Mailbox
The `ReceivedMessageCallback` in the `Mailbox` class invokes `ClassificationExecutionContext::execute` for each incoming email.

#### Testing
The `ClassifierManagerTest` class has been updated with methods for testing classification task registration, assignment, and execution.

### DIFF
--- a/store/src/java/com/zimbra/cs/ml/ClassificationTask.java
+++ b/store/src/java/com/zimbra/cs/ml/ClassificationTask.java
@@ -1,0 +1,158 @@
+package com.zimbra.cs.ml;
+
+import java.util.Set;
+
+import com.google.common.base.Objects;
+import com.google.common.collect.Sets;
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.util.ZimbraLog;
+import com.zimbra.cs.ml.callback.ClassificationCallback;
+import com.zimbra.cs.ml.callback.ExclusiveClassCallback;
+import com.zimbra.cs.ml.callback.OverlappingClassCallback;
+import com.zimbra.cs.ml.classifier.Classifier;
+import com.zimbra.cs.ml.classifier.ClassifierManager;
+import com.zimbra.cs.ml.schema.ClassifierInfo;
+
+/**
+ * This class represents a known task to be performed with the classification system.
+ */
+public abstract class ClassificationTask<C extends Classifiable> {
+
+    private String taskName;
+    private ClassificationType taskType;
+    protected ExclusiveClassCallback<C> exclusiveCallback;
+
+    public static enum ClassificationType {
+        BINARY, MULTI_LABEL;
+    }
+
+    public ClassificationTask(String taskName, ClassificationType taskType) {
+        this.taskName = taskName;
+        this.taskType = taskType;
+    }
+
+    public String getTaskName() {
+        return taskName;
+    }
+
+    public ClassificationType getTaskType() {
+        return taskType;
+    }
+
+    public ClassificationTask<C> withExclusiveClassCallback(ExclusiveClassCallback<C> callback) {
+        this.exclusiveCallback = callback;
+        return this;
+    }
+
+    public boolean supportsClassifier(Classifier<C> classifier) {
+        if (exclusiveCallback == null) {
+            return false;
+        }
+        Set<String> classifierExclusiveClasses = Sets.newHashSet(classifier.getInfo().getExclusiveClasses());
+        return classifierExclusiveClasses.containsAll(exclusiveCallback.getExclusiveClasses());
+    }
+
+    protected abstract void checkCanRegister() throws ServiceException;
+
+    public void register() {
+        try {
+            checkCanRegister();
+            ZimbraLog.ml.info("registering classification task %s", taskName);
+            ClassifierManager.getInstance().registerClassificationTask(this);
+        } catch (ServiceException e) {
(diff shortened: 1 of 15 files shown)


store/src/java/com/zimbra/cs/ml/ClassificationTask.java
```diff
@@ -0,0 +1,158 @@
...
+        } catch (ServiceException e) {
+            ZimbraLog.ml.error("unable to register classification task %s", taskName, e);
+        }
+    }
+
+    /**
+     * Given a classifier, return the appropriate callback
+     */
+    public ClassificationCallback<C> resolveCallback(Classifier<C> classifier) throws ServiceException {
+        String[] exclusiveClasses = classifier.getInfo().getExclusiveClasses();
```
grishick: I am missing the logic in this method. It seems to always return the same class member (exclusiveCallback) as a simple getter would (e.g. getExclusiveCallback), but the comment says that it returns "appropriate callback" given a classifier. Is there supposed to be any other logic that chooses which callback to return in addition to null check?

ilyusha: Ah, I can see how this is a bit unclear. This method is in the abstract `ClassificationTask` class, which handles the common use case of resolving an exclusive class callback. `BinaryClassificationTask` overrides this method to first try to resolve an overlapping class callback, calling `super.resolveCallback(classifier);` if an appropriate overlapping callback is not found.

ilyusha:
@tmclane I'm actually not sure, but I'll keep it as 8.8.6 since that's what the other NextGen-related attributes have been marked as. We can always change them prior to release.

PR #495, opened 2018-01-19
ZCS-3887:Enhancement of SMTP log format

ykomiyam:
Add the following data to the log event when the message is sent via SMTP:
 - Size (if the size info has already calculated): "size="
 - Envelope From: "sender="
 - Number of envelope to: "nrcpts="
 - Envelope To: "to="

This event will be logged by SendMsgRequest (send message from WebClient)
BounceMsgRequest (redirect message from WebClient), and redirecting by
the Sieve redirect filter.

The event text may be long; to restrict the number of characters in one
log event, the administrator can set the max characters from log4j
configuration:
```
log4j.appender.LOGFILE.layout.ConversionPattern=%d %-5p [%t] [%z] %c{1} - %.10000m%n
```

### DIFF
--- a/store/src/java/com/zimbra/cs/mailbox/MailSender.java
+++ b/store/src/java/com/zimbra/cs/mailbox/MailSender.java
@@ -817,28 +817,79 @@ public class MailSender {
         }
     }
 
-    public void logMessage(MimeMessage mm, String smtpHost, ItemId origMsgId, Collection<Upload> uploads, String replyType) {
+    public void logMessage(MimeMessage mm, Address[] rcptAddresses, String smtpHost, ItemId origMsgId, Collection<Upload> uploads, String replyType) {
         // Log sent message info
-        if (ZimbraLog.smtp.isInfoEnabled()) {
-            StringBuilder msg = new StringBuilder("Sending message");
-            if (smtpHost != null) {
-                msg.append(" to MTA at ").append(smtpHost);
+        if (!ZimbraLog.smtp.isInfoEnabled()) {
+            return;
+        }
+        StringBuilder msg = new StringBuilder("Sending message");
+        appendMsgMTA(msg, smtpHost);
+        appendMsgMessageID(msg, mm, origMsgId);
+        appendReplyType(msg, replyType);
+        appendUploads(msg, uploads);
+        appendSize(msg, mm);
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -817,28 +817,79 @@ private Object getTargetMailbox(OperationContext octxt, Account authuser, boolea
...
+            msg.append(", uploads=" + uploads);
+        }
+    }
+
+    private void appendSize(StringBuilder msg, MimeMessage mm) {
+        int size;
+        try {
+            size = mm.getSize();
+        } catch (MessagingException e) {
+            size = 0;
```
fatkudu: Might as well just return here.

store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -817,28 +817,79 @@ private Object getTargetMailbox(OperationContext octxt, Account authuser, boolea
...
+            return;
+        }
+        StringBuilder msg = new StringBuilder("Sending message");
+        appendMsgMTA(msg, smtpHost);
+        appendMsgMessageID(msg, mm, origMsgId);
+        appendReplyType(msg, replyType);
+        appendUploads(msg, uploads);
+        appendSize(msg, mm);
+        appendEnvelopeFrom(msg, mEnvelopeFrom);
+        appendEnvelopeTo(msg, rcptAddresses);
```
Prashantsurana: There are security and personal information disclosure. 
This may add Bcc address in logs. 
@plobbes  do you think it is good idea to log such information?

ykomiyam: Thanks @Prashantsurana for the comment.  In the first place, this fix is used by the system administrator (especially the ISP operator) to track down the message flow from the WebClient action through the MBS to the external MTA. On the MTA side, for example Postfix, all the RCPT TO addresses including Bcc have been already logged.

plobbes: As @ykomiyam mentioned, bcc is logged as part of RCPT TO for MTAs. Logging these as part of the 'to' addresses on the mailbox servers does not introduce a new/additional exposure beyond what we already do in the product.

PR #499, opened 2018-01-22
ZCS-3826:imap range queries

fatkudu:


plobbes:
This may be a little out in left field, but do we (or can we add) some negative test cases for things we don't expect to happen. Specifically a client specifying a range where min and max are transposed (so max < min) and permutations where perhaps min and max are negative values too? I took just a quick peek but didn't see anything along those lines, but I could see a rogue client and/or fuzzer doing something like that.

fatkudu:
The PMD fixes I did were mostly in old code.  It still complains that The method sequenceAsSearchTerm() has an NPath complexity of 312 which is mostly my new changes.  I've seen a lot worse and easy mechanisms to reduce the complexity evade me at the moment.

PR #500, opened 2018-01-23
ZCS-4004 - ensure mailbox not modified by another thread

gordyt:
This ticket covered updating the `updateAccessTime` method.  Found a second method (`reload`) that needed the same update.


gordyt:
@tmclane It's funny you mention that.  I had actually deleted those comments originally, then thought better of it and added them back in.

PR #502, opened 2018-01-24
ZCS-4004 Update to initial fix

gordyt:
Replaced call to `this.getMailboxOrNull`, as we need to have a `MailboxStore`.  Original implementation was breaking the `TestImapViaEmbeddedRemote` SOAP tests.

Verified that unit tests and all of the enabled IMAP-related SOAP tests pass.


tmclane:
That change seems good. I wonder if all the usages of mbox shouldn't be replaced with 'mailbox' instead though to clean it up further.

gordyt:
merged and closed

gordyt:
@tmclane It was that way originally (see first commits for this ticket). Problem was that `mailbox` is volatile, so if it is null'd out in another thread, the code wasn't seeing that.  But by assigning it to another variable, you can then check for null.

PR #503, opened 2018-01-24
Bugfix/pwsaltlen

wdomburg:
While we were on-boarding a new client we discovered that Zimbra would not authenticate against their pre-hashed SSHA passwords even though we could bind to the address via LDAP.  Further investigation showed that the PasswordUtil class only works on passwords created with a salt length of 4 or 8 characters.

This patch goes beyond the scope of our immediate problem by supporting arbitrary salt lengths for all password schemes and modifies the interface so calling code no longer has to explicitly check individual password schemes.  The also simplifies adding new schemes in the future; e.g. support for SHA3-512 could be added with a single line to one class.

### DIFF
--- a/store/src/java/com/zimbra/cs/account/auth/AuthMechanism.java
+++ b/store/src/java/com/zimbra/cs/account/auth/AuthMechanism.java
@@ -245,26 +245,18 @@ public abstract class AuthMechanism {
                         namePassedIn(authCtxt), "missing "+Provisioning.A_userPassword);
             }
 
-            if (PasswordUtil.SSHA512.isSSHA512(encodedPassword)) {
-                if (PasswordUtil.SSHA512.verifySSHA512(encodedPassword, password)) {
-                    return; // good password, RETURN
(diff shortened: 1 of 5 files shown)


wdomburg:
Added credential reload logic.  Removed isEncodedPassword and isValidEncodedPassword methods since they are unnecessary with the new PasswordUtil#verify method (which returns null for unencoded passwords, and will otherwise verify against the appropriate encoding).

wdomburg:
Just to note: not entirely happy about using a null from the PasswordUtil#verify to signal a password without a recognized password-scheme.  Could alternately create and throw an UnrecognizedPasswordScheme exception, though I normally chafe at using exceptions for flow control as well.

The caller could alternately use 

Scheme scheme = PasswordUtil.getScheme(password);

And do a null check on that instead, before calling "scheme.verify(password)".

fatkudu:
@wdomburg Is it worth adding a Jira ticket for this?

PR #505, opened 2018-01-30
ZCS-4007: Contact Frequency Changes

ilyusha:
This PR changes the contact frequency graph API to allow clients to specify arbitrary range start times.

Previously, the client could only specify a combination of "d", "w", or "m" flags:
 - "d" returned the current month in day increments
 - "w" returned the last 6 months in week increments
 - "m" returned the _current_ year in month increments

This has been updated so that the client can specify one or more `spec` request elements, each consisting of:
 - A `range` string,  in the format `#(d|w|m)`; for example, "30d" or "25w".
 - An `interval`, which can be one of `d|day|w|week`. Note that aggregations by month are no longer supported. 

Given N time units (day, week, or month), the graph range will encompass N _complete_ units, as well as part of the current unit. In other words, requesting 6 months of data on January 15th would return data from August 1st until January 15th, with buckets for each of the 6 complete months and a partial bucket for January.

#### Update 1/31
Made the `range` parameter case-sensitive (only lower-case `d|w|m` is accepted)
Restricted the `interval` parameter to case-sensitive `d|w`

### DIFF
--- a/soap/src/java/com/zimbra/soap/mail/message/GetContactFrequencyRequest.java
+++ b/soap/src/java/com/zimbra/soap/mail/message/GetContactFrequencyRequest.java
@@ -1,9 +1,14 @@
 package com.zimbra.soap.mail.message;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlRootElement;
 
 import com.zimbra.common.soap.MailConstants;
+import com.zimbra.soap.mail.type.ContactFrequencyGraphSpec;
 
 @XmlRootElement(name=MailConstants.E_GET_CONTACT_FREQUENCY_REQUEST)
 public class GetContactFrequencyRequest {
@@ -12,13 +17,12 @@ public class GetContactFrequencyRequest {
      */
     @XmlAttribute(name=MailConstants.A_EMAIL, required=true)
     private String contactEmail;
+
     /**
-     * @zm-api-field-description list of frequency graphs to return. time range values should be concatenated together
-     * Values are one or more of "d(day)w(week)m(month)".
-     * request can have any combination of values "dwm" like "dw", "wm", "mdw", etc.
+     * @zm-api-field-description List of contact frequency graph specifications to return
      */
-    @XmlAttribute(name=MailConstants.A_CONTACT_FREQUENCY_BY, required=true)
-    private String frequencyBy;
+    @XmlElement(name=MailConstants.A_CONTACT_FREQUENCY_GRAPH_SPEC, type=ContactFrequencyGraphSpec.class, required=true)
+    private List<ContactFrequencyGraphSpec> specs = new ArrayList<>();
 
     /**
      * @zm-api-field-description offset in minutes from UTC to user's current timezone.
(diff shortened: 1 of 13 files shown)


soap/src/java/com/zimbra/soap/mail/message/GetContactFrequencyRequest.java
```diff
@@ -13,12 +18,10 @@
     @XmlAttribute(name=MailConstants.A_EMAIL, required=true)
     private String contactEmail;
     /**
-     * @zm-api-field-description list of frequency graphs to return. time range values should be concatenated together
-     * Values are one or more of "d(day)w(week)m(month)".
-     * request can have any combination of values "dwm" like "dw", "wm", "mdw", etc.
+     * @zm-api-field-description
```
fatkudu: I would have something in the description - perhaps "list of contact frequency graphs to return"
May be worth generating the docs under soap/ and checking that they look OK.

soap/src/java/com/zimbra/soap/mail/type/ContactFrequencyData.java
```diff
@@ -14,19 +13,19 @@
...
-    public ContactFrequencyData(String frequencyBy, List<ContactFrequencyDataPoint> dataPoints) {
-        setFrequencyBy(frequencyBy);
+    public ContactFrequencyData(ContactFrequencyGraphSpec graphSpec, List<ContactFrequencyDataPoint> dataPoints) {
+        setGraphSpec(graphSpec);
         setDataPoints(dataPoints);
     }
 
-    @XmlAttribute(name=MailConstants.A_CONTACT_FREQUENCY_BY, required=true)
-    private String frequencyBy;
+    @XmlElement(name=MailConstants.A_CONTACT_FREQUENCY_GRAPH_SPEC, required=true)
```
fatkudu: Codacity likes to have fields before constructors etc.  Worth following that convention just to shut Codacity up?  Or if we don't think it is worth the hassle, perhaps we should change the Codacity rules.

store/src/java/com/zimbra/cs/event/analytics/contact/ContactAnalytics.java
```diff
@@ -1,21 +1,57 @@
...
+        DAY("d"), WEEK("w");
+
+        private String shortRepr;
+
+        private ContactFrequencyGraphInterval(String shortRepr) {
+            this.shortRepr = shortRepr;
+        }
+        public static ContactFrequencyGraphInterval of(String str) throws ServiceException {
+            for (ContactFrequencyGraphInterval interval: ContactFrequencyGraphInterval.values()) {
+                if (interval.name().equalsIgnoreCase(str) || interval.shortRepr.equalsIgnoreCase(str)) {
```
fatkudu: I'm guessing this accepts "DAY" or "d"?  Is there a reason to accept both?  If so, does that apply to ContactFrequencyGraphTimeRangeUnit too?
My preference would be to keep the accepted values to a minimum - I'd actually prefer to not ignore case too...  That way can keep test cases to a minimum and use simpler code + perhaps use enums in APIs

ilyusha: Yeah good point, there's not really any reason for this. I'll keep this as case-sensitive d|w

soap/src/java/com/zimbra/soap/mail/message/GetContactFrequencyRequest.java
ilyusha: I think I was going to put something there, and then forgot. good catch.

ilyusha:
@tmclane It was one of Gren's recommendations - I tend to agree with it, as it makes the API spec more precise.

PR #506, opened 2018-01-31
ZCS-3810 imap uid search should return more than 1000 items

fatkudu:
Also implemented search sorting by ID - likely to be more efficient than sorting by date where there may be more items with the same key.


gordyt:
Also I did deploy your PR and ran all of the IMAP-related SOAP tests (all either passed or were skipped).

fatkudu:
@gordyt thanks for the reminder.  I had intended to squash that commit but forgot.  However, I'm kind of liking the other variant you talked about elsewhere where changes are made incrementally until enough approvals are received, then squashed before final merge.  Just need to remember!

PR #507, opened 2018-02-01
ZCS-3705:SortBy read/unread asc/desc does not work with mailbox

sneha-patil-synacor:
Added following sortBy support:
readAsc,readDesc,unreadAsc, unreadDesc

Updated soap doc
Added junit test

ant test-all passed


ilyusha:
I'm concerned that this sort order doesn't have a corresponding entry in `LuceneQueryOperation::toLuceneSort`, which may result in incorrect behavior in index-first query execution plans.  As-is, the enum will fall through to `DATE_DESCENDING`.

More importantly, I'm not sure how sorting on mutable fields would work with index-first queries, since updating the sort value is reflected in the DB but not the index. We already support sorting by `FLAG`, which which is mutable, so there might already be a bug here unless I'm missing something. 

The simple solution might be to simply force db-first query execution for these sort orders. There's already a precedent for this, since an `UNREAD` flag search will always be executed db-first (see `DbSearchConstraints.Leaf.tryDbFirst()`).

PR #508, opened 2018-02-01
ZCS-3870:message correction and logging

sneha-patil-synacor:
Making the soap fault message more generic
Added logging for different failure reasons
If the image is not resized since its equal or less than the resize dimensions, update the actual image to LDAP

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/SaveProfileImage.java
+++ b/store/src/java/com/zimbra/cs/service/mail/SaveProfileImage.java
@@ -3,11 +3,13 @@ package com.zimbra.cs.service.mail;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
 import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.io.IOUtils;
 
 import com.zimbra.common.localconfig.DebugConfig;
 import com.zimbra.common.localconfig.LC;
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/service/mail/SaveProfileImage.java
```diff
@@ -128,10 +128,12 @@ private static void updateLDAP(long inputSize, InputStream thumbnailIn, String c
                 Provisioning.getInstance().modifyAttrs(mbox.getAccount(), prefs, true,
                     zsc.getAuthToken());
             } else {
-                throw ServiceException.FAILURE("Unable to update thumbnail image in LDAP", null);
+                ZimbraLog.mailbox.warn("Unable to resize profile image");
+                throw ServiceException.FAILURE("FAILURE_MESSAGE", null);
             }
         } else {
-            throw ServiceException.FAILURE("Unable to update thumbnail image in LDAP", null);
+            ZimbraLog.mailbox.warn("Profile image is larger than maximimum allowed size for resizing");
```
plobbes: nits:
1. typo maximimum → maximum
2. given this is a log only message and it sounds quite odd anyway, why not just log specifically what we're comparing it against?
For example:
```"Profile image is larger than localconfig max_image_size_to_resize"```
or
```"Profile image is larger than maximum size allowed (max_image_size_to_resize)"```
We could even log the actual size...
```"Profile image is larger than maximum size allowed (max_image_size_to_resize): %d > %d", inputSize, LC.max_image_size_to_resize.intValue()```

sneha-patil-synacor: Fixed

PR #510, opened 2018-02-01
ZCS-3390: Server does not advance galsync token after SyncGalResponse…

Prashantsurana:
ZCS-3390: Server does not advance galsync token after SyncGalResponse returns only <deleted> nodes

### DIFF
--- a/store/src/java-test/com/zimbra/cs/gal/GalSyncTokenTest.java
+++ b/store/src/java-test/com/zimbra/cs/gal/GalSyncTokenTest.java
@@ -39,12 +39,12 @@ public class GalSyncTokenTest {
         GalSyncToken token1 = new GalSyncToken(t1);
         GalSyncToken token2 = new GalSyncToken(t2);
         token2.merge(token1);
-        Assert.assertEquals(t2, token2.toString());
+        Assert.assertEquals("GalSync tokens not merged correctly.", t2, token2.toString());
         t1 = "20180131045916.000Z:b1010a37-e08d-45d4-b69b-1ea411a75138:13";
         t2 = "20180131045916.000Z:b1010a37-e08d-45d4-b69b-1ea411a75138:14";
         token1 = new GalSyncToken(t1);
         token2 = new GalSyncToken(t2);
         token1.merge(token2);
-        Assert.assertEquals(t2, token1.toString());
+        Assert.assertEquals("GalSync tokens not merged correctly.", t2, token1.toString());
     }
 }
--- a/store/src/java-test/com/zimbra/cs/gal/GalSyncTokenTest.java
+++ b/store/src/java-test/com/zimbra/cs/gal/GalSyncTokenTest.java
@@ -31,4 +31,20 @@ public class GalSyncTokenTest {
         Assert.assertEquals(fullTimestamp, token.getLdapTimestamp());
         Assert.assertEquals(dateWithHour + "Z", token.getLdapTimestamp("yyyyMMddHHmmss'Z'"));
     }
+
+    @Test
+    public void testSyncTokenMerge() throws ServiceException {
+        String t1 = "20180131045916.000Z:b1010a37-e08d-45d4-b69b-1ea411a75138:11";
+        String t2 = "20180131045916.000Z:b1010a37-e08d-45d4-b69b-1ea411a75138:12";
+        GalSyncToken token1 = new GalSyncToken(t1);
+        GalSyncToken token2 = new GalSyncToken(t2);
+        token2.merge(token1);
+        Assert.assertEquals(t2, token2.toString());
+        t1 = "20180131045916.000Z:b1010a37-e08d-45d4-b69b-1ea411a75138:13";
+        t2 = "20180131045916.000Z:b1010a37-e08d-45d4-b69b-1ea411a75138:14";
+        token1 = new GalSyncToken(t1);
+        token2 = new GalSyncToken(t2);
+        token1.merge(token2);
+        Assert.assertEquals(t2, token1.toString());
+    }
 }
--- a/store/src/java/com/zimbra/cs/gal/GalSearchControl.java
+++ b/store/src/java/com/zimbra/cs/gal/GalSearchControl.java
@@ -553,7 +553,7 @@ public class GalSearchControl {
                 TypedIdList tIdList =  mbox.getTombstones(changeId);
                 deleted = tIdList.getAllIds();
                 int deletedChangeId = tIdList.getMaxModSequence();
-                ZimbraLog.gal.info("deleted change id = %s", deletedChangeId);
+                ZimbraLog.gal.debug("deleted change id = %s", deletedChangeId);
                 if (deletedChangeId > changeId) {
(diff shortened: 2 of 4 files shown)


store/src/java/com/zimbra/cs/gal/GalSearchControl.java
```diff
@@ -550,7 +550,15 @@ private void doLocalGalAccountSync(Account galAcct) throws ServiceException {
         List<Integer> deleted = null;
         if (callback.getResponse() != null && !callback.getResponse().getAttributeBool(MailConstants.A_QUERY_MORE) && changeId > 0) {
             try {
-                deleted = mbox.getTombstones(changeId).getAllIds();
+                TypedIdList tIdList =  mbox.getTombstones(changeId);
+                deleted = tIdList.getAllIds();
+                int deletedChangeId = tIdList.getMaxModSequence();
+                ZimbraLog.gal.info("deleted change id = %s", deletedChangeId);
```
Prashantsurana: I can move this to "debug" .

store/src/java/com/zimbra/cs/mailbox/util/TypedIdList.java
```diff
@@ -224,6 +224,16 @@ public boolean contains(Integer id) {
         return ids;
     }
 
+    public int getMaxModSequence() {
+        int mod = 0 ;
+        for (List<ItemInfo> set : type2ids.values()) {
+            for (ItemInfo itemInfo : set) {
+                mod = itemInfo.getModSequence();
```
fatkudu: Doesn't this need to only set it if it the new value is larger than the current one?

store/src/java/com/zimbra/cs/gal/GalSyncToken.java
```diff
@@ -208,8 +208,17 @@ public boolean isEmpty() {
     public void merge(GalSyncToken that) {
         ZimbraLog.gal.debug("merging token "+this+" with "+that);
```
fatkudu: I know this hasn't changes but worth changing to %s style here and elsewhere?  Especially as this is at debug level, so this usually creates unused objects.

store/src/java/com/zimbra/cs/gal/GalSyncToken.java
```diff
@@ -208,8 +208,17 @@ public boolean isEmpty() {
     public void merge(GalSyncToken that) {
```
fatkudu: Would benefit from a description explaining what merge is intended to achieve.

store/src/java/com/zimbra/cs/gal/GalSyncToken.java
```diff
@@ -208,8 +208,17 @@ public boolean isEmpty() {
...
+        for (String aid : that.mChangeIdMap.keySet()) {
+            String strThatVal = that.mChangeIdMap.get(aid);
+            String strThisVal = this.mChangeIdMap.get(aid);
+            if (StringUtils.isNotBlank(strThatVal) && StringUtils.isNotBlank(strThisVal) &&
+                    StringUtils.isNumeric(strThatVal) && StringUtils.isNumeric(strThisVal)) {
+                int thisVal = Integer.parseInt(this.mChangeIdMap.get(aid));
+                int thatVal = Integer.parseInt(that.mChangeIdMap.get(aid));
+                strThatVal = thisVal > thatVal ? strThisVal : strThatVal;
+            }
+            mChangeIdMap.put(aid, strThatVal);
```
fatkudu: Why the preference for value from that (when the if branch isn't executed)?  My intuition would be to take the value from whichever won in getEarlierTimestamp?

store/src/java-test/com/zimbra/cs/gal/GalSyncTokenTest.java
```diff
@@ -31,4 +31,20 @@ public void testSyncTokenTimestampZulu() throws ServiceException {
...
     }
+
+    @Test
+    public void testSyncTokenMerge() throws ServiceException {
+        String t1 = "20180131045916.000Z:b1010a37-e08d-45d4-b69b-1ea411a75138:11";
+        String t2 = "20180131045916.000Z:b1010a37-e08d-45d4-b69b-1ea411a75138:12";
+        GalSyncToken token1 = new GalSyncToken(t1);
+        GalSyncToken token2 = new GalSyncToken(t2);
+        token2.merge(token1);
+        Assert.assertEquals(t2, token2.toString());
```
fatkudu: Please use the variants which have a description as the first argument to make it easier to see from the report what is failing when a test fails.

store/src/java/com/zimbra/cs/gal/GalSyncToken.java
Prashantsurana: The thing come into picture when both getEarlierTimestamp is same. but the deleted item in tombstone having higher mod sequence than mail item.

store/src/java/com/zimbra/cs/mailbox/util/TypedIdList.java
Prashantsurana: Database query is order by sequence number.

store/src/java/com/zimbra/cs/gal/GalSyncToken.java
Prashantsurana: will change.

store/src/java/com/zimbra/cs/gal/GalSyncToken.java
Prashantsurana: I have added test case to explain.

store/src/java-test/com/zimbra/cs/gal/GalSyncTokenTest.java
Prashantsurana: will do.

store/src/java/com/zimbra/cs/gal/GalSyncToken.java
Prashantsurana: First token set after processing changes from mail_item table. 
Then token set after processing changes from tombstones.

store/src/java/com/zimbra/cs/gal/GalSyncToken.java
fatkudu: Could you add that as a comment please?

store/src/java/com/zimbra/cs/mailbox/util/TypedIdList.java
fatkudu: Perhaps a comment to that effect to avoid future people worrying about it?

store/src/java/com/zimbra/cs/gal/GalSyncToken.java
```diff
@@ -206,10 +206,19 @@ public boolean isEmpty() {
...
+        }
         ZimbraLog.gal.debug("result: "+this);
```
fatkudu: I was hoping this would get changed to %s style too.

store/src/java/com/zimbra/cs/gal/GalSyncToken.java
Prashantsurana: Done

PR #514, opened 2018-02-06
ZCS-3705:SortBy read asc/desc does not work with mailbox

sneha-patil-synacor:
Added following sortBy support:
readAsc,readDesc

Updated soap doc
Added junit test

ant test-all passed

### DIFF
--- a/store/src/java/com/zimbra/cs/db/DbSearch.java
+++ b/store/src/java/com/zimbra/cs/db/DbSearch.java
@@ -141,6 +141,8 @@ public final class DbSearch {
                 return db.concat("(1 + " + db.sign(db.bitAND("mi.flags", String.valueOf(Flag.BITMASK_HIGH_PRIORITY))) +
                         " - " + db.sign(db.bitAND("mi.flags", String.valueOf(Flag.BITMASK_LOW_PRIORITY))) + ")",
                         db.lpad("mi.id", 10, "0"));
+            case UNREAD: // 0 or 1 or 2
+                return "mi.unread";
             case DATE:
             default:
                return "mi.date";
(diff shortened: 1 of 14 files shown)


fatkudu:
Added @ilyusha as a reviewer as I know he had some comments on the previous pull request.

store/src/java/com/zimbra/cs/db/DbSearch.java
```diff
@@ -173,6 +175,10 @@ static String orderBy(SortBy sort, boolean alias) {
         if (sort.getKey() == SortBy.Key.NONE) { // no ORDER BY for NONE
             return "";
         }
+        if (sort.getKey() == SortBy.Key.UNREAD) {
+            return " ORDER BY " + (alias ? SORT_COLUMN_ALIAS : toSortField(sort))
+                + (sort.getDirection() == SortBy.Direction.DESC ? " DESC" : "") + ", mi.id DESC";
```
fatkudu: Why mi.id DESC by default?  Interestingly, this is very close to a change I propose in https://github.com/Zimbra/zm-mailbox/pull/513.  Maybe I should change that to default to DESC or perhaps use the same direction as the first sort.  Would be interested in your thoughts.

sneha-patil-synacor: to keep chronological order(latest on top) always for group of read and unread mails.
So, first sort by unread flag(Asc/Desc) and sort by id(Desc)

fatkudu: Note that I've committed my changes, so you will need to rebase - sorry.

ilyusha:
After looking more carefully at the search code, here is what I found:
In an index-first query operation, a large chunk of the index result set is consumed (potentially the whole thing). This is done by iteratively expanding the lucene result limit until no more matches are found (See `LuceneQueryOperation::getNextResultsChunk`). After each call to this method in `DBQueryOperation::luceneFirstGetNextChunk`, the newly found index results are used as filters on a new DB query, and the final DB results are appended to a running list.

What I'm worried about is that while the _actual_ order of the underlying index result set isn't used, as the DB query enforces its own sort, consuming it _in batches_ may run the risk of returning an incorrect final sort of the concatenated result list when the DB sort differs from the original index sort.

For a simple example, assume 10 index docs [1,2,...,10] that we want to return in lexicographic order. However, they are fetched from the index in batches of 4 in numeric order. That would give us four batches [1,2,3,4], [5,6,7,8], [9,10] passed to the DB query. Each batch is sorted lexicographically, resulting in the results being [1,2,3,4,5,6,7,8,9,10]. However, the true results should be [1,10,2,3,4,5,6,7,8,9].

It's entirely possible that I missed something in the search code or that I misunderstood a key step. If this really is a problem, however, I think it's outside of the scope for this ticket and should be addressed separately.

PR #516, opened 2018-02-07
bugfix/ZCS-4221 - Address uncontrolled heap growth in ImapSessionManager::sessions

tmclane:
- Updated PR to instead provide a concrete implementation of `getListeners(ItemIdentifier ident) ` for `LocalImapMessageStore`.
- Updated `testDeleteFolderNotificationActiveFolder` to capture a disconnect as a `success`.  Test was incorrectly not disconnecting when a selected folder was deleted from another session.

From PR comment:

The problem that this resolved is as follows:

The `closeFolder` method in `ImapSessionManager` asks the
`ImapMailboxStore` for listeners.  The implementation,
defined in `ImapMailboxStore`, worked correctly for remote
IMAP sessions, but not local ones.

This commit moved the implementation that was defined in
ImapMailboxStore to RemoteImapMailboxStore and added the
appropriate implementation to LocalImapMailboxStore.

Once this was done the closeFolder method could correctly detach
the appropriate sessions, which would caused them to be
unloaded from `sessions`.

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/ImapMailboxStore.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapMailboxStore.java
@@ -111,18 +111,7 @@ public abstract class ImapMailboxStore {
             throws ServiceException;
     public abstract void registerWithImapServerListener(ImapListener listener);
     public abstract void unregisterWithImapServerListener(ImapListener listener);
-
-    public List<ImapListener> getListeners(ItemIdentifier ident) {
-        String acctId = ident.accountId != null ? ident.accountId : getAccountId();
-        try {
-            ImapServerListener listener = ImapServerListenerPool.getInstance().getForAccountId(acctId);
-            return Lists.newArrayList(listener.getListeners(acctId, ident.id));
-        } catch (ServiceException se) {
-            ZimbraLog.imap.debug("Problem getting listeners for folder=%s acct=%s from ImapServerListener",
-                    ident, acctId, se);
-            return Collections.emptyList();
-        }
-    }
+    public abstract List<ImapListener> getListeners(ItemIdentifier ident);
 
     public List<ImapListener> getListeners(FolderStore folder) {
         try {
--- a/store/src/java/com/zimbra/cs/imap/LocalImapMailboxStore.java
+++ b/store/src/java/com/zimbra/cs/imap/LocalImapMailboxStore.java
@@ -17,6 +17,7 @@
 package com.zimbra.cs.imap;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
@@ -40,6 +41,7 @@ import com.zimbra.cs.mailbox.Mailbox;
 import com.zimbra.cs.mailbox.Metadata;
 import com.zimbra.cs.mailbox.MetadataList;
 import com.zimbra.cs.mailbox.OperationContext;
+import com.zimbra.cs.session.Session;
 import com.zimbra.cs.util.AccountUtil;
 
 public class LocalImapMailboxStore extends ImapMailboxStore {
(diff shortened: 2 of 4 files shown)


gordyt:
Note:

- Verified `zm-mailbox/store` unit tests pass
- Verified all enabled IMAP-related SOAP tests pass

If this PR is approved, will create a patch for 8.8.6GA that contains the fix.

tmclane:
Note: These `sessions` are used only to determine which `folders` should be paged in and out and are not used directly by any IMAP code.

store/src/java/com/zimbra/cs/imap/ImapSessionManager.java
```diff
@@ -113,19 +112,29 @@ protected static ImapSessionManager getInstance() {
...
     }
 
+    /**
+     * Computes a key to use for caching the ImapListener `session` in the `sessions` in-memory cache
+     *
+     * @param session ImapListener
+     * @return a cache key to deterministically identify an IMAP folder for an Account
+     */
+    protected String sessionsCacheKey(ImapListener session) {
+        return String.format("%s:%d", session.getAuthenticatedAccountId(), session.getFolderId());
```
fatkudu: How does this work for mountpoints and other users folders (under /home)
My intuition is that it doesn't and you end up wiping out entries in the cache for a folder in the authenticated users mailbox with ones that happen to have the same folder Id from the other user's mailbox.

store/src/java/com/zimbra/cs/imap/ImapSessionManager.java
```diff
@@ -171,9 +182,12 @@ public void run() {
 
                 // As we are more likely to decide to drop or page out sessions we process later, we want to start
                 // with the most recently used ones as they are more likely to be useful again.
-                Iterator<ImapListener> mostRecentToLeastRecentIterator = sessions.descendingKeySet().iterator();
+                Iterator<String> mostRecentToLeastRecentIterator = sessions.descendingKeySet().iterator();
                 while (mostRecentToLeastRecentIterator.hasNext()) {
-                    ImapListener session = mostRecentToLeastRecentIterator.next();
+                    ImapListener session =  sessions.get(mostRecentToLeastRecentIterator.next());
+                    ZimbraLog.imap.debug("Session: %s interactive? %s nonInteractiveCutoff=%d cutoff=%d lastAccessed=%d",
```
fatkudu: May not be a new issue but I would be happier if we had code handling session being null here.

store/src/java/com/zimbra/cs/imap/ImapSessionManager.java
```diff
@@ -241,7 +254,9 @@ public void run() {
                         }
                     });
                 }
+                sessions.remove(sessionsCacheKey(session));
```
fatkudu: This is potentially removing a different object from the cache which happens to have been created for the same folder and may be for a newly opened select - are we happy with that?

tmclane: We were attempting to prevent the number of objects from growing without bounds.  I was attempting to find out the connection details such that each IMAP connection itself would compute keys that matched up to itself and only itself.  This would prevent what you were saying but so far I haven't been successful.  

Considering these values are only placed in this cache and then never referenced but they are used for deciding which session to page out or not.  This would just lead to early eviction possibly which would cause a reload and not a failure I wouldn't expect.

fatkudu: Just looked and found another remove method `remove(key, value)` - that would probably be a safer one to use?

store/src/java/com/zimbra/cs/imap/ImapSessionManager.java
tmclane: Handled.

tmclane:
I approve of this re-write of my PR 👍

tmclane:
Patch has been generated and is in testing.  Closing this PR.

PR #517, opened 2018-02-09
ZCS-3890:Request Proxy modification to support JWT

shrikant-prasad:
When request flow is using JWT and it has to be proxied to another server, jwt and salt will be populated in soap envelope context instead of auth token. 
Added new class ZJWToken which extends ZAuthToken and overrides all it's public methods to handle the proxy logic as per jwt flow.

### DIFF
--- a/common/src/java/com/zimbra/common/auth/ZJWToken.java
+++ b/common/src/java/com/zimbra/common/auth/ZJWToken.java
@@ -1,0 +1,67 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2018 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ * ***** END LICENSE BLOCK *****
+ */
+package com.zimbra.common.auth;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import com.zimbra.common.soap.Element;
+import com.zimbra.common.soap.HeaderConstants;
+import com.zimbra.common.util.StringUtil;
+import com.zimbra.common.util.ZimbraCookie;
+
+public class ZJWToken extends ZAuthToken {
+
+    private String salt;
+
+    public ZJWToken(String value, String salt) {
+        super(value, null);
+        this.salt = salt;
+    }
+
+    @Override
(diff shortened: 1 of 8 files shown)


store/src/java/com/zimbra/cs/service/util/JWTUtil.java
```diff
@@ -126,7 +129,7 @@ public static final String getSalt(Element soapCtxt, Map engineCtxt) {
                     javax.servlet.http.Cookie cookies[] =  req.getCookies();
                     if (cookies != null) {
                         for (int i = 0; i < cookies.length; i++) {
-                            if (cookies[i].getName().equals(Constants.ZM_JWT_COOKIE)) {
+                            if (cookies[i].getName().equals(ZimbraCookie.COOKIE_ZM_JWT)) {
```
sneha-patil-synacor: can use ZimbraCookie.COOKIE_ZM_JWT.equals() to avoid NPE

common/src/java/com/zimbra/common/util/ZimbraCookie.java
```diff
@@ -27,6 +27,7 @@
     public static final String COOKIE_ZM_AUTH_TOKEN       = "ZM_AUTH_TOKEN";
     public static final String COOKIE_ZM_ADMIN_AUTH_TOKEN = "ZM_ADMIN_AUTH_TOKEN";
     public static final String COOKIE_ZM_TRUST_TOKEN      = "ZM_TRUST_TOKEN";
+    public static final String COOKIE_ZM_JWT      = "ZM_JWT";
```
ykomiyam: Just a cosmetic thing: You may want to align the "= ZM_JWT" to other declarations above.

shrikant-prasad: done

store/src/java/com/zimbra/cs/service/util/JWTUtil.java
shrikant-prasad: done

common/src/java/com/zimbra/common/auth/ZJWToken.java
```diff
@@ -0,0 +1,67 @@
...
+    private String salt;
+
+    public ZJWToken(String value, String salt) {
+        super(value, null);
+        this.salt = salt;
+    }
+
+    @Override
+    public Element encodeSoapCtxt(Element ctxt) {
+        Element ejwToken = ctxt.addNonUniqueElement(HeaderConstants.E_JWT_TOKEN);
```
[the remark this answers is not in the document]
shrikant-prasad: if ctxt is null, different flow is executed in which this method is not used.

store/src/java/com/zimbra/cs/service/util/JWTUtil.java
```diff
@@ -126,7 +129,7 @@ public static final String getSalt(Element soapCtxt, Map engineCtxt) {
                     javax.servlet.http.Cookie cookies[] =  req.getCookies();
                     if (cookies != null) {
                         for (int i = 0; i < cookies.length; i++) {
-                            if (cookies[i].getName().equals(Constants.ZM_JWT_COOKIE)) {
+                            if (ZimbraCookie.COOKIE_ZM_JWT.equals(cookies[i].getName())) {
```
ykomiyam: Don't we need to check the the value of cookies[x].getName() in case-insensitively?

PR #521, opened 2018-02-11
Bug 34202 - STARTTLS support for outbound mail sent by mailbox server

MakiWatanabe:
This fix add a new ldap attribute: zimbraSmtpStartTlsMode

> 
> Configures the starttls mode on sending messages from mailboxd to the
> MTA host specified with zimbraSmtpHostname. ( on - uses starttls if
> the MTA supports. off - doesn't use starttls. only - requires starttls
> to send messages. )
> 
>                type : enum
>               value : on,off,only
>            callback :
>           immutable : false
>         cardinality : single
>          requiredIn :
>          optionalIn : domain,globalConfig,server
>               flags : serverInherited
>            defaults : off
>                 min :
>                 max :
>                  id : 3022
>     requiresRestart :
>               since : 8.8.6
>     deprecatedSince :

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9629,7 +9629,7 @@ TODO: delete them permanently from here
     <desc>Account thumbnail photo</desc>
 </attr>
 
-<attr id="4001" name="zimbraSmtpStartTlsMode" type="enum" value="on,off,only" cardinality="single" optionalIn="globalConfig,server,domain" flags="serverInherited" since="8.8.6">
+<attr id="3022" name="zimbraSmtpStartTlsMode" type="enum" value="on,off,only" cardinality="single" optionalIn="globalConfig,server,domain" flags="serverInherited" since="8.8.6">
   <globalConfigValue>only</globalConfigValue>
   <globalConfigValueUpgrade>off</globalConfigValueUpgrade>
   <desc>Configures the starttls mode on sending messages from mailboxd to the MTA host specified with zimbraSmtpHostname.
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -1405,7 +1405,7 @@ TODO - add support for multi-line values in globalConfigValue and defaultCOSValu
   <desc>Data for class that scans attachments during compose</desc>
 </attr>
 
-<attr id="240" name="zimbraPrefCalendarInitialView" type="enum" value="day,week,workWeek,month,list,year" cardinality="single" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable">
+<attr id="240" name="zimbraPrefCalendarInitialView" type="enum" value="day,week,workWeek,month,list" cardinality="single" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable">
   <defaultCOSValue>workWeek</defaultCOSValue>
   <desc>initial calendar view to use</desc>
 </attr>
@@ -9627,6 +9627,15 @@ TODO: delete them permanently from here
 
 <attr id="3021" name="thumbnailPhoto" type="binary" cardinality="single" optionalIn="account" since="8.8.6">
     <desc>Account thumbnail photo</desc>
+</attr>
+
+<attr id="3022" name="zimbraSmtpStartTlsMode" type="enum" value="on,off,only" cardinality="single" optionalIn="globalConfig,server,domain" flags="serverInherited" since="8.8.6">
+  <globalConfigValue>off</globalConfigValue>
+  <desc>Configures the starttls mode on sending messages from mailboxd to the MTA host specified with zimbraSmtpHostname.
+        ( on   - uses starttls if the MTA supports.
+          off  - doesn't use starttls.
+          only - requires starttls to send messages. )
+  </desc>
 </attr>
 
 </attrs>
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9629,7 +9629,7 @@ TODO: delete them permanently from here
     <desc>Account thumbnail photo</desc>
 </attr>
 
-<attr id="3022" name="zimbraSmtpStartTlsMode" type="enum" value="on,off,only" cardinality="single" optionalIn="globalConfig,server,domain" flags="serverInherited" since="8.8.6">
+<attr id="4001" name="zimbraSmtpStartTlsMode" type="enum" value="on,off,only" cardinality="single" optionalIn="globalConfig,server,domain" flags="serverInherited" since="8.8.6">
   <globalConfigValue>only</globalConfigValue>
   <globalConfigValueUpgrade>off</globalConfigValueUpgrade>
   <desc>Configures the starttls mode on sending messages from mailboxd to the MTA host specified with zimbraSmtpHostname.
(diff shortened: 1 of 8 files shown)


MakiWatanabe:
I have found unexpected file diff on store/conf/attrs/zimbra-attrs.xml line 1408.
My base zimbra-attrs.xml was possibly outdated.

store/src/java-test/com/zimbra/cs/mailbox/MailSenderTest.java
```diff
@@ -378,4 +380,42 @@ public void sendExternalMessage() throws Exception {
         ZimbraLog.smtp.setLevel(Level.trace);
         MailSender.relayMessage(mm);
     }
+    
+
+    @Test
+    public void setupStartTlsMode() throws Exception {
+    		Provisioning prov = Provisioning.getInstance();
+    		// domain is required to run setupSmtpStartTlsMode() without NPE
+    		Domain domain = prov.createDomain("example.com", new HashMap<String, Object>());
```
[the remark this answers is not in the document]
MakiWatanabe: Without "example.com" domain instance in Provisioning, MailSender.setupSmtpStartTlsMode throws NPE on getting the domain attribute.

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9629,4 +9629,13 @@ TODO: delete them permanently from here
     <desc>Account thumbnail photo</desc>
 </attr>
 
+<attr id="3022" name="zimbraSmtpStartTlsMode" type="enum" value="on,off,only" cardinality="single" optionalIn="globalConfig,server,domain" flags="serverInherited" since="8.8.6">
```
chinmay15: Is this really needed on server and domain both?
It can be optionalIn="globalConfig,server" flags="serverInherited" or 
optionalIn="globalConfig,domain" flags="domainInherited"

MakiWatanabe: This attribute is relevant to zimbraSmtpHostname, zimbraSmtpPort, and zimbraSmtpTimeout.
These are all defined with optionalIn="globalConfig,server,domain".
I think we should better to keep it consistent.

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9629,4 +9629,13 @@ TODO: delete them permanently from here
...
+  <globalConfigValue>off</globalConfigValue>
```
plobbes: I think we want to default to **only** for new installs and **on** for upgrades (would require an upgrade step for this - assuming we don't want to default to **only** for upgrades as well).

store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -1402,4 +1408,33 @@ public static void unregisterPreSendMailListener(PreSendMailListener listener) {
...
+    /**
+     * Setup JavaMail session properties for STARTTLS based the ldap configuration
+     * @param account
+     * @param smtpSession
+     * @throws ServiceException
+     */
+    public static void setupStartTlsMode(Account account, Session smtpSession) throws ServiceException {
+        SmtpStartTlsMode startTlsMode = Provisioning.getInstance().getDomain(account).getSmtpStartTlsMode();
+        if (startTlsMode == null) {
+        		startTlsMode = Provisioning.getInstance().getLocalServer().getSmtpStartTlsMode();
```
plobbes: nit: (more tabs...) it appears indentation is off here (and below), I'd expect 4 spaces past the start of **if** block

store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -1202,11 +1206,13 @@ private String getNextHost() {
     }
 
     private void sendMessageToHost(String hostname, MimeMessage mm, Address[] rcptAddresses)
-    throws MessagingException {
-        mSession.getProperties().setProperty("mail.smtp.host", hostname);
+    throws MessagingException, ServiceException {
+    	    mSession.getProperties().setProperty("mail.smtp.host", hostname);
```
plobbes: nit: shouldn't have tabs in the source code, but looks like one snuck in here

store/conf/attrs/zimbra-attrs.xml
MakiWatanabe: Changed globalConfigValue to "only" and add globalConfigValueUpgrade as "off".

store/src/java/com/zimbra/cs/mailbox/MailSender.java
MakiWatanabe: Fixed eclipse config and applied the format.

store/src/java/com/zimbra/cs/mailbox/Notification.java
```diff
@@ -523,6 +523,10 @@ private void notifyIfNecessary(Account account, Message msg, String rcpt)
                 ZimbraLog.mailbox.warn("error encoutered looking up return path configuration, using null return path instead", se);
             }
             smtpSession.getProperties().setProperty("mail.smtp.from", envFrom);
+            
+            MailSender.setupStartTlsMode(account, smtpSession);
+            ZimbraLog.smtp.debug("Sending message %s to SMTP host %s with properties: %s",
+            		mm.getMessageID(), smtpSession.getProperties().getProperty("mail.smtp.host"), smtpSession.getProperties());
```
ykomiyam: There are several other places where the Transport.send() is called.  For example, in the  com.zimbra.cs.mailbox.Notification.interceptIfNecessary(). Don't we need to take care of those calls as well?

plobbes: Definitely looks like we need to cover those. Maybe we can find an easier way to fit this into the existing call flows without touching each call point?

MakiWatanabe: Ok, this is bigger animal than I expected.  There are total 11 references for Transport.send() or Transport.sendMessage() in zm-mailbox, excluding test code and smtpInject tool.
I have updated 2 of them .
How about to unify the all smtp submissions into single method in MailSender class?
(Transfer.send proxy)

MakiWatanabe: Transport.send (and transport.sendMessage ) takes MimeMessage as argument.
MimeMessage is created with Session.
And the Session is created by JMSession.get*Session methods.
In these methods, JavaMail SMTP properties (host, port, timeout, etc) are configured.
- JMSession.getSmtpSession(account)
  - This method doesn't configure STARTTLS.
  - I will move the starttls related config logic from MailSender to this method.
- JMSession.getRelaySession()
  - This method refers "zimbraShareNotificationMtaConnectionType" (CLEARTEXT,SSL, or STARTTLS).
  - This method is used on connecting zimbraShareNotificationMtaHostname host to send share notification. If the attr is not defined, it will fallback to getSmtpSession(null).
- JMSession.getSession(DataSource)
  - This method is used on connecting smtp host defined in the datasource.
  - This method refers local config "javamail_smtp_enable_starttls". There are ldap attributes named "zimbraDataSourceConnectionType", but it is used only for POP and IMAP.
  - Should I add the logic here also?

store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -1162,7 +1166,7 @@ protected void updateRepliedStatus(OperationContext octxt, Account authuser, boo
                     } else {
                         throw e;
                     }
-                }
+				}
```
chinmay15: Correct indentation.

MakiWatanabe:
After discussion with @chinmay15, I will remove this branch in a few days.
I have created a new branches on snickers:

- bugfix/NICPS-59-snickers: only for zimbra-attr.xml changes and its getters/setters

I will create another child branch for the feature implementation, on bugfix/NICPS-59-snickers.

MakiWatanabe:
Need PM approval

PR #531, opened 2018-02-20
ZCS-4213: Shared Solr Index Architecture

ilyusha:
This PR introduces support for a horizontally-scalable shared-index search architecture to ZCS.

### Motivation
Prior to this change, each account on a mailbox server was issued its own index. This was fine when then index was built directly on top of Lucene, since a mailbox index was simply a Lucene directory. With the transition to a Solr backend, however, this became a problem for several reasons:
 - A Solr core has a non-trivial memory overhead, so having one core per account would quickly exhaust server resources for large customers
 - SolrCloud with Zookeeper is not meant to scale to such a large number of collections. With too many nodes to manage, the system becomes brittle. As a rule of thumb, it is advised to keep the number of collections per SolrCloud cluster below 1000.

### Goals
The primary goals of migrating to a multi-tenant indexing architecture were:

#### Data Isolation
We need to retain data isolation in a shared index. In other words, queries for a given mailbox shouldn't be influenced by data from other mailboxes.

#### Mailbox Co-Location
The new indexes must support sharding to allow for horizontal scalability. For efficient searches, we need to be able to co-locate documents for a given mailbox on the same shard; queries should be routed to only the appropriate shard to avoid distributed searches.

#### Filters
Each index document now includes an `acct_id` field. Each search query includes a corresponding filter query to limit the results to only the specified account. This is constructed as a filter on the `terms` query component: `{!terms f=acct_id}<acct id>`; this is more efficient that than a term query, as no scoring is performed.

#### CompositeId Routing
Solr's `compositeId` routing mechanism lets us ensure that all documents for an account end up on the same shard. This mechanism relies on the format of the _solrId_ field; in short, all documents for which this id begins with `<accountId>!` get routed to the same shard. At query time, the `_route_=<accountId>!` parameter is set to ensure that the query only goes to this same shard.
Note that this does not let us specify _which_ shard a document is routed to; it only ensures co-location. That said, this mechanism opens up the possibility of _shard splitting_ to scale an existing collection over time.

These changes are primarily encapsulated in the `SolrRequestHandler` and `SolrCollectionLocator` classes and subclasses. In fact, this PR consolidates the `SolrIndex` and `SolrCloudIndex` classes into a single class by delegating the handling of all Solr backend details to these helper classes.

The event logging and event store systems have been updated to work with this architecture as well, although far fewer changes have been necessary here; this is because this code was written to rely on `SolrRequestHelper` to begin with. The only big change here has been to generate `compositeId`-style event IDs for co-location in the event index.

For more information on `compositeId` routing, see https://lucene.apache.org/solr/guide/6_6/shards-and-indexing-data-in-solrcloud.html#ShardsandIndexingDatainSolrCloud-DocumentRouting

### New LDAP Attributes
Four new LDAP attributes have been added.
 - `zimbraMailboxIndexName` and `zimbraEventIndexName` specify the name of the solr collections to be used for indexing mailbox and event data, respectively
- `zimbraMailboxIndexInitialNumShards` and `zimbraEventIndexInitialNumShards` specify the number of shards that these indexes are initially created with.

`zimbraEventIndexName` replaces the `zimbraEventSolrIndexType` attribute, since we no longer have a use case for the former.

It should be noted that these attribute currently _need_ to be set at setup time - they do not provide defaults. This is because they have an unusual scope of `account/cos/domain/server` to allow for a hierarchical collection structure for flexible Solr configuration, but `zimbra-attrs.xml` doesn't allow for defaults to be set for this combination of scopes.

### Issues
Some second-order changes are necessary due to some parts of the codebase working under the assumption that each mailbox has its own index.

#### Terms Enumeration
The `TermFieldEnumeration` API is used for efficiently enumerating term values in the index. This underpins `BrowseRequest` and is also used by the `MailboxIndex::existsInContacts` method. Unfortunately, this is no longer a workable approach since term enumeration cannot be filtered, meaning that we cannot enumerate terms for a single account. The workaround was to implement `SolrFacetEnumeration`, which achieves similar functionality with faceting.

The only limitation of this faceting approach is that we cannot specify a `startTerm` value like we could with term enumeration. Luckily, the only functionality that relied on this was the `existsInContacts` method, which has now been re-implemented as a standard Solr query. The `ZimbraIndexReader::getTermsForField` interface method has been updated to no longer take a `startTerm` parameter.

Another small change necessary to make this possible was to slightly change the handling of `from`/`to`/`cc` fields in the index. In a nutshell, we need to search and facet over different fields, with `ReversedWildcardFilterFactory` applied to the search field. This is necessary to avoid facet results including reversed tokens. See https://github.com/Zimbra/zm-solr/pull/4 more details.

#### Deleted Index APIs
The following interface methods have been deleted:
 - `Indexer::compact`, since index compaction is a low-level operation that no longer applies
 - `Indexer::maxDocs` and `ZimbraIndexReader::numDeleteDocs`, since the number of deleted documents for an account can no longer be determined
 - `IndexStore::getLatestIndexGeneration`, since the concept of "index generation" does not apply at the account level.

#### GetIndexStatsRequest
The concept of "index stats" doesn't apply with a shared architecture, since the info returned by `GetIndexStatsRequest` can only be computed at the level of the collection. While the SOAP endpoint has not been deleted, it no longer returns any data. The API team has been notified of this deprecation.

#### DeleteCalendarResourceRequest
A slight change had to be made to `DeleteCalendarResourceRequest` SOAP handler. Specifically, the `deleteMailbox` call has been moved to before the `deleteCalendarResource` call; this was necessary because deleting the underlying account makes it impossible to look up the location of the underling index using the `zimbraMailboxIndexName` attribute.

#### Other Miscellaneous Changes
 - renamed some method names for clarity
 - updated event tests to better clean up after themselves to avoid accumulating orphan ZooKeeper threads
 - updated some unit tests that relied on deprecated index methods


store/src/java/com/zimbra/cs/event/SolrEventDocument.java
```diff
@@ -37,41 +36,48 @@ public SolrInputDocument getDocument() {
...
+            idParts = new Object[] {eventType, dsId};
             break;
         case MESSAGE:
             msgId = (Integer) event.getContextField(EventContextField.MSG_ID);
-            accountLevelIdentifier = String.format("%s:%d", eventType, msgId);
+            idParts = new Object[] {eventType, msgId};
             break;
         case MSG_AND_SENDER:
             msgId = (Integer) event.getContextField(EventContextField.MSG_ID);
             String sender = (String) event.getContextField(EventContextField.SENDER);
```
tmclane: I presume 'getContextField' doesn't have a method to return a default value instead of null?

PR #532, opened 2018-02-21
ZCS-4223:CertAuth broken: failing with exception

shrikant-prasad:
Upgrading the bcprov jar version in store repo to 1.55 which is packaged in jetty/common/lib to remove the NoClassDefFoundError as earlier the repo was built with 1.46 version and there were significant changes in bouncy castle classes in version 1.47.

Testing Done: certauth working fine without any exceptions and user is logged in successfully.


plobbes:
Shouldn't we be going with the latest release? It appears 1.59 should be used instead of 1.55:
- https://www.bouncycastle.org/releasenotes.html

PR #533, opened 2018-02-22
Implementation for NICPS-59/Bug 34202 - STARTTLS support for outbound mail

MakiWatanabe:
Problem analysis
-----------------
SMTP submission is implemented at many classes in mailboxd independently.
With Java Mail, we can use one of following methods to send out SMTP mails.

a. static void send(Message msg)
b. static void send(Message, Address[] addrs)
c. static void send(Message msg, Address[] addrs, String user, String password)
d. static void send(Message msg, String user, String password)
e. abstract void sendMessage(Message msg, Address[] addrs)

I investigated the code refers these methods, and found only a, b, and e are used.
Please refer to the attached diagram.

( image removed to run Codacy )

Fix strategy
------------
In most of the cases, mailboxd calls JMSession.getSmtpSession() to get SMTP Session object.
I add a new method configureStartTls() to set starttls properties based on the zimbraSmtpStartTlsMode value, and call it from getSmtpSession.

When mailboxd calls JMSession.getSession(DataSource) to get session, it uses following own ldap attributes to setup JavaMail properties:
- zimbraDataSourceSmtpHost
- zimbraDataSourceSmtpPort
- zimbraDataSourceSmtpConnectionType

And a localconfig:
- javamail_smtp_enable_starttls

I kept it unchanged.

Update 2018.2.28
------------------
To configure mail.smtp.ssl.trust Java Mail property, add a new ldap attribute zimbraSmtpStartTlsTrustedHosts in PR #540 .

Linked PR
----------
https://github.com/Zimbra/zm-mailbox/pull/540 to add new ldap attribute

### DIFF
--- a/store/src/java/com/zimbra/cs/util/JMSession.java
+++ b/store/src/java/com/zimbra/cs/util/JMSession.java
@@ -34,6 +34,7 @@ import javax.mail.Session;
 import com.google.common.base.Joiner;
 import com.zimbra.common.account.ZAttrProvisioning.DataSourceAuthMechanism;
 import com.zimbra.common.account.ZAttrProvisioning.ShareNotificationMtaConnectionType;
+import com.zimbra.common.account.ZAttrProvisioning.SmtpStartTlsMode;
 import com.zimbra.common.localconfig.LC;
 import com.zimbra.common.net.SocketFactories;
 import com.zimbra.common.service.ServiceException;
@@ -139,6 +140,7 @@ public final class JMSession {
         boolean isAuthRequired = ds.isSmtpAuthRequired();
         String smtpUser = ds.getSmtpUsername();
         String smtpPass = ds.getDecryptedSmtpPassword();
+        ds.getDomain();
         if (DataSourceAuthMechanism.XOAUTH2.name().equalsIgnoreCase(ds.getAuthMechanism())) {
             smtpPass = ds.getDecryptedOAuthToken();
         }
@@ -237,6 +239,7 @@ public final class JMSession {
         }
 
         Properties props = getJavaMailSessionProperties(server, domain);
+        configureStartTls(props,server,domain);
         Session session = Session.getInstance(props);
         setProviders(session);
         if (LC.javamail_smtp_debug.booleanValue()) {
@@ -347,10 +350,38 @@ public final class JMSession {
         if (domain != null) {
             props.setProperty("mail.host", domain.getName());
         }
-
+        
         return props;
     }
 
+    public static Properties configureStartTls (Properties props, Server server, Domain domain) {
+        SmtpStartTlsMode startTlsMode = null;
+        if (domain != null) {
+            startTlsMode = domain.getSmtpStartTlsMode();
+        }
+        if (startTlsMode == null) {
+            startTlsMode = server.getSmtpStartTlsMode();
+        }
+
+        if (startTlsMode.isOff()) {
+            props.setProperty("mail.smtp.starttls.enable", "false");
+        } else if (startTlsMode.isOnly()) {
+            props.setProperty("mail.smtp.starttls.enable", "true");
+            props.setProperty("mail.smtp.starttls.required", "true");
+            props.setProperty("mail.smtp.ssl.trust", "*");
+        } else {
+            if (!startTlsMode.isOn()) {
+                // Should not be reached here
+                ZimbraLog.smtp.warn("invalid value configured for %s. fallback to \"on\".", Provisioning.A_zimbraSmtpStartTlsMode);
+            }
+            props.setProperty("mail.smtp.starttls.enable", "true");
+            props.setProperty("mail.smtp.starttls.required", "false");
+            props.setProperty("mail.smtp.ssl.trust", "*");
+        }
+        
+        return props;
+    }
+    
     /**
      * Add OAuth2 properties to a JAVA Mail Session
      * @param oauthToken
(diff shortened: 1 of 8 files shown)


store/src/java/com/zimbra/cs/util/JMSession.java
```diff
@@ -347,10 +351,38 @@ private static Properties getJavaMailSessionProperties(Server server, Domain dom
...
+            props.setProperty("mail.smtp.starttls.required", "true");
+            props.setProperty("mail.smtp.ssl.trust", "*");
+        } else {
+            if (!startTlsMode.isOn()) {
+                // Should not be reached here
+                ZimbraLog.smtp.warn("invalid value configured for %s. fallback to \"on\".", Provisioning.A_zimbraSmtpStartTlsMode);
+            }
+            props.setProperty("mail.smtp.starttls.enable", "true");
+            props.setProperty("mail.smtp.starttls.required", "false");
+            props.setProperty("mail.smtp.ssl.trust", "*");
```
plobbes: I believe this is disabling certificate verification, right? I'm not sure that should be our default setting, although this is certainly something that could be configurable.

MakiWatanabe: The code only talk to Zimbra MTA (specified by zimbraSmtpHostname), I didn't think we need to verify the server cert on TLS. However I remove these lines to set mail.smtp.ssl.trust to avoid hard coding.

MakiWatanabe:
Codacy failed as "Codacy was unable to analyse your pull request." but it seems there are no problems in "Details"->"View logs".

plobbes:
FWIW, we may want to consider looking into support of LC ssl_allow_untrusted_certs and ssl_allow_mismatched_certs like some other components to allow things to continue to work even when certs are messed up

MakiWatanabe:
Do you think we should support another ldap attribute, something like zimbraSmtpSSLTrust with following spec?

If this attribute is defined, set the value to mail.smtp.ssl.trust.
\* = trust all host
a whitespace separated list of host  = trust these hosts

If this attribute is not defined, we won't define mail.smtp.ssl.trust property.

MakiWatanabe:
@plobbes , here is my current view on the SSL Trust issue.

1. We MUST define mail.smtp.ssl.trust in the code. Without it, the code uses JVM truststore at /opt/zimbra/common/etc/java/cacert. It only includes well-known CAs plus the self-signed cert of the local server. In multi-node installation, we will need to add self-signed certs of all MTA nodes to the truststore.  It will be painful and error plone operation for users.

2. mail.smtp.ssl.trust can be configured as "*"(trust any hosts), or "a white space separated list of trusted hosts".  I believe we should make "*" as the default value in following reasons:
We can create "trusted host list" on the fly from ldap attribute (zimbraSmtpHostname or zimbraShareNotificationMtaHostname), but we need to do it EACH TIME on creating SmtpSession. It almost never happens that zimbraSmtpHostname or zimbraShareNotificationMtaHostname points fake nodes.
zimbraDataSourceSmtpHost is foreign host in general, but getSession(Datasource) code doesn't set mail.smtp.ssl.trust, so the server cert is verified with well-known CA certs in JVM's truststore.

To conclude, I believe we should add a new ldap attribute (ex. zimbraSmtpStartTlsTrustHosts) to configure mail.smtp.ssl.trust property, and make the default to "*".

store/src/java/com/zimbra/cs/util/JMSession.java
```diff
@@ -351,6 +353,43 @@ private static Properties getJavaMailSessionProperties(Server server, Domain dom
...
+            props.setProperty("mail.smtp.starttls.enable", "false");
+        } else if (startTlsMode.isOnly()) {
+            props.setProperty("mail.smtp.starttls.enable", "true");
+            props.setProperty("mail.smtp.starttls.required", "true");
+            if (sslTrustedHosts != null) {
+                props.setProperty("mail.smtp.ssl.trust", sslTrustedHosts);
+            }
+        } else {
+            if (!startTlsMode.isOn()) {
+                ZimbraLog.smtp.warn("invalid value configured for %s. fallback to \"on\".", Provisioning.A_zimbraSmtpStartTlsMode);
```
plobbes: nit: change `"invalid value configured for %s. fallback to \"on\"."` to ...
something more consistent with other parts of the product (ha - if only we were consistent :) ) here's one way we've logged similar issues:
`.../AttributeInfo.java:                    ZimbraLog.misc.warn("Invalid value '%s' for property %s of attribute %s.  Defaulting to %d."`
so perhaps:
`ZimbraLog.smtp.warn("Invalid value '%s' for %s. Defaulting to 'on'.", sslTrustedHosts, Provisioning.A_zimbraSmtpStartTlsMode);`
or just
`ZimbraLog.smtp.warn("Invalid value for %s. Defaulting to 'on'.", Provisioning.A_zimbraSmtpStartTlsMode);`

plobbes:
@MakiWatanabe Regarding the statement "... JVM truststore at /opt/zimbra/common/etc/java/cacert. It only includes well-known CAs plus the self-signed cert of the local server. In multi-node installation, we will need to add self-signed certs of all MTA nodes to the truststore. It will be painful and error plone operation for users."

In theory, this is the right way to do things and **should** be manageable. However, given this is a new feature for ZCS, we are not introducing any new, or major risks by defaulting to "*". I suspect our customers would find this acceptable. For those that don't, they can set this to "" and ensure the certificates are managed properly and have the more ideal solution from the security perspective.

store/src/java/com/zimbra/cs/util/JMSession.java
MakiWatanabe: I really appreciate your help. Will be fixed in next commit.

MakiWatanabe:
@plobbes , OK, I'll change the default to "undefined" = use JVM's truststore.

@plobbes , @chinmay15 , I was advised to retract this PR until #540 will be accepted and merged, then rebase this branch.
I'll close this PR once, and reopen later. Sorry for the messy branch management.

PR #534, opened 2018-02-22
ZCS-4012:SSL Server Private Key Leakage

sneha-patil-synacor:
Added a debug config 'restricted_server_ldap_attributes'. The value contains comma separated list of ldap attributes which needs to be excluded from soap response. Default value is 'zimbraSSLPrivateKey'

Added junit test.
ant test-all passed.


fatkudu:
How does this affect the working of Admin Console - does the lack of this attribute cause any problems?

plobbes:
FWIW, here are some references to how we treat similar issues in other parts of our code:
- https://github.com/Zimbra/zm-mailbox/blob/develop/store/src/java/com/zimbra/cs/account/Provisioning.java#L434
- https://github.com/Zimbra/zm-mailbox/blob/develop/store/src/java/com/zimbra/cs/service/admin/GetDomainInfo.java#L129

There might be an argument for handling this in a similar fashion.

fatkudu:
I think Phil's suggestion is well worth thinking about, although an LDAP attribute which contains a list of attrs that should be either dropped or treated that way seems to have value.  Either way though,  Admin Console needs testing and the attribute needs documenting that adding attrs may affect admin tools.  Definitely need release note that this change will affect output of SOAP requests and any custom admin code that relies on those may need updating.

sneha-patil-synacor:
@fatkudu @plobbes 
Made the change consistent for zimbraSSLPrivateKey. It will return the value as VALUE_BLOACKED instead of removing it.
Checked with Admin console, it doesn't seem to be using this value anywhere. Raised ticket to mention this change in release note as well : ZCS-4357

PR #535, opened 2018-02-23
ZCS-4249: Fixed Imap Search Bug with Solr

ilyusha:
The bug was caused by the lack of special character escaping in the `SolrIndex::docFreq` method, which is run to determine whether to execute a query index-first. This breaks IMAP queries that target the `l.field` index field, such as `bcc`, that indexes email headers. The problem occurs because such queries compile to an invalid Solr search string like `l.field:bcc:foo*` (notice the unescaped second colon). This bug can also be reproduced in the web client with the query `#bcc:foo*`.
This doesn't happen with non-wildcard queries because those get expressed as `terms` component filters like `{!terms f=l.field}bcc:foo`

The solution is simply to add character-escaping logic to the `SolrIndex::docFreq` method, similar to how it is handled in the `SolrIndex::search` method.

Additionally, this PR removes several instances of the `acct_id` filter being applied twice to the same query. In the course of development, the `SolrRequestHelper::newQuery` convenience method was added that adds the account filter if necessary; several methods in `SolrIndex` unnecessarily added another identical filter.

I should note that header queries don't actually support wildcards at this time - the query analysis chain for the `l.field` Solr field swallows the asterisk. I have not verified whether this is a regression or if we never supported this functionality, but if this is an issue, then it should be addressed in a separate bug.

### DIFF
--- a/store/src/java/com/zimbra/cs/index/solr/SolrIndex.java
+++ b/store/src/java/com/zimbra/cs/index/solr/SolrIndex.java
@@ -496,12 +496,9 @@ public class SolrIndex extends IndexStore {
                 SolrQuery q = solrHelper.newQuery(accountId).setQuery("*:*").setRows(0);
                 if (SolrUtils.isWildcardQuery(term.text()) || SolrUtils.containsWhitespace(term.text())) {
                     //can't use a terms filter
-                    q.addFilterQuery(termToQuery(term));
+                    q.addFilterQuery(termToQuery(escapeSpecialChars(term)));
                 } else {
                     addTermsFilter(q, Arrays.asList(term));
-                }
-                if (solrHelper.needsAccountFilter()) {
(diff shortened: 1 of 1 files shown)


fatkudu:
Would you mind updating the commit message with the text from this pull request please?

ilyusha:
@fatkudu I agree that account-level filtering should ideally be enforced better - perhaps even with type safety (something like a `FilteredSolrQuery` perhaps). I didn't want to spend time on refactoring with this bug, bug hopefully we can address this soon.

plobbes:
@ilyusha sorry for the late comment (and my lack of digging deeper into the code at this point), but is there any need to do escaping for https://github.com/Zimbra/zm-mailbox/blob/d8c6f6bd2efefb4ff0b59452dc550a27da0bb2b2/store/src/java/com/zimbra/cs/index/solr/SolrIndex.java#L434 too? Or is that a different use case altogether?

PR #538, opened 2018-02-24
ZCS-3948 Account enumeration

rupalid:
Added code to return 404 status code when auth is BASIC irrespective of whether the accounts exists or authentication fails.

Added unit test

1285 | 0 | 0 | 10 | 100.00% | 286.098

### DIFF
--- a/store/src/java/com/zimbra/cs/service/UserServlet.java
+++ b/store/src/java/com/zimbra/cs/service/UserServlet.java
@@ -279,7 +279,7 @@ public class UserServlet extends ZimbraServlet {
             && (req.getMethod().equalsIgnoreCase("POST") || req.getMethod().equalsIgnoreCase("PUT"))) {
             resp.sendError(HttpServletResponse.SC_UNAUTHORIZED, L10nUtil.getMessage(MsgKey.errMustAuthenticate, req));
         } else {
-            if (ctxt.getAuthAccount() instanceof GuestAccount && ctxt.basicAuthAllowed() ) {
+            if (ctxt != null && ctxt.getAuthAccount() instanceof GuestAccount && ctxt.basicAuthAllowed() ) {
                 resp.sendError(HttpServletResponse.SC_UNAUTHORIZED, L10nUtil.getMessage(MsgKey.errMustAuthenticate, req));
             } else {
                 resp.sendError(HttpServletResponse.SC_NOT_FOUND, message);
--- a/store/src/java/com/zimbra/cs/service/UserServlet.java
+++ b/store/src/java/com/zimbra/cs/service/UserServlet.java
@@ -278,13 +278,11 @@ public class UserServlet extends ZimbraServlet {
         } else if (ctxt != null && ctxt.cookieAuthHappened && !ctxt.isCsrfAuthSucceeded()
             && (req.getMethod().equalsIgnoreCase("POST") || req.getMethod().equalsIgnoreCase("PUT"))) {
             resp.sendError(HttpServletResponse.SC_UNAUTHORIZED, L10nUtil.getMessage(MsgKey.errMustAuthenticate, req));
-        } else {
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/service/UserServlet.java
```diff
@@ -278,7 +279,11 @@ private void sendError(UserServletContext ctxt, HttpServletRequest req, HttpServ
             && (req.getMethod().equalsIgnoreCase("POST") || req.getMethod().equalsIgnoreCase("PUT"))) {
             resp.sendError(HttpServletResponse.SC_UNAUTHORIZED, L10nUtil.getMessage(MsgKey.errMustAuthenticate, req));
         } else {
-            resp.sendError(HttpServletResponse.SC_NOT_FOUND, message);
+            if (ctxt.getAuthAccount() instanceof GuestAccount && ctxt.basicAuthAllowed() ) {
```
fatkudu: The above code hints that `ctxt` can be null under some circumstances.

store/src/java/com/zimbra/cs/service/UserServlet.java
```diff
@@ -278,7 +279,11 @@ private void sendError(UserServletContext ctxt, HttpServletRequest req, HttpServ
...
+                resp.sendError(HttpServletResponse.SC_UNAUTHORIZED, L10nUtil.getMessage(MsgKey.errMustAuthenticate, req));
+            } else {
+                resp.sendError(HttpServletResponse.SC_NOT_FOUND, message);
```
fatkudu: Doesn't this still allow account enumeration?

gordyt: Yea at first glance it appears we can still send different responses depending on if the account exists or not.  I didn't see any notes in the ticket specifying exactly what the desired response should be (irregardless of account existence).

store/src/java/com/zimbra/cs/service/UserServlet.java
rupalid: Fixed

store/src/java/com/zimbra/cs/service/UserServlet.java
rupalid: Actually this is to specifically handle the case with basic auth where account does not exist or it failed validation. It is this case that has been mentioned in the report.  Also for the case whether authentication succeeds but  mail item cannot be found , it still makes sense to return a 404 with message "item not found"

store/src/java/com/zimbra/cs/service/UserServlet.java
```diff
@@ -279,7 +279,7 @@ private void sendError(UserServletContext ctxt, HttpServletRequest req, HttpServ
             && (req.getMethod().equalsIgnoreCase("POST") || req.getMethod().equalsIgnoreCase("PUT"))) {
             resp.sendError(HttpServletResponse.SC_UNAUTHORIZED, L10nUtil.getMessage(MsgKey.errMustAuthenticate, req));
         } else {
-            if (ctxt.getAuthAccount() instanceof GuestAccount && ctxt.basicAuthAllowed() ) {
+            if (ctxt != null && ctxt.getAuthAccount() instanceof GuestAccount && ctxt.basicAuthAllowed() ) {
```
fatkudu: Think it would be cleaner codewise to handle ctxt==null before this if branch (or in the if branch)

store/src/java/com/zimbra/cs/service/UserServlet.java
fatkudu: Hmmm.  This seems to work based on smoke and mirrors, if you understand all the code paths to how you got here then this is how it should behave and there aren't even any comments explaining the assumptions at that point.  I don't have a warm fuzzy feeling about that.  Perhaps a generic `sendError` handler like this is not appropriate?

rupalid: I cannot claim to know all the code pathways, but this is the solution  I narrowed down after trying various scenarios and debugging to see what pathways are followed.

PR #542, opened 2018-02-28
ZCS-4247: Bugfix in DbSearchHistory

ilyusha:
The DbSearchHistory::setSavedSearchStatus method, which updates a row in the `searches` DB table, would throw and error if the return value of the stmt.executeUpdate() call is zero. This happens if the update doesn't actually affect any rows, meaning that the search has not been registered in search history. This is actually a valid scenario, however, since it's possible to create a search folder for a query that isn't logged in search history, such as "in:inbox". This fix removes the return value check.


tmclane:
I approve as wel

PR #547, opened 2018-03-07
ZCS-3871:Handle removal of account profile image

sneha-patil-synacor:
Renamed the SaveProfileImage API to ModifyProfileImage
1. ModifyProfileImage with image data - same as how the SaveProfileImage is currently working - add profile image if not present AND replace profile image if already present
2. ModifyProfileImage with no image data - will delete the account profile image if already present

Added Junit test
ant test-all passed
Updated soap doc


store/src/java/com/zimbra/cs/service/mail/ModifyProfileImage.java
```diff
@@ -200,4 +219,38 @@ private int updateDatabase(Mailbox mbox, OperationContext octxt, String imageFol
             MailItem.Type.DOCUMENT, null, customMetaData, false);
         return imgDocItem.getId();
     }
+    
+    private static void removeFromLDAP(Mailbox mbox, ZimbraSoapContext zsc) throws IOException, ServiceException {
+        HashMap<String, Object> prefs = new HashMap<String, Object>();
+        prefs.put(Provisioning.A_thumbnailPhoto, null);
+        Provisioning.getInstance().modifyAttrs(mbox.getAccount(), prefs, true,
+            zsc.getAuthToken());
+        return;
```
shrikant-prasad: no need of return.

soap/src/java/com/zimbra/soap/JaxbUtil.java
```diff
@@ -988,8 +988,8 @@
             com.zimbra.soap.mail.message.SaveDraftResponse.class,
             com.zimbra.soap.mail.message.SaveIMAPSubscriptionsRequest.class,
             com.zimbra.soap.mail.message.SaveIMAPSubscriptionsResponse.class,
-            com.zimbra.soap.mail.message.SaveProfileImageRequest.class,
-            com.zimbra.soap.mail.message.SaveProfileImageResponse.class,
+            com.zimbra.soap.mail.message.ModifyProfileImageRequest.class,
+            com.zimbra.soap.mail.message.ModifyProfileImageResponse.class,
```
fatkudu: Perhaps try and retain the sort order of lines

store/src/java/com/zimbra/cs/service/mail/ModifyProfileImage.java
```diff
@@ -118,10 +129,18 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
                 prefs.put(Provisioning.A_thumbnailPhoto, null);
                 Provisioning.getInstance().modifyAttrs(mbox.getAccount(), prefs, true,
                     zsc.getAuthToken());
+            } else if (removeFromLDAPSuccess && !removeFromDBSuccess) {
+                // Database removal failed; revert to LDAP
+                HashMap<String, Object> prefs = new HashMap<String, Object>();
+                prefs.put(Provisioning.A_thumbnailPhoto, ldapBackup);
```
fatkudu: Do we really need to store the default thumbnail in the prefs for every account?  Seems to me, the absense of it should imply "use the default".

store/src/java/com/zimbra/cs/service/mail/ModifyProfileImage.java
```diff
@@ -200,4 +219,38 @@ private int updateDatabase(Mailbox mbox, OperationContext octxt, String imageFol
...
+            zsc.getAuthToken());
+        return;
+    }
+    
+    private void removeFromDatabase(Mailbox mbox, OperationContext octxt, String imageFolderName)
+        throws ServiceException {
+        int folderId;
+        try {
+            Folder imgFolder = mbox.getFolderByName(octxt, Mailbox.ID_FOLDER_ROOT, imageFolderName);
+            folderId = imgFolder.getId();
```
fatkudu: Could imgFolder be null here - perhaps for old mailboxes that existed before this feature?

sneha-patil-synacor: NO_SUCH_FOLDER is being handled in catch

store/src/java/com/zimbra/cs/service/mail/ModifyProfileImage.java
sneha-patil-synacor: Done

soap/src/java/com/zimbra/soap/JaxbUtil.java
sneha-patil-synacor: Fixed

sneha-patil-synacor:
@fatkudu 
The reason for having it in ldap as well as mailbox is:
1. LDAP for GAL support - only thumbnail
2. Mailbox for rest support - full image
We are only storing thumbnail version of image to LDAP because of LDAP size restriction. And the actual image will reside on the mailbox for web client use.
We can not store full image into LDAP, as it will create huge ldap data dump.

store/src/java/com/zimbra/cs/service/mail/ModifyProfileImage.java
sneha-patil-synacor: this code is basically restoring the thumbnail image to LDAP in case of database image removal fails. To make sure that if the image is deleted, it is deleted from LDAP as well as database, otherwise not deleted from anywhere.

store/src/java/com/zimbra/cs/service/mail/ModifyProfileImage.java
fatkudu: Fair enough, looks like `getFolderByName` can't return null.

PR #550, opened 2018-03-08
ZCS-4347:BuildInfo.TYPE uses zimbra-license.jar

fatkudu:
We're preparing to make NETWORK edition an overlay on top of FOSS
edition.  To achieve this, it no longer makes sense to embed
the flavor of Zimbra in the zm-store jar.

This change means that BuildInfo.TYPE will choose whether it is FOSS
or NETWORK depending in the absence or presence of:
    /opt/zimbra/lib/ext/zimbra-license/zimbra-license.jar
This seems like a more sensible marker than what is currently used
by zmcontrol (zmbackupquery).  I may change zmcontrol too.

The only thing I could find in the build sandbox that uses
BuildInfoGenerated is BuildInfo, hence why I'm removing TYPE from
that.

### DIFF
--- a/store/src/java/com/zimbra/cs/util/BuildInfo.java
+++ b/store/src/java/com/zimbra/cs/util/BuildInfo.java
@@ -28,8 +28,11 @@ import com.zimbra.cs.db.Versions;
 
 public final class BuildInfo {
 
+    public static final String TYPE_NETWORK = "NETWORK";
+    public static final String TYPE_FOSS = "FOSS";
+
+    public static final String TYPE;     /* whether this is a FOSS or NETWORK installation */
     public static final String VERSION;
-    public static final String TYPE;
     public static final String RELEASE;
     public static final String DATE;
     public static final String HOST;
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/util/BuildInfo.java
```diff
@@ -85,6 +87,11 @@
         }
     }
 
+    private static String getType() {
```
Prashantsurana: Instead of hardcoded location. Can we use reflection and see if some class or java package of zimbra-license present in the classpath.

tmclane: That's a good idea!

fatkudu: My suspicion is that will be no more reliable and might be more brittle if a caller doesn't have a classpath including the licensing extensions.  So, I would prefer not to do that unless there is a compelling reason for it.  I kind of think that any future change which involves a rename of the jar file will very quickly discover this code and change it appropriately.

Prashantsurana: How about checking /opt/zimbra/bin/zmlicense utility? 
keeping flexibility on jar names or extension location.

fatkudu: Yes, I think that would be better.  Will make that change.

PR #553, opened 2018-03-08
ZCS-4242: Upgrade to Solr 7

ilyusha:
This PR bumps the `Solr/Lucene` ivy dependencies to `7.2.1` and resolves a few issues caused by API changes that comes with the version upgrade.

Relevant Lucene 7 API changes:
- `TopDocs.totalHits` is now a `long` instead of an `int`. For our use case, it's safe to just cast it to an integer
- `Field.fieldType()` now returns `IndexableFieldType`. This affects the `TermInfo` class, though I'm not sure if this class is even used anymore.

### DIFF
--- a/store/src/java/com/zimbra/cs/index/ZimbraTopDocs.java
+++ b/store/src/java/com/zimbra/cs/index/ZimbraTopDocs.java
@@ -65,7 +65,7 @@ public class ZimbraTopDocs {
      * Create equivalent ZimbraTopDocs object to a Lucene TopDocs object
      */
     public static ZimbraTopDocs create(TopDocs luceneTopDocs) {
-        return new ZimbraTopDocs(luceneTopDocs.totalHits,
+        return new ZimbraTopDocs((int) luceneTopDocs.totalHits,
                 ZimbraScoreDoc.listFromLuceneScoreDocs(luceneTopDocs.scoreDocs), luceneTopDocs.getMaxScore());
     }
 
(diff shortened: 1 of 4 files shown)


store/src/java/com/zimbra/cs/index/ZimbraTopDocs.java
```diff
@@ -65,7 +65,7 @@ public static ZimbraTopDocs create(int totalHits, List<ZimbraScoreDoc> scoreDocs
      * Create equivalent ZimbraTopDocs object to a Lucene TopDocs object
      */
     public static ZimbraTopDocs create(TopDocs luceneTopDocs) {
-        return new ZimbraTopDocs(luceneTopDocs.totalHits,
+        return new ZimbraTopDocs((int) luceneTopDocs.totalHits,
```
fatkudu: How difficult would it be use long instead?  Would seem better to fit with the underlying type.  Just a minor push back as I assume it would be kind of scary if we blew an int here.

ilyusha: I actually originally tried to propagate the long, but it seems to be more trouble than it's worth if we want to avoid the down-cast anywhere. To quote https://issues.apache.org/jira/browse/LUCENE-7872:

```Even though a single index cannot have more than 2B documents, TopDocs.merge might merge TopDocs instances that have more than 2B documents in total.```

In other words, an `int` is enough to hold the number of results from a single shard. And since we are co-locating mailbox documents, it's not actually possible at this time to have an overflow here, since we'd never be merging results from different shards.

PR #554, opened 2018-03-09
Task/zcs 4381 - generate an ldap schema file + add zimbraLDAPSchemaVersion attribute

tmclane:
A mechanism was needed to be added such that a given `zimbra-store` package could know which version of the LDAP schema it was built against.

This PR adds a new ant target `create-version-ldap` which generates a file which is utilized by scripts in the `zm-build` repository.

### DIFF
--- a/pkg-builder.pl
+++ b/pkg-builder.pl
@@ -350,6 +350,7 @@ sub stage_zimbra_common_mbox_conf_attrs()
    cpy_file( "store/conf/attrs/amavisd-new-attrs.xml", "$stage_base_dir/opt/zimbra/conf/attrs/amavisd-new-attrs.xml" );
    cpy_file( "store/conf/attrs/zimbra-attrs.xml",      "$stage_base_dir/opt/zimbra/conf/attrs/zimbra-attrs.xml" );
    cpy_file( "store/conf/attrs/zimbra-ocs.xml",        "$stage_base_dir/opt/zimbra/conf/attrs/zimbra-ocs.xml" );
+   cpy_file( "store/build/dist/conf/attrs/zimbra-attrs-schema",   "$stage_base_dir/opt/zimbra/conf/zimbra-attrs-schema" );
 
    return ["store/conf/attrs"];
 }
--- a/store/build.xml
+++ b/store/build.xml
@@ -86,6 +86,7 @@
     <mkdir dir="${base.dir}/${slapd.conf.dir}"/>
     <mkdir dir="${base.dir}/${slapd.etc.dir}/schema"/>
     <mkdir dir="${base.dir}/bin"/>
+    <mkdir dir="${base.dir}/conf/attrs"/>
     <mkdir dir="${base.dir}/conf/msgs"/>
     <mkdir dir="${base.dir}/db"/>
     <mkdir dir="${base.dir}/lib/ext"/>
@@ -392,6 +393,14 @@
       </classpath>
        <arg line="-o ${dist.dir}"/>
     </java>
+  </target>
+  <target name="create-version-ldap" depends="build-init" description="Creates ldap schema version: zimbra-attrs-schema">
+    <exec executable="git" failonerror="false" output="${dist.dir}/conf/attrs/zimbra-attrs-schema">
+      <arg value="log"/>
+      <arg value="-1"/>
+      <arg value="--pretty=format:%at"/>
+      <arg value="conf/attrs/zimbra-attrs.xml"/>
+    </exec>
   </target>
   <target name="setup-for-soap-tests" description="Does necessary setup for RunUnitTestsRequests">
     <exec executable="sudo" failonerror="false">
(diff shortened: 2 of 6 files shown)


store/build.xml
```diff
@@ -393,6 +393,14 @@
        <arg line="-o ${dist.dir}"/>
     </java>
   </target>
+  <target name="create-version-ldap" depends="build-init" description="Creates ldap schema version: zimbra-attrs-schema.txt">
+    <exec executable="git" failonerror="false" output="conf/attrs/zimbra-attrs-schema">
```
fatkudu: As it isn't stored under git I would prefer if this was created under a build/ directory.

tmclane: I'll work on getting this to happen.  I meant to at least add it to the .gitignore file but this is a better idea!

tmclane:
@fatkudu Moved it to the 'build' directory.

store/build.xml
```diff
@@ -393,6 +393,14 @@
        <arg line="-o ${dist.dir}"/>
     </java>
   </target>
+  <target name="create-version-ldap" depends="build-init" description="Creates ldap schema version: zimbra-attrs-schema">
+    <exec executable="git" failonerror="false" output="${dist.dir}/zimbra-attrs-schema">
```
fatkudu: I would still expect this to go under somewhere like conf/attrs/ under ${dist.dir} - may need to create that directory if it isn't there already.

store/build.xml
```diff
@@ -393,6 +394,14 @@
        <arg line="-o ${dist.dir}"/>
     </java>
   </target>
+  <target name="create-version-ldap" depends="build-init" description="Creates ldap schema version: zimbra-attrs-schema">
+    <exec executable="git" failonerror="false" output="${dist.dir}/conf/attrs/zimbra-attrs-schema">
```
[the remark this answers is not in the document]
Prashantsurana: Make sure this file should be part of zimbra-common-mbox-conf-attrs*.deb/rpm package.

pkg-builder.pl
```diff
@@ -350,6 +350,7 @@ ()
    cpy_file( "store/conf/attrs/amavisd-new-attrs.xml", "$stage_base_dir/opt/zimbra/conf/attrs/amavisd-new-attrs.xml" );
    cpy_file( "store/conf/attrs/zimbra-attrs.xml",      "$stage_base_dir/opt/zimbra/conf/attrs/zimbra-attrs.xml" );
    cpy_file( "store/conf/attrs/zimbra-ocs.xml",        "$stage_base_dir/opt/zimbra/conf/attrs/zimbra-ocs.xml" );
+   cpy_file( "store/conf/attrs/zimbra-attrs-schema",        "$stage_base_dir/opt/zimbra/conf/attrs/zimbra-attrs-schema" );
```
tmclane: Note: This change causes the following to appear:

`
zimbra@zcs-dev-two:~$ zmprov gcf zimbraLDAPSchemaVersion

[] WARN: while loading attrs, ignoring not .xml file: /opt/zimbra/conf/attrs/zimbra-attrs-schema
zimbraLDAPSchemaVersion: 1520374873`

fatkudu: Hmm.  That suggests the attrs/ sub-directory probably isn't a good location for the file.  Either that or some clever stuff is needed to suppress the error.

tmclane: I was afraid that would be the answer.  Any other suggestions? 
I thought maybe a hidden file would work:

`[] WARN: while loading attrs, ignoring not .xml file: /opt/zimbra/conf/attrs/.zimbra-attrs-schema`

Alas it was not to be.

fatkudu: How about in /opt/zimbra/conf?

tmclane: Offending code in AttributeManager.java:288

       for (File file : files) {
            if (!file.getPath().endsWith(".xml")) {
                ZimbraLog.misc.warn("while loading attrs, ignoring not .xml file: %s", file);
                continue;
            }
            if (!file.isFile()) {
                ZimbraLog.misc.warn("while loading attrs, ignored non-file: %s", file);
            }

Any reason we couldn't just drop those lines entirely? If the code doesn't care about anything but XML files is there any reason to warn someone and ignore the file when it could just be ignored with no warning.

tmclane: Looks like we have some choices now:

- move the file to /opt/zimbra/conf
- update xml finder code to ignore non-xml files silently
- rename the file to `.zimbra-attrs-schema` and update code to ignore files with a leading '.' character

tmclane: FYI: Going with option 1: move the file to /opt/zimbra/conf

store/build.xml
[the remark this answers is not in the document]
tmclane: This is done.

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9629,4 +9629,9 @@ TODO: delete them permanently from here
     <desc>Account thumbnail photo</desc>
 </attr>
 
+<attr id="3022" name="zimbraLDAPSchemaVersion" type="string" cardinality="single" requiredIn="globalConfig" since="8.8.8">
```
fatkudu: Note that you are going to need to choose a new id for this.  I think this one is taken now...

tmclane:
Rebased and pushed.

PR #562, opened 2018-03-16
ZCS-4380 Change name of zimbra jars pkg

rupalid:
Change name of zimbra jars pkg from zimbra-core-jar to zimbra-common-core-jar

### DIFF
--- a/pkg-builder.pl
+++ b/pkg-builder.pl
@@ -66,7 +66,7 @@ my %PKG_GRAPH = (
       hard_deps => [
          "zimbra-mbox-war",
          "zimbra-mbox-conf",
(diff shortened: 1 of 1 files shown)


pkg-builder.pl
```diff
@@ -190,7 +190,7 @@ ($)
       stage_fun  => sub { &stage_zimbra_common_mbox_docs(@_); },
    },
    
-   "zimbra-core-jar" => {
+   "zimbra-common-core-jar" => {
```
Prashantsurana: Minor: good if you change name of staging function

rupalid: done

PR #563, opened 2018-03-19
NICPS-18 Modify optionalIn of some OAuth ldap attributes

MakiWatanabe:
Requirement
---
We want to sync Zimbra calendar app to Google Calendar.

Fix
---
We will add OAuth2 authorization mechanism to CalDav Datasource.
In this PR, I'd like to modify the definition of existing ldap attributes.
We already have OAuth2 related ldap attributes since 8.7.0, but the scope are not appropriate.

1. zimbraDataSourceOAuthClientId - client_id is the id for each OAuth client application. This value should be stored in global config or domain level, rather than in each DataSource object.

1. zimbraDataSourceOAuthClientSecret - client_secret is a string used with client_id to identify the OAuth client application. This value should be stored in global config or domain level, rather than in the each DataSource object.

1. zimbraDataSourceOAuthRefreshTokenUrl - refresh token Url is the OAuth service endpoint which accept "refresh token" request. This value should be stored in global config or domain level, rather than in the each DataSource object.

1. zimbraDataSourceOAuthToken - This is for access_token, which is a token string to authorize the access to the hosted resources. This value should be stored for each user, so we should store it in DataSource object. access_token for Google services expires in 3600s. In large scale deployment, we should better to store it in the ephemeral storage.

1. zimbraDataSourceOAuthRefreshToken - Refresh token is a token string used to refresh access_token. This value should be stored for each user, and won't be updated often. We should store it in DataSource object. In this PR, I don't modify the definition of this attribute.

Testing done
---
Testing done by developer

Testing to be done by QA
---
Testing needed to be done by QA

Linked PR
---
NA

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -8461,7 +8461,7 @@ TODO: delete them permanently from here
   <desc>whether or not an account represents a Mobile Gateway app</desc>
 </attr>
 
-<attr id="1761" name="zimbraDataSourceOAuthToken" type="string" cardinality="single" optionalIn="imapDataSource" since="8.7.0">
+<attr id="1761" name="zimbraDataSourceOAuthToken" type="string" cardinality="single" optionalIn="dataSource,imapDataSource" flags="ephemeral" since="8.7.0">
   <desc>OAuth token for authentication using OAuth</desc>
 </attr>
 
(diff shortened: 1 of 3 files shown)


store/conf/attrs/zimbra-attrs.xml
```diff
@@ -8461,7 +8461,7 @@ TODO: delete them permanently from here
   <desc>whether or not an account represents a Mobile Gateway app</desc>
 </attr>
 
-<attr id="1761" name="zimbraDataSourceOAuthToken" type="string" cardinality="single" optionalIn="imapDataSource" since="8.7.0">
+<attr id="1761" name="zimbraDataSourceOAuthToken" type="string" cardinality="single" optionalIn="dataSource,imapDataSource" flags="ephemeral" since="8.7.0">
```
fatkudu: We will need a lot of testing and probably changes in this area when we add new ephemeral attributes.
Suspect it will work fine when first switching on using an ephemeral store but if that is already in use, adding new ones may/will probably lose all values stored in LDAP for that new attribute (that may not be a big issue in this case but it needs thinking about)

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -8946,19 +8946,19 @@ TODO: delete them permanently from here
   <desc>Subject prefix for the spam training messages used to sent to the zimbraSpamIsSpamAccount/zimbraSpamIsNotSpamAccount account.</desc>
 </attr>
 
-<attr id="2021" name="zimbraDataSourceOAuthClientId" type="string" cardinality="single" optionalIn="dataSource,imapDataSource" since="8.7.0,9.0.0">
+<attr id="2021" name="zimbraDataSourceOAuthClientId" type="string" cardinality="single" optionalIn="globalConfig,domain,dataSource,imapDataSource" since="8.7.0,9.0.0">
```
fatkudu: Probably need upgrade testing to make sure that anything working before an upgrade continues to work.

store/conf/attrs/zimbra-attrs.xml
MakiWatanabe: I understand. It seems the attribute is not used in zm-mailbox now, but it will be better to  create a new attribute for NIC project. Thanks.

store/conf/attrs/zimbra-attrs.xml
MakiWatanabe: @fatkudu , I'll retract this PR, and discuss about the design in project. Thank you for your input.

PR #566, opened 2018-03-22
ZBUG-184 Fixing issue with span tag

rupalid:
Added new regex to filter out patter containing  repetition of (') character.

Added unit test.

Injected faulty mime and verified that email is loaded.


plobbes:
Sorry, didn't get time to test and this review was done really quickly. Just looking at that regex, it sure seems like you could come up with ways to work around that. For example '<space>'<space>'<space>'... (spaces between each single quote) or perhaps alternating '&apos;'&apos;'... which may work depending upon whether or not character translations have already occurred. Pretty sure I'm just scraping the top of the ice berg here.

PR #568, opened 2018-03-23
ZCS-1695 Moving stack trace to debug level

rupalid:
Moving stack trace to debug level

### DIFF
--- a/common/src/java/com/zimbra/common/util/Log.java
+++ b/common/src/java/com/zimbra/common/util/Log.java
@@ -287,6 +287,14 @@ public class Log {
     public void warn(Object o, Throwable t) {
         getLogger().warn(o, t);
     }
+    
+    public void warnQuietly(Object o, Throwable t) {
+        if (isDebugEnabled()) {
+            getLogger().warn(o, t);
+        } else {
+            getLogger().warn(String.format("%s : %s", o, t.getMessage()));
+        }
+    }
 
     public void warn(String format, Object ... objects) {
         if (isWarnEnabled()) {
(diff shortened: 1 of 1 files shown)


common/src/java/com/zimbra/common/util/Log.java
```diff
@@ -287,6 +287,14 @@ public void warn(Object o) {
     public void warn(Object o, Throwable t) {
         getLogger().warn(o, t);
     }
+    
+    public void warnQuietly(Object o, Throwable t) {
+        if (isDebugEnabled()) {
+            getLogger().warn(o, t);
```
chinmay15: Is .warn correct here? Just confirming because in above if "isDebugEnabled()" is checked.

rupalid: Fixed

common/src/java/com/zimbra/common/util/Log.java
```diff
@@ -287,6 +287,14 @@ public void warn(Object o) {
     public void warn(Object o, Throwable t) {
         getLogger().warn(o, t);
     }
+    
+    public void warnQuietly(Object o, Throwable t) {
+        if (isDebugEnabled()) {
+            getLogger().debug(o, t);
```
fatkudu: Please change this (back) to warn. (same for other one lower down)
This method is intended to be used for warnings, so the logged result should reflect that, grepping for WARN should find it.  The fact that a more detailed warning is emitted because debug is enabled is irrelevant.

rupalid: Fixed

PR #569, opened 2018-03-23
ZCS-1200:Need way to query server for total number of GAL entries

shrikant-prasad:
added optional attribute 'getCount' in SyncGalRequest. response will have 'remain' attribute when getCount= true and it will contain count of remaining items".

### DIFF
--- a/soap/src/java/com/zimbra/soap/account/message/SyncGalRequest.java
+++ b/soap/src/java/com/zimbra/soap/account/message/SyncGalRequest.java
@@ -64,6 +64,14 @@ public class SyncGalRequest {
     private ZmBoolean idOnly;
 
     /**
+     * @zm-api-field-tag getCount.
+     * @zm-api-field-description Flag whether count of remaining records should be returned in response or not.
+     * Note: <b>getCount</b> works only when idOnly is set to true and GAL sync account is configured/enabled.
+     */
+    @XmlAttribute(name=AccountConstants.A_GET_COUNT /* getCount */, required=false)
+    private ZmBoolean getCount;
+
+    /**
      * @zm-api-field-tag limit
      * @zm-api-field-description Page size control for SyncGalRequest. The maximum entries that can be returned for
  every SyncGal Request can be controlled by specifying this limit.
@@ -89,6 +97,7 @@ public class SyncGalRequest {
             .add("token", token)
             .add("galAccountId", galAccountId)
             .add("idOnly", idOnly)
+            .add("getCount", getCount)
             .add("limit", limit);
     }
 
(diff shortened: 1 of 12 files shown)


store/src/java/com/zimbra/cs/db/DbMailItem.java
```diff
@@ -2778,6 +2779,54 @@ public static UnderlyingData getByHash(Mailbox mbox, String hash) throws Service
         }
     }
 
+    public static int getModifiedItemsCount(Mailbox mbox, MailItem.Type type, long lastSync,
+            int sinceDate, Set<Integer> visible, int lastDeleteSync)
```
chinmay15: remove unused param lastDeleteSync

store/src/java/com/zimbra/cs/db/DbMailItem.java
```diff
@@ -2778,6 +2779,54 @@ public static UnderlyingData getByHash(Mailbox mbox, String hash) throws Service
...
+    public static int getModifiedItemsCount(Mailbox mbox, MailItem.Type type, long lastSync,
+            int sinceDate, Set<Integer> visible)
+    throws ServiceException {
+        if (Mailbox.isCachedType(type)) {
+            throw ServiceException.INVALID_REQUEST("folders and tags must be retrieved from cache", null);
+        }
+        DbConnection conn = mbox.getOperationConnection();
+        PreparedStatement stmt = null;
+        try {
+            String typeConstraint = type == MailItem.Type.UNKNOWN ? "type NOT IN " + NON_SYNCABLE_TYPES : typeIn(type);
```
fatkudu: There are a lot of intermediate Strings being created here to create the statement.  How about sticking with `StringBuilder` syntax and doing something like:

    buf.append("SELECT COUNT(id)")
        .append(" FROM ")
        .append(getMailItemTableName(mbox))
        .append(" WHERE ")
        .append(IN_THIS_MAILBOX_AND)
        .append("mod_metadata > ? AND ");
        .append(sinceDate > 0 ? "date > ? AND " : "")
        .append(MailItem.Type.UNKNOWN ? "type NOT IN " + NON_SYNCABLE_TYPES : typeIn(type))
    if (visible != null && visible.size() > 0) {
        String folderString = visible.stream().map(i -> i.toString()).collect(Collectors.joining(", "));
        buf.append(" AND folder_id IN ( ").append(folderString).append(" )");
    }

soap/src/java/com/zimbra/soap/account/message/SyncGalRequest.java
```diff
@@ -63,6 +63,13 @@
     @XmlAttribute(name=AccountConstants.A_ID_ONLY /* idOnly */, required=false)
     private ZmBoolean idOnly;
 
+    /**
+     * @zm-api-field-tag getCount
+     * @zm-api-field-description Flag whether count of remaining records should be returned in response or not.
```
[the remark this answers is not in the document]
shrikant-prasad: Done

store/src/java/com/zimbra/cs/db/DbMailItem.java
shrikant-prasad: Done

PR #570, opened 2018-03-26
ZCS-4276 mailbox locking

fatkudu:
Believe this is a useful state, although still want to refactor mailbox transactions and get rid of Mailbox.lock() which doesn't actually do that any more...


gordyt:
Thank you @fatkudu!  I am going to do some local testing with this branch now.

PR #575, opened 2018-04-02
Bugfix/zcs 4691 failed mailbox unlocks

fatkudu:
Discovered that if you don't set a lease time for locks (we don't currently) then the default is 30 seconds.  Made changes to use longer lease times.  Also added diagnostics so that we can get a feel for how long we actually do locks.  Lock objects get an id so that can track their behavior.  At least temporarily, logging some stuff at info level.  Still need robustness against failed unlocks in case we go past the lease times.
The other listener related change is slightly speculative and needs a good review to make sure logic is OK for competing threads entering/leaving maintenance and adding/removing listeners.

### DIFF
--- a/store/src/java/com/zimbra/cs/mailbox/Mailbox.java
+++ b/store/src/java/com/zimbra/cs/mailbox/Mailbox.java
@@ -1105,42 +1105,47 @@ public class Mailbox implements MailboxStore {
             return;
         }
         assert(session.getSessionId() != null);
+        if (maintenance != null) {
+            throw MailServiceException.MAINTENANCE(mId);
+        }
+        ZimbraLog.mailbox.debug("adding listener: %s", session);
+        if (!mListeners.contains(session)) {
+            mListeners.add(session);
+        }
+        /* check whether beginMaintenance happened whilst adding a listener.  If it did then
+         * undo it.  This avoids getting a write lock */
+        if (maintenance != null) {
+            purgeListeners();
+            throw MailServiceException.MAINTENANCE(mId);
+        }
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -1105,74 +1105,79 @@ public void addListener(Session session) throws ServiceException {
...
+        if (maintenance != null) {
+            throw MailServiceException.MAINTENANCE(mId);
+        }
+        ZimbraLog.mailbox.debug("adding listener: %s", session);
+        if (!mListeners.contains(session)) {
+            mListeners.add(session);
+        }
+        /* check whether beginMaintenance happened whilst adding a listener.  If it did then
+         * undo it.  This avoids getting a write lock */
+        if (maintenance != null) {
```
tmclane: Could we combine this maintenance check with the previous one?
Or perhaps remove the first one and let them all be handled here?

fatkudu: writes to mListener involve discarding the previous array and creating a new one, so avoiding this is worth doing - so I'd rather not just rely on the 2nd if.  With 2 ifs, the 2nd check should  hardly ever succeed (but it will sometimes)  I consider it is better to have this 2 stage check than to lock the mailbox.

tmclane: Fair enough. It was just a thought.

fatkudu:
Closed in favor of https://github.com/Zimbra/zm-mailbox/pull/577 which targets feature/ha

PR #576, opened 2018-04-03
ZBUG-184, ZBUG-201 changes for patch

rupalid:
Porting changes for the following bugs to 8.8.8.p1 branch


mgarbin:
hi guys, i will inform you that there is another bug that cause high cpu consume and loop.

File store/src/java/com/zimbra/cs/html/DefangFilter.java
function "private static String sanitizeStyleValue(String value)"
The problem is inside this code: 
value = STYLE_UNWANTED_FUNC.matcher(value).replaceAll("");

Try the last attachment inside this https://bugzilla.zimbra.com/show_bug.cgi?id=108928
A solution is to set a timeout to the matcher function like this https://www.exratione.com/2017/06/preventing-unbounded-regular-expression-operations-in-java/ or set limit word limit inside the regex ...

PR #580, opened 2018-04-04
ZCS-4455: DbLogWriter Connection Handling

ilyusha:
This PR addresses a problem with The `DbLogWriter` class, hopefully resolving the `com.zimbra.cs.redolog.logger.DbLogWriter com.zimbra.cs.redolog.logger.DbLogWriter failed` error.

Prior to this change, `DbLogWriter` maintained a single `DbConnection` instance in the `conn` variable. This connection was created at `RedoLogManager` startup and only closed in the `stop` method, which is only called during system shutdown. This connection was responsible for handling all DB calls for the duration of the `DbLogWriter` lifetime; it is entirely possible that the intermittent error is due to it timing out or being otherwise corrupted.

This change makes it so that every database call in this class uses a new `DbConnection` instance (for reference, this is how all other DB-bound classes function). For convenience, this is done via a new inner class `DbLogWriterConn`, which implements `Closeable` so that the methods can use a try-with-resources block. 

A side effect of the previous design was that the `DbLogWriter::log` and `getSize` methods had to be synchronized. With this update, this is no longer the case. Lastly, several instances of `RuntimeException` being thrown has been removed, replaced by an `IOException` declared at the interface level.

### DIFF
--- a/store/src/java-test/com/zimbra/cs/redolog/logger/DbLogWriterTest.java
+++ b/store/src/java-test/com/zimbra/cs/redolog/logger/DbLogWriterTest.java
@@ -43,7 +43,6 @@ public class DbLogWriterTest {
     @Test
     public void openCloseLog() throws Exception {
         logWriter.open();
-        assertTrue("Connection is open successfully", logWriter.isOpen());
         assertTrue("Table has to have header in place when open the redolog first time",
                 (logWriter.getSize() == LogHeader.HEADER_LEN));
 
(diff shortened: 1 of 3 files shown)


tmclane:
I've been testing this on the zm-docker 'feature/ha' environment and was able to reproduce it before applying this change.
After application of this change it appears it does not happen any longer.  I am going to continue testing to see if the DbLogWriter  continues to work. 

I am not expecting any issues.

tmclane:
No failures whatsoever even after waiting overnight.

PR #583, opened 2018-04-10
Bugfix/guava

sneha-patil-synacor:
Updating Guava library to 23.0 version
Changes Done:
1. Objects.toStringHelper is deprecated. Used MoreObjects.toStringHelper instead.
2. Objects.firstNonNull is deprecated. Used MoreObjects.firstNonNull instead.
3. NullOutputStream is deprecated. Used ByteStreams.nullOutputStream() instead.
4. com.google.common.collect.Iterators.emptyIterator() is deprecated. Used Collections.emptyIterator() instead.
5. InternetDomainName.fromLenient(String domain) is deprecated. Used from(String) instead.
6. MapMaker().makeComputingMap is deprecated. Used CacheBuilder instead.
7. closeQuietly(Closeable) is deprecated. using try with resources wherever possible and where it isn't using new IOUtil.closeQuietly method.

ant test-all passed.


client/ivy.xml
```diff
@@ -10,7 +10,7 @@
   <dependency org="jaxen" name="jaxen" rev="1.1-beta-10"/>
   <dependency org="dom4j" name="dom4j" rev="1.5.2" />
   <dependency org="log4j" name="log4j" rev="1.2.16" />
-  <dependency org="com.google.guava" name="guava" rev="13.0.1" />
+  <dependency org="com.google.guava" name="guava" rev="23.0" />
```
fatkudu: I assume that this commit does 2 things:
* updates the dependency in ivy.xml
* uses MoreObjects instead of Objects

In order to save reviewers time scanning through screeds of changes that the compiler is better at validating,
please change the commit comment with useful information.

sneha-patil-synacor: Done

fatkudu: "and few other changes" 👎 So, my assumption was incorrect.  Sorry, but you will need to find other reviewers.

store/src/java/com/zimbra/cs/service/formatter/ArchiveFormatter.java
```diff
@@ -52,13 +52,14 @@
 import javax.mail.internet.MimePart;
 import javax.servlet.http.HttpServletResponse;
 
+import org.mockito.internal.util.io.IOUtil;
```
fatkudu: This probably isn't what you intended.

store/src/java/com/zimbra/cs/service/mail/Search.java
```diff
@@ -107,19 +107,17 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
+            params)) {
             // must use results.getSortBy() because the results might have ignored our sortBy
             // request and used something else...
             response.addAttribute(MailConstants.A_SORTBY, results.getSortBy().toString());
             putHits(zsc, octxt, response, results, params, memberOfMap);
-            return response;
-        } finally {
-            Closeables.closeQuietly(results);
-        }
+        } catch (IOException e) {
```
fatkudu: Why are you swallowine IOException here when weren't before?

store/src/java/com/zimbra/qa/unittest/TestUtil.java
```diff
@@ -456,12 +455,14 @@ public static ZOutgoingMessage getOutgoingMessage(String recipient, String subje
...
-        while (r.hasNext()) {
-            ZimbraHit hit = r.getNext();
-            ids.add(new Integer(hit.getItemId()));
+        try (ZimbraQueryResults r = mbox.index.search(new OperationContext(mbox), query, types,
+            SortBy.DATE_DESC, 100)) {
+            while (r.hasNext()) {
+                ZimbraHit hit = r.getNext();
+                ids.add(new Integer(hit.getItemId()));
+            }
+        } catch (IOException e) {
```
fatkudu: Why swallowing IOException?

store/src/java/com/zimbra/cs/service/formatter/ArchiveFormatter.java
sneha-patil-synacor: Good catch. Fixed.

store/src/java/com/zimbra/cs/service/mail/Search.java
sneha-patil-synacor: since that was the behavior with Closeables.closeQuietly(). It used to swallow any kind of Exception during close operation. When we are using try with resources, it makes it mandatory to catch the IOException if the resource close() method is declared to throw it.

fatkudu: Ah.  Ok.  Thanks for the clarification.

store/src/java/com/zimbra/cs/zookeeper/CuratorManager.java
```diff
@@ -23,6 +23,7 @@
 import java.util.Set;
 import java.util.TreeSet;
 
+import org.apache.commons.compress.utils.IOUtils;
```
fatkudu: Not org.apache.commons.io.IOUtils?  or even com.zimbra.cs.util.IOUtil?

PR #584, opened 2018-04-10
ZCS-4777:redisson 3.6.5 instead of 3.5.6

fatkudu:
Use newest redisson jar.  Hoping may help with some problems where we
lock something, then think we don't have a lock.

Tested after:

* docker cp redisson-3.6.5.jar zm-docker_zmc-mailbox.1.d1kr7qimm8re818k9gpft7gtb:/opt/zimbra/jetty_base/webapps/service/WEB-INF/lib/redisson-3.5.6.jar (a hack for now)
* Update zm-mailbox jars
* zmcontrol restart

### DIFF
--- a/store/build.xml
+++ b/store/build.xml
@@ -271,7 +271,7 @@
     <ivy:install organisation="ant-tar-patched" module="ant-tar-patched" revision="1.0" settingsRef="dev.settings" from="chain-resolver" to="build-tmp" overwrite="true" transitive="true" type="jar"/>
 
     <!-- REDIS -->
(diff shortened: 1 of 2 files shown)


tmclane:
Is there a ticket that needs to be opened to get the installer updated?

fatkudu:
@tmclane I assumed that too at first but I think the answer is "no".  The file ends up in services.war (built in zm-mailbox/store via "ant war") and is installed as a result of that.  Just rebuilt zm-docker based on my branch and it is there:
/opt/zimbra/jetty_base/webapps/service/WEB-INF/lib/redisson-3.6.5.jar

tmclane:
Nice!

PR #597, opened 2018-04-20
ZCS-3055 Different date shown for Recurring Appointment Instance

fatkudu:
ZWC expects each exception in a message to have a unique item ID
    associated with it.  If this doesn't happen, the details of the first
    exception are used.

    Pseudo exceptions created when a reply is received specifying a
    particular PARTSTAT to single instance of the main series
    were not obeying this rule.  Instead 0 was used for both the item ID
    and the component number.  I had hoped that using the item ID of the
    calendar entry and a different component number would fix the issue,
    as intuitively, this is what I expected those values to mean but
    that doesn't match reality or what happens when exceptions are created
    in the calendar as a result of ZWC actions (or CalDAV actions)

    * public `Mailbox.ItemIdGetter` interface and private
      `Mailbox.NextItemIdGetter` implementation added as a preferred
      workaround to making `getNextItemId` protected
    * Eliminated a few wrapper `processNewInvite` and `processNewInviteReply`
      methods

### DIFF
--- a/store/src/java/com/zimbra/qa/unittest/TestJaxb.java
+++ b/store/src/java/com/zimbra/qa/unittest/TestJaxb.java
@@ -20,9 +20,15 @@ package com.zimbra.qa.unittest;
 import java.io.IOException;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Calendar;
 import java.util.Date;
+import java.util.GregorianCalendar;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.TimeZone;
 
 import javax.xml.bind.Marshaller;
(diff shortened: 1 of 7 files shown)


store/src/java/com/zimbra/qa/unittest/TestJaxb.java
```diff
@@ -135,6 +144,127 @@ private static String getCN(ZMailbox mbox) throws ServiceException {
...
+        inviteComp.setIsAllDay(false);
+        inviteComp.setName(subject);
+        inviteComp.setLocation(location);
+        inviteComp.setOrganizer(CalOrganizer.createForAddress(organizerZMbox.getName()));
+        return inviteComp;
+    }
+
+    public static MimePartInfo makeTextAndHtmlAlternatives(String text) {
+        MimePartInfo mimePart = MimePartInfo.createForContentType("multipart/alternative");
+        mimePart = MimePartInfo.createForContentType("multipart/alternative");
```
gordyt: Should these two lines be combined?

fatkudu: Yes :)  Thanks.

fatkudu: Fixed now.

PR #599, opened 2018-04-23
ZCS-4842: Redis MailItem caching

ilyusha:
This PR introduces caching abstractions and Redis implementations for three types of mailbox state previously kept in memory.

#### ItemCache
`ItemCache` was originally a map-like in-memory cache for MailItem instances. This is now an abstract class instantiated through a Factory. There are two implementations provided:
* `LocalItemCache` provides essentially the same in-memory functionality as before (set for unit test runs)
* `RedisItemCache` serializes the mailtems to their underlying metadata snapshots and stores them in Redis.
Since both implementations operate on `Map` interfaces internally, they both subclass an abstract `MapItemCache` to avoid code duplication.

#### FolderTagCache
The `ItemCache` abstraction described above also provides folder/tag caching functionality previously performed (optionally) by memcached. This cache is different from the ItemCache in that it caches snapshots of the _entire_ folder/tag state for a mailbox. This gets loaded at the beginning at every transaction to build the folder hierarchy; the cache allows us to avoid querying the DB for this data every time, which is more expensive. 

#### MailboxState
The third caching abstraction is for _mailbox state_ - that is, fields stored on the `MailboxData` object, and corresponding to the mailbox row in the DB. This includes things like mailbox size, unread message count, etc.
Caching this is facilitated by a wrapper `MailboxState` class, which is initialized with a `MailboxData` instance. This class exposes getters/setters for fields previously accessed directly on `MailboxData`; for example, Mailbox calls to `mData.size` are replaced by `state.getSize()`. This ensures that the returned value is accurate.
When the state object is initialized, the fields on the provided `MailboxData` instance are compared the corresponding values in the cache backend. If a value does not exist in the cache, then it is set to its `MailboxData` value (this happens the first time the mailbox is initialized cluster-wide); otherwise, the local value is replaced by the value in the cache, as the cached value is more recent.
Note that the `MailboxState` class supplants the `IdProvider` mechanism introduced recently to ensure monotonically increasing IDs.
`LocalMailboxState` is used for unit tests to avoid introducing a Redis dependency.

#### Explicit caching
An extra call to cache the MailItem had to be added to the `Mailbox::addMessageInternal` code path. This is because the first time the message is cached, it does not yet have an index ID. With an in-memory cache, this was not an issue, since modifying the item later updated the cache entry directly (since it's the same object), but this is not the case with an external cache.


store/src/java/com/zimbra/cs/mailbox/cache/MapItemCache.java
```diff
@@ -0,0 +1,83 @@
...
+
+    protected abstract MailItem fromCacheValue(T value);
+
+    @Override
+    public void put(MailItem item) {
+        int id = item.getId();
+        mapById.put(id, toCacheValue(item));
+        String uuid = item.getUuid();
+        if (uuid != null) {
+            uuid2id.put(uuid, id);
```
tmclane: Should an exception be raised If the item has no uuid and can't be stored?

ilyusha: I don't think so - this is actually exactly now the original `ItemCache` worked, minus the `toCacheValue` call.

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -924,7 +910,7 @@ public String getAccountId() {
     }
 
     public MailboxData getData() {
-        return mData;
+        return mData; //TODO: should this return the latest state synced with Redis?
```
gordyt: @ilyusha Looks like the only code that I saw that is using this is MailItem and the only things it is looking up from that is the mailbox id and account id.  I think is OK as is.

ilyusha: thanks for looking into that!

PR #601, opened 2018-04-24
Feature/zcs 4841

tmclane:
Update FileUploadServlet to create deterministic filenames such that any mailbox replica can operate on a given upload id.

* remove the temp directory cleanup on startup since replicas may come up and down at any time during the operation of the swarm.


tmclane:
I'll make a ticket.  Note that the only time this will happen is if in the window of time that an upload is in the temporary location the replica managing it dies.  I don't foresee mailbox replicas dying all that often .

PR #602, opened 2018-04-24
ZCS-3055,ZCS-4932 Declining single instance of all-day series handled strangely

fatkudu:
Hold off reviewing.  Not working in Harare time.
ZWC generates faulty responses to single instances of all-day series.  Specifying a timezone and DATE-TIME (with zeroes in the time part) in the RECURRENCE-ID.
The previous fix to ZCS-3055 handled this in a similar but slightly different way to previous code which may have introduced problems.
With this change, processing responses in order to update the calendar should handle more of
these faulty responses in a way which reflects what was intended and for those it doesn't fix, behaviour should be the same as prior to the original ZCS-3055 fix

### DIFF
--- a/common/src/java/com/zimbra/common/calendar/ParsedDateTime.java
+++ b/common/src/java/com/zimbra/common/calendar/ParsedDateTime.java
@@ -282,7 +282,7 @@ public final class ParsedDateTime {
     }
 
     @Override
-    public Object clone() {
+    public ParsedDateTime clone() {
         GregorianCalendar cal = (GregorianCalendar) mCal.clone();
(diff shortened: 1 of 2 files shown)


fatkudu:
Was hoping would be able to create an exception where the reply intended.  Hit 2 problems:

*  This code doesn't quite work as it still uses the recurrence-id from the reply and it should
    use one that works for the series.
*  Even if we could get that working, would probably end up creating another spurious exception
    next time round.

Abandoning this as too complicated.  Better to get ZWC and other clients to actually do the right thing.

PR #603, opened 2018-04-24
ZCS-4362, ZCS-4363, ZCS-4364: Distributed Sessions via Pub/Sub

ilyusha:
This PR enables soap session functionality in an multi-node environment, restoring mailbox notifications via response headers. This is done through several enhancements:
1. A new pub/sub mechanism that allows sessions to receive notifications from anywhere in the cluster
2. Migration of several instances of session state to Redis
3. An update the session modification sequence logic that (hopefully) ensures that the client does not receive duplicate notifications from multiple nodes

#### The problem
To understand this PR, it's first important to understand how soap sessions handled mailbox notifications prior to HA and how the new architecture breaks it. 
The two relevant components of a SoapSession are:
* `SoapSession.changes`, which is a `QueuedNotifications` object that contains an increasing sequence number
* `SoapSession.sentChanges`, which is a list of past `QueuedNotifications` whose members are evicted when the client sends confirmation that the changes have been successfully applied

Over time, the session is notified by the mailbox of `PendingModifications`, which it accumulates in the current `QueuedNotifications` object. When it comes time to package these notifications to send back to the client, this `QueuedNotifications` object is appended to the `sentChanges` list and the _entire_ list is encoded in the response header (though usually this list only has one element). A new `QueuedNotifications` instance is then created, with an incremented modification sequence, ready to accept further notifications.

With stateless distributed mailbox workers, this mechanism break down in the following ways:
* The modification sequence is no longer monotonically increasing from the client's perspective, since mailbox worker A is not aware of the sequence number on worker B. This is bad since the client will not apply changes with a sequence number lower than its known watermark.
* A session registered on one mailbox server will not be notified of changes made by a request processed on a different server, leading to missing notifications.
* If a client sends a request to a newly-started mailbox worker, the worker will not have the referenced session in its `SessionCache`, resulting in a new session being created instead.
* Since the `sentChanges` array is stored in-memory, the client is not guaranteed to receive previous notifications in the event that something went wrong, since a subsequent request is likely to go to a different server unaware of the change history.
* IDs for sessions generated on different servers may conflict, since session IDs are generated in an increasing sequence.

#### SessionDataProvider
The `SessionDataProvider` abstraction is a key component in resolving some of the issues described above, providing a way to share certain session information across the cluster. It currently provides access the following:
* The latest session ID by session type. This ensures that there are no session ID conflicts.
* The latest soap session modification sequence, by session ID
* A check for whether a soap session with a given ID exists on the cluster
* The list of sent notifications by session, replacing the  in-memory `sentChanges` list.

#### NotificationPubSub
A major change in this PR is that sessions are no longer registered directly on the `Mailbox` object. Instead, a `NotificationPubSub` abstract class is used to notify sessions of these changes. This is done with the `Publisher` and `Subscriber` helper classes: the mailbox pushes changes to the `Publisher`, which are received by sessions listening on the associated `Subscriber.` For simplicity, much of the code in `NotificationPubSub` base class was lifted directly from the corresponding `Mailbox` methods.

The primary implementation is `RedisPubSub`, which publishes `PendingLocalModifications` to a Redis channel named after the mailbox's account ID. The `RedisSubscriber` listens on this channel, reconstructs the modifications object, and notifies the sessions registered on its server. As an optimization, and to avoid introducing a race condition, sessions that are registered on the same node as the publisher are notified _directly_ using the `NotificationPubSub::notifyLocal` method.

There are two cases in which sessions registered on a subscriber are _not_ notified of changes received through the Redis channel:
* The changes originated on the same mailbox node as the listener. Since these sessions were already notified using `notifyLocal`, processing the change would result in duplicate notifications. This is done by checking the object hash of the `Mailbox` that originated the change to the hash of the Session's mailbox.
* The changes originated on a different mailbox node, but by an instance of the same session. Processing these would also result in duplicate notifications.

#### SoapSession Changes
Putting this all together, soap session logic has been changed in the following ways:
* If a client request is handled by a server on which the referenced soap session has not been cached, `DocumentHandler` will first check to see if the session is registered _globally_ - that is, if a modification sequence exists for that session ID in Redis. If so, a `SoapSession` will be instantiated with the passed-in ID, and cached in the local `SessionCache` as usual. If not, an entirely new session will be created (as was the original behavior).
* The notification sequence for each `QueuedNotifications` object is kept consistent for a given session across the cluster; this ensures that the client receives notifications in a monotonically increasing order regardless of which server the response comes from.
* The list of sent notifications previously kept in-memory is now stored in Redis. This means that any server can trim this list upon acknowledgement from the client or re-send notifications if necessary.
* Since an instance of a session will be notified of a change on _each server_ where it is cached, but only one of these servers will actually send that change back to the client, we need a way to invalidate changes on the _other_ servers after this happens. This is done by comparing the modification sequence on the `QueuedNotifications` object to the _latest_ sequence number in Redis; if at any point the sequence on the server falls behind the Redis value, it means that another server has already sent these changes to the client and the local copy can be discarded. This check happens in two places: prior to updating `QueuedNotifications` with a new change, and prior to encoding the existing changes into the response headers.

#### Open Issues
A limitation of the Redis pub/sub mechanism is we no longer have access to session-specific metadata, or the count of how many listeners there are, by type, across all workers. This can be implemented as a separate registry that gets updated with metadata upon session registration. Methods such as `Mailbox::getListeners` and `Mailbox::getNumListeners` are now on the `Publisher` class, but currently only return the locally-registered sessions. That said, this limitation should not affect `SoapSession` operation.

Another limitation has to do with mailbox maintenance. Previously, putting a mailbox into maintenance mode would unregister all listeners; this is no longer the case, since unregistering listeners on other nodes would require passing a signal along the notification channel. While this is not difficult, it is somewhat outside the scope of this ticket. This can be changed once mailbox maintenance is more clearly defined.

Lastly, it is entirely possible that there are still gaps or race conditions in this distributed session mechanism that I overlooked. Such bugs may result in the client either receiving duplicate notifications, or not receiving them at all. These issues (should they come up) will of course be addressed, but it should be mentioned that a refresh of the UI should bring the client back in sync with the server.

### DIFF
--- a/store/src/java/com/zimbra/cs/session/RedisSessionDataProvider.java
+++ b/store/src/java/com/zimbra/cs/session/RedisSessionDataProvider.java
@@ -1,0 +1,172 @@
+package com.zimbra.cs.session;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.redisson.api.RAtomicLong;
+import org.redisson.api.RList;
+import org.redisson.api.RMap;
+import org.redisson.api.RedissonClient;
+
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.util.ZimbraLog;
+import com.zimbra.cs.mailbox.Mailbox;
+import com.zimbra.cs.mailbox.RedissonClientHolder;
+import com.zimbra.cs.session.PendingLocalModifications.PendingModificationSnapshot;
+import com.zimbra.cs.session.SoapSession.QueuedNotifications;
+
+public class RedisSessionDataProvider extends SessionDataProvider {
+
+    private static final String SOAP_SESSION_SEQ_MAP_NAME = "SOAP_SESSION_SEQUENCES";
+    private static final String SESSION_ID_SEQ_NAME = "SESSION_ID_SEQUENCE";
+    private RMap<String, Integer> soapSequenceMap;
+    private RAtomicLong idCounter;
+    private Map<String, RedisNotificationQueue> notificationMap;
+
+    public RedisSessionDataProvider() {
+        RedissonClient client = RedissonClientHolder.getInstance().getRedissonClient();
+        soapSequenceMap = client.getMap(SOAP_SESSION_SEQ_MAP_NAME);
+        idCounter = client.getAtomicLong(SESSION_ID_SEQ_NAME);
+        notificationMap = new HashMap<>();
+    }
+
+    @Override
+    protected long getNextIdSequence() {
+        return idCounter.addAndGet(1);
+    }
+
+    @Override
+    public int getNextSoapSessionSequence(String soapSessionId) {
+        return soapSequenceMap.addAndGet(soapSessionId, 1);
+    }
+
+
+    @Override
+    public int getCurrentSoapSessionSequence(String soapSessionId) {
+        Integer val = soapSequenceMap.putIfAbsent(soapSessionId, 1);
+        return val == null ? 1 : val;
+    }
+
+    @Override
+    public void cleanup(String sessionId) {
+        soapSequenceMap.remove(sessionId);
+        if (notificationMap.containsKey(sessionId)) {
+            notificationMap.get(sessionId).clear();
+        }
+    }
+
+    @Override
+    public boolean soapSessionExists(String sessionId) {
+        return soapSequenceMap.containsKey(sessionId);
+    }
+
+    @Override
+    public NotificationQueue getSoapNotifications(SoapSession session) {
+        RedisNotificationQueue queue = notificationMap.get(session.getSessionId());
+        if (queue == null) {
+            queue = new RedisNotificationQueue(session);
+            notificationMap.put(session.getSessionId(), queue);
+        }
+        return queue;
+    }
+
+    public static class RedisNotificationQueue extends SessionDataProvider.NotificationQueue {
+
+        private RList<SerializableNotification> list;
+        private Mailbox mbox;
+        private String authAcctId;
+
+        public RedisNotificationQueue(SoapSession session) {
+            RedissonClient client = RedissonClientHolder.getInstance().getRedissonClient();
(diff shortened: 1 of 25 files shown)


store/src/java/com/zimbra/cs/session/RedisSessionDataProvider.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        @Override
+        public void add(QueuedNotifications notifications) {
+            try {
+                PendingModificationSnapshot snapshot = notifications.mMailboxChanges.toSnapshot();
+                int seq = notifications.getSequence();
+                list.add(new SerializableNotification(authAcctId, snapshot, seq));
+            } catch (ServiceException e) {
+                ZimbraLog.session.error("unable to push QueuedNotifications to redis");
```
fatkudu: worth including the exception in the log?

store/src/java/com/zimbra/cs/session/RedisSessionDataProvider.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        @Override
+        public int size() {
+            return list.size();
+        }
+
+        @Override
+        public void purge(int sequence) {
+            //should this be made fully atomic with a Lua script?
```
fatkudu: Perhaps mark with TODO?

store/src/java/com/zimbra/cs/session/SoapSession.java
```diff
@@ -471,7 +472,7 @@ void clearMailboxChanges() {
 
     // read/write access to all these members requires synchronizing on "mSentChanges"
     protected int forceRefresh;
-    protected LinkedList<QueuedNotifications> sentChanges = new LinkedList<QueuedNotifications>();
+    //protected LinkedList<QueuedNotifications> sentChanges = new LinkedList<QueuedNotifications>();
```
fatkudu: Remove if not needed any more?

PR #612, opened 2018-04-27
ZCS-4147: Explicitly indexing 'none' and 'any' attachment types

ilyusha:
Attachment queries like `has:attachement` have been broken with Solr due to how we now index attachment types. The fix on the mailbox side is to explicitly add "any" or "none" tokens to each index document, instead of relying on the tokenizer to do this.

See https://github.com/Zimbra/zm-solr/pull/9 for the zm-solr side of this fix, and for a more in-depth explanation of the issue.

### DIFF
--- a/store/src/java/com/zimbra/cs/mime/ParsedMessage.java
+++ b/store/src/java/com/zimbra/cs/mime/ParsedMessage.java
@@ -1020,8 +1020,14 @@ public final class ParsedMessage {
 
 
         // Get the list of attachment content types from this message and any TNEF attachments
-        for (String attachment: Mime.getAttachmentTypeList(messageParts)) {
-        	doc.addAttachments(attachment);
+        Set<String> attachmentTypes = Mime.getAttachmentTypeList(messageParts);
+        if (attachmentTypes.isEmpty()) {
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/mime/ParsedMessage.java
```diff
@@ -1020,8 +1020,14 @@ private IndexDocument getMainBodyLuceneDocument(StringBuilder fullContent)
...
         // Get the list of attachment content types from this message and any TNEF attachments
-        for (String attachment: Mime.getAttachmentTypeList(messageParts)) {
-        	doc.addAttachments(attachment);
+        Set<String> attachmentTypes = Mime.getAttachmentTypeList(messageParts);
+        if (attachmentTypes.isEmpty()) {
+            doc.addAttachments(LuceneFields.L_ATTACHMENT_NONE);
+        } else {
+            for (String attachment: attachmentTypes) {
+                doc.addAttachments(attachment);
+                doc.addAttachments(LuceneFields.L_ATTACHMENT_ANY);
```
plobbes: I doubt this matters much, but shouldn't this line be outside the for loop?

PR #614, opened 2018-04-30
Bugfix/zcs 4847 back port 3055 to 8.8.8.p3

fatkudu:
Note that created hotfix/8.8.8.p3 branch where 8.8.8.p1 was.  I assume p3 is the right terminology based on message from Sandesh today (i.e. NOT p2)

Hmm.  Actually 8.8.8 patch 2 has apparently just been release but I'm confused as I can't see any evidence for branches or tags referencing it in zm-mailbox or zm-build???

### DIFF
--- a/soap/src/java/com/zimbra/soap/mail/message/CreateAppointmentExceptionRequest.java
+++ b/soap/src/java/com/zimbra/soap/mail/message/CreateAppointmentExceptionRequest.java
@@ -17,14 +17,15 @@
 
 package com.zimbra.soap.mail.message;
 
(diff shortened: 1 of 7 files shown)


gordyt:
@fatkudu I don't see anything related to `8.8.8.p2` either.  @Prashantsurana is that correct?

gordyt:
@fatkudu No need to keep ticket open as you have already provided the necessary commits for p3

fatkudu:
Note.  Created 8.8.8.p3 tag where this has been merged to.  Presumably that will need to move when other fixes are done.

PR #615, opened 2018-05-01
ZCS-4963: Redis Subscribe Timeout Fix

ilyusha:
The original Redis pub/sub system for notifications created a dedicated subscription channel for each mailbox. It actually turns out that the number of subscriptions is limited by two factors, defined in the Redisson client config:
- `subscriptionConnectionPoolSize`, which defaults to 50
- `subscriptionsPerConnection`, which defaults to 5
It's no surprise then that Redission refused new subscription connections when the number of created accounts surpassed 250, leading to timeouts.
While it is possible to raise these values, it seems unwise to assume that we can safely increase the number of connections without limit. 
A workaround introduced in this PR is to have _shared_ subscription channels. We map mailboxes to one of _n_ channels in a deterministic way (_n_ being configurable in LC), and attach the account ID to the message being published. On the receiving end, the listener can route the incoming notification messages appropriately.

This is done using the new `NotificationChannel` static inner class. `RedisPubSub` has a map of these keyed by their channel names, now called `NOTIFICATION-CHANNEL-{N}`, and each instance maintains a map of account IDs mapped to their `RedisSubscriber` objects. When a `NotificationChannel` receives a message, it looks up the appropriate subscriber in this map and calls `notifyListeners` on it.

This PR also contains a few other small semi-related changes:
 - The `Mailbox.trimItemCache()` method no longer calls hasListeners() to determine the cache size, as this is no longer a meaningful check
- The `RedisItemCache.values()` method reads all hash values in a single operation instead of relying on the built-in `RMap` values() iterator, which actually fetches the values one-by-one
- The logging call in `SoapSession` notifying of a stale notification is now logged at a `debug` level, as it happens very often

### DIFF
--- a/store/src/java/com/zimbra/cs/mailbox/RedisPubSub.java
+++ b/store/src/java/com/zimbra/cs/mailbox/RedisPubSub.java
@@ -1,32 +1,42 @@
 package com.zimbra.cs.mailbox;
 
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
 
 import org.redisson.api.RTopic;
 import org.redisson.api.RedissonClient;
 import org.redisson.api.listener.MessageListener;
 
+import com.zimbra.common.localconfig.LC;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.cs.session.PendingLocalModifications;
+import com.zimbra.cs.session.Session;
 import com.zimbra.cs.session.PendingLocalModifications.PendingModificationSnapshot;
 import com.zimbra.cs.session.Session.SourceSessionInfo;
 import com.zimbra.cs.session.Session.Type;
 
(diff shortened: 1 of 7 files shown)


store/src/java/com/zimbra/cs/mailbox/RedisPubSub.java
```diff
@@ -62,38 +81,104 @@ public int getNumListeners(Type type) {
...
+            this.listenerId = channel.addListener(this);
+            ZimbraLog.mailbox.info("beginning listening on Redis notifification channel %s", name);
+            active = true;
+        }
+
+        public void addSubscriber(RedisSubscriber subscriber) {
+            beginListening();
+            String acctId = subscriber.getMailbox().getAccountId();
+            subscriberMap.put(acctId, subscriber);
+            ZimbraLog.mailbox.info("added account %s to Redis notifification channel %s", acctId, name);
```
fatkudu: spelling - notifification

PR #616, opened 2018-05-03
ZCS-4928:fix ambiguous Log.[warn|error]Quietly

fatkudu:
Fix compilation issue:
    [javac]                 ZimbraLog.ews.errorQuietly("Error generating the salt", e);
    [javac]                              ^
    [javac]   both method errorQuietly(Object,Throwable) in Log and method errorQuietly(String,Object) in Log match


fatkudu:
Merged because compilation errors are bad...

PR #617, opened 2018-05-03
NICPS-187 Add new ldap attributes needed for custom templates

JSreeGarapati:


store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9675,4 +9675,13 @@ TODO: delete them permanently from here
   <defaultCOSValue>TRUE</defaultCOSValue>
   <desc>This attribute is used to enable/disable zimbra headers on email printouts</desc>
 </attr>
+<attr id="5008" name="zimbraUserDesignation" type="string" cardinality="single" optionalIn="account" flags="accountInfo" since="8.8.6">
```
chinmay15: put 8.8.8 as "since" for all 3 attributes.

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9675,4 +9675,16 @@ TODO: delete them permanently from here
   <defaultCOSValue>TRUE</defaultCOSValue>
   <desc>This attribute is used to enable/disable zimbra headers on email printouts</desc>
 </attr>
+
+<attr id="5008" name="zimbraUserDesignation" type="string" cardinality="single" optionalIn="account" flags="accountInfo" since="8.8.8">
```
plobbes: The attribute names and descriptions are likely too general for inclusion in the product (assuming we're going to make this part of the standard ZCS product and not a one-off fork, right?). The attribute naming should help to indicate which attributes are related. The attribute descriptions should also indicate how they are used and the relationships with other attributes.

In this case, perhaps a name like 'zimbraCustomTemplateUserDesignation' is better? Or perhaps others have ideas on a common 'prefix' for these three attributes?

Likewise the description should ideally give us a clue of intended use(s). Perhaps a description similar to these might be better?
- 'Designation to be used with custom templates.' or
- 'Designation for the user. May be used as a template variable in Custom Templates.'

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9675,4 +9675,16 @@ TODO: delete them permanently from here
...
+	<desc>This attribute contains user designation</desc>
+</attr>
+
+<attr id="5009" name="zimbraUserSignature" type="binary" cardinality="single" optionalIn="account" flags="accountInfo" since="8.8.8">
+	<desc>This attribute contains user signature</desc>
+</attr>
+
+<attr id="5010" name="zimbraTemplateID" type="string" cardinality="single" optionalIn="account" flags="accountInfo" since="8.8.8">
```
plobbes: - ditto on the attr name/desc

Also, the assumptions on the internal docs state, "End Users can have multiple templates and will select which templates to use via a drop down during message composition."  The use of _cardinality="single"_ however would indicate that we're only allowing a single template per account. Which is correct?

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9675,4 +9675,16 @@ TODO: delete them permanently from here
   <defaultCOSValue>TRUE</defaultCOSValue>
   <desc>This attribute is used to enable/disable zimbra headers on email printouts</desc>
 </attr>
+
+<attr id="5008" name="zimbraUserDesignation" type="string" cardinality="single" optionalIn="account" flags="accountInfo" since="8.8.8">
+	<desc>This attribute contains user designation</desc>
+</attr>
+
+<attr id="5009" name="zimbraUserSignature" type="binary" cardinality="single" optionalIn="account" flags="accountInfo" since="8.8.8">
+	<desc>This attribute contains user signature</desc>
```
plobbes: - ditto on the attr name/desc

PR #622, opened 2018-05-10
ZCS-4833:Support signing of the profile configuration

shrikant-prasad:
Sign the mobileconfig using certitifcate if the values are set in ldap else return the unsigned file.

As per below discussion, have removed the certificate ldap attributes added in first commit and re-used zimbraSSLCertificate/zimbraSSLPrivateKey pair.


fatkudu:
I agree with @Prashantsurana. If we can avoid admin pain and keep things simple it is good to do.

store/src/java/com/zimbra/cs/service/formatter/MobileConfigFormatter.java
```diff
@@ -225,39 +241,48 @@ public void formatCallback(UserServletContext context) throws ServiceException {
...
-                        FileBufferedWriter.MAX_BUFFER_SIZE);
+            try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
                 DOMSource domSource = new DOMSource(document);
-                StringWriter writer = new StringWriter();
-                StreamResult responseStream = new StreamResult(writer);
+                Result responseStream = new StreamResult(baos);
                 transformer.transform(domSource, responseStream);
-                ZimbraLog.misc.debug(writer.toString());
-                fileBufferedWriter.write(writer.toString());
+                byte[] signedConfig = signConfig(server, baos.toByteArray());
```
Prashantsurana: Can we check if private key is there on domain if not then use the server one?

store/src/java/com/zimbra/cs/service/formatter/MobileConfigFormatter.java
```diff
@@ -250,20 +259,26 @@ public void formatCallback(UserServletContext context) throws ServiceException {
...
-        char[] expPass = StringUtil.isNullOrEmpty(certPass) ? "".toCharArray() : certPass.toCharArray();
-        if (digiCert != null) {
-            try {
-                signedConfig = DataSigner.signData(config, digiCert, expPass);
-            } catch (UnrecoverableKeyException | KeyStoreException | NoSuchAlgorithmException | CertificateException
-                    | OperatorCreationException | IOException | CMSException e) {
-                ZimbraLog.misc.debug("exception occurred during signing config", e);
+        if (server != null) {
+            String certStr = server.getSSLCertificate();
+            String pvtKeyStr = server.getSSLPrivateKey();
```
Prashantsurana: Can we check if SSL private key is present of domain then fallback on server?

store/src/java/com/zimbra/cs/service/formatter/MobileConfigFormatter.java
shrikant-prasad: Done.

PR #627, opened 2018-05-18
Bugfix/zcs 5035

sneha-patil-synacor:
Added support for following scenarios:
1. If email is changed in sendCode, user should be able to send verification email. If it is not changed, the sendCode will not send the code again.
2. recovery email address should not same as primary address in sendCode.

Updated soap doc.
Updated junit tests.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/SetRecoveryEmail.java
+++ b/store/src/java/com/zimbra/cs/service/mail/SetRecoveryEmail.java
@@ -1,0 +1,240 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2018 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ * ***** END LICENSE BLOCK *****
+ */
+
+package com.zimbra.cs.service.mail;
+
+import java.text.DateFormat;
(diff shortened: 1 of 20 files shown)


store/src/java/com/zimbra/cs/service/mail/SetRecoveryEmail.java
```diff
@@ -231,8 +248,9 @@ protected void reset(Mailbox mbox, ZimbraSoapContext zsc) throws ServiceExceptio
     }
 
     private static void validateEmail(String email, Account account) throws ServiceException {
-        String[] addresses = account.getMailAddress();
-        if (Arrays.asList(addresses).contains(email)) {
+        String accountName = account.getName();
+        String aliases[] = account.getAliases();
+        if (Arrays.asList(aliases).contains(email) || accountName.equals(email)) {
             throw ServiceException
                 .FAILURE("Recovery address should not be same as primary email address.", null);
```
swapnilpingle89: How about having failure message as "Recovery address should not be same as primary/aliases email address."?
User might enter alias email and argue it is not same as primary email.

store/src/java/com/zimbra/cs/service/mail/SetRecoveryEmail.java
```diff
@@ -231,10 +248,11 @@ protected void reset(Mailbox mbox, ZimbraSoapContext zsc) throws ServiceExceptio
     }
 
     private static void validateEmail(String email, Account account) throws ServiceException {
-        String[] addresses = account.getMailAddress();
-        if (Arrays.asList(addresses).contains(email)) {
+        String accountName = account.getName();
+        String aliases[] = account.getAliases();
+        if (Arrays.asList(aliases).contains(email) || accountName.equals(email)) {
             throw ServiceException
```
rupalid: Do you want to log at the debug level what the 2 email addresses were.

piyush4360:
For point 2, recovery email address not be same as primary email address, 

Are we verifying the aliases also for that user?

sneha-patil-synacor:
@piyush4360 Yes, we are.

sneha-patil-synacor:
Added logging. Added new logger called zimbra.passwordreset.

store/src/java/com/zimbra/cs/service/mail/SetRecoveryEmail.java
sneha-patil-synacor: Done

PR #628, opened 2018-05-18
ZCS-5021:Implementing a JWE token to store data for recovery email setup

shrikant-prasad:
added utility methods to store the data related to recovery email setup(email, code, expiryTime and resend code) in a single encrypted string and decode the values back from that string.
fixed three issues:
1. When code is sent for first time, resendCount should be zero.
2. When existing code is re-sent again, it's expiry time should not be updated.
3. when resend op is called, it was behaving as send and giving exception that code has already been sent to this recovery email address.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/SetRecoveryEmail.java
+++ b/store/src/java/com/zimbra/cs/service/mail/SetRecoveryEmail.java
@@ -37,13 +37,14 @@ import com.zimbra.common.mime.MimeConstants;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.soap.Element;
 import com.zimbra.common.util.L10nUtil;
-import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.common.util.L10nUtil.MsgKey;
 import com.zimbra.common.util.StringUtil;
+import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.cs.account.Account;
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/service/mail/SetRecoveryEmail.java
```diff
@@ -238,20 +249,17 @@ protected void resendCode(Account account, Mailbox mbox, ZimbraSoapContext zsc,
...
                 resendCount = resendCount + 1;
                 HashMap<String, Object> prefs = new HashMap<String, Object>();
-                prefs.put(Provisioning.A_zimbraRecoveryEmailVerificationData,
-                    email + ":" + code + ":" + newExpiryTime + ":" + resendCount);
-                Provisioning.getInstance().modifyAttrs(mbox.getAccount(), prefs, true,
-                    zsc.getAuthToken());
-                sendRecoveryEmailVerificationCode(authAccount, account, email, code, newExpiryTime,
+                dataMap.put(RESEND_COUNT, String.valueOf(resendCount));
+                String verificationDataStr = JWEUtil.getJWE(dataMap);
+                prefs.put(Provisioning.A_zimbraRecoveryEmailVerificationData, verificationDataStr);
```
gordyt: Could replace `verificationDataStr` with `JWEUtil.getJWE(dataMap)` and delete previous line.

PR #636, opened 2018-05-28
ZCS-4798 

chinmay15:
Problem : Editing recovery email using zmprov leads to send email to wrong email.
Fix : verify preference email with ecrypted email in code and send email to the one set in preference.
Testing done : manually tested scenario


chinmay15:
As per discussion with @Prashantsurana, this fix is not required once we fix zcs-5122. Since closing without merging.

PR #639, opened 2018-05-31
ZCS-5085 Restoring log message

rupalid:
Change the logic that display the discard message

Verified that all unit test pass

Created a filter with discard action and verified the message is discarded and message is logged.

### DIFF
--- a/store/src/java/com/zimbra/cs/filter/ZimbraMailAdapter.java
+++ b/store/src/java/com/zimbra/cs/filter/ZimbraMailAdapter.java
@@ -281,9 +281,11 @@ public class ZimbraMailAdapter implements MailAdapter, EnvelopeAccessors {
 
             // If the Sieve script has no actions, JSieve generates an implicit keep.  If
             // the script contains a single discard action, JSieve returns an empty list.
-            if (getActions().size() == 0) {
+            if (this.discardActionPresent) {
                 ZimbraLog.filter.info("Discarding message with Message-ID %s from %s",
                     messageId, Mime.getSender(handler.getMimeMessage()));
+            }
+            if (getActions().size() == 0) {
                 handler.discard();
                 return;
             }


store/src/java/com/zimbra/cs/filter/ZimbraMailAdapter.java
```diff
@@ -281,9 +281,11 @@ public void executeAllActions() throws SieveException {
 
             // If the Sieve script has no actions, JSieve generates an implicit keep.  If
             // the script contains a single discard action, JSieve returns an empty list.
-            if (getActions().size() == 0) {
+            if (this.discardActionPresent) {
                 ZimbraLog.filter.info("Discarding message with Message-ID %s from %s",
                     messageId, Mime.getSender(handler.getMimeMessage()));
+            }
+            if (getActions().size() == 0) {
```
fatkudu: If `discardActionPresent` is false at this point but this if branch gets executed, `handler.discard()` will happen but the message will not be logged.  Is that the intention?

On the flip side, if `discardActionPresent` is false but the size is not zero, we will log that the message is being discarded but will NOT call `handler.discard()`.

It isn't entirely clear to me from the bug description whether `handler.discard()` as well as the logging should be executed.

rupalid: discardActionPresent is currently being  set true for ereject, reject and discard.
An action is only added for reject and ereject.

rupalid: actions is only empty if there is a discard action. For all other cases an implicit keep is added while evaluating. So we cannot have a case where discardActionPresent is false and action is empty. Also handler.discard() is an empty method.

fatkudu: `ExistingMessageHandler.discard()` is not empty.  Perhaps that doesn't matter given what else you said but just want to check.

rupalid: Yes ExistingMessageHandler.discard() is not empty but considering that we have a implicit  keep action if no action is specified in the sieve filter, we would not hit the if (getActions().size() == 0)  for other actions.

PR #647, opened 2018-06-06
ZCS-5122:ForgotPassword implementation modifications

shrikant-prasad:
Following changes  are made in this fix:

1. on change of zimbraPrefPasswordRecoveryAddress (through any action), all attributes related to password recovery should reset their values.
2. for sendCode operation of RecoveryAccount API, resend count should not be reset when new code is sent after expiry of old one. When max attempts has been reached, forgotPassword feature should be suspended.
3. in all forgot password api, check the state of forgotPassword feature. If it's suspended, check the time when feature was suspended and if it has passed the suspension time, make the feature enabled. If suspension time is not over, throw exception.


store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9640,12 +9640,12 @@ TODO: delete them permanently from here
   <globalConfigValue>1518163473</globalConfigValue>
 </attr>
 
-<attr id="3024" name="zimbraFeatureResetPasswordEnabled" type="boolean" cardinality="single" optionalIn="account,cos,domain" flags="accountCosDomainInherited,accountInfo" since="8.8.9">
-  <defaultCOSValue>FALSE</defaultCOSValue>
-  <desc>whether or not password reset feature is enabled</desc>
+<attr id="3024" name="zimbraFeatureResetPasswordStatus" type="enum" value="enabled,suspended,disabled" cardinality="single" optionalIn="account,cos,domain" flags="accountCosDomainInherited,accountInfo" since="8.8.9">
```
gordyt: Just a note... Some of these attribute IDs conflict with ones we are currently using in the [zm-mailbox develop branch in ZimbraOS](https://github.com/ZimbraOS/zm-mailbox/blob/develop/store/conf/attrs/zimbra-attrs.xml), so if we will have to pull this code into the _ZimbraOS_ repos will have to handle that.

shrikant-prasad: changed the ids to resolve conflict.

PR #648, opened 2018-06-08
ZCS-5122: add recovery attempts remain in response

shrikant-prasad:
1. return recovery attempts remaining before feature suspension in recoverAccount response.
2. clear the recovery code data when feature status is changed.ex: suspended to enabled.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/RecoverAccount.java
+++ b/store/src/java/com/zimbra/cs/service/mail/RecoverAccount.java
@@ -90,20 +90,22 @@ public final class RecoverAccount extends MailDocumentHandler {
 
         RecoverAccountResponse resp = new RecoverAccountResponse();
         switch (op) {
-            case GET_RECOVERY_EMAIL:
-                recoveryEmail = StringUtil.maskEmail(recoveryEmail);
-                ZimbraLog.account.debug("%s Recovery email: %s", LOG_OPERATION, recoveryEmail);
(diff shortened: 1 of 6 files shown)


store/src/java/com/zimbra/cs/service/mail/RecoverAccount.java
```diff
@@ -90,20 +90,22 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
+        case GET_RECOVERY_EMAIL:
+            recoveryEmail = StringUtil.maskEmail(recoveryEmail);
+            ZimbraLog.account.debug("%s Recovery email: %s", LOG_OPERATION, recoveryEmail);
+            resp.setRecoveryEmail(recoveryEmail);
+            break;
+        case SEND_RECOVERY_CODE:
+            String storedCodeString = user.getResetPasswordRecoveryCode();
+            ZonedDateTime currentDate = ZonedDateTime.now(ZoneId.systemDefault());
+            Map<String, String> recoveryCodeMap = new HashMap<String, String>();
+            int maxAttempts = user.getPasswordRecoveryMaxAttempts();
```
fatkudu: How would this get reset?

shrikant-prasad: Once max attempts is reached, feature is in suspended state.It can be reset by admin or when request is triggered after suspension time is over.

PR #649, opened 2018-06-08
ZCS-4802

chinmay15:
**Problem:** check if passwored is locked and password strength
**Fix:** checked zimbraPasswordLocked and password strength

### DIFF
--- a/store/src/java/com/zimbra/cs/service/account/ResetPassword.java
+++ b/store/src/java/com/zimbra/cs/service/account/ResetPassword.java
@@ -24,6 +24,7 @@ import com.zimbra.common.soap.AccountConstants;
 import com.zimbra.common.soap.Element;
 import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.cs.account.Account;
+import com.zimbra.cs.account.AccountServiceException;
 import com.zimbra.cs.account.AuthToken;
 import com.zimbra.cs.account.Provisioning;
 import com.zimbra.cs.service.AuthProvider;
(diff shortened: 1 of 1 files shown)


chinmay15:
merged in feature/PasswordReset branch

commit 6b9bd29967cae18a73264818752139af98b9618b
Date:   Fri Jun 8 18:32:57 2018 +0530

    ZCS-4802: check if passwored is locked and password strength

store/src/java/com/zimbra/cs/service/account/ResetPassword.java
```diff
@@ -47,8 +48,13 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
         AuthProvider.validateAuthToken(prov, at, false);
 
         Account acct = at.getAccount();
+        boolean locked = acct.getBooleanAttr(Provisioning.A_zimbraPasswordLocked, false);
+        if (locked) {
+            throw AccountServiceException.PASSWORD_LOCKED();
+        }
         ResetPasswordUtil.validateFeatureResetPasswordStatus(acct);
         String newPassword = req.getPassword();
+        prov.checkPasswordStrength(acct, newPassword);
```
shrikant-prasad: instead of explicitly calling this method, could have just used prov.setPassword(acct, newPassword, true); where password is being set.

chinmay15:
checkPasswordStrength() is public. Also the setPassword(), you mentioned is eventually calling same checkPasswordStrength(), so i don't think there is any harm in calling this method explicitly...

PR #659, opened 2018-06-15
ProxyConfGen new !{explode server(serviceName)}

ZeXtrasJay:
Added !{explode server(serviceName)} to explode zimbra-docs configurations.
see https://github.com/Zimbra/zm-nginx-conf/pull/9 for the related configuration

### DIFF
--- a/store/src/java/com/zimbra/cs/util/ProxyConfGen.java
+++ b/store/src/java/com/zimbra/cs/util/ProxyConfGen.java
@@ -1757,6 +1757,30 @@ class WebXmppBoshEnablerVar extends ProxyConfVar {
 }
 
 /**
+ * Simplified container object for a server
+ * @author Davide Baldo
+ */
+class ServerAttrItem {
+    public String zimbraId;
+    public String hostname;
+    public String[] services;
+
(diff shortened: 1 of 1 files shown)


ZeXtrasJay:
Seems like the failure is unrelated to the patch: "Test com.zimbra.common.mime.MimeHeaderTest failed"

fatkudu:
I suspect this is a companion commit to another commit, hence adding potential reviewers.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
```diff
@@ -1756,6 +1756,30 @@ public void update() throws ServiceException {
     }
 }
 
+/**
+ * Simplified container object for a server
+ * @author Davide Baldo
```
plobbes: Nit: remove @author tag - going forward, it would seem perhaps better if we stop using/relying on @author and just use git history. See also:
- https://stackoverflow.com/questions/17269843/javadoc-author-tag-good-practices
- https://www.vojtechruzicka.com/stop-using-javadoc-author-tag/

gordyt:
Rebased this PR on `develop` and merged.

PR #661, opened 2018-06-18
ZCS-5202,ZCS-5112 JSON XmlElementWrapper support broken

fatkudu:
This is a workaround for a presumed bug in Jackson (see https://github.com/FasterXML/jackson-modules-base/issues/47) where it doesn't work with our way of mapping `@XmlElementWrapper` / `@XmlElement` annotated properties to JSON - we use the name in `@XmlElementWrapper` rather than the usual Jackson preference for the name in `@XmlElement`.
In theory, there is a feature which should switch things to work the way we want them to work but that doesn't work, at least in specific scenarios we use where we have the same name at the `@XmlElement` level for objects of the same class for 2 different fields.
After that change, can now select "edit properties" of the inbox in ZWC and also configure message retention.


fatkudu:
I would hold off merging this until we flip the switch everywhere...

PR #672, opened 2018-06-29
Feature/nicps 282

nileshpawar1:
created new LDAP variable which will store domain-specific template ids


nileshpawar1:
by mistake created for devlope, it should be for snickers

PR #680, opened 2018-07-05
ZCS-4845_3 Changes in store for httpclient jar upgrade

rupalid:
Verified that unit test pass. Verified that there are no compilation errors.
Completed sanity test on FOSS Build with new http client library changes.


rupalid:
@gordyt  Agreed. Once we are done with changes to all repos and QA is complete, the changes should be migrated to ZimbraOS.

PR #682, opened 2018-07-06
Bugfix/zbug 481

rupalid:
Added white list of tags and attributes in XHtmlhandler for sanitizing mime.
Added unit test

Verified with the .svg files provided in the bug that an email is not sent and we do not get a popup.
Verified that email with pdf and html attachment and content are correctly rendered.

### DIFF
--- a/common/src/java/com/zimbra/common/localconfig/DebugConfig.java
+++ b/common/src/java/com/zimbra/common/localconfig/DebugConfig.java
@@ -234,7 +234,7 @@ public final class DebugConfig {
         "a,abbr,acronym,blockquote,div,font,h1,h2,h3,h4,h5,h6,img,li,ol,p,span,table,td,th,tr,ul");
 
     public static final String xhtmlWhitelistedAttributes = value("defang_xhtml_whitelisted_attributes",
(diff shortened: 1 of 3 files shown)


common/src/java/com/zimbra/common/localconfig/DebugConfig.java
```diff
@@ -230,6 +230,12 @@ private DebugConfig() {
             "defang_style_unwanted_import",
             "@import(\\s)*((\'|\")?(\\s)*(http://|https://)?([^\\s;]*)(\\s)*(\'|\")?(\\s)*;?)");
 
+    public static final String xhtmlWhitelistedTags = value("defang_xhtml_whitelisted_tags",
+        "a,abbr,acronym,blockquote,div,font,h1,h2,h3,h4,h5,h6,img,li,ol,p,span,table,td,th,tr,ul");
+
+    public static final String xhtmlWhitelistedAttributes = value("defang_xhtml_whitelisted_attributes",
+        "abbr,align,alt,border,cellpadding,cellspacing,cite,class,color,colspan,height,href,id,,name,rel,rev,rowspan,size,src,style,title,target,valign,width");
```
desouzas: Does the extra comma between `id` and `name` affect anything?

jdavid-synacor: No, it shouldn't but I removed it.

PR #687, opened 2018-07-13
ZCS-5577 Receiving Invalid request exception for CreateCosRequest.

chinmay15:
**Problem:** CreateCosRequest fails with invalid request exception.

**Fix:**
- Invalid request exception was thrown because the entry object is null while creating cos.
- This was not found on local as the new attribute was introduced as part ZCS-90 and It is not initialized to default value on dev environment.
- Added null check for entry object in the callback class from where the exception was being thrown.

**Testing done:**
- Tested on zdev-vm004.eng.zimbra.com by replacing store jar.
- Tested by creating cos from command line using zmprov command.
- Tested by creating cos from admin console.

**Testing to be done by QA:**
- Cos creation should not fail because of this attribute.
- Account creation should not fail because of this attribute.


chinmay15:
merged in the develop branch

commit 6aeb46a2d9ed164ed4ea3641fc49ea162745934a
Date:   Fri Jul 13 14:49:55 2018 +0530

    ZCS-5577 Receiving Invalid request exception for CreateCosRequest.

PR #690, opened 2018-07-18
ZCS-5361:Primary Email address Changes in postfix and Mailbox code

shrikant-prasad:
1. added new cache object for the zimbraOldMailAddress in AccountCache.
2. modified the LDAP filter string to search the value of zimbraOldMailAddress as well.
3. added debug config parameter for sleep between account rename and alias creation to enable QA to test mail delivery to old address during the change primary email process.

### DIFF
--- a/store/src/java/com/zimbra/cs/account/cache/AccountCache.java
+++ b/store/src/java/com/zimbra/cs/account/cache/AccountCache.java
@@ -1,7 +1,7 @@
 /*
  * ***** BEGIN LICENSE BLOCK *****
  * Zimbra Collaboration Suite Server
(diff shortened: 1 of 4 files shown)


store/src/java/com/zimbra/cs/account/cache/AccountCache.java
```diff
@@ -1,7 +1,7 @@
 /*
  * ***** BEGIN LICENSE BLOCK *****
  * Zimbra Collaboration Suite Server
- * Copyright (C) 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2016 Synacor, Inc.
```
chinmay15: no need to remove 2016

PR #703, opened 2018-08-01
ZCS-5598: Bug 109012 - Account Enumeration [CWE-203]

jdavid-synacor:
This change removes the "WWW-Authenticate: BASIC realm="Zimbra"  from the response header for existing users in the test curl calls, 401 responses.

Test calls:
curl -k -i -u 'admin:password' 'https://[zimbra.hostname]/service/home/~'
curl -k -i -u 'doesnotexistrandom737:password' 'https://[zimbra.hostname]/service/home/~'

For the following calls both 401 headers should not have the following row, 
"WWW-Authenticate: BASIC realm="Zimbra"


rupalid:
@jdavid-synacor  Can you list the testing done. The behavior for valid user with HTTP 200 response.

PR #722, opened 2018-08-21
NICPS-381: Changes in protocol value being passed to authAccount

tarusharora:
Hi @zimsuchitgupta, as you are pretty much aware of the context here. Please review this change.


tarusharora:
As discussed in standup, pushing this change to snickers branch, due to its priority.

PR #619, opened 2018-05-08
Fixed blob store if it's compressed size is equal to uncompressed one

DinoPuller:
There is a bug on blob handling if it's compressed size is equal to uncompressed one.
* Reproduction: 
1. create a primary compressed volume with a threshold of 1 byte (or really small)
2. upload to briefcase attached file
3. try to see the image just uploaded, you will obtain an error.

* Solution:
In BlobBuilder.finish() if blob is compressed and it's size is equal to original version (uncompressed), uncompress it.

et bis

PR #725, opened 2018-08-23
Fixed blob store if it's compressed size is equal to uncompressed one (bis)

DinoPuller:
https://github.com/Zimbra/zm-mailbox/pull/619


store/src/java/com/zimbra/cs/store/BlobBuilder.java
```diff
@@ -235,6 +238,27 @@ public Blob finish() throws IOException, ServiceException {
         if (digest != null) {
             blob.setDigest(ByteUtil.encodeFSSafeBase64(digest.digest()));
             blob.setRawSize(totalBytes);
+            File file = blob.getFile();
+            File uncompresedFile = null;
+            if (blob.isCompressed() && totalBytes == file.length())
+            {
+              ZimbraLog.store.info("Blob compression is useless avoid it");
```
fatkudu: Make this debug level - how about:
`ZimbraLog.store.debug("Blob compression is useless avoid it for %s size=%s", file, totalBytes)`

PR #727, opened 2018-08-27
ZCS-5708 Modify HAB group

rupalid:
Added Admin API to move a HAB group from one parent to another across OU and domains
Verified the test cases mentioned in test rail are passing.

### DIFF
--- a/store/src/java/com/zimbra/cs/account/soap/SoapProvisioning.java
+++ b/store/src/java/com/zimbra/cs/account/soap/SoapProvisioning.java
@@ -229,6 +229,7 @@ import com.zimbra.soap.admin.message.GetUCServiceResponse;
 import com.zimbra.soap.admin.message.HABOrgUnitRequest;
 import com.zimbra.soap.admin.message.HABOrgUnitRequest.HabOp;
 import com.zimbra.soap.admin.message.HABOrgUnitResponse;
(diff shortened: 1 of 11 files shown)


store/src/java/com/zimbra/cs/account/soap/SoapProvisioning.java
```diff
@@ -3131,4 +3134,15 @@ public void getHab(String rootHabGroupId, Account account) throws ServiceExcepti
         invokeOnTargetAccount(req, account.getId());
     }
 
+    /**
+     * @param string
+     * @param string2
+     * @param string3
+     * @throws ServiceException
+     */
```
desouzas: Remove this.

store/src/java/com/zimbra/cs/account/soap/SoapProvisioning.java
```diff
@@ -3131,4 +3134,15 @@ public void getHab(String rootHabGroupId, Account account) throws ServiceExcepti
...
+    public void modifyHabGroup(String habGroupId, String currentParentGroupId, String targetParentGroupId) throws ServiceException {
```
desouzas: This method could be formatted.

rupalid: fixed

PR #728, opened 2018-08-28
ZCS-5791 make a zmprov command for CreateHABGroupRequest

naijenn:
CLI using zmprov for creating HAB Group using CreateHABGroupRequest.

The command depends on three required values: HAB Group name, HAB Org Unit Name, and the account@domain name. Usage message shows if all three are params are not added. Additional parameters include boolean for dynamic and distibutionListInfo attributes.

A HAB Org Unit must be created in advance (see ZCS-5396 jira ticket for additional info).

Additional notes will be added to the ZCS-5791 ticket.


store/src/java/com/zimbra/cs/account/ProvUtil.java
```diff
@@ -1774,6 +1779,20 @@ private void modifyHabGroup(String[] args)  throws ServiceException {
         ((SoapProvisioning) prov).modifyHabGroup(args[1], args[2], args[3]);
     }
 
+    private void doCreateHabGroup(String args[]) throws ServiceException {
+        if (!(prov instanceof SoapProvisioning)) {
+            throwSoapOnly();
+        }
+        if(args.length < 4) {
+            usage();
+            return true;
```
shrikant-prasad: @naijenn  compilation error: please fix it.

PR #743, opened 2018-08-31
ZCS-4904:End all session using EndSession API

shrikant-prasad:
Added two attributes in EndSessionRequest: all and excludeCurrent.

- When all is 1, all soap sessions will be cleared.
- When all is 1 and excludeCurrent is also 1, all soap sessions except the current session will be cleared.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/account/EndSession.java
+++ b/store/src/java/com/zimbra/cs/service/account/EndSession.java
@@ -1,7 +1,7 @@
 /*
  * ***** BEGIN LICENSE BLOCK *****
  * Zimbra Collaboration Suite Server
- * Copyright (C) 2008, 2009, 2010, 2013, 2014, 2016 Synacor, Inc.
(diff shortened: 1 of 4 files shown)


store/src/java/com/zimbra/cs/service/account/EndSession.java
```diff
@@ -40,23 +47,56 @@
...
+        Account account = getAuthenticatedAccount(zsc);
+        if (clearAllSessions) {
+            String currentSessionId = null;
+            if (excludeCurrrentSession && zsc.hasSession()) {
+                Session currentSession = getSession(zsc);
+                currentSessionId = currentSession.getSessionId();
+            }
+            Collection<Session> sessionCollection = SessionCache.getSoapSessions(account.getId());
+            if (sessionCollection != null) {
+                List<Session> sessions = new ArrayList<Session>(sessionCollection);
```
rupalid: Why not use for (Session session: sessions)

shrikant-prasad: was getting ConcurrentModificationException while using for (Session session: sessions)

PR #747, opened 2018-09-04
ZCS-5848:Create object class in ldap for address list

shrikant-prasad:
created zimbraAddressList object class. It will have following attributes:

- objectClass
- uid
- zimbraId
- displayName
- zimbraAddressListMemberFilter
- zimbraIsAddressListActive

### DIFF
--- a/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
+++ b/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
@@ -2961,6 +2961,14 @@ public class ZAttrProvisioning {
     public static final String A_zimbraActiveSyncEhcacheMaxDiskSize = "zimbraActiveSyncEhcacheMaxDiskSize";
 
     /**
+     * filter to decide members of the address list
+     *
+     * @since ZCS 8.8.10
+     */
+    @ZAttr(id=3074)
+    public static final String A_zimbraAddressListMemberFilter = "zimbraAddressListMemberFilter";
+
+    /**
      * access control mechanism for admin access acl: ACL based access
      * control (a.k.a. delegated admin). global: allows only global admins.
      *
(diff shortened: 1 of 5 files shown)


common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
```diff
@@ -70,6 +70,22 @@ public static AccountStatus fromString(String s) throws ServiceException {
...
+        private AddressListStatus(String value) { mValue = value; }
+        public String toString() { return mValue; }
+        public static AddressListStatus fromString(String s) throws ServiceException {
+            for (AddressListStatus value : values()) {
+                if (value.mValue.equals(s)) return value;
+             }
+             throw ServiceException.INVALID_REQUEST("invalid value: "+s+", valid values: "+ Arrays.asList(values()), null);
+        }
+        public boolean isActive() { return this == active;}
+        public boolean isInactive() { return this == inactive;}
```
chinmay15: is isInactive() really needed? there is isActive() just above isInactive() which can be used to get whether adress list is active or not.

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -212,7 +212,7 @@ TODO - add support for multi-line values in globalConfigValue and defaultCOSValu
     <desc>RFC2256: destination indicator</desc>
 </attr>
 
-<attr name="displayName" type="string" optionalIn="account,distributionList,group,groupDynamicUnit" requiredIn="calendarResource" flags="accountInfo,domainAdminModifiable" callback="DisplayName">
+<attr name="displayName" type="string" optionalIn="account,distributionList,group,groupDynamicUnit,dynamicContactGroup" requiredIn="calendarResource" flags="accountInfo,domainAdminModifiable" callback="DisplayName">
```
chinmay15: let's stick to 1 keyword either "address list" or "dynamic contact group".

common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
shrikant-prasad: these methods are auto generated by the generate-getters ant target.

PR #748, opened 2018-09-05
ZCS-5871 updated is:unread condition to fix non null error

naijenn:
Was able to replicate error, and then updated condition to prevent the non null exception.

For testing remove sortby value in searchRequest while querying "is:unread"

```Java
<soap:Body>
	<SearchRequest xmlns="urn:zimbraMail" types="message" needExp="1" recip="0" limit="100" offset="0">
		<query>is:unread</query>
	</SearchRequest>
</soap:Body>
```


chinmay15:
please verify results with different sort by options and without sort by.

PR #750, opened 2018-09-07
ZCS-5389: API to get all address lists for domain

shrikant-prasad:
API to get all active address list in user's domain.
removed displayName and added description ldap attribute for addressList objectClass.


shrikant-prasad:
moved handler to zm-network-store repo

PR #761, opened 2018-09-14
ZCS-5754 - Tamil translations 20180914

mihaipopahtp:


mihaipopahtp:
@reedesl, 
Thank you for the note. These will be taken into account and fixed if needed during linguistic testing.

PR #762, opened 2018-09-15
add tyep in params

chinmay15:
**Problem:** if type is "all", !(#zimbraAccountCalendarUserType:RESOURCE) should not be added in gal and ldap query.

**Fix:** missed to add type in params, so account was being set as default value and !(#zimbraAccountCalendarUserType:RESOURCE) was being added in the query.

**Testing done:**
- Tested create address list api manually for different values of "type" and queries are generated according to type.

**Testing to be done by QA:**
- verify gal and ldap queries are generated according to type and given conditions.


chinmay15:
merged in feature/addresslist

commit e8f5356755d51fdd04ca29f11dd95296b45c43cb (HEAD -> feature/addresslist, origin/feature/addresslist)
Date:   Sun Sep 16 01:41:20 2018 +0530

    add tyep in params

PR #767, opened 2018-09-21
ZCS-5388 Modify address list api.

desouzas:
Modify address list api.

Looking for feedback on best method to retrieve the existing address list's domain and search type. This will allow domain/type to be unchanged when left out of the request.

**Testing Done**
Basic modifications of name/desc/filter.

**Testing to be done by QA**
Filter modification testing.


store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
```diff
@@ -11474,4 +11475,43 @@ public AddressList getAddressList(String zimbraId) throws ServiceException {
         }
         return list;
     }
+
+    @Override
+    public void modifyAddressList(String addressListId, Domain domain, String name, String desc,
+        Map<String, String> attrs) throws ServiceException {
```
rupalid: Stephan, can you check if only prov.modifyAttrs woill no server the purpose. Do we need to have this method

desouzas: I can do all of this logic in the document handler, but looking at the other modify methods in this class - this seemed the appropriate place to use zlc.

desouzas: Also worth mentioning, prov.modifyAttrs tries and fails to cast AddressList to an ldap entry so would need to make some small changes.

desouzas:
Do not merge this yet. Discussing solution for fetching domain from target address list if we do not want to change the domain. Likely will have to extract it from the end of the address list dn.

desouzas:
Rebased to develop.

PR #779, opened 2018-10-01
ZCS-5825:fix ghgm issues

shrikant-prasad:
1.removed seniorityIndex from toString as it should not be displayed in zmprov ghgm response.
2.removed caching of ghgm result for static group as cache is not updated when seniorityIndex is modified.


shrikant-prasad:
added condition to include seniorityIndex in toString if it's non-zero.

PR #782, opened 2018-10-03
NICPS-431: ldap schema changes to add new attribute type intl_email

ajinkya81:
Part 1 - LDAP schema changes to support RFC6530 i.e. UTF8 characters in local part of email address .
Part 2 - updating java mail library to make compatible to support utf8 in local part of email address  
Screenshot ->
http://synscreen.buf.synacor.com/view/93030f.png

### DIFF
--- a/common/src/java/com/zimbra/common/io/TcpServerInputStream.java
+++ b/common/src/java/com/zimbra/common/io/TcpServerInputStream.java
@@ -30,6 +30,7 @@ import java.nio.charset.StandardCharsets;
 public class TcpServerInputStream extends BufferedInputStream {
 
     StringBuilder buffer;
+    BufferedReader br;
     protected static final int CR = 13;
     protected static final int LF = 10;
 
@@ -54,9 +55,9 @@ public class TcpServerInputStream extends BufferedInputStream {
      */
 	public String readLine() throws IOException {
 		buffer.delete(0, buffer.length());
-		BufferedReader br = new BufferedReader(new InputStreamReader(this, StandardCharsets.UTF_8));
+		br = new BufferedReader(new InputStreamReader(this, StandardCharsets.UTF_8));
 		while (true) {
-			int ch = read();
+			int ch = br.read();
 			if (ch == -1) {
 				return null;
 			} else if (ch == CR) {
@@ -67,4 +68,8 @@ public class TcpServerInputStream extends BufferedInputStream {
 			buffer.append((char)ch);
 		}
 	}
+
+	public void close ()  throws IOException {
+	   br.close();
+	} 
 }
--- a/common/src/java/com/zimbra/common/io/TcpServerInputStream.java
+++ b/common/src/java/com/zimbra/common/io/TcpServerInputStream.java
@@ -69,7 +69,8 @@ public class TcpServerInputStream extends BufferedInputStream {
 		}
 	}
 
-	public void close ()  throws IOException {
-	   br.close();
+	 public void close ()  throws IOException {
+        br.close();
+		super.close();
 	} 
 }
--- a/common/src/java/com/zimbra/common/io/TcpServerInputStream.java
+++ b/common/src/java/com/zimbra/common/io/TcpServerInputStream.java
@@ -57,23 +57,27 @@ public class TcpServerInputStream extends BufferedInputStream {
      *         the end of the stream has been reached
      * @throws IOException
      */
-	public String readLine() throws IOException {
-		buffer.delete(0, buffer.length());
-		while (true) {
-			int ch = br.read();
-			if (ch == -1) {
-				return null;
-			} else if (ch == CR) {
-				continue;
-			} else if (ch == LF) {
-				return buffer.toString();
-			}
-			buffer.append((char)ch);
-		}
-	}
+    public String readLine() throws IOException {
+        buffer.delete(0, buffer.length());
+        while (true) {
+            int ch = br.read();
+            if (ch == -1) {
+                return null;
+            } else if (ch == CR) {
+                continue;
+            } else if (ch == LF) {
+                return buffer.toString();
+            }
+            buffer.append((char)ch);
+        }
+    }
 
-	 public void close ()  throws IOException {
-        br.close();
-		super.close();
-	} 
+    public void close ()  throws IOException {
+        try {
+           br.close();
+           super.close();
+        } catch (IOException $e) {
+            // @TODO --> handle / log exception
+        }
+    }
 }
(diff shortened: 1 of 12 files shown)


common/src/java/com/zimbra/common/io/TcpServerInputStream.java
```diff
@@ -50,18 +53,24 @@ public TcpServerInputStream(InputStream in, int size) {
...
-                continue;
-            } else if (ch == LF) {
-                return buffer.toString();
-            }
-            buffer.append((char)ch);
-        }
-    }
+	public String readLine() throws IOException {
+		buffer.delete(0, buffer.length());
+		br = new BufferedReader(new InputStreamReader(this, StandardCharsets.UTF_8));
```
Prashantsurana: I think there is no need to create BufferReader object on each readLine. 
Also we are closing just once.

Move BufferReader to both constructors.

Thanks

Prashantsurana (changes requested):

store/src/java/com/zimbra/cs/lmtpserver/LmtpAddress.java
```diff
@@ -119,7 +123,7 @@ public String getDomainPart() {
     }
 
     public boolean isValid() {
-	return mIsValid;
+		return true;
```
shrikant-prasad: why are we by-passing the validity check?

jkester1986: ^ @ajinkya81 did you see this question? This function is always going to return `true` now

common/src/java/com/zimbra/common/io/TcpServerInputStream.java
```diff
@@ -50,18 +57,23 @@ public TcpServerInputStream(InputStream in, int size) {
...
+				continue;
+			} else if (ch == LF) {
+				return buffer.toString();
+			}
+			buffer.append((char)ch);
+		}
+	}
+
+	 public void close ()  throws IOException {
+        br.close();
```
jkester1986: there seem to be a lot of inconsistencies on spaces vs tabs throughout your changes. It's more noticeable in some spots than others because of how the code lines up (or doesn't) - here in particular it's very noticeable because. Please fix in all relevant locations

jkester1986 (changes requested):

common/src/java/com/zimbra/common/io/TcpServerInputStream.java
```diff
@@ -64,4 +71,13 @@ public String readLine() throws IOException {
             buffer.append((char)ch);
         }
     }
+
+    public void close ()  throws IOException {
+        try {
+           br.close();
+           super.close();
+        } catch (IOException $e) {
+            // @TODO --> handle / log exception
```
jeastman: Let's not leave a @TODO comment here, since there is nothing to do when handing this exception. I would simply ignore the exception and leave a comment in the code to that effect.

store/src/java/com/zimbra/cs/account/AttributeInfo.java
```diff
@@ -336,6 +336,11 @@ protected void checkValue(String value, Map attrsToModify) throws ServiceExcepti
                 throw AccountServiceException.INVALID_ATTR_VALUE(
                         mName+" is longer than max allowed: "+mMaxDuration, null);
             return;
+        case TYPE_INTL_EMAIL:
+            if (value.length() > mMax)
+                throw AccountServiceException.INVALID_ATTR_VALUE(
+                        mName+" value length("+value.length()+") larger than max allowed: "+mMax, null);
+            validEmailAddress(value, false);
```
jeastman: Is there a "return;" missing after this line?

store/src/java/com/zimbra/cs/account/names/NameUtil.java
```diff
@@ -16,14 +16,17 @@
  */
 package com.zimbra.cs.account.names;
 
-import javax.mail.internet.AddressException;
+//import javax.mail.internet.AddressException;
```
jeastman: Let's not leave a commented our header in the code.

store/src/java/com/zimbra/cs/account/names/NameUtil.java
```diff
@@ -16,14 +16,17 @@
...
+
+import java.io.UnsupportedEncodingException;
 import javax.mail.internet.InternetAddress;
 
 import com.google.common.base.Strings;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.util.EmailUtil;
 import com.zimbra.cs.account.Config;
 import com.zimbra.cs.account.Provisioning;
+//import com.zimbra.common.util.ZimbraLog;
```
jeastman: Let's not leave a commented out header in the code.

store/src/java/com/zimbra/cs/account/names/NameUtil.java
```diff
@@ -59,10 +62,19 @@ private static boolean containsNonLDH(String input) {
...
-                throw ServiceException.INVALID_REQUEST("invalid email address", null);
-        } catch (AddressException e) {
+            int index = addr.indexOf('@');
+            String localPart;
+            if (index == -1) {
+               localPart = addr;
+            } else {
+               localPart = addr.substring(0, index);
+            }
+            //InternetAddress ia = new InternetAddress(addr, true);
```
jeastman: Let's not leave old code commented out in new code.

store/src/java/com/zimbra/cs/lmtpserver/LmtpAddress.java
```diff
@@ -119,7 +120,7 @@ public String getDomainPart() {
     }
 
     public boolean isValid() {
-	return mIsValid;
+        return mIsValid;
```
jeastman: I would suggest doing whitespace only changes in their own commit.

store/src/java/com/zimbra/cs/lmtpserver/LmtpAddress.java
```diff
@@ -40,6 +40,7 @@
     private String mRemoteServer; // if mOnLocalServer is false
 
     public LmtpAddress(String arg, String[] allowedParameters, String rcptDelim) {
+
```
jeastman: I would suggest doing whitespace only changes in their own commit.

jeastman (changes requested):

Prashantsurana (approved):
Looks good to me. 
did you update new java-mail in zm-zcs-libs? 
What about updating ldap core schema? 
Is it worth adding into zimbra.schema?

PR #784, opened 2018-10-04
ZCS-5909:Improved mail sender persona

shrikant-prasad:
When Persona Reply_To is configured, email address in Return-Path should be the email address 
 configured in reply-to rather than primary email address.

For this expected behaviour, zimbraSmtpRestrictEnvelopeFrom should be set to False.


ghen2:
Hi

We just found out about this undocumented behaviour.  Setting the envelope-sender to the Reply-To address is not a good idea, as this could be a 3rd party address (like gmail.com) and violate SPF or DMARC policy for that domain.   Only From and/or Sender -which should be local domains- should be considered as envelop sender, not Reply-To.  Can you reconsider this behaviour, or make it an additional configuration knob?

PR #793, opened 2018-10-16
Returning raw messages for "show original" requests setting a charset

zx-fulvio:
When a user requests the raw message via the "show original" web
interface link a text/plain page is returned, without any charset
indication in the ``Content-Type`` header of the response.

That often enough leads to text encoded as UTF-8 or other encodings to
be processes as Windows CP-1252, with incorrect characters displayed to the user.

For example an e-mail containing a fragment such as
```
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: 8bit

dd 🙵🙵🙵🙵 
```
might be displayed as
```
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: 8bit

dd ðŸ™µðŸ™µðŸ™µðŸ™µ 
```
which confuses the user, although it does not change the actual contents of the message.

This commit introduces a selection of the charset for the ``Content-Type`` header of the
response using, in the following order, (a) the value of an optional
"charset" parameter in the request or (b) the value, if set, of the
zimbraPrefMailDefaultCharset for the target account or (c) a sane(r) default of "utf-8".

### DIFF
--- a/store/src/java/com/zimbra/cs/service/formatter/NativeFormatter.java
+++ b/store/src/java/com/zimbra/cs/service/formatter/NativeFormatter.java
@@ -165,6 +165,11 @@ public final class NativeFormatter extends Formatter {
             handleMessagePart(context, mp, msg);
         } else {
             context.resp.setContentType(MimeConstants.CT_TEXT_PLAIN);
+            context.resp.setCharacterEncoding(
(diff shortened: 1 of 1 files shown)


Prashantsurana:
@fatkudu  @rupalid  please review.

PR #794, opened 2018-10-16
Postfix long queue IDs support: removed case alterations to queue IDs

zx-fulvio:
In order to support the long queue ID format of Postfix, as described at http://www.postfix.org/postconf.5.html#enable_long_queue_ids, the queue ID cannot be converted to upper/lower case since it contains both upper and lower case characters.

Soon following PRs for the scripts handling the interaction with postfix.


Prashantsurana:
Please review.

fatkudu:
Thanks for this.  Has been integrated for testing https://github.com/Zimbra/zm-mailbox/pull/860

PR #806, opened 2018-10-26
Hotfix/zcs 6214 additional quota provider

zxAlka:


fatkudu:
This seems completely unrelated to additional quota provider work.
Is it required for patch 3 (code freeze today) in which case it seems late to the party?

PR #817, opened 2018-11-21
NICPS-474

jkester1986:


common/src/java/com/zimbra/common/soap/Element.java
```diff
@@ -304,6 +304,11 @@ public Element getOptionalElement(QName qname) {
      *  with the given name exist, returns an empty <tt>List</tt> */
     public abstract List<Element> listElements(String name);
 
+    public abstract Element getMetaSection(String name);
```
fatkudu: At first look, this seems rather strange.  Element is a very low level library.  It looks like you're adding info specific to a particular application to it.  Not in favor of that without an extremely good reason.  For instance, if any pre-existing usages use the name "meta" this is likely to break them and if not will prevent that name being used in the future.

jkester1986: This is to get the metadata off of an element, using the section name, if it exists, which you can see here: https://github.com/Zimbra/zm-mailbox/pull/817/files#diff-7e24ba5b0dd46d67e8b06ddde7f2f495R1833.

This is a client-specific change, in a client-specific repo. As nothing is calling this function besides ZMessage.java, I'm not sure how this would break something?

This is so we can get the metadata we need up into a print file. It's a quick-and-dirty solution for something the client is expecting for a build this week. Ideally, I would have liked to call GetCustomMetadata right from the print file, however there are no beans/tags set up for that in taglib, and from what I've been told no one really knows that repo. Alternatively, I could call GetCustomMetadata in ZAppointmentHit (which is where the data for the appointments for print comes from), but I don't know how to make that call in there because I don't know how to get access to the context. But again, quick solution for something the client wants, and it's client-specific

jkester1986: @fatkudu I ended up moving the function into ZMessage.java instead of messing with Element.java

PR #819, opened 2018-11-30
ZCS-6286 Fixed the code added to populate detailed data

rupalid:
Used DetailedMerged view instead of Detailed while making the request. Used the merged view to create information about intervals
Iterated over the calendar event array to find matching interval and populate event data
In response only returned all parameters for detailed view for others it is start and end time

Tested by creating users on exchange side with different permission on FB info and requesting free busy for them

### DIFF
--- a/store/src/java/com/zimbra/cs/fb/ExchangeEWSFreeBusyProvider.java
+++ b/store/src/java/com/zimbra/cs/fb/ExchangeEWSFreeBusyProvider.java
@@ -923,6 +923,8 @@ public class ExchangeEWSFreeBusyProvider extends FreeBusyProvider {
         for (Request re : req) {
             
             int i = 0;
+            long startTime = req.get(0).start;
+            long endTime = req.get(0).end;
             for (FreeBusyResponseType attendeeAvailability : results) {
                 if (attendeeAvailability.getFreeBusyView() != null) {
                     	String fbResponseViewType = attendeeAvailability.getFreeBusyView().getFreeBusyViewType().get(0);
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/fb/ExchangeEWSFreeBusyProvider.java
```diff
@@ -920,63 +921,55 @@ public boolean handleMailboxChange(String accountId) {
...
                     if (re.email == emailAddress) {
-                        if (ResponseClassType.SUCCESS != attendeeAvailability.getResponseMessage().getResponseClass()) {
-                            ZimbraLog.fb.warn("%s :: %s",attendeeAvailability.getResponseMessage().getResponseCode(), attendeeAvailability.getResponseMessage().getMessageText());
+                        if (ResponseClassType.ERROR == attendeeAvailability.getResponseMessage().getResponseClass()) {
+                            ZimbraLog.fb.debug("Unable to fetch free busy for %s  error code %s :: %s",
+                                emailAddress,
+                                attendeeAvailability.getResponseMessage().getResponseCode(),
+                                attendeeAvailability.getResponseMessage().getMessageText());
                             if (attendeeAvailability.getResponseMessage().getResponseCode().equals(ResponseCodeType.ERROR_NO_FREE_BUSY_ACCESS)) {
-                            	FreeBusy npFreeBusy = FreeBusy.nodataFreeBusy(name, req.get(0).start, req.get(0).end);
```
shrikant-prasad: better to move the code common in both if else block, out of the if block.

                                FreeBusy npFreeBusy = FreeBusy.nodataFreeBusy(name, req.get(0).start, req.get(0).end);
                                ret.add(npFreeBusy);

store/src/java/com/zimbra/cs/fb/ExchangeEWSFreeBusyProvider.java
```diff
@@ -920,63 +921,55 @@ public boolean handleMailboxChange(String accountId) {
...
-                            	npFreeBusy.setHasPermission(false);
-                            	ret.add(npFreeBusy);
+                                	FreeBusy npFreeBusy = FreeBusy.nodataFreeBusy(name, req.get(0).start, req.get(0).end);
```
shrikant-prasad: req.get(0).start and  req.get(0).end has been used at several places. better to assign these values on top and use the variables.

PR #821, opened 2018-12-03
NICPS-494: IDN - LMTP - support Hindi Email with special characters

ajinkya81:
When we send email to above users (having special hindi characters like joined letter) LMTP throws error "Syntax error in recipient address"
for e.g. अजिंक्य@test.com, उत्तम@test.com

Verify this work using this fix.

### DIFF
--- a/store/src/java/com/zimbra/cs/lmtpserver/LmtpAddress.java
+++ b/store/src/java/com/zimbra/cs/lmtpserver/LmtpAddress.java
@@ -352,7 +352,7 @@ public class LmtpAddress {
     }
 
     private boolean isLetter(int c) {
-      return Character.isAlphabetic(c);
+      return (Character.isAlphabetic(c) || (Character.UnicodeBlock.of(c) == Character.UnicodeBlock.DEVANAGARI));
     }
 
     private boolean skipAddress() {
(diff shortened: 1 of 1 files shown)


jeastman:
Would it be possible to make this more generic, rather than Hindi specific?

I was thinking along the lines of how isAlphabetic() works. It looks at the type (isType()) result and classifies it as alphabetic if the type is one of the following:

UPPERCASE_LETTER
LOWERCASE_LETTER
TITLECASE_LETTER
MODIFIER_LETTER
OTHER_LETTER
LETTER_NUMBER

If we were to extend this to include the types which are failing (but we want to include), would this be a more generic solution?

Some types to consider would be (but may not be correct):

MODIFIER_SYMBOL
NON_SPACING_MARK
OTHER_SYMBOL
SURROGATE

I'm not sure which "characters" are failing the test in our Hindi case, so we made need to see the type which is returned in those cases. Just seems to me that special-casing Hindi may cause us future maintenance problems.

ajinkya81:
Thanks for the feedback, 
I have verified that the disallowed character type was "NON_SPACING_MARK" and so included it now and removed special hindi case.

jeastman:
Looks good. Good job in finding the proper type of character.

PR #823, opened 2018-12-06
Bug 109088: Avoid DoS via catastrophic backtracking in HtmlPurifier

mss:
This issue was introduced with the fix for ZCS-5696/Bug 109020.  It probably fixes some XSS issue; since the bugs are private nobody can tell.

But unfortunately did the fix also introduce a case of catastrophic backtracking with the regex
```
.*javascript\s*
```

This patch drops the initial greedy match since it is unneeded.  It could also drop the trailing greedy match since it is unneeded as well but instead appends a colon since that's probably the intention here.

It also drops the check for `alert()` because... seriously guys?  An `alert()` is just a symptom of a test XSS, no real XSS will ever do an `alert()` but you run this regex on any HTML content.

### DIFF
--- a/common/src/java/com/zimbra/common/localconfig/DebugConfig.java
+++ b/common/src/java/com/zimbra/common/localconfig/DebugConfig.java
@@ -265,8 +265,7 @@ public final class DebugConfig {
     public static final String defangImgSkipOwaspSanitize = value("defang_img_skip_owasp_sanitize", "^cid:.*@");
     public static final String defangOwaspValidImgTag = value("owasp_valid_img_tag", "<\\s*img");
     public static final String defangStyleUnwantedStrgPattern = value("defang_style_unwanted_strg_pattern", "\\s*(('){2,})");
-    public static final String defangOwaspAlert = value("defang_owasp_alert_tag", "alert\\((.*)\\)");
-    public static final String defangOwaspJavaScript = value("defang_owasp_javascript_tag", ".*javascript\\s*");
+    public static final String defangOwaspJavaScript = value("defang_owasp_javascript", "javascript\\s*:");
 
     /*
      * Default maximum size of convertd response. This reduces OOME in case of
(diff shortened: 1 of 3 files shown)


fatkudu:
I seem to remember it is possible to do damage with the contents of `alert()`.  If we don't catch that, then people targeting us who would have used javascript may try to use `alert()` instead knowing that we catch javascript but not alert.  So, I'm a bit wary of the removed autotests (and by implication by the looks, one of the changes) - if they covered a potential hole.

plobbes:
@mss is correct that looking for `alert()` isn't the way we should be addressing things. Ultimately, we should have a 'whitelist' of allowed elements/attrs and others should be stripped.  We want to remove attrs that allow javascript to potentially be executed (i.e. https://www.w3schools.com/js/js_htmldom_events.asp).  I feel I should mention DOMPurify here, although I know we're just trying to address a single issue and not necessarily look at the bigger picture.

mss:
Thanks for looking into this.

I think there is generally something fishy about this code. The REs (also the one before) look like an optimization to skip the defanging by the `owasp-java-html-sanitizer`. It should also be possible to just get rid of the RE matching and just apply the defanging to all content but I tried to stay close to the current behaviour.

It looks like the new REs as added by ZCS-5696 just copied the code before from the check before and modified it. At least the constant name `VALID_SCRIPT_TAG` doesn't make sense since it doesn't look for a `<script>` tag (as the check above looks for an `` tag) but a call of JavaScript. Which is why I renamed it.

Additionally, the current behaviour as added by ZCS-5696 can actually cause the defanger to run twice if both RE blocks (the existing and the new one) match. This might be intended behaviour or not. Even if it is, it might be cheaper to just unconditionally run the defanger twice.

Another thing I mentioned in [bug 109088](https://bugzilla.zimbra.com/show_bug.cgi?id=109088) is that the used version of `owasp-java-html-sanitizer` is five years or so out of date. I am sure that it doesn't catch/defang things which the current version do.

So IMO somebody should have a closer look at this part of the code but I kept my patch small to just get rid of the bad RE (and on top the one which just adds wrong confidence).

PR #826, opened 2018-12-19
ZCS-6141 Annotations for freeBusy api.

desouzas:
* Added graphql annotations for freeBusy

See associated PR:
Zimbra/zm-gql#44

**Testing done**
Free busy checks for different users

**Testing to be done by QA**
See jira/testrail.

### DIFF
--- a/soap/src/java/com/zimbra/soap/mail/type/FreeBusySlot.java
+++ b/soap/src/java/com/zimbra/soap/mail/type/FreeBusySlot.java
@@ -22,9 +22,16 @@ import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlAttribute;
 
+import com.zimbra.common.gql.GqlConstants;
 import com.zimbra.common.soap.MailConstants;
 
+import io.leangen.graphql.annotations.GraphQLIgnore;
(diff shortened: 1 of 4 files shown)


soap/src/java/com/zimbra/soap/mail/type/FreeBusySlot.java
```diff
@@ -134,6 +147,7 @@ public void setSubject(String subject) {
 		this.subject = subject;
 	}
 
+	@GraphQLIgnore
```
shrikant-prasad: why are these fields ignored?

soap/src/java/com/zimbra/soap/mail/type/FreeBusyUserSpec.java
```diff
@@ -51,8 +56,11 @@
     public FreeBusyUserSpec() {
     }
 
+    @GraphQLInputField(name=GqlConstants.FOLDER_ID, description="Calendar folder ID; if omitted, get f/b on all calendar folders")
     public void setFolderId(Integer folderId) { this.folderId = folderId; }
+    @GraphQLInputField(name=GqlConstants.ID, description="Zimbra ID Either name or id must be specified")
```
shrikant-prasad: instead of using this class, better to create new one in zm-gql/models/input and in that we can have selector (byName or byId) and the value.

PR #833, opened 2019-01-08
ZCS-6232:GetContacts returns DL members even if expansion/viewing is …

sneha-patil-synacor:
Issue:
GetContacts SOAP command returns DL members even if expansion/viewing is disabled for user

Fix:
Do not add members to soap response element if the user does not have DL expansion right

Added junit test

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/ToXML.java
+++ b/store/src/java/com/zimbra/cs/service/mail/ToXML.java
@@ -96,8 +96,10 @@ import com.zimbra.cs.account.Provisioning;
 import com.zimbra.cs.account.Server;
 import com.zimbra.cs.account.accesscontrol.GranteeType;
 import com.zimbra.cs.account.accesscontrol.ZimbraACE;
+import com.zimbra.cs.account.accesscontrol.Rights.User;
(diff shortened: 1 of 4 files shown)


store/src/java/com/zimbra/cs/service/mail/ToXML.java
```diff
@@ -874,6 +894,22 @@ public static Element encodeContact(Element parent, ItemIdFormatter ifmt, Operat
...
+    public static boolean hasDLViewRight(String email, Account mRequestedAcct,
+        Account mAuthedAcct) {
+        if (mRequestedAcct != null && mAuthedAcct != null) {
+            GalGroup.GroupInfo groupInfo = GalGroupInfoProvider.getInstance().getGroupInfo(email,
+                true, mRequestedAcct, mAuthedAcct);
+            boolean canExpand = true;
+            if (groupInfo != null) {
+                canExpand = (GalGroup.GroupInfo.CAN_EXPAND == groupInfo);
+            }
+            if (!canExpand) {
```
rupalid: just return canExpand

sneha-patil-synacor: Fixed

PR #835, opened 2019-01-11
ZCS-6512 Defanger high CPU utilization fix

rupalid:
Fixed the regex causing HIGH CPU utilization
Tested with all mimes that were reported in ZBUG-736
Added unit test


fatkudu:
Looks like this is related to external contribution https://github.com/Zimbra/zm-mailbox/pull/823.
Worth looking at the conversation there?

rupalid:
> Looks like this is related to external contribution #823.
> Worth looking at the conversation there?

Yes I had a look at it, it more of a workaround. The new solution works with a single regex which fixes the original issues mentioned in ZCS-5696

PR #855, opened 2019-02-28
Changes neede for JDK 11

rupalid:
Modified the JVM parameters and cacerts location to work with JDK 11.
Tested ZCS with these changes and JDK 11
Verified that all ZCS services start without any issue and ZCS is functional


common/src/java/com/zimbra/common/localconfig/LC.java
```diff
@@ -575,17 +575,11 @@
             " -Djava.awt.headless=true" +
             " -Dsun.net.inetaddr.ttl=${networkaddress_cache_ttl}" +
             " -Dorg.apache.jasper.compiler.disablejsr199=true" +
-            " -XX:+UseConcMarkSweepGC" +
+            " -XX:+UseG1GC" +
```
fatkudu: Do we need perf testing of different garbage collection strategies?

rupalid: Sure we have done with G1GC, will do  with other GC too..

PR #861, opened 2019-03-15
ZBUG-910:SSRF vulnerability-ProxyServlet

sneha-patil-synacor:
Issue:
Auth token not getting validated because of flaw in code execution flow. That was leading to SSRF vulnerability in ProxyServlet.

Testing done:
Executed the steps to repro SSRF vulnerability, the simple http server mentioned in the steps no more getting the targeted request


plobbes:
It appears we missed the part about `getServerPort()` being tainted (user supplied data), no?
`  return req.getServerPort() == LC.zimbra_admin_service_port.intValue();`

sneha-patil-synacor:
@plobbes 
req.getLocalPort() is going to be 8443 as the request is actually received on that port for URL https://host-ip/service/proxy
Proxy servlet allows following ports : 80, 443, 7070
If we change it from getServerPort to getLocalPort, the local port isAdminRequest() will never match the zimbra_admin_service_port value is 7071

PR #862, opened 2019-03-15
Warn log level for exceeded maximum upload size

Victor-D:
Log a warning level message in case of exceeded maximum upload size


fatkudu:
Hi Victor.  Thanks for the suggested change.  Could you explain why it is important to you that this is logged at WARN level please?  By default, it would hit the log files either way.

Victor-D:
Hello,

In this case I found this hidden one because Zimbra Migration Tool was not importing PST files properly.

On content stores I periodically check for errors and warning the only way I know :
cat /opt/zimbra/log/mailbox.log | grep -v INFO

If warnings are not logged as "WARN" I can't find them hidden with other INFO logs.

That's why I suggest raising this log level.

Victor

fatkudu:
Closing this as have new integration pull request - see https://github.com/Zimbra/zm-mailbox/pull/886

PR #863, opened 2019-03-15
ZBUG-913 SSRF vulnerability - restrict feed manager target addresses.

desouzas:
Issue:
* FeedManager has no target address restriction of user supplied RSS feed locations, resulting in potential misuse of the feed resource for SSRF to internal addresses. 

Solution:
* Restrict feed address from all private addresses by default with a localconfig (for now) blacklist which supports CIDR notation and single addresses - no wildcards.

Potential Solution Issue:
*  _Update: Private can be configured with the blacklist._ Restricting private addresses may cause problems in a corporate environment where an RSS service is available in a private ip range for employee use - this solution would prevent access if no public range ip is available for the service.

Other:
IPv6 feeds are not supported due to HttpURL parsing. (unchanged)

**Testing Done**
* Tested in webclient with original issue instructions on ticket.
* Tested various addresses along with hostname hacks, and random valid feed addresses.

**Testing to be done by QA**
See jira ticket + maybe closed loop server with http proxy.

### DIFF
--- a/common/src/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/java/com/zimbra/common/localconfig/LC.java
@@ -1296,8 +1296,11 @@ public final class LC {
     public static final KnownKey zm_oauth_classes_handlers_yahoo = KnownKey.newKey("com.zimbra.oauth.handlers.impl.YahooOAuth2Handler");
     public static final KnownKey zm_oauth_classes_handlers_google = KnownKey.newKey("com.zimbra.oauth.handlers.impl.GoogleOAuth2Handler");
     public static final KnownKey zm_oauth_classes_handlers_facebook = KnownKey.newKey("com.zimbra.oauth.handlers.impl.FacebookOAuth2Handler");
-    
-    
+
+    // Feed Manager comma-separated ip blacklist
+    public static final KnownKey zimbra_feed_manager_ip_blacklist = KnownKey.newKey("");
+
+
     static {
         // Automatically set the key name with the variable name.
         for (Field field : LC.class.getFields()) {
--- a/common/src/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/java/com/zimbra/common/localconfig/LC.java
@@ -1297,8 +1297,8 @@ public final class LC {
     public static final KnownKey zm_oauth_classes_handlers_google = KnownKey.newKey("com.zimbra.oauth.handlers.impl.GoogleOAuth2Handler");
     public static final KnownKey zm_oauth_classes_handlers_facebook = KnownKey.newKey("com.zimbra.oauth.handlers.impl.FacebookOAuth2Handler");
 
-    // Feed Manager comma-separated ip blacklist
-    public static final KnownKey zimbra_feed_manager_ip_blacklist = KnownKey.newKey("");
+    // Feed Manager comma-separated blacklist. addresses can be CIDR notation, or single ip. no wild-cards (default blacklist private networks)
+    public static final KnownKey zimbra_feed_manager_blacklist = KnownKey.newKey("10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,fd00::/8");
 
 
     static {
--- a/common/src/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/java/com/zimbra/common/localconfig/LC.java
@@ -1300,7 +1300,6 @@ public final class LC {
     // Feed Manager comma-separated blacklist. addresses can be CIDR notation, or single ip. no wild-cards (default blacklist private networks)
     public static final KnownKey zimbra_feed_manager_blacklist = KnownKey.newKey("10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,fd00::/8");
 
-
     static {
         // Automatically set the key name with the variable name.
         for (Field field : LC.class.getFields()) {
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/service/FeedManager.java
```diff
@@ -185,6 +193,28 @@ public void cleanup() {
         }
     }
 
+    /**
+     * Returns true if target address is private, link-local, loopback, or individually blacklisted.
+     * @param url The target
+     * @return True if address is blocked for feed manager
+     */
+    protected static boolean isBlockedFeedAddress(HttpURL url) {
+        List<String> blacklist = Arrays.asList(LC.zimbra_feed_manager_ip_blacklist.value().split(","));
```
plobbes: Can we support CIDR notation to allow for easier blocking of a large range of addresses?

common/src/java/com/zimbra/common/localconfig/LC.java
```diff
@@ -1296,8 +1296,11 @@
     public static final KnownKey zm_oauth_classes_handlers_yahoo = KnownKey.newKey("com.zimbra.oauth.handlers.impl.YahooOAuth2Handler");
     public static final KnownKey zm_oauth_classes_handlers_google = KnownKey.newKey("com.zimbra.oauth.handlers.impl.GoogleOAuth2Handler");
     public static final KnownKey zm_oauth_classes_handlers_facebook = KnownKey.newKey("com.zimbra.oauth.handlers.impl.FacebookOAuth2Handler");
-    
-    
+
+    // Feed Manager comma-separated ip blacklist
+    public static final KnownKey zimbra_feed_manager_ip_blacklist = KnownKey.newKey("");
```
chinmay15: I think we should add "localhost" and "127.0.0.1" as default values. Others can be added later by admin. What do you say?

desouzas: Those are covered by link local.

store/src/java/com/zimbra/cs/service/FeedManager.java
desouzas: Done and moved private addresses to default config rather than forced off.

common/src/java/com/zimbra/common/localconfig/LC.java
```diff
@@ -1296,8 +1296,11 @@
     public static final KnownKey zm_oauth_classes_handlers_yahoo = KnownKey.newKey("com.zimbra.oauth.handlers.impl.YahooOAuth2Handler");
     public static final KnownKey zm_oauth_classes_handlers_google = KnownKey.newKey("com.zimbra.oauth.handlers.impl.GoogleOAuth2Handler");
     public static final KnownKey zm_oauth_classes_handlers_facebook = KnownKey.newKey("com.zimbra.oauth.handlers.impl.FacebookOAuth2Handler");
-    
-    
+
+    // Feed Manager comma-separated blacklist. addresses can be CIDR notation, or single ip. no wild-cards (default blacklist private networks)
+    public static final KnownKey zimbra_feed_manager_blacklist = KnownKey.newKey("10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,fd00::/8");
+
+
```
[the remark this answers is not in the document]
desouzas: Ha, I actually added that back because it was originally there.

fdalton:
Hi!

As side effect of this fix, is not possible anymore to import .ics calendars that are located in the same network where zimbra is. This fix should also affect calendars?

rupalid:
> Hi!
> 
> As side effect of this fix, is not possible anymore to import .ics calendars that are located in the same network where zimbra is. This fix should also affect calendars?

How are u importing .ics calendars. This code does not come into play for Calendar imports through Webclient-->Prreferences-->Import

fdalton:
> > Hi!
> > As side effect of this fix, is not possible anymore to import .ics calendars that are located in the same network where zimbra is. This fix should also affect calendars?
> 
> How are u importing .ics calendars. This code does not come into play for Calendar imports through Webclient-->Prreferences-->Import

Hello!

I'm importing through the an ICS URL, that is hosted in the same network where zimbra is. You can access this menu through Webcliente->Calendar->Options->Add External Calendar->Add Extenal Calendar (others)-> Type: iCAL subscription.

mvonballmo:
I have a suggestion for an improvement: the logging should reflect that the address was blocked due to blacklisting of feed URLs. Currently, Zimbra just writes INVALID_REQEUST into the log, which was not helpful in determining why our internal feed suddenly started failing after an upgrade.

It took us quite a while to verify that our ICS was valid and that there were no networking problems, until we finally stumbled on this issue and "learned" of the new restriction. If the log had said something along the lines of "ACCESS DENIED: 192.168.xx.xx is a blacklisted feed URL", that would have been very helpful. Even more helpful would be to write "ACCESS DENIED: 192.168.xx.xx is a blacklisted feed URL (see zmconfig zimbra_feed_manager_blacklist for more information)".

I can make a new issue to extend the logging, if you think that would be helpful.

fjwcash:
So, if there's a blacklist, why isn't there a whitelist?

If we have a Zimbra server in 10.100.100.0/24 and a web server with an ICS feed in the same subnet, are we really expected to remove 10.0.0.0/8 from the blacklist, and replace it with a bazillion separate 10.x.x.x subnets, leaving 10.100.100.0/24 out of the list?  Really?

That entry in zmlocalconfig is going to look just nasty!

Please add a zimbra_feed_manager_whitelist entry to zmlocalconfig.

If the requested feed matches an IP or subnet in that list, then allow it.

If it doesn't match, then check the blacklist.  If it matches an IP or subnet in there, block the request (but use an actually useful error message).

If it doesn't match anything in either the whitelist or the blacklist, then allow it.


The error message says "file not found" not "denied by blacklist", so we just spent 2 days trying to figure out why our web server wasn't allowing the Zimbra server to access the .ics file anymore.  A properly worded error message would have saved us a lot of tech time.  And a simple whitelist entry would make it much easier to work with.

PR #868, opened 2019-03-29
ZCS-6944:Defanger-CPU spikes for some messages

sneha-patil-synacor:
Issue: Some messages causing CPU spike since they had large style value 

Fix: Changed the unwanted function removal regex. Added limit for style value(debug configurable parameter/default 10000 characters)

Testing done:
All junit tests are passing

### DIFF
--- a/store/src/java/com/zimbra/cs/html/DefangFilter.java
+++ b/store/src/java/com/zimbra/cs/html/DefangFilter.java
@@ -25,6 +25,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.apache.commons.lang.StringEscapeUtils;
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/html/DefangFilter.java
```diff
@@ -465,13 +466,30 @@ public void startDocument(XMLLocator locator, String encoding,
...
+                }
+            }
+            matcher.appendTail(stringBuffer);
+            value = stringBuffer.toString();
+            // strip off any @import
+            ZimbraLog.mailbox.info("sanitizeStyleValue END");
+            value = STYLE_UNWANTED_IMPORT.matcher(value).replaceAll("");
+        } else {
+            ZimbraLog.mailbox.info("style value is too long:%d characters. Trimming it.", value.length());
+            value = "";
```
rupalid: Sneha you may want to return the original value, if users have emails with value for style larger than 10K but still valid they may complain

PR #874, opened 2019-04-03
ZCS-6944:sanitize data in small chunks

sneha-patil-synacor:
Couple of changes during sanitization of style value:
1. Remove the value if its larger than what is defined in debug config
2. Execute the unwanted function sanitization for style value in parts

Testing done:
All junit tests passing

### DIFF
--- a/store/src/java/com/zimbra/cs/html/DefangFilter.java
+++ b/store/src/java/com/zimbra/cs/html/DefangFilter.java
@@ -21,6 +21,7 @@ import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.security.SecureRandom;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/html/DefangFilter.java
```diff
@@ -466,29 +467,49 @@ public void startDocument(XMLLocator locator, String encoding,
...
+            sanitizedValue = STYLE_UNWANTED_STRG_PATTERN.matcher(sanitizedValue).replaceAll("");
+            sanitizedValue = COMMENT.matcher(sanitizedValue).replaceAll("");
             // strip off any @import
-            ZimbraLog.mailbox.info("sanitizeStyleValue END");
-            value = STYLE_UNWANTED_IMPORT.matcher(value).replaceAll("");
+            sanitizedValue = STYLE_UNWANTED_IMPORT.matcher(sanitizedValue).replaceAll("");
         } else {
-            ZimbraLog.mailbox.info("style value is too long:%d characters.", value.length());
+            ZimbraLog.mailbox.info("style value is too long:%d characters. Removing it.", value.length());
+            sanitizedValue = "";
```
shrikant-prasad: instead of empty string, sanitised value after trimming the string should be sent.

PR #877, opened 2019-04-06
ZCS-6944 Reworking the change

rupalid:
Reworked the change introduced by Sneha baed on discussion with Prashant Surana.

Used StringBuilder instead on string concatenation
Applied the logic of chunking and sanitizing to all style values.

Verified all unit test pass
Verified that the mimes do not cause a CPU spike


store/src/java/com/zimbra/cs/html/DefangFilter.java
```diff
@@ -465,50 +465,48 @@ public void startDocument(XMLLocator locator, String encoding,
...
-            sanitizedValue = "";
+            matcher.appendTail(stringBuffer);
+            valuePart = stringBuffer.toString();
+            data.append(valuePart);
+            random = r.nextInt(range) + range;
+            startIndex = endIndex;
         }
+
+        sanitizedValue = data.toString();
+       // remove comments
```
desouzas: Spacing issue.

PR #892, opened 2019-04-30
ZCS7167 Fixing issues found in SOAP automation

rupalid:
Fixed the code that created the Sockets.
For FeedManager changed the code for reading content type from HttpResponse

Verified that the following error is no longer seen
[] INFO: I/O exception (java.net.SocketException) caught when processing request to {s}->https://localhost:7071: Socket is closed
[] INFO: Retrying request to {s}->https://localhost:7071

Verified by running the failing SOAP automation test case

/opt/qa/soapvalidator/data/soapvalidator/MailClient/Folders/folders_url.xml

### DIFF
--- a/common/src/java/com/zimbra/common/net/SecureProtocolSocketFactoryWrapper.java
+++ b/common/src/java/com/zimbra/common/net/SecureProtocolSocketFactoryWrapper.java
@@ -1,7 +1,7 @@
 /*
  * ***** BEGIN LICENSE BLOCK *****
  * Zimbra Collaboration Suite Server
- * Copyright (C) 2010, 2013, 2014, 2016, 2018 Synacor, Inc.
+ * Copyright (C) 2010, 2013, 2014, 2016, 2018, 2019 Synacor, Inc.
  *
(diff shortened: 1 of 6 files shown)


common/src/java/com/zimbra/common/net/SecureProtocolSocketFactoryWrapper.java
```diff
@@ -79,7 +75,18 @@ public Socket createLayeredSocket(Socket socket,  String target, int port, HttpC
             if (context != null) {
                 HttpClientContext clientContext = HttpClientContext.adapt(context);
                 HttpHost  host = clientContext.getTargetHost();
-                return factory.createSocket(host.getHostName(), host.getPort());
+
+                if(host.getPort() == -1) {
```
fatkudu: Very similar looking code in 3 places - can they be made into a utility method somewhere?

rupalid: Will do

PR #895, opened 2019-05-07
ZCS-6903:Fix test cases with OWASP-5

shrikant-prasad:
Modified owasp defanger code to fix following unit tests:

- testBug83999 - block same host form post
- testBug102637 - decimal character references
- testBug102910 - cid with @ being escaped -- test expectation fixed, @ will be escaped.
- testBug73037  - URL with spaces and URL like '//Shared_srv/folder/file with spaces.txt'
- testBug7387 - base href with src val

### DIFF
--- a/store/src/java/com/zimbra/cs/html/owasp/policies/BaseElementPolicy.java
+++ b/store/src/java/com/zimbra/cs/html/owasp/policies/BaseElementPolicy.java
@@ -1,4 +1,5 @@
 package com.zimbra.cs.html.owasp.policies;
+
 /*
  * ***** BEGIN LICENSE BLOCK *****
  * Zimbra Collaboration Suite Server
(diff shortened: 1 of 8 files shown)


store/src/java/com/zimbra/cs/html/owasp/policies/BaseElementPolicy.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class BaseElementPolicy implements ElementPolicy {
+
+    @Override
+    public String apply(String elementName, List<String> attrs) {
+
+        final int hrefIndex = attrs.indexOf("href");
+        if (hrefIndex != -1) {
+            String hrefValue = attrs.get(hrefIndex + 1);
+            OwaspHtmlSanitizer.zThreadLocal.set(hrefValue);
```
sneha-patil-synacor: isn't there any way of owasp handling it internally? Since using base url is known vulnerability.

shrikant-prasad: Couldn't find any way of handling this through owasp policies.

PR #922, opened 2019-06-18
ZBUG-977: updating allowed list of classes to deserialize

dasiyogesh:
Made fix for ImapFolder de-serialize exception by adding list of classes for comparing against valid list.


store/src/java/com/zimbra/cs/imap/MemcachedImapCache.java
```diff
@@ -111,8 +115,16 @@ public Object serialize(ImapFolder folder) throws ServiceException {
         @Override
         public ImapFolder deserialize(Object obj) throws ServiceException {
             ObjectInputStream in = null;
+
             try {
-                in = new SecureObjectInputStream(new ByteArrayInputStream((byte[]) obj), ImapFolder.class.getName());
+                Set<String> validClassNames = new HashSet<String>();
```
rupalid: Move this Class level

PR #926, opened 2019-06-28
ZBUG-1017: update business logic for email address check

ajinkya81:
Made fixes so that the existence of zimbraDataSourceEmailAddress is checked not only createDataSource but also at modifyDataSource

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/CreateDataSource.java
+++ b/store/src/java/com/zimbra/cs/service/mail/CreateDataSource.java
@@ -1,7 +1,7 @@
 /*
  * ***** BEGIN LICENSE BLOCK *****
  * Zimbra Collaboration Suite Server
- * Copyright (C) 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016 Synacor, Inc.
+ * Copyright (C) 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2019 Synacor, Inc.
  *
  * This program is free software: you can redistribute it and/or modify it under
  * the terms of the GNU General Public License as published by the Free Software Foundation,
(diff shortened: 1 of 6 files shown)


store/src/java/com/zimbra/cs/service/mail/CreateDataSource.java
```diff
@@ -252,6 +254,21 @@ static void validateFolderId(Account account, Mailbox mbox, Element eDataSource,
             }
         }
     }
+    /**
+     * Confirms that the zimbraDataSourceEmailAddress attribute is unique
+     * 
+     */
+    default static void validateEmailAddr(Account account, Element eDataSource)
```
chinmay15: don't use default access, use public

dasiyogesh: validateEmailAddr() sounds more like email address validation than data source validation, may be changed to validateEmailDataSource()

ajinkya81: fixed

ajinkya81: fixed

store/src/java/com/zimbra/cs/service/mail/CreateDataSource.java
```diff
@@ -252,8 +254,23 @@ static void validateFolderId(Account account, Mailbox mbox, Element eDataSource,
             }
         }
     }
+    /**
+     * Confirms that the zimbraDataSourceEmailAddress attribute is unique
+     * 
+     */
+    public static void validateDataSourceEmail(Account account, Element eDataSource)
```
jrjena136: As this `validateDataSourceEmail(Account account, Element eDataSource)` method is used only for `ModifyDataSource`Class and indirectly called from `processCommonOptionalAttrs()` so its better you can define this method in that class as well. unnecessarily you are creating this in `CreateDataSource` class and calling it from `ModifyDataSource` class

chinmay15: better to move to utility class.

ajinkya81: fixed

store/src/java/com/zimbra/cs/service/mail/ModifyDataSource.java
```diff
@@ -251,4 +254,20 @@ public static void processCommonOptionalAttrs(Map<String, Object> dsAttrs, Eleme
         	dsAttrs.put(Provisioning.A_zimbraDataSourceAttribute, attrList.toArray(new String[0]));
         }
     }
+
+    /**
+     * Confirms that the zimbraDataSourceEmailAddress attribute is unique
+     *
+     */
+    private static void validateDataSourceEmail(Account account, Element eDataSource)
```
chinmay15: remove static

chinmay15: my bad. calling method is static

ajinkya81:
rebased & Merged into develop branch.

https://github.com/Zimbra/zm-mailbox/commit/bc4dbb4e7963d05cad96bbb347cfa9a73547428e

PR #929, opened 2019-07-04
ZCS-7498:Creating signature through ZCO puts extra space within sign

sneha-patil-synacor:
Issue: Signature created through ZCO had extra spaces within signature lines. It was due to style which was being removed during html sanitization process. The OWASP doesn't yet support sanitization of style element.

Fix: Added support for style element sanitization using antisamy library. 

Manual testing done. The soap response now contains the sanitized style in html content.


ajinkya81:
Looks good to me, Do we need to set owasp flag to default true again?

store/src/java/com/zimbra/cs/html/owasp/OwaspPolicy.java
```diff
@@ -65,10 +65,11 @@
     static {
         try {
             load(null);
-        } catch (DocumentException de) {
-            throw new RuntimeException(de);
-        } catch (Exception ce) {
-            throw new RuntimeException(ce);
+        } catch (Exception e) {
+            if (ZimbraLog.mailbox.isDebugEnabled()) {
```
rupalid: we do not need this, it happens as part of ZimbraLog framework

store/src/java/com/zimbra/cs/html/owasp/policies/StyleTagReceiver.java
```diff
@@ -0,0 +1,98 @@
...
+                        .scan(STYLE_OPENING_TAG + text + STYLE_CLOSING_TAG, policy, AntiSamy.DOM)
+                        .getCleanHTML();
+                    sanitizedStyle = sanitizedStyle.replace(STYLE_OPENING_TAG, "")
+                        .replace(STYLE_CLOSING_TAG, "");
+                } catch (ScanException | PolicyException e) {
+                    ZimbraLog.mailbox.warn("Failed to sanitize html style element");
+                    sanitizedStyle = "";
+                }
+            }
+            wrapped.text(sanitizedStyle);
```
rupalid: We should return the original style here.

sneha-patil-synacor:
> Can we add a test case or two for the style tag since that seems to be the primary focus here?
> 
> Possibly useful refs:
> 
>     * https://github.com/nahsra/antisamy/blob/master/src/test/java/org/owasp/validator/html/test/AntiSamyTest.java
> 
>     * https://security.stackexchange.com/questions/60712/is-xss-through-style-attribute-possible-in-modern-browsers

@plobbes Thanks for the references. Will definitely add new test cases.

PR #935, opened 2019-07-15
ZCS-7568 Fixing exception while moving email to another folder

rupalid:
Fixing "java.lang.IllegalArgumentException: MIME type may not contain reserved characters
        at org.apache.http.util.Args.check(Args.java:36)" while moving item from shared folder to user folder.

Verified that the above exception does not happen an email or email with attachment is moved.

### DIFF
--- a/client/src/java/com/zimbra/client/ZMailbox.java
+++ b/client/src/java/com/zimbra/client/ZMailbox.java
@@ -2610,7 +2610,7 @@ public class ZMailbox implements ToZJSONObject, MailboxStore {
         int statusCode;
         try {
             MultipartEntityBuilder builder = MultipartEntityBuilder.create();
-            builder.addBinaryBody("upfile", in, ContentType.create(contentType), name);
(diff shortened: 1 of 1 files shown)


client/src/java/com/zimbra/client/ZMailbox.java
```diff
@@ -2610,7 +2610,7 @@ public String uploadContentAsStream(String name, InputStream in, String contentT
         int statusCode;
         try {
             MultipartEntityBuilder builder = MultipartEntityBuilder.create();
-            builder.addBinaryBody("upfile", in, ContentType.create(contentType), name);
+            builder.addBinaryBody("upfile", in, ContentType.DEFAULT_BINARY, name);
```
desouzas: Removing `contentType` here leaves it unused in this method. Is losing that information acceptable?

rupalid: Maybe wit the old library it was necessary. I tested various mimes with content type as binary and did not see any issues. I am going to keep it in method signature as it would have a cascading effect on the calling methods. Trying to keep code change to a minimum

PR #947, opened 2019-08-10
zcs-7680 : Ability to have an option to block login using the alias

jrjena136:
**Issue:**
When the accounts has aliases, accounts kept getting locked out due to bots trying to login with alias instead of actual email address. Currently there is no option to block the alias login.

**Fix:**
Added a Local config which will be `true` by default that means alias login is enabled.
If we want to block alias login modify the Local config value to `false`. After that all alias login will be blocked.

**Test:**
- Manual testing done. Created an alias for an account. default local config value is set to `true`, so the user successfully logged in.
- when the local config value modified to `false`, all alias login blocked.

**Testing to be done by QA:**
- Test that login with email address works.
- Login with alias account works.
- when local config set to `false`, only alias login will blocked but login with email address would work.


rupalid:
Jyoti, can you add a. unit test. for this.

PR #949, opened 2019-08-12
ZCS-7686:Make GetFolderRequest return "permissions" data

sneha-patil-synacor:
GetFolderRequest is not returning "permissions" data

Fix:
Adding perm attribute for mounted folders


rupalid:
Sneha, please check why the permission was being returned only for remote folders. Also it make sense to only do this for mounted folders/shared folders

PR #950, opened 2019-08-15
zcs-7680 - Added alias check at common place and added unit test

jrjena136:
**_Issue :_** 
- Alias login was successful though `LC` value `alias_login_enabled` set to `false` in case of `IMAP`, `POP` and `EWS` as well because auth provisioning is different for different protocol. 
- We had only a check in `Auth.java` where only `ZWC` login request was getting blocked for alias but all other auth request were getting successful. 
- Another problem was that when we send the whole email address in request and `LC` set to `false` the login request was getting failed where as when we send only the display name like `test` instead of `test@test.com` it was getting successful.

**_Fix:_**
- Ultimately when auth request comes from any other client other than ZWC like IMAP, POP and EWS, it encounters `prov.authAccount()` call which verifies the password in `LDAPProvisioning.java`. 
- Now I have moved the alias check logic inside `verifypassword()` method. Also created an utility method in `AccountUtil.java` which check for alias login allowed or not before verification of password.
- Also put a check for the display name if we does not get the whole email address in request as well.
- Added unit tests and modified logic in `MockProvisioning.java` for unit tests.

**_Test:_**
- Manual testing done for ZWC, EWS, IMAP, POP and Active-sync as well. Alias login is blocked in all cases when `LC` value set to `false`;
- All Unit tests are passing.

**_Testing to be done by QA:_**
- Test Auth request for different auth mode like ZWC, EWS, IMAP, POP, Active-sync.
- Auth request should pass with actual account and alias account when `LC` value set to `true`.
- Auth request should pass for actual account and fail for alias account when `LC` value set to `false`.

### DIFF
--- a/store/src/java-test/com/zimbra/cs/account/MockProvisioning.java
+++ b/store/src/java-test/com/zimbra/cs/account/MockProvisioning.java
@@ -32,11 +32,13 @@ import com.zimbra.common.account.Key.AccountBy;
 import com.zimbra.common.account.Key.AlwaysOnClusterBy;
 import com.zimbra.common.account.Key.ShareLocatorBy;
 import com.zimbra.common.account.Key.UCServiceBy;
+import com.zimbra.common.localconfig.LC;
 import com.zimbra.common.account.ProvisioningConstants;
 import com.zimbra.common.mime.MimeConstants;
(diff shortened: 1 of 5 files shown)


rupalid:
account name is the primary email address user1@domain.com and not just user1

PR #951, opened 2019-08-19
ZBUG-1087: Java - Kerberos authentication is broken on upgrade

ajinkya81:
Java 1.9 has a change with kerbros authentication which breaks Zimbra's Kerberos Auth
https://bugs.openjdk.java.net/browse/JDK-8173069

Applied the patch given in jira ticket
Fixed the exception raised due to the failed logout attempt, the verify password function returns false and a login is not attempted.
Zimbra 8.7 was JDK 1.7 and Zimbra 8.8 is JDK 1.11. This issue started with JDK 1.9.

### DIFF
--- a/store/src/java/com/zimbra/cs/account/krb5/Krb5Login.java
+++ b/store/src/java/com/zimbra/cs/account/krb5/Krb5Login.java
@@ -14,7 +14,6 @@
  * If not, see <https://www.gnu.org/licenses/>.
  * ***** END LICENSE BLOCK *****
  */
-
 package com.zimbra.cs.account.krb5;
 
 import com.zimbra.common.service.ServiceException;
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/account/krb5/Krb5Login.java
```diff
@@ -48,13 +47,21 @@
     public static void verifyPassword(String principal, String password) throws LoginException {
         LoginContext lc = null;
         try {
+            ZimbraLog.account.debug("Kerberos(krb5) principal login with password: " + principal + " and password: " + password);
```
rupalid: Change this to parametrized (%s)...

Prashantsurana:
Can you keep formatting changes and code changes in separate commit? 
I think formatting changed may not even required.

ajinkya81:
> de changes in separate commits.
> I think formatting changes may even not requir

I have skipped formatting changes, pushed only relevant code changes.

store/src/java/com/zimbra/cs/account/krb5/Krb5Login.java
ajinkya81: Done

store/src/java/com/zimbra/cs/account/krb5/Krb5Login.java
```diff
@@ -48,19 +48,27 @@
...
             if (lc != null) {
                 try {
+                    //Only log out if a login occured. If a login didn't occur, then a logout throws a NPE.
+                    if (lc.getSubject() != null
+                            && !lc.getSubject().getPrincipals().isEmpty()) {
+                        ZimbraLog.account.debug("Kerberos(krb5) Login Context subject and principal are not null. Safely logging out.");
                     lc.logout();
-                } catch(LoginException le) {
+                    } else {
+                        ZimbraLog.account.debug("Kerberos(krb5) Login Context subject and principal are null. Cannot safely logging out.");
```
rupalid: Typo safely log out

PR #955, opened 2019-08-28
ZBUG-1075:A delegated domain admin can retrieve other domains users

sneha-patil-synacor:
Issue:
A delegated domain admin can retrieve other domains users

Fix:
- Added a LC flag 'enable_delegated_admin_ldap_access' to either allow/block access/modification to LDAP entries through SOAP API
- By default, the flag value would be true to have consistency with previous behavior.

### DIFF
--- a/common/src/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/java/com/zimbra/common/localconfig/LC.java
@@ -1291,6 +1291,8 @@ public final class LC {
     // owasp handler
     public static final KnownKey zimbra_use_owasp_html_sanitizer = KnownKey.newKey(true);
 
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/account/accesscontrol/ACLAccessManager.java
```diff
@@ -81,8 +81,7 @@ private AdminRight actualRightForAdminLoginAs(Account target) {
 
     @Override
     public boolean isDomainAdminOnly(AuthToken at) {
```
rupalid: Add a unit test for this method

PR #958, opened 2019-09-05
ZBUG-711:convertd exhausts system memory when .har file is indexed

sneha-patil-synacor:
Issue:
The convertd native process is stuck and CPU spikes if the .har file is uploaded to briefcase or in attachment

Fix:
When .har file is uploaded its content type was being set to 'application/octet-stream' by default(retrieved from HttpServletRequest instance)
But, the file actually contains data in json format. the convertd is trying to convert the document based on 'application/octet-stream' and failing.
Changing the content type of the uploaded .har document to 'application/json' resolves the issue.


Victor-D:
Is the system memory overused also when a webmail user receives an outgoing message with this kind of .har attachment ? I ask because it could be a security issue.

PR #964, opened 2019-09-26
Bugfix/zbug 1163

nileshpawar1:
SPAM Learn - NullPointerException while running zmtrainsa

- Null pointer exception was coming when running the command "zmtrainsa" from the command line
- Set gm.setURI  for HttpGet before calling HttpClientUtil.executeMethod
- To test this, run the command "zmtrainsa" from the command line with zimbra user and it should not show Null pointer exception.

### DIFF
--- a/store/src/java/com/zimbra/cs/util/SpamExtract.java
+++ b/store/src/java/com/zimbra/cs/util/SpamExtract.java
@@ -25,6 +25,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/util/SpamExtract.java
```diff
@@ -222,6 +224,7 @@ public static void main(String[] args) throws ServiceException, HttpException, S
 
     private static void extract(String authToken, Account account, Server server, String query, File outdir, boolean delete, boolean raw) throws ServiceException, HttpException, SoapFaultException, IOException {
         String soapURL = getSoapURL(server, false);
+        String serverURL = getServerURL(server, false).toString();
```
jrjena136: you can use the existing variable `restURL` here.

PR #967, opened 2019-10-01
Fix for addictional quota counted multiple times.

ZeXtrasJay:
The mailbox quota is stored by the DbMailbox at every end of transaction and loaded during mailbox initiaization.
The stored quota accounts for both mail items quota and addictional quota but the addictional quota is provided at run-time and should not be stored at all.
This fix split the size of the mailbox into mail items and addictional quota and store in database only the mail items size.
In the affected mailboxes a recalculation of the mailbox quota is needed.


Prashantsurana:
@rupalid  please review and schedule this for next patch.

chinmay15:
LGTM

PR #971, opened 2019-10-30
ZBUG-987 Fixing NPE

rupalid:
Fixed a NPE seen when configured with AD and contact group has members with no email.

Added uni test


store/src/java-test/com/zimbra/cs/mailbox/ContactTest.java
```diff
@@ -339,6 +339,15 @@ public void testZCS6232() throws Exception {
         Assert.assertFalse(ToXML.hasDLViewRight("mydl@zimbra.com", account, account));
     }
 
+    @Test
+    public void testZCS6232WithNullEmail() throws Exception {
+        Account account = Provisioning.getInstance().getAccountByName("test6232@zimbra.com");
+        // mocking the group not to have view permission
```
desouzas: Not a big deal, but this comment is wrong.

PR #972, opened 2019-11-06
ZBUG-1212 Fix Repetitive copy  from Apple mail client configured with Imap

rupalid:
Added code to add the list of processed sequences  and check when a new COPY UID command is received. If the sequence are already processed do no processing and return a OK to client.

This is only when emails are moved to Trash.

Sending the PR to get feedback on approach, will clean code once all testing is done

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/CopyCommand.java
+++ b/store/src/java/com/zimbra/cs/imap/CopyCommand.java
@@ -17,10 +17,16 @@
 
 package com.zimbra.cs.imap;
 
+import java.util.Set;
+
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.util.ZimbraLog;
+
 public class CopyCommand extends ImapCommand {
 
     private ImapPath destPath;
     private String sequenceSet;
+    private  Set<String> processedList;
 
     public CopyCommand(String sequenceSet, ImapPath destPath) {
         super();
@@ -54,6 +60,7 @@ public class CopyCommand extends ImapCommand {
                 return false;
             }
         } else if (!sequenceSet.equals(other.sequenceSet)) {
+            ZimbraLog.imap.info("IMAP::: Seq 1 " + sequenceSet + "\n" + "other: " + other.sequenceSet);
             return false;
         }
         if (destPath == null) {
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/imap/CopyCommand.java
```diff
@@ -54,15 +60,79 @@ public boolean equals(Object obj) {
...
         } else if (!destPath.equals(other.destPath)) {
+            ZimbraLog.imap.info("Dest path " + destPath + " 2: " +other.destPath );
             return false;
         }
         return true;
     }
+
+    public boolean isCopyToTrash() {
+        try {
+            return this.destPath.getFolder().getFolderIdAsString().contains("3");
```
chinmay15: you compare id with Mailbox.ID_FOLDER_TRASH constant

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -137,7 +138,7 @@
         "SORT", "THREAD=ORDEREDSUBJECT", "UIDPLUS", "UNSELECT", "WITHIN", "XLIST"
     ));
 
-    private static final long MAXIMUM_IDLE_PROCESSING_MILLIS = 15 * Constants.MILLIS_PER_SECOND;
+    private static final long MAXIMUM_IDLE_PROCESSING_MILLIS = 5 * Constants.MILLIS_PER_SECOND;
```
chinmay15: keep 15 only

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -4458,6 +4481,23 @@ protected boolean doCOPY(String tag, String sequenceSet, ImapPath path, boolean
...
+                if (processedDeletes.get(mbox.getAccountId()) == null) {
+                    processedList.add(sequenceSet);
+                    processedDeletes.put(mbox.getAccountId(), processedList);
+                } else {
+                    processedList = processedDeletes.get(mbox.getAccountId());
+                    processedList.add(sequenceSet);
+                    processedDeletes.put(mbox.getAccountId(), processedList);
+                }
+            } catch (ServiceException e) {
+
```
chinmay15: log exception on debug level

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -149,6 +150,10 @@
     private static final String IMAP_DELETE_RIGHTS = "xted";
     private static final String IMAP_ADMIN_RIGHTS  = "a";
 
+    private static Map<String, Set<String>> processedDeletes =
```
desouzas: Do we sticky for IMAP or will this only work if the request hits this same service? (i.e. single node)

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -4333,18 +4339,35 @@ private boolean doSTORE(String tag, String sequenceSet, List<String> flagNames,
...
+            commandP.setProcessedList(processedDeletes.get(mbox.getAccountId()));
+
+        }  catch ( ServiceException e) {
+
+        }
+
+        try {
+            checkCommandThrottle(commandP);
+        } catch (ImapThrottledException e) {
+            sendOK(tag, copyuid + command + " completed");
```
desouzas: Why are we sending completion if we are throttling them, can we add a comment above this explaining why?

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -4333,18 +4339,35 @@ private boolean doSTORE(String tag, String sequenceSet, List<String> flagNames,
...
-
         ImapFolder i4folder = getSelectedFolder();
         if (i4folder == null) {
             throw new ImapSessionClosedException();
         }
         MailboxStore mbox = i4folder.getMailbox();
+
+        try {
+            commandP.setProcessedList(processedDeletes.get(mbox.getAccountId()));
+
```
desouzas: Extra space, and log the exception below.

store/src/java/com/zimbra/cs/imap/CopyCommand.java
```diff
@@ -54,15 +60,79 @@ public boolean equals(Object obj) {
...
+        } catch (ServiceException e) {
+            return false;
+        }
+    }
```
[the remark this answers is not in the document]
rupalid: Agree we should percolate up the Exception but in this case the method Signature does not allow me to do that and the exception would be a rare case scenario here. If it happens there is something really wrong with mailbox service. So keeping false and reprocessing looks ok to me.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
rupalid: This is classic it always goes to the same node. Fixed map

PR #976, opened 2019-11-14
zbug-903:Problem with sendAs delegate rights and Save a copy of sent messages to my Sent folder

jrjena136:
**Issue :** 

The option "Save a copy of sent messages to my Sent folder" doesn't work with a persona in the web client.

- UserA has given userB "SendAs" or "Send of Behalf". and selected the option "Save a copy of sent messages to my Sent folder"
- UseB has created a persona for userA and sent an email to userC.  Sent email did't save in userA's sent folder.

**Approach and Fix:**

When the `SendMsgRequest` comes for `sendAs` and `sendOnBehalf`  from the logged in user account, the request header contains the actual user accountId and in this case we get the target mailbox is the userA, but in case of Persona the header contains the logged in user accountId which is userB, so the target mailbox is always the logged in user mailbox and when the message sent successfully it stores in the target mailbox sent folder.

Every user account has a default persona(identity). So when we add a persona, it creates another identity. So if the SendMsgRequest comes from persona, checking if the identity has `sendAs` or `sendOnBehalf`  rights and then getting the account from the identity mail address and target mailbox for the account. Now the sent message will be stored in the actual user's sent folder

**Test:**
Manually testing done. QA needs to verify all the required scenarios.

### DIFF
--- a/store/src/java/com/zimbra/soap/SoapEngine.java
+++ b/store/src/java/com/zimbra/soap/SoapEngine.java
@@ -31,6 +31,9 @@ import javax.xml.stream.XMLStreamReader;
 import org.eclipse.jetty.continuation.ContinuationSupport;
 
 import com.google.common.base.Strings;
+import com.zimbra.common.account.Key;
+import com.zimbra.common.account.ProvisioningConstants;
+import com.zimbra.common.account.Key.AccountBy;
 import com.zimbra.common.localconfig.LC;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.soap.Element;
@@ -51,8 +54,10 @@ import com.zimbra.cs.account.AccessManager;
 import com.zimbra.cs.account.Account;
 import com.zimbra.cs.account.AccountServiceException;
 import com.zimbra.cs.account.AccountServiceException.AuthFailedServiceException;
+import com.zimbra.cs.account.accesscontrol.generated.RightConsts;
 import com.zimbra.cs.account.AuthToken;
 import com.zimbra.cs.account.GuestAccount;
+import com.zimbra.cs.account.Identity;
 import com.zimbra.cs.account.Provisioning;
 import com.zimbra.cs.mailbox.Mailbox;
 import com.zimbra.cs.mailbox.MailboxManager;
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/service/mail/SendMsg.java
```diff
@@ -136,6 +141,23 @@ public Element handle(Element request, Map<String, Object> context, QName respQn
...
+                */
+               if (identityId != null) {
+                   Identity identity = Provisioning.getInstance().get(octxt.getAuthenticatedUser(), Key.IdentityBy.id, identityId);
+                   if (identity != null &&
+                           !identity.getAttr(Provisioning.A_zimbraPrefIdentityName).equals(ProvisioningConstants.DEFAULT_IDENTITY_NAME) &&
+                           !StringUtil.isNullOrEmpty(identity.getAttr(Provisioning.A_zimbraPrefFromAddressType)) &&
+                           (identity.getAttr(Provisioning.A_zimbraPrefFromAddressType).equals(RightConsts.RT_sendAs) ||
+                                   identity.getAttr(Provisioning.A_zimbraPrefFromAddressType).equals(RightConsts.RT_sendOnBehalfOf))) {
+                       Account targetAccount = Provisioning.getInstance().get(AccountBy.name, identity.getAttr(Provisioning.A_zimbraPrefFromAddress));
+                       mbox = MailboxManager.getInstance().getMailboxByAccount(targetAccount);
```
desouzas: Wouldn't this cause issues down below (fetching draft etc) if the account isn't local to this mailbox?

jrjena136: resolved. instead of doing the same modified in `soapengine.java` to proxy the request automatically.

desouzas: Good call.

store/src/java/com/zimbra/soap/SoapEngine.java
```diff
@@ -327,6 +332,22 @@ private Element dispatch(String path, Element envelope, Map<String, Object> cont
...
+                if (identityId != null) {
+                    Identity identity = Provisioning.getInstance().get(DocumentHandler.getOperationContext(zsc, context)
+                            .getAuthenticatedUser(), Key.IdentityBy.id, identityId);
+                    if (identity != null &&
+                            !identity.getAttr(Provisioning.A_zimbraPrefIdentityName).equals(ProvisioningConstants.DEFAULT_IDENTITY_NAME) &&
+                            !StringUtil.isNullOrEmpty(identity.getAttr(Provisioning.A_zimbraPrefFromAddressType)) &&
+                            (identity.getAttr(Provisioning.A_zimbraPrefFromAddressType).equals(RightConsts.RT_sendAs) ||
+                                    identity.getAttr(Provisioning.A_zimbraPrefFromAddressType).equals(RightConsts.RT_sendOnBehalfOf))) {
+                        Account targetAccount = Provisioning.getInstance().get(AccountBy.name, identity.getAttr(Provisioning.A_zimbraPrefFromAddress));
+                        zsc.setmRequestedAccountId(targetAccount.getId());
```
desouzas: Maybe not a big deal but if `requestedAccountId` is already set to something else, this would override that.

jrjena136: Only if the above condition satisfied, then only overriding the value of `requestedAccountId`

jrjena136:
commit 0cade97e83f3ba0ed7e197c217cb65a82f018d78 (HEAD -> develop, origin/develop, origin/HEAD)
Date:   Thu Nov 14 15:26:50 2019 +0530

    zbug-903:Problem with sendAs delegate rights and Save a copy of sent messages to my Sent folder

    resolving CR comments

PR #987, opened 2019-11-28
ZCS-8246 Fixing NPE with CSRF check

rupalid:
When a SOAP request is sent with Expired auth token and CSRF token enabled. The CSRF token validation fails as auth token is set to null in ZSC when auth token is expired.

Verified that NPE is not thrown
Added unit test

1332 | 0 | 0 | 10 | 100.00% | 245.740

### DIFF
--- a/store/src/java-test/com/zimbra/cs/servlet/util/CsrfUtilTest.java
+++ b/store/src/java-test/com/zimbra/cs/servlet/util/CsrfUtilTest.java
@@ -184,7 +184,7 @@ public class CsrfUtilTest {
             assertEquals(false, validToken);
 
 
-        } catch (ServiceException  e) {
+        } catch (Exception  e) {
             fail("Should not throw exception.");
         }
     }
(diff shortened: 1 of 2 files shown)


store/src/java-test/com/zimbra/cs/servlet/util/CsrfUtilTest.java
```diff
@@ -171,6 +171,24 @@ public final void testIsValidCsrfTokenForAccountWithMultipleTokens() {
...
+                "test@zimbra.com");
+            AuthToken authToken = new ZimbraAuthToken(acct);
+
+            String csrfToken1 = CsrfUtil.generateCsrfToken(acct.getId(),
+                AUTH_TOKEN_EXPR, CSRFTOKEN_SALT, authToken);
+            boolean validToken = CsrfUtil.isValidCsrfToken(csrfToken1, null);
+            assertEquals(false, validToken);
+
+
+        } catch (ServiceException  e) {
```
sneha-patil-synacor: should be Exception instead of ServiceException

PR #992, opened 2019-12-04
ZCS-8018 Removing spring-asm , not required

rupalid:
Removing spring-asm as it is not required by apache cxf and newer version of spring


ozkanogun:
Hello i have a question about this pull request. This pull request seems really old, but spring-asm jar is still written inside of the "store/build.xml" of 8.8.15.p45 and later versions of 8.8.15. And also when i install 8.8.15.p45 spring-asm-3.0.7.RELEASE.jar comes inside of the /opt/zimbra/lib/jars and /opt/zimbra/jetty_base/webapps/service/WEB-INF/lib folders. Is spring-asm necesserray in 8.8.15 or is it forgotten there?

PR #996, opened 2019-12-09
ZCS-8014: Third party vulnerabilities in dom4j

nileshpawar1:
Issue: Few SOAP test cases were failing due to the dom4j library upgrade, as newly updated library added an exception.

Fix: Update the codebase to catch the exception.

### DIFF
--- a/common/src/java/com/zimbra/common/soap/Element.java
+++ b/common/src/java/com/zimbra/common/soap/Element.java
@@ -257,7 +257,13 @@ public abstract class Element implements Cloneable {
 
     public QName getQName() {
         String uri = getNamespaceURI(mPrefix);
-        return uri == null ? QName.get(mName) : QName.get(getQualifiedName(), uri);
+        QName returnValue = null;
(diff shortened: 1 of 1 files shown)


common/src/java/com/zimbra/common/soap/Element.java
```diff
@@ -257,7 +257,13 @@ public String getQualifiedName() {
 
     public QName getQName() {
         String uri = getNamespaceURI(mPrefix);
-        return uri == null ? QName.get(mName) : QName.get(getQualifiedName(), uri);
+        QName returnValue = null;
+        try {
+            returnValue = (uri == null) ? QName.get(mName) : QName.get(getQualifiedName(), uri);
+        } catch (Exception e) {
```
dasiyogesh: Can we catch specific exception that is thrown by dom4j?

nileshpawar1: Done

PR #1006, opened 2020-01-07
ZCS 8290: Upgrade TLS in Zimbra classic

nileshpawar1:
TLS1.0 and TLS1.1 support are going to expire in all major browsers, so we are making it off by default.

If users want to use it, they can on it by setting the values of some LDAP variables manually.


rupalid:
@nileshpawar1  please add testing notes

PR #1015, opened 2020-01-27
ZCS-8474: Add new Feature enable attributes for Zimbra 9

nileshpawar1:
Add attribute at Cos and account level
zimbraFeatureMobileAppEnabled
zimbraFeatureModernDesktopEnabled
 
This should be returned in the GetInfo Response license block

Testing Instructions:
Check the response of "GetInfoRequest" soap call, it should return values of above attributes in the license block

`zmsoap -z -m t1@`zmhostname` GetInfoRequest/ @sections="mbox,prefs,attrs,props,idents,sigs,dsrcs,children" -t account`

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9770,9 +9770,9 @@ TODO: delete them permanently from here
   <deprecateDesc>deprecated with attribute zimbraFeatureModernDesktopEnabled</deprecateDesc>
 </attr>
 
-<attr id="3081" name="zimbraFeatureModernMobileUIEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInherited" since="8.9.0">
+<attr id="3081" name="zimbraFeatureMobileAppEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInherited" since="8.9.0">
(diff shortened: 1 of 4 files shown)


store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9769,4 +9769,13 @@ TODO: delete them permanently from here
   <desc>Whether to allow a user to access Zimbra X desktop</desc>
 </attr>
 
+<attr id="3081" name="zimbraFeatureModernMobileUIEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInherited" since="8.9.0">
```
rupalid: check if the id is already used in Zimbra X,  deprecate the earlier zimbraFeatureZXDesktopEnabled

nileshpawar1: added deprecate for attribute zimbraFeatureZXDesktopEnabled

desouzas: 3081 is in use by the common password config. This may be backported eventually - the collision might be annoying.

https://github.com/ZimbraOS/zm-mailbox/pull/212/files#diff-3b031f41b9ac7ae803e3e7cf389d1733R9981

PR #1016, opened 2020-01-30
PREAPPS-4182: Change login options available to the user

madhavdhuppe:
-Replaced zimbrax with modern.


rupalid:
@madhavdhuppe  @silentsakky  when you make a change in zimbra-attrs.xml . Please also run ant generate-getters in store repo  and checkin the auto generated files. Otherwise when we add new attrs the changes from you are change are also added to out auto generated files.

PR #1017, opened 2020-01-30
ZBUG-1315: An issue with data type range values for Mysql database ta…

nileshpawar1:
Field 'mod_content' has a data type int (10) unsigned, so can accept values range from 0 to 4294967295 in db.

Problem: Currently when "mod_content" value goes beyond integer range (above 2147483647) then in java, we accessing that value using getInt() method which is not capable to hold unsigned int range

Fix: Used getLong() instead of getInt() as mentioned here - https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-reference-type-conversions.html
and did respective changes in the code base.

Testing done:
1. Manually reproduce the issue as per steps mentioned in the ticket and verified it is fixed.
2. Checked back and restore functionality using following steps
    1. Took full backup using the command   `zmbackup -a all -f`
    2. Restore the data from backup for few accounts and check the logs.
       `zmrestore -a t1@`zmhostname`
       `zmrestore -a admin@`zmhostname`


ajinkya81:
@nileshpawar1  please check if we also need to make similar change for mod_metadata field?

PR #1019, opened 2020-02-06
ZBUG-1311: Fix for zimbraSmtpHostname should honour domain level value (if set) instead of server level

ajinkya81:
- Updated code to use domain level value (if set) for zimbraSmtpHostname.
- Along with Read receipt and receiving notifcation, also fixed the other files listed in ticket description.

**Testing Done:**
Verified Read Receipt and Receiving Notification with domain level value set and empty both.
As per jira ticket comment OOO issue is not reproducible.

**Testing to be done by QA:**
Verify fixes done in other files as per ticket description.

### DIFF
--- a/store/src/java/com/zimbra/cs/account/ldap/AutoProvision.java
+++ b/store/src/java/com/zimbra/cs/account/ldap/AutoProvision.java
@@ -33,6 +33,7 @@ import javax.activation.DataHandler;
 import javax.activation.DataSource;
 import javax.mail.Address;
 import javax.mail.MessagingException;
+import javax.mail.Session;
 import javax.mail.Transport;
 import javax.mail.internet.AddressException;
 import javax.mail.internet.InternetAddress;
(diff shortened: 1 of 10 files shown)


store/src/java/com/zimbra/cs/account/ldap/AutoProvision.java
```diff
@@ -432,7 +433,8 @@ protected void sendNotifMessage(Account acct, String password)
         String toAddr = acct.getName();
 
         try {
-            SMTPMessage out = new SMTPMessage(JMSession.getSmtpSession());
+            Domain domain = Provisioning.getInstance().getDomain(acct);
```
rupalid: Can we make this into a single utility method...

ajinkya81: can you please provide me an example for this as reference?

chinmay15: Below 2 lines are repeated in multiple classes
```
Domain domain = Provisioning.getInstance().getDomain(owner);
SMTPMessage report = new SMTPMessage(JMSession.getSmtpSession(domain));
```
We should create a utility method which accept account as arg and return `SMTPMessage` object as output.

ajinkya81: ok, any suggestions on which class file I can add this method?

chinmay15: AccountUtil I think, because we are fetching details from Account object.

ajinkya81: Done above changes.

PR #1039, opened 2020-03-17
feature/NICPS-719

tarun-synacor:
New Zimbra SOAP API for fetching login details from elasticsearch

### DIFF
--- a/store/src/java/com/zimbra/cs/service/account/GetLoginDetails.java
+++ b/store/src/java/com/zimbra/cs/service/account/GetLoginDetails.java
@@ -1,0 +1,148 @@
+package com.zimbra.cs.service.account;
+import java.util.Map;
+
+import com.google.common.base.Strings;
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.soap.AccountConstants;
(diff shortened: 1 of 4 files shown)


Prashantsurana:
zimbra provides extension framework to handle such things. 
The code here will diverge from core. and will have recurring maintenance efforts. 
There are many extension there in Zimbra repo. You can refer those. 
https://blog.zimbra.com/2010/04/extending-zimbra-with-server-extensions/

store/src/java/com/zimbra/cs/service/account/GetLoginDetails.java
```diff
@@ -0,0 +1,148 @@
...
+		
+
+        Element response = getResponseElement(zc);
+        StringBuffer strBuff = new StringBuffer();
+       
+        response.addAttribute("ES", json_response);
+
+		return response;
+	}
+	static String body(String user_email, String from_time, String to_time,String attr) {
```
zimsuchitgupta: I think we should not do this hard coded. We should try to wrap this all within objects.

tarun-synacor:
Closing the pull request as implementation would be done with a different approach

PR #1040, opened 2020-03-18
ZCS-8785:Fix for issue with data type range values for Mysql

sneha-patil-synacor:
Issue:
mail_item table column 'mod_content' has a data type int (10) unsigned 
However, in code 'mod_content' field is using as int
This was causing out of range exception when fetching mod_content value if its greater than java integer range.

Fix:
Changing it to long.
Keeping old get getSavedSequence() method signature as it is to avoid code breaks in other clients.
Introduced new getSavedSequenceLong() method which will return long value of mod_content

Testing done:
Manual testing done by following steps in ZBUG-1315. Web client is loaded without any out of range exception.

Testing to be done by QA:
Test NG modules for ZCS-8696

### DIFF
--- a/store/src/java/com/zimbra/cs/mailbox/Folder.java
+++ b/store/src/java/com/zimbra/cs/mailbox/Folder.java
@@ -1690,8 +1690,8 @@ public class Folder extends MailItem implements FolderStore {
      * @see Folder#getSavedSequence()
      **/
     @Override
-    public int getUIDValidity() {
-        return Math.max(getSavedSequence(), 1);
+    public long getUIDValidity() {
+        return Math.max(getSavedSequenceLong(), 1);
     }
 
     @Override
--- a/store/src/java/com/zimbra/cs/mailbox/MailItem.java
+++ b/store/src/java/com/zimbra/cs/mailbox/MailItem.java
@@ -308,7 +308,7 @@ public abstract class MailItem implements Comparable<MailItem>, ScheduledTaskRes
         public String metadata;
         public int modMetadata;
         public int dateChanged; /* Seconds since 1970-01-01 00:00:00 UTC */
-        public int modContent;
+        public long modContent;
         public String uuid;
 
         public String getSubject() {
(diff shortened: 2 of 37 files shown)


store/src/java/com/zimbra/cs/mailbox/Folder.java
```diff
@@ -1691,7 +1691,7 @@ public boolean isIMAPSubscribed() {
      **/
     @Override
     public int getUIDValidity() {
-        return Math.max(getSavedSequence(), 1);
+        return (int)Math.max(getSavedSequenceLong(), 1);
```
desouzas: I guess this is no worse than the cast below it, but what is the expected behavior if the sequence is bigger than int max here?

store/src/java/com/zimbra/cs/mailbox/MailItem.java
```diff
@@ -986,6 +986,10 @@ public long getDate() {
      *  content was modified.  For immutable objects (e.g. received messages),
      *  this will be the same change ID as when the item was created. */
     public int getSavedSequence() {
+        return (int) mData.modContent;
```
desouzas: Same question here, what's the expected behavior on loss if modContent is bigger than int max? Any reason we're not changing the method signatures all the way through?

sneha-patil-synacor: Its being used by NG module. If we change the return type from int to long, their code breaks since long can not fit into int. Eventually, they should use the new getSavedSequenceLong() in their code. But, this is to avoid any immediate functionality errors.

store/src/java/com/zimbra/cs/mailbox/Folder.java
sneha-patil-synacor: This we can change with the hope that its not used anywhere else except internal IMAP code. changing into to long is no harm if its method argument, but if its return type, then it can break the things.

PR #1044, opened 2020-03-25
ZCS-8911 Allow extensions to support HTTP Patch method.

desouzas:
* Patch support is needed for oauth extension.


Prashantsurana:
@rupalid can you please merge and tag JIRA for the next patch.

PR #1058, opened 2020-04-16
ZBUG-1499 Reverting to convertd client

rupalid:
Reverted to using convertd extraction client.
Made the client class configurable

verified extraction of attachments works with convertd.
verified reindexing and search works


common/src/java/com/zimbra/common/localconfig/LC.java
```diff
@@ -1430,6 +1430,8 @@
     // account listener rollback handler on failure
     public static final KnownKey rollback_on_account_listener_failure = KnownKey.newKey(true);
 
+    public static final KnownKey attachment_extraction_client_class = KnownKey.newKey("com.zimbra.cs.convert.LegacyConverterClient");
```
Prashantsurana, line 1433: can you please add comment what is value of the key if one need to switch to tika?

PR #1060, opened 2020-04-22
ZBUG-1459:Abuse of File upload functionality in Zimbra webmail

sneha-patil-synacor:
Issue:
A user can upload malicious files in the Contact section of the mailbox as avatar image for a contact.

Fix: 
Added new LC parameter to have blacklisted content type during any file upload. The value can be regex value to blacklist certain category of files.
Default value is set to 'application/x-mx*' to blacklist all exe files.

Testing done:
Manual testing done. The exe file upload now throws 403 forbidden error.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/FileUploadServlet.java
+++ b/store/src/java/com/zimbra/cs/service/FileUploadServlet.java
@@ -21,6 +21,7 @@ import java.io.FileFilter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintWriter;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -32,6 +33,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.TimerTask;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import javax.mail.util.SharedByteArrayInputStream;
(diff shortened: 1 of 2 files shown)


jrjena136:
@sneha-patil-synacor please check if it works fine for pdf file as well. because recently there was some issue with tika library.

rupalid:
Sneha, Please test this with a variety of non executable files too, tika if does not have the parser will throw exception(ZBUG-1499). Verify that is not happening. Is there a way to avoid using Tika for this. Do we have things in core Java

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -712,17 +714,35 @@ public void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOEx
...
+        }
+        if (mimeType == null) {
+            mimeType = up.contentType;
+        }
+        String contentTypeBlacklist = LC.zimbra_file_content_type_blacklist.value();
+        List<String> blacklist = new ArrayList<String>();
+        if (!StringUtil.isNullOrEmpty(contentTypeBlacklist)) {
+            blacklist.addAll(Arrays.asList(contentTypeBlacklist.trim().split(",")));
+        }
+        for (String blacklistedContentType : blacklist) {
```
rupalid, line 736: we can simplify this using streams.
 if (blackist.stream().anyMatch(ctype -> ctype.equals(blacklistedContentType)) {
           
        }

sneha-patil-synacor:
@rupalid Java does have Files.probeContentType(), but it has some limitations. This method returns null if the extension is removed from file. Here, even if the tika throws exception, we are catching it and reading the content type from upload object, which populates the content type based on few criterias such as file extension, first few bytes of file and browser-specified content type.

sneha-patil-synacor:
@jrjena136 it works with pdf. Tika does throw exception, but it reads its from Upload object and proceeds.

sneha-patil-synacor:
Removing tika and using existing MimeDetect.getMimeDetect().detect() utility to detect file content type. Changed the order of content detection criteria in Upload class to detect the file type based magic rule first rather than extension.

rupalid:
Sneha add a unit test if possible

PR #1066, opened 2020-05-22
ZBUG-148 Fixing Tika extractor

rupalid:
Removed the Usage of Tika in core mailbox.
Reworked fix for ZBUG-711 and verified ZBUG-711 
Verified ZBUG-1499

Imported 1GB mailbox and reindexed, no issues seen

### DIFF
--- a/store/src/java/com/zimbra/cs/service/FileUploadServlet.java
+++ b/store/src/java/com/zimbra/cs/service/FileUploadServlet.java
@@ -21,7 +21,6 @@ import java.io.FileFilter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintWriter;
-import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -718,12 +717,11 @@ public void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOEx
...
         items.add(fi);
         Upload up = new Upload(acct.getId(), fi, filename);
-        
+
         if (filename.endsWith(".har")) {
             File file = ((DiskFileItem) fi).getStoreLocation();
             try {
-                Tika tika = new Tika();
-                String mimeType = tika.detect(file);
+                String mimeType = MimeDetect.getMimeDetect().detect(file);
```
sneha-patil-synacor, line 724: We would also need to verify ZBUG-711 with this change

rupalid: Yes I have and added it to testing notes

PR #1067, opened 2020-05-25
ZCS-6022 Fixing Exception handling

rupalid:
Fix exception seen in logs  when sorting emails by Unread in a multi node environment when sharer and sharee are on different mbox server.

This exception is seen when search response hit does not contain a 'u' attr.

Verified the fix by sorting by Unread/All in Message and Conversation view in multinode env.

### DIFF
--- a/store/src/java/com/zimbra/cs/index/ZimbraHit.java
+++ b/store/src/java/com/zimbra/cs/index/ZimbraHit.java
@@ -23,6 +23,7 @@ import com.google.common.base.MoreObjects;
 import com.zimbra.common.mailbox.MailItemType;
 import com.zimbra.common.mailbox.ZimbraQueryHit;
 import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.soap.Element;
 import com.zimbra.common.soap.MailConstants;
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/index/ZimbraHit.java
```diff
@@ -379,7 +380,21 @@ else if (result < 0)
      */
     public static int getReadStatus(ZimbraHit zh) throws ServiceException {
         if (zh instanceof ProxiedHit) {
-            return ((ProxiedHit) zh).getElement().getAttributeInt(MailConstants.A_UNREAD);
+            try {
+                return ((ProxiedHit) zh).getElement().getAttributeInt(MailConstants.A_UNREAD);
+            } catch (ServiceException e) {
+                // This is  message hit
+                Element msgHit = null;
+                try {
```
sneha-patil-synacor, line 388: can we move this try block itself in its parent try block at line 383 instead of handling this in catch block

rupalid: no because for some messages you do get u="0" only when we do not get u=0 we fallback to f=ua

PR #1068, opened 2020-05-25
zbug-903:Problem with 'send as' delegate rights and 'Save a copy of sent messages to my Sent folder

jrjena136:
Issue:
Draft is not getting deleted when message is send using persona account and draft folder is count is getting increased.

Fix:
- renamed method `isMessageSentUsingPersona` to `isMessageSentUsingOwnersPersona`.
- `AccountUtil.isMessageSentUsingOwnersPersona()` method was always returning false because `fromAddrPrefType` and `fromAddrPrefValue` value is always null as these are the properties set on identity. updated the method to get the values from identity.
- The above modification is required because in case of `sendAs and SOB` sendMsg request contains draft id which is always appended with auth user account id but in case of persona it is only the item id because of which it was not getting deleted.

Test:
Tested that all functionality is working fine and drafts are getting deleted after message is successfully sent.

### DIFF
--- a/store/src/java/com/zimbra/soap/ZimbraSoapContext.java
+++ b/store/src/java/com/zimbra/soap/ZimbraSoapContext.java
@@ -415,6 +415,30 @@ public final class ZimbraSoapContext {
             }
 
             mMountpointTraversed = eAccount.getAttributeBool(HeaderConstants.A_MOUNTPOINT, false);
+        } else if (body != null && requestName.equals(MailConstants.SEND_MSG_REQUEST)) {
+            // To handle SendMsgRequest sent using zmsoap command where header does not exists.
+            // Check if from Address is not equal to auth user email address then set `mRequestedAccountId`
+            // as the from address accountId.
+            if (mAuthToken == null) {
+                throw ServiceException.AUTH_REQUIRED();
+            }
+            Account authAccount = mAuthToken.getAccount();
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/soap/ZimbraSoapContext.java
```diff
@@ -415,6 +415,30 @@ else if (format.equals(HeaderConstants.TYPE_JAVASCRIPT))
...
+            if (mAuthToken == null) {
+                throw ServiceException.AUTH_REQUIRED();
+            }
+            Account authAccount = mAuthToken.getAccount();
+            String fromAddress = AccountUtil.extractFromAddress(body);
+            String identityId = body.getElement(MailConstants.E_MSG).getAttribute(MailConstants.A_IDENTITY_ID, null);
+            if (!StringUtil.isNullOrEmpty(fromAddress) && AccountUtil.isMessageSentUsingOwnersPersona(identityId, authAccount, mAuthToken)) {
+                Account account = prov.get(AccountBy.name, fromAddress, mAuthToken);
+                if (account == null) {
+                    if (!mAuthToken.isAdmin()) {
```
rupalid, line 431: Why the 2 different exceptions

jrjena136: I have kept it same as above

PR #1079, opened 2020-07-21
ZBUG-1605 Fixing NPE

rupalid:
While  replying  to  an email  if the original message  is included  as  an  attachment it has no name, hence the NPE.
Added  check for NPE. Verified  scenario for  replying, forwarding  and sending  attachment.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/formatter/ArchiveFormatter.java
+++ b/store/src/java/com/zimbra/cs/service/formatter/ArchiveFormatter.java
@@ -636,7 +636,7 @@ public abstract class ArchiveFormatter extends Formatter {
                             throw MailServiceException.NO_SUCH_PART(part);
                         }
                         name = Mime.getFilename(mp);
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/service/formatter/ArchiveFormatter.java
```diff
@@ -637,7 +636,7 @@ private ArchiveOutputStream saveItem(UserServletContext context, MailItem mi,
                             throw MailServiceException.NO_SUCH_PART(part);
                         }
                         name = Mime.getFilename(mp);
-                        if (!Normalizer.isNormalized(name, Normalizer.Form.NFC)) {
+                        if (name != null && !Normalizer.isNormalized(name, Normalizer.Form.NFC)) {
```
chinmay15, line 639: StringUtil.isNullOrEmpty() can be used here

PR #1080, opened 2020-07-22
ZCS-9569 Return  empty response for success  and dryrun=1

rupalid:
When dryrun is true and new password is valid and satisfy the rules return an empty response.
Tested with dryrun true and dryrun  false


store/src/java/com/zimbra/cs/service/account/ChangePassword.java
```diff
@@ -109,8 +109,10 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
         AuthToken at = AuthProvider.getAuthToken(acct);
```
Prashantsurana, line 109: Can we move this in if block this will avoid generating one auth token.

AuthToken at = AuthProvider.getAuthToken(acct);

PR #1081, opened 2020-07-27
ZCS-1666 1670

rupalid:
Refactored the code for handling  Persona  to be in SendMsg.
Removed previous changes

Also now when delegate owner account is inactive and the mail is being sent as a Persona of the delegate owner account, the mail will be saved to sent folder of delegated account.

verified  ZBUG-1660 ZBUG-1659 ZBUG-1665 ZBUG-1670 ZBUG-903 ZBUG-1209 ZBUG-1292 ZBUG-1334 ZCS-9483 ZCS-9464 and ZBUG-1666

Will add unit test


rupalid:
> Is this new constraint added ?
> before this, was it working even account is not active.

SendAS and SOBO was not not  working before too if owner account  is not active.  After we  added  fix for persona(ZBUG-903) it was working for Peronsa SendAs/SOBO only  and not  for SendAS and SOBO.

PR #1084, opened 2020-08-24
Bugfix/zbug 1694

rupalid:
Added code to split up processing of folders for IMAP LIST command into chunks.
Tested with  different chunk size and defaulted to 2000.
No errors seen on the backend

Observation: Thunderbird takes  a  long time  to process the received response.
Tried  with command line, verified all folders and subfolders are received and no  exception  in the logs

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/ImapHandler.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapHandler.java
@@ -52,6 +52,7 @@ import com.google.common.base.Charsets;
 import com.google.common.base.Joiner;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.zimbra.client.ZFolder;
 import com.zimbra.client.ZSharedFolder;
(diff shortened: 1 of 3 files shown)


jrjena136 (approved):

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -2229,41 +2233,84 @@ private boolean doLIST(String tag, String referenceName, Set<String> mailboxName
...
-            for (Object match : mountMatches.values()) {
-                if (match instanceof String[]) {
-                    for (String response : (String[]) match) {
-                        sendUntagged(response);
-                    }
-                } else {
-                    sendUntagged((String) match);
-                }
-            }
+        if (ownerSelected.size() > paginationSize) {
```
chinmay15, line 2250: can it be placed as else block for above if?

chinmay15 (approved):
LGTM

store/src/java/com/zimbra/cs/imap/ImapHandler.java
rupalid: The code sendUntagged was outside the try/catch block, so maintaining it.

PR #1089, opened 2020-09-04
ZBUG-1670: Sending email using owner persona does not delete draft

jrjena136:
**Issue:** 
- When sending email using persona in case of multi node setup does not delete draft and fills up the draft folder.
- This issue only occurs when users are on different server.

**Fix:**
- In case of multi node environment, when 2 user are on different server and lets say userA has given delegate right to userB and userB has created a persona for userA using userA email address, the request is getting proxied to the target server to execute the request and response is returned. As draft is always saved in logged in user's draft folder it was not able to delete the item.
- Instead of returning directly after proxying request, capture the request, delete the draft and then return the response.

**Test:**
- Tested on multi node environment that, sending mail using owner's persona is working fine and the drafts are getting deleted properly.
- set `zimbraAllowAnyFromAddress TRUE` on userB and create persona for any other account which is exist or not, sending mail is working fine with persona and drafts are getting deleted.
- create a persona using other user email address whose status is not active, sending mail is working fine and drafts are getting deleted.
- create a persona for logged in user account. sending email is working fine and drafts are getting deleted.


chinmay15 (approved):
LGTM

dasiyogesh (approved):

desouzas (approved):

PR #1092, opened 2020-09-15
zcs-9810:Fixing IMAP regression caused by ZBUG-1694

jrjena136:
Fixing ZBUG-1694 regression. As part of ZBUG-1694, a new LC was added to fix the pagination issue when there are huge number of folders present on an account. So the block 
```
if (ownerSelected.size() <= paginationSize) {
```
was causing the issue because when the size is less then pagination size server was not able to send the folder list which caused the regression [ZCS-9810](https://jira.corp.synacor.com/browse/ZCS-9810) and [ZCS-9820](https://jira.corp.synacor.com/browse/ZCS-9820).
- Done some code refactoring. There was multiple try-catch blocks. so removed those as the exception is already handled in single try block. Moved those codes into same try block
- There was high memory usage when connecting from multiple IMAP sessions. so externally nullifying all the objects so that those objects could be immediately garbage collected.
- Added the fix, so that in both the cases it will send the folder list to the client.

### DIFF
--- a/common/src/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/java/com/zimbra/common/localconfig/LC.java
@@ -1441,8 +1441,11 @@ public final class LC {
 
     // enable blocking common passwords
     public static final KnownKey zimbra_block_common_passwords_enabled = KnownKey.newKey(false);
- // enable blocking common passwords
+    // imap folder pagination size
     public static final KnownKey zimbra_imap_folder_pagination_size =  KnownKey.newKey(2000);
+
+    // imap folder pagination enabled
+    public static final KnownKey zimbra_imap_folder_pagination_enabled =  KnownKey.newKey(false);
 
     static {
         // Automatically set the key name with the variable name.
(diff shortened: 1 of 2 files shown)


dasiyogesh (approved):

chinmay15 (approved):

jrjena136:
merged to develop

*   3d8b0e0040 - (HEAD -> develop, origin/develop, origin/HEAD) Merge branch 'bugfix/zcs-9810' into develop (7 minutes ago) <Jyotiranjan>
|\
| * 2bc1da714d - zcs-9810: added new lc to enable/disable pagination (12 hours ago) <Jyotiranjan>
| * 0f3af40c49 - zcs-9810: code refactored (3 days ago) <Jyotiranjan>
| * f939f951b1 - fix for mem usage (4 days ago) <Jyotiranjan>
| * 7932821e39 - zcs-9810:Fixing IMAP regression caused by ZBUG-1694 (3 weeks ago) <Jyotiranjan>

PR #1099, opened 2020-10-22
ZCS-9842: Fixed when request has dryRun=true password should not be changed.

log2akshat:
**Problem:**
When the `ResetPassword` request has `dryRun=true` password is also getting changed.

**Fix:**
Fixed when the request has `dryRun=true` password is not getting changed now.

**Testing Done:**
- Tested when `dryRun` option is provided it satisfies the password rule when `COS` settings are changed and is **NOT** changing the password.
- Tested when the `dryRun` option is not provided then in that case only password is getting changed.

### DIFF
--- a/store/src/java-test/com/zimbra/cs/account/MockProvisioning.java
+++ b/store/src/java-test/com/zimbra/cs/account/MockProvisioning.java
@@ -953,9 +953,8 @@ public final class MockProvisioning extends Provisioning {
     }
 
 	@Override
-	public SetPasswordResult resetPassword(Account acct, String newPassword, boolean dryRun) throws ServiceException {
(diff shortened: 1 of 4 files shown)


jrjena136:
@log2akshat do we really need to change the existing methods. password would change or not, it should be dependent on the `dryRun` flag.

log2akshat:
@jrjena136 Yes, we need to change the method as this method `oapProvisioning.resetPassword` I have written in my last PR which is causing the problem and this method is getting used only when `dryRun` option is provided.

chinmay15 (approved):

madhavdhuppe (approved):

PR #1101, opened 2020-11-07
ImapRequest: Use standard Java date-time library

basilgello:
This closes https://bugzilla.zimbra.com/show_bug.cgi?id=109337

The recent update of Apple's Mail.app (an iOS default email client)
encodes the IMAP INTERNALDATE field ignoring the leading zero in
days:

    '6-Nov-2020 01:23:45 +0300'

instead of compliant:

    '06-Nov-2020 01:23:45 +0300'

This breaks the IMAP APPEND command and, in turn, mail.app fails to
send message via SMTP later.

Simplify the 'ImapRequest.readDate' using 'SimpleDateFormat'.

Signed-off-by: Vasyl Gello <vasek.gello@gmail.com>


basilgello:
@plobbes @silentsakky please look at the commit.

PR #1103, opened 2020-11-12
 ZBUG-1456 Connection pool shut down 3.0

chinmay15:
**Problem:** `InternalEventHandler.handleRefresh()` and `InternalEventHandler.handleModify()` stuck in loop calling each other resulting `StackOverflowError`.

**Approach and Fix:**
- This is side effect of fix for [ZBUG-527](https://jira.corp.synacor.com/browse/ZBUG-527) and was missed in [ZBUG-1080](https://jira.corp.synacor.com/browse/ZBUG-1080)
- Issue not reproducible locally or in lab.
- `ZMailbox.populateFolderCache()` method creates refresh event and calls `handleRefresh()` method on handlers since `mUserRoot` is not set and the loop starts.
- So set `mUserRoot` before calling `mailbox.getAccountId()`

**Testing to be done by QA:**
- Reproduce issue before fix, if possible. Steps to reproduce are given on ticket by Mukesh.
- Verify shared folder syncing on IMAP client.(sharee and sharer should be on different mailbox nodes)
- Make sure nothing is broken.

PR #1104, opened 2020-11-22
ZBUG-1456

desouzas:
**Problem**
* Notification cycling due to a combination of the following issues resulting in a stackoverflow on the client and/or potentially a connection pool shutdown from too many noop/getInfo requests.
* Early request auth failures when using ZMailbox without explicitly specified auth token at creation
* accountId is unset if not explicitly specified, and may remain unset due to:
  * invalid auth in early requests preventing the requests from retrieving the accountId for a period of time
  * notification cycling if many events are happening on the server, preventing the getAccountInfo invoke finally block from returning.

**Solution**
* Prevent notification cycle by disabling forced noop invoke when handling notifications. The other fixes mask the need for invokes during response context handling.
* Fix missing auth on early requests for non-delegated auth ZMailboxes by saving auth token before processing response context
* Fix missing accountId by batching a getInfo request into first soap request if necessary - saves an extra request and prevents receiving more notifications during a second request that will be handled out of order to the current set being handled.

**Testing**
* Reproduced issue prior to fix by triggering moderate amount of modify events on several threads (mark as read 50-100k messages, move 50-100k messages between folders, add 6 messages per second to inbox), while requesting a new mailbox along with all folders, etc on another. See ticket for details.
* Verified issue could not be reproduced in specified fashion after fix.

**Testing to be done by QA**
* Reproduce issue before fix. Steps to reproduce are given on ticket by Prashant.
* Verify issue cannot be reproduced after fix. Steps to reproduce are given on ticket by Prashant.
* Verify shared folder syncing on IMAP client. (sharee and sharer should be on different mailbox nodes)
* Test zmmailbox cli utility with direct auth (-m, -p flags), admin auth (-a, or -z flags) for impact. Use -d flag for soap request/response inspection.
* Test classic webclient in a variety of use-cases.
* Make sure nothing is broken.

See #1103.

Side note: the notification handling is/was problematic as mailbox#invoke handles them in `try { transport.invoke } finally { handleResponseContext ... }` fashion. Since they're processed on the stack, if `handleResponseContext` invokes and receives more notifications, the newer notifications would be handled before the previous stack layer finishes handling theirs. This probably needs a queue on the client side if invokes are eventually unavoidable again while handling notifications

### DIFF
--- a/client/src/java/com/zimbra/client/ZMailbox.java
+++ b/client/src/java/com/zimbra/client/ZMailbox.java
@@ -41,6 +41,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.TimeZone;
 import java.util.UUID;
+import java.util.function.Consumer;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -737,10 +738,7 @@ public class ZMailbox implements ToZJSONObject, MailboxStore {
         }
         if (options.getAuthToken() != null) {
             if (options.getAuthAuthToken()) {
-                mAuthResult = authByAuthToken(options);
-                initCsrfToken(mAuthResult.getCsrfToken());
-                initAuthToken(mAuthResult.getAuthToken());
-                initTrustedToken(mAuthResult.getTrustedToken());
+                authByAuthToken(options);
             } else {
                 initAuthToken(options.getAuthToken());
             }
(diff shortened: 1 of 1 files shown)


KenMGJ (approved):

Prashantsurana (approved):
LGTM

chinmay15 (approved):

PR #1108, opened 2020-12-21
bugfix/zcs-10142: Send if powerpaste service is enabled and running in GetInfo response

jrjena136:
Like zimbra-x it is required to identify whether powerpaste is installed and running on the server so that the feature would work on UI. We do not have LDAP attributes for zimbra9 as we have added for zimbra-x.

Added condition to check if powerpaste service is installed and running in the GetInfo Response.

This will be returned as part of GetInfo response
`<powerpasteEnabled>1</powerpasteEnabled>`
1 - for enabled
0 - disabled

### DIFF
--- a/store/src/java/com/zimbra/cs/service/account/GetInfo.java
+++ b/store/src/java/com/zimbra/cs/service/account/GetInfo.java
@@ -16,12 +16,10 @@
  */
 package com.zimbra.cs.service.account;
 
+import java.io.File;
 import java.io.IOException;
 import java.net.URL;
-import java.nio.file.Files;
 import java.nio.file.InvalidPathException;
-import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.HashSet;
@@ -35,7 +33,6 @@ import com.google.common.collect.Sets;
 import com.zimbra.common.account.Key;
 import com.zimbra.common.account.Key.AccountBy;
 import com.zimbra.common.account.ProvisioningConstants;
-import com.zimbra.common.localconfig.KnownKey;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.soap.AccountConstants;
 import com.zimbra.common.soap.Element;
@@ -266,9 +263,9 @@ public class GetInfo extends AccountDocumentHandler  {
         // so check if powerpaste service is installed and running by checking the installed directory and connectivity
         // if yes return powerpasteEnabled = true else return powerpasteEnabled = false
         if (checkIfPowerpasteInstalled()) {
-            response.addAttribute("powerpasteEnabled", true);
+            response.addAttribute("powerpasteEnabled", true, Element.Disposition.CONTENT);
         } else {
-            response.addAttribute("powerpasteEnabled", false);
+            response.addAttribute("powerpasteEnabled", false, Element.Disposition.CONTENT);
         }
 
         return response;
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/service/account/GetInfo.java
```diff
@@ -461,4 +475,26 @@ protected Element encodeChildAccount(Element parent, Account child, boolean isVi
...
     }
+
+    private boolean checkIfPowerpasteInstalled() {
+        String libLocation = "/opt/zimbra/common/lib/pasteitcleaned";
+        String extLoc = "/opt/zimbra/lib/ext/powerpaste-ext/zm-powerpaste-extension.jar";
+        try {
+            File lib = new File(libLocation);
+            File ext = new File(extLoc);
+            if (lib.exists() && lib.isDirectory() && lib.canRead()) {
+                if (ext.exists() && ext.isFile() && ext.canRead()) {
```
dasiyogesh, line 486: Can we do this check by checking the handler registered like its done in ExtensionDispatcherServlet?

jrjena136: That can not be done here. and also until we  restart mailbox the extension would still present in the list. so I think it is better to check in that way.

store/src/java/com/zimbra/cs/service/account/GetInfo.java
```diff
@@ -461,4 +475,26 @@ protected Element encodeChildAccount(Element parent, Account child, boolean isVi
...
+                    URL url = new URL("http://localhost:5000");
+                    if (url.openConnection() != null) {
```
chinmay15, line 488: As per api doc, https://docs.oracle.com/javase/7/docs/api/java/net/URL.html#openConnection(),
```
A new instance of URLConnection is created every time when invoking the URLStreamHandler.openConnection(URL) method of the protocol handler for this URL.

It should be noted that a URLConnection instance does not establish the actual network connection on creation. This will happen only when calling URLConnection.connect().
```
I think connect method needs to be called on url connection to check whether the connection is available or not. https://docs.oracle.com/javase/7/docs/api/java/net/URLConnection.html#connect()

chinmay15 (changes requested):

store/src/java/com/zimbra/cs/service/account/GetInfo.java
jrjena136: Will fix this

store/src/java/com/zimbra/cs/service/account/GetInfo.java
```diff
@@ -254,6 +261,16 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
         for (GetInfoExt extension : extensions) {
             extension.handle(zsc, response);
         }
+
+        // we do not have any ldap attrs to define whether powerpaste service is installed and running
+        // so check if powerpaste service is installed and running by checking the installed directory and connectivity
+        // if yes return powerpasteEnabled = true else return powerpasteEnabled = false
+        if (checkIfPowerpasteInstalled()) {
```
sreejan-chowdhury, line 265: Can we write the following block in a single line?
like : `response.addAttribute("powerpasteEnabled", checkIfPowerpasteInstalled());`

jrjena136: can be done

chinmay15 (approved):
LGTM

dasiyogesh (approved):

sreejan-chowdhury (approved):
LGTM

PR #1128, opened 2021-03-19
TSS-18004: Fix for PTD - Message not displaying correctly

log2akshat:
**Problem:** 
PTD - Message not displaying correctly. [[TSS-18004](https://jira.corp.synacor.com/browse/TSS-18004)]

**Analysis:**
It was reported that some of the messages were not getting displayed correctly. On further analysis, it was found that
the particular attached message was having a lot of malformed inline styling in the original message body which contains a
lot of unclosed double quotes which was escaping large chunks of the real message. The mailer which the customer has used seems like `sendgrid` which has malformed the html.

Looking through the Antisamy policy which we are using turned out that it cannot be fixed through that the reason is that the
`antisamy` policy which we are using with `owasp` can able to handle various operations like validate, truncate, filter, remove based on the element types, tag names, tag values, attributes, and the `css` rules but not with the malformed `qoutes`.

Looking inside the `owasp` java sanitizer code turns out that the `owasp` uses `HtmlLexer` for the sanitization process which traverses element by element so before sanitizing the html it takes the value of the attribute and strips of any vulnerable kinds of stuff based on what policies we have applied. But in our case, the problem is not that `OWASP` html filter is strict with the quotes rather it is skipping the texts because it is considering the text as the attribute name because of the malformed double quotes which were present inside the span tag and it thinks of it as the new attribute has started and intercepts whatever is there as its value before the next double-quotes.

For example in case of the following html:
```
<span style="font-size:11pt;font-family:Arial;font-variant-ligatures:normal;font-variant-east-asian:normal;font-variant-position:normal;vertical-align:baseline"">Dear Member,</span></p><br style="-webkit-text-size-adjust:auto;" />
```
The attribute style is properly interpreted then the next malformed double  quote it has interpreted as the `!attrsReadyForName` and it has interpreted the key as: `">Dear Member,</span></p><br style="` and detected its value as `-webkit-text-size-adjust:auto;"` and since the value parts look vulnerable to `owasp` so it has quietly dropped.

And as per comments from the `owasp` community _`Unfortunately we cannot throw compilation errors for invalid HTML markup.`_

On analyzing the already reported issues of `antisamy` and `owasp` it is found that the user has to handle the invalid HTML markup.

**Fix:**
Looked at some libraries that can fix the malformed html for that tried using `jsoup` which is used for parsing, cleaning, extract, manipulate and make the output tidy. Used its `Cleaner` class and got the expected results. Later it is realized that `jsoup` is also capable of doing the sanitization so to remove any future collision between the two `jsoup` is dropped.

Finally, used the java port of the old `HTML Tidy` which is originally written in C and is a command-line utility that is widely used for cleaning the malformed markup. So, used the java port `JTidy` which is updated to work with HTML5 new tags. Used its `Tidy` class with other options for cleaning the malformed html. The advantage of using this library is that it only fixes the markup but does not do any other process of sanitization or manipulation as `jsoup` does. The cleaned
html is now passed to the `sanitize()` method for further sanitization process using `antisamy` and `owasp`.

Also, added debug config for enabling and disabling the `JTidy` library.

**Testing Done:**
- Injected the malformed html mime, attached in the ticket and verified it is displaying properly.

### DIFF
--- a/store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
+++ b/store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
@@ -16,6 +16,9 @@
  */
 package com.zimbra.cs.html.owasp;
 
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.UnsupportedEncodingException;
 import java.util.concurrent.Callable;
 
 import org.owasp.html.Handler;
@@ -23,6 +26,8 @@ import org.owasp.html.HtmlSanitizer;
 import org.owasp.html.HtmlSanitizer.Policy;
 import org.owasp.html.HtmlStreamRenderer;
 import org.owasp.html.PolicyFactory;
+import org.w3c.tidy.Tidy;
+
 import com.zimbra.common.util.StringUtil;
 import com.zimbra.cs.html.owasp.policies.StyleTagReceiver;
 
@@ -53,8 +58,9 @@ public class OwaspHtmlSanitizer implements Callable<String> {
      * 
      * @return the sanitized form of the <code>html</code>; <code>null</code> if
      *         <code>html</code> was <code>null</code>
+     * @throws UnsupportedEncodingException
      */
-    public String sanitize() {
+    public String sanitize() throws UnsupportedEncodingException {
         OwaspThreadLocal threadLocalInstance = new OwaspThreadLocal();
         threadLocalInstance.setVHost(vHost);
         OwaspHtmlSanitizer.zThreadLocal.set(threadLocalInstance);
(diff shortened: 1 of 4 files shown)


chinmay15 (approved):
LGTM

store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
```diff
@@ -77,12 +83,27 @@ public void handle(final String x) {
...
+        tidy.setOutputEncoding("UTF-8");
+        tidy.setPrintBodyOnly(true);
+        tidy.setXHTML(true);
+        ByteArrayInputStream inputStream = new ByteArrayInputStream(str.getBytes("UTF-8"));
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        tidy.parseDOM(inputStream, outputStream);
+        if (outputStream.toString("UTF-8").isEmpty()) {
+            return str;
+        }
+        return outputStream.toString("UTF-8");
```
Prashantsurana, line 104: Did you test multibyte non-utf8 characters?

log2akshat: Will check and update.

log2akshat: **Testing with multibyte non-utf8 characters e-mails:**
EUC-JP

Shift_JIS

-------------

**Malformed multibyte non-utf8 characters e-mails:**
**Without using JTidy:**
Multi-Byte_Malformed_Japanese_Without_JTidy

-------------

**Using JTidy:**
Multi-Byte_Malformed_Japanese_With_JTidy

-------------

**Malformed JIS e-mails:**
Malformed JIS e-mails

PR #1129, opened 2021-03-20
ZBUG-1967 Problem synchronizing shared calendar via caldav when a '+' character is present on the EVENT

chinmay15:
**Problem:** when  a '+' character is present in UID in shared appointment, it breaks caldav syncing.

**Fix and Approach:**
- when '+' character is present in uid, the url created to get ics gets broken.
- url is decoded after getting proxy response and before sending it to client.
- removed code which was decoding href after getting response from proxy request
- added encoding of '@' to fix bugzilla bug#106438

**Testing done:**
- Tested on mac calendar client and thunderbird, both do sync all appointments in shared calendar including appointment with '+' character in UID and `+`, `@` in calendar name.

**Testing to be done by QA:**
- verify given scenario
- verify nothing is broken

### DIFF
--- a/store/src/java/com/zimbra/cs/dav/service/DavServlet.java
+++ b/store/src/java/com/zimbra/cs/dav/service/DavServlet.java
@@ -20,7 +20,7 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.net.URLDecoder;
+import java.net.URLEncoder;
 import java.util.Collection;
 import java.util.Enumeration;
 import java.util.HashMap;
@@ -52,6 +52,7 @@ import com.zimbra.client.ZMailbox;
 import com.zimbra.common.account.Key;
 import com.zimbra.common.account.Key.AccountBy;
 import com.zimbra.common.httpclient.HttpClientUtil;
+import com.zimbra.common.mime.MimeConstants;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.soap.W3cDomUtil;
 import com.zimbra.common.soap.XmlParseException;
@@ -818,8 +819,8 @@ public class DavServlet extends ZimbraServlet {
                             }
                             Element href = ((Element)responseObj).element(DavElements.E_HREF);
                             String v = href.getText();
-                            v = URLDecoder.decode(v);
                             // Bug:106438, because v contains URL encoded value(%40) for '@' the comparison fails
+                            newPrefix = newPrefix.replace("@", URLEncoder.encode("@", MimeConstants.P_CHARSET_UTF8));
                             if (v.startsWith(newPrefix)) {
                                 href.setText(prefix + v.substring(newPrefix.length()+1));
                             }


store/src/java/com/zimbra/cs/dav/service/method/CalendarMultiget.java
```diff
@@ -48,9 +51,24 @@ public void handle(DavContext ctxt) throws ServiceException, DavException {
...
-				hrefs.add(((Element)obj).getText());
+        for (Object obj : query.elements(DavElements.E_HREF)) {
+            if (obj instanceof Element) {
+                String text = ((Element)obj).getText();
+                // just check for '+' and encode
+                // encoding already encoded string will create issues
+                if (text.contains("+")) {
+                    try {
+                        ZimbraLog.dav.debug("going to encode '+' characters in href text=%s", text);
+                        text = text.replace("+", URLEncoder.encode("+", MimeConstants.P_CHARSET_UTF8));
```
Prashantsurana, line 62: why do not URL encode the complete text ? 

Remove + sign in all logger statement use  %s instead.

Prashantsurana (changes requested):

store/src/java/com/zimbra/cs/dav/service/method/CalendarMultiget.java
chinmay15:
```
// just check for '+' and encode
// encoding already encoded string will create issues
```
encoding already encoded string will create issues. please see below example

e.g. 
`+` gets encoded as `%2B`
but if string already contains `%2B`, it will get encoded to `%252B`


all the new added logs are having `%s`

log2akshat (approved):

dasiyogesh (approved):

Prashantsurana:
As discussed please try encoding the uid while sending to the client.

store/src/java/com/zimbra/cs/dav/service/DavServlet.java
```diff
@@ -818,8 +819,8 @@ private boolean isProxyRequest(DavContext ctxt, DavMethod m) throws IOException,
                             }
                             Element href = ((Element)responseObj).element(DavElements.E_HREF);
                             String v = href.getText();
-                            v = URLDecoder.decode(v);
                             // Bug:106438, because v contains URL encoded value(%40) for '@' the comparison fails
+                            newPrefix = newPrefix.replace("@", URLEncoder.encode("@", MimeConstants.P_CHARSET_UTF8));
```
Prashantsurana, line 823: replace or replaceall ?

chinmay15: as per discussion, replace is okay

Prashantsurana: Thanks

Prashantsurana (approved):

thefoolwhocodes (approved):

chinmay15:
merged in develop

commit 4058af2249093abffe05ca4afd9346dad1c69f06 (HEAD -> develop, origin/develop, origin/HEAD)
Merge: c7a5377e3f a6100e289a
Date:   Tue Mar 23 16:20:55 2021 +0530

    Merge branch 'bugfix/zbug-1967' into develop

PR #1130, opened 2021-03-23
ZCS-10383: Integrating zulip with zimbra 9

dasiyogesh:
Integrating Zulip with Zimbra9. 
Porting the required code from ZimbraX to Zimbra 9 to get zulip working with Z9.

### DIFF
--- a/common/src/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/java/com/zimbra/common/localconfig/LC.java
@@ -1372,7 +1372,7 @@ public final class LC {
 
     public static final KnownKey conversation_ignore_maillist_prefix = KnownKey.newKey(true);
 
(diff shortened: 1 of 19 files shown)


common/src/java/com/zimbra/common/localconfig/LC.java
```diff
@@ -1372,6 +1372,11 @@
 
     public static final KnownKey conversation_ignore_maillist_prefix = KnownKey.newKey(true);
 
+    // base host for zulip
+    public static final KnownKey zulip_base_host = KnownKey.newKey("https://%s.zimbracloud.com");
```
silentsakky, line 1376: we should call it zimbra.com instead of zimbracloud.com

PR #1132, opened 2021-03-25
ZCS-10322: LDAP Attribute for supporting Attachment Type Restrictions

log2akshat:
**Problem:**
LDAP Attribute for supporting Attachment Type Restrictions.

**Approach for Implementation:**
Added new ldap attributes at account and cos levels.
`zimbraFileUploadBlockedFileTypes` - Commonly blocked file types for uploading.
`zimbraFeatureFileTypeUploadRestrictionsEnabled` - Enable/Disable blocked file types set in `zimbraFileUploadBlockedFileTypes` for uploading.

Added ldap attribute `zimbraFeatureFileTypeUploadRestrictionsEnabled` for enabling and disabling the blocked file types set in the `zimbraFileUploadBlockedFileTypes` for uploading. If this ldap attribute is disabled then the atachment is uploaded without scanning the ldap attribute `zimbraFileUploadBlockedFileTypes` and the any type of the attachments are allowed to be uploaded.

If the ldap attribute `zimbraFeatureFileTypeUploadRestrictionsEnabled` is enabled then the file extensions mentioned in the ldap attribute are scanned one by one and if any of the matching blocked extensions are found during attachement upload then `BLOCKED_FILE_TYPE_UPLOAD` service exception is thrown and the attachment is disallowed from uploading. And suppose in case while uploading the attachment the extension is not found in the blocked ldap attribute `zimbraFileUploadBlockedFileTypes` then after that it's `content-type` is detected using the `tika-core` library and based on it's content type the file extension is detected and matched from the already fetched list of the blocked extension and if a match is found then `BLOCKED_FILE_TYPE_UPLOAD` service exception is thrown and the attachment is disallowed from uploading. 

This is also helpful in the situation when the user tries to upload an attachment after removing the extension.

**Updates:**
- Removed alreday blacklisted `application/x-ms*`from `zmlocalconfig` and fixed the issue of `.har` file content-type was not able to get detected from the mime type, used tika to get the content-type of the har file.
- Added debug statements for detected content-types and file extension.
- Added logic for handling the content-types specified in the `zimbraFileUploadBlockedFileTypes` ldap attribute.
- Added `custom-mimetypes.xml` in `store-conf` which will be copied over to the `/opt/zimbra/conf` directory where we can be able override the existing mimetypes and can be able to map the `content-type` to any of the extensions with a higher magic priority. So, in this change we are using `compositeDetector` which comprises of the default detector as well as the custom detector. If the `custom-mimetype.xml` is missing then in that case the `mediaType` will fallback to the default detector and will use the default `tika-mimetypes.xml` file.
- Removed the older way of `contentType` detection using the `MimeDetect` class and removed the unnecessary code for handling the `.har` file as it was based upon only the extension. As, if we remove the extension of the `.har` file [ZBUG-711](https://jira.corp.synacor.com/browse/ZBUG-711) still exist and making the web-client unresponsive and also in CPU spike. While it is correctly handled by the `tika` with it's `content-type` detected as `application/pdf` contrary to earlier where it is getting detected as `application/octet-stream` but the file actually contains `html` content and the `convertd` is trying to convert the document based on `application/octet-stream` and failing.
- 30-April-2021: Bumped the version of newly introduced `LDAP` attributes zimbraFeatureFileTypeUploadRestrictionsEnabled` and `zimbraFileUploadBlockedFileTypes` from `9.0.0` to `9.1.0`.

**Testing Done:**
- Tested when the ldap attribute `zimbraFeatureFileTypeUploadRestrictionsEnabled` is set to `FALSE` and verified that  any type of the attachments are allowed to be uploaded.
- Tested when the ldap attribute `zimbraFeatureFileTypeUploadRestrictionsEnabled` is set to `TRUE` and verified that the  file extensions mentioned it are blocked during the attachment upload and throws `BLOCKED_FILE_TYPE_UPLOAD` service exception.
- Verified that on removing the file extension from the blocked file type user is not able to upload the attachment.
- Verified by changing the extension of an `exe` file to `pdf` and and verified that the attachment upload throws `BLOCKED_FILE_TYPE_UPLOAD` service exception and is disallowed from uploading.
- Verified all the above scenarios for the uploads to the `e-mail` as well as for the `briefcase`.
- Verified no content types based on `application/x-ms*` are getting uploaded with or without the extensions.
- Verified `custom-mimetypes.xml` content-types are getting override to the default mime types.
- Verified that on uploading `.har` file with or without extension is not making the web-client unresponsive and is able to properly index and search the contents of the `.har` file.

**Related PR:**
[zm-zcs-lib](https://github.com/Zimbra/zm-zcs-lib/pull/72)
[zm-build](https://github.com/Zimbra/zm-build/pull/180)

### DIFF
--- a/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
+++ b/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
@@ -6495,6 +6495,15 @@ public class ZAttrProvisioning {
     public static final String A_zimbraFeatureExternalFeedbackEnabled = "zimbraFeatureExternalFeedbackEnabled";
 
     /**
+     * Enable/Disable blocked file types set in
+     * zimbraFileUploadBlockedFileTypes for uploading
+     *
+     * @since ZCS 9.0.0
+     */
+    @ZAttr(id=3092)
+    public static final String A_zimbraFeatureFileTypeUploadRestrictionsEnabled = "zimbraFeatureFileTypeUploadRestrictionsEnabled";
+
+    /**
      * filter prefs enabled
      */
     @ZAttr(id=143)
@@ -7193,6 +7202,14 @@ public class ZAttrProvisioning {
      */
     @ZAttr(id=1362)
     public static final String A_zimbraFileShareLifetime = "zimbraFileShareLifetime";
+
+    /**
+     * Commonly blocked file types for uploading
+     *
+     * @since ZCS 9.0.0
+     */
+    @ZAttr(id=3091)
+    public static final String A_zimbraFileUploadBlockedFileTypes = "zimbraFileUploadBlockedFileTypes";
 
     /**
      * Maximum size in bytes for file uploads
--- a/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
+++ b/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
@@ -6498,7 +6498,7 @@ public class ZAttrProvisioning {
      * Enable/Disable blocked file types set in
      * zimbraFileUploadBlockedFileTypes for uploading
      *
-     * @since ZCS 9.0.0
+     * @since ZCS 9.1.0
      */
     @ZAttr(id=3092)
     public static final String A_zimbraFeatureFileTypeUploadRestrictionsEnabled = "zimbraFeatureFileTypeUploadRestrictionsEnabled";
@@ -7206,7 +7206,7 @@ public class ZAttrProvisioning {
     /**
      * Commonly blocked file types for uploading
      *
-     * @since ZCS 9.0.0
+     * @since ZCS 9.1.0
      */
     @ZAttr(id=3091)
     public static final String A_zimbraFileUploadBlockedFileTypes = "zimbraFileUploadBlockedFileTypes";
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9850,6 +9850,7 @@ TODO: delete them permanently from here
   <defaultCOSValue>wsf</defaultCOSValue>
   <defaultCOSValue>wsh</defaultCOSValue>
   <defaultCOSValue>xl</defaultCOSValue>
+  <defaultCOSValue>application/x-ms*</defaultCOSValue>
   <desc>Commonly blocked file types for uploading</desc>
  </attr>
 <attr id="3092" name="zimbraFeatureFileTypeUploadRestrictionsEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" since="9.0.0">
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9851,6 +9851,7 @@ TODO: delete them permanently from here
   <defaultCOSValue>wsh</defaultCOSValue>
   <defaultCOSValue>xl</defaultCOSValue>
   <defaultCOSValue>application/x-ms*</defaultCOSValue>
+  <defaultCOSValue>application/x-dosexec</defaultCOSValue>
   <desc>Commonly blocked file types for uploading</desc>
  </attr>
 <attr id="3092" name="zimbraFeatureFileTypeUploadRestrictionsEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" since="9.0.0">
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9822,4 +9822,38 @@ TODO: delete them permanently from here
   <defaultCOSValue>TRUE</defaultCOSValue>
   <desc>whether to use Web Client feature</desc>
 </attr>
+<attr id="3091" name="zimbraFileUploadBlockedFileTypes" type="string" cardinality="multi" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" since="9.0.0">
+  <defaultCOSValue>asd</defaultCOSValue>
+  <defaultCOSValue>bat</defaultCOSValue>
+  <defaultCOSValue>chm</defaultCOSValue>
+  <defaultCOSValue>cmd</defaultCOSValue>
+  <defaultCOSValue>com</defaultCOSValue>
+  <defaultCOSValue>dll</defaultCOSValue>
+  <defaultCOSValue>do</defaultCOSValue>
+  <defaultCOSValue>exe</defaultCOSValue>
+  <defaultCOSValue>hlp</defaultCOSValue>
+  <defaultCOSValue>hta</defaultCOSValue>
+  <defaultCOSValue>js</defaultCOSValue>
+  <defaultCOSValue>jse</defaultCOSValue>
+  <defaultCOSValue>lnk</defaultCOSValue>
+  <defaultCOSValue>ocx</defaultCOSValue>
+  <defaultCOSValue>pif</defaultCOSValue>
+  <defaultCOSValue>reg</defaultCOSValue>
+  <defaultCOSValue>scr</defaultCOSValue>
+  <defaultCOSValue>shb</defaultCOSValue>
+  <defaultCOSValue>shm</defaultCOSValue>
+  <defaultCOSValue>shs</defaultCOSValue>
+  <defaultCOSValue>vbe</defaultCOSValue>
+  <defaultCOSValue>vbs</defaultCOSValue>
+  <defaultCOSValue>vbx</defaultCOSValue>
+  <defaultCOSValue>vxd</defaultCOSValue>
+  <defaultCOSValue>wsf</defaultCOSValue>
+  <defaultCOSValue>wsh</defaultCOSValue>
+  <defaultCOSValue>xl</defaultCOSValue>
+  <desc>Commonly blocked file types for uploading</desc>
+ </attr>
+<attr id="3092" name="zimbraFeatureFileTypeUploadRestrictionsEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" since="9.0.0">
+  <defaultCOSValue>TRUE</defaultCOSValue>
+  <desc>Enable/Disable blocked file types set in zimbraFileUploadBlockedFileTypes for uploading</desc>
+ </attr>
 </attrs>
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9822,7 +9822,7 @@ TODO: delete them permanently from here
   <defaultCOSValue>TRUE</defaultCOSValue>
   <desc>whether to use Web Client feature</desc>
 </attr>
-<attr id="3091" name="zimbraFileUploadBlockedFileTypes" type="string" cardinality="multi" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" since="9.0.0">
+<attr id="3091" name="zimbraFileUploadBlockedFileTypes" type="string" cardinality="multi" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" since="9.1.0">
   <defaultCOSValue>asd</defaultCOSValue>
   <defaultCOSValue>bat</defaultCOSValue>
   <defaultCOSValue>chm</defaultCOSValue>
@@ -9854,7 +9854,7 @@ TODO: delete them permanently from here
   <defaultCOSValue>application/x-dosexec</defaultCOSValue>
   <desc>Commonly blocked file types for uploading</desc>
  </attr>
-<attr id="3092" name="zimbraFeatureFileTypeUploadRestrictionsEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" since="9.0.0">
+<attr id="3092" name="zimbraFeatureFileTypeUploadRestrictionsEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" since="9.1.0">
   <defaultCOSValue>TRUE</defaultCOSValue>
   <desc>Enable/Disable blocked file types set in zimbraFileUploadBlockedFileTypes for uploading</desc>
  </attr>
--- a/store/src/java/com/zimbra/cs/service/FileUploadServlet.java
+++ b/store/src/java/com/zimbra/cs/service/FileUploadServlet.java
@@ -47,6 +47,7 @@ import org.apache.commons.fileupload.FileUploadException;
 import org.apache.commons.fileupload.disk.DiskFileItem;
 import org.apache.commons.fileupload.disk.DiskFileItemFactory;
 import org.apache.commons.fileupload.servlet.ServletFileUpload;
+import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.http.Header;
 import org.apache.http.HttpException;
@@ -55,6 +56,13 @@ import org.apache.http.HttpStatus;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.impl.client.HttpClientBuilder;
+import org.apache.tika.config.TikaConfig;
+import org.apache.tika.detect.Detector;
+import org.apache.tika.exception.TikaException;
+import org.apache.tika.io.TikaInputStream;
+import org.apache.tika.metadata.Metadata;
+import org.apache.tika.mime.MediaType;
+import org.apache.tika.mime.MimeType;
 
 import com.google.common.base.Strings;
 import com.zimbra.client.ZMailbox;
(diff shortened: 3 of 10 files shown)


common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
```diff
@@ -6494,6 +6494,15 @@ public static TwoFactorAuthSecretEncoding fromString(String s) throws ServiceExc
...
     public static final String A_zimbraFeatureExternalFeedbackEnabled = "zimbraFeatureExternalFeedbackEnabled";
 
+    /**
+     * Enable/Disable blocked file types set in
+     * zimbraFileUploadBlockedFileTypes for uploading
+     *
+     * @since ZCS 9.0.0
+     */
+    @ZAttr(id=3092)
+    public static final String A_zimbraFeatureFileTypeUploadRestrictionsEnabled = "zimbraFeatureFileTypeUploadRestrictionsEnabled";
```
silentsakky, line 6504: are we planning to release this change in kepler patches? ldap changes are not possible to be released with patch system

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -180,6 +188,27 @@
...
+            if (acct.getCOS().isFeatureFileTypeUploadRestrictionsEnabled()) {
+                blockedFileTypes = acct.getCOS().getMultiAttr(Provisioning.A_zimbraFileUploadBlockedFileTypes);
+                for(String blockedExt : blockedFileTypes){
+                    if(blockedExt.equalsIgnoreCase(extension)) {
+                        throw ServiceException.BLOCKED_FILE_TYPE_UPLOAD(
+                                String.format("Blocked attachment during uploading %s filetype ", blockedExt), null);
+                    }
+                }
+
+                List<String> blockedExtensionList = new ArrayList<>(Arrays.asList(blockedFileTypes));
```
Prashantsurana, line 205: you are creating this List so that you can use blockedExtensionList.contains instead of for loop. 
You are doubling the heap use.

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -188,6 +217,25 @@
         public long getSize()           { return file == null ? 0 : file.getSize(); }
         public BlobInputStream getBlobInputStream()        { return blobInputStream; }
 
+        public String getExtension(FileItem filename) {
+            String fileExtension = null;
+            try {
+                TikaConfig tikaConfig = TikaConfig.getDefaultConfig();
```
Prashantsurana, line 223: please see https://github.com/Zimbra/zm-mailbox/commit/99af5db1c47aa25a043c1f152744f6235fd86074 
Make sure we test address all regressions around Tika.

Prashantsurana (changes requested):

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -180,6 +188,27 @@
                 if (contentType == null)
                     contentType = MimeConstants.CT_APPLICATION_OCTET_STREAM;
             }
+
+            Account acct = Provisioning.getInstance().getAccount(acctId);
+            String extension = FilenameUtils.getExtension(filename).trim();
+            String [] blockedFileTypes = null;
+
+            if (acct.getCOS().isFeatureFileTypeUploadRestrictionsEnabled()) {
```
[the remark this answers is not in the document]
chinmay15: account inheritance is in place so `.getCOS()` is not required here

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -180,6 +188,27 @@
...
+                blockedFileTypes = acct.getCOS().getMultiAttr(Provisioning.A_zimbraFileUploadBlockedFileTypes);
```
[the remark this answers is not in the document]
chinmay15: account inheritance is in place so `.getCOS()` is not required here

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
chinmay15: you can use `Arrays.stream(arr).anyMatch()` to do `contains` check on array.

chinmay15 (changes requested):

common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
chinmay15: if this feature is targeted for patch release, please use LC and not ldap attribute.

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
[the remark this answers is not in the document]
log2akshat: Will do in all...

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
log2akshat: This will not create an independent list, for the existing array we are getting the a list view. It's just a wrapper which is used to make available array available as a list and no data is copied or created so heap is not wasted.

common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
log2akshat: @Prashantsurana Can you please suggest.

chinmay15 (approved):
LGTM

common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
Prashantsurana: We can go with LDAP attributes.

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -180,6 +191,30 @@
                 if (contentType == null)
                     contentType = MimeConstants.CT_APPLICATION_OCTET_STREAM;
             }
+
+            Account acct = Provisioning.getInstance().getAccount(acctId);
+            String [] blockedFileTypes = null;
+            if (acct.isFeatureFileTypeUploadRestrictionsEnabled()) {
+                blockedFileTypes = acct.getMultiAttr(Provisioning.A_zimbraFileUploadBlockedFileTypes);
```
Prashantsurana, line 198: you can check if acct.getFileUploadBlockedFileTypes();  can be used or not?

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -180,6 +191,30 @@
...
+                        throw ServiceException.BLOCKED_FILE_TYPE_UPLOAD(
+                                String.format("Blocked attachment during uploading %s filetype ", blockedExt), null);
+                    }
+                }
+
+                List<String> blockedExtensionList = new ArrayList<>(Arrays.asList(blockedFileTypes));
+                mLog.debug("Start - Using Tika library for retrieving the extension.");
+                String fileExtension = getExtension(file);
+                mLog.debug("End - Using Tika library for retrieving the extension.");
+                if (blockedExtensionList.contains(fileExtension)) {
```
Prashantsurana, line 210: is this case insensitive?

Prashantsurana (approved):
LGTM

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -188,6 +223,25 @@
...
+            String fileExtension = null;
+            try {
+                TikaConfig tikaConfig = TikaConfig.getDefaultConfig();
+                Detector detector = tikaConfig.getDetector();
+                TikaInputStream stream = TikaInputStream.get(filename.getInputStream());
+                Metadata metadata = new Metadata();
+                metadata.add(Metadata.RESOURCE_NAME_KEY, filename.getName());
+                MediaType mediaType = detector.detect(stream, metadata);
+                MimeType mimeType = tikaConfig.getMimeRepository().forName(mediaType.toString());
+                fileExtension = mimeType.getExtension().split("\\.")[1];
```
chinmay15, line 236: you can use Apache commons' `StringUtils.strip(String, String)`

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -138,6 +150,7 @@
             uuid      = localServer + UPLOAD_PART_DELIMITER + LdapUtil.generateUUID();
             name      = FileUtil.trimFilename(filename);
             file      = attachment;
+            extension = FilenameUtils.getExtension(filename).trim();
```
thefoolwhocodes, line 153: This line can cause NPE.

thefoolwhocodes: Because getExtension can return null.

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -180,6 +193,27 @@
                 if (contentType == null)
                     contentType = MimeConstants.CT_APPLICATION_OCTET_STREAM;
             }
+
+            Account acct = Provisioning.getInstance().getAccount(acctId);
+            String [] blockedFileTypes = null;
+            if (acct.isFeatureFileTypeUploadRestrictionsEnabled()) {
+                blockedFileTypes = acct.getFileUploadBlockedFileTypes();
+                List<String> blockedExtensionList = new ArrayList<>(Arrays.asList(blockedFileTypes));
+                if (blockedExtensionList.stream().anyMatch(extension::equalsIgnoreCase)) {
```
thefoolwhocodes, line 202: I would recommend a size check here before triggering stream + match.

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -180,6 +193,27 @@
...
+                    throw ServiceException.BLOCKED_FILE_TYPE_UPLOAD(
+                            String.format("Blocked attachment during uploading %s filetype ", extension), null);
+                }
+
+                mLog.debug("Start - Using Tika library for retrieving the extension.");
+                String fileExtension = getExtension(file);
+                mLog.debug("End - Using Tika library for retrieving the extension.");
+                if (blockedExtensionList.stream().anyMatch(fileExtension::equalsIgnoreCase)) {
```
thefoolwhocodes, line 210: I would recommend a size check here before triggering stream + match.

thefoolwhocodes, line 210: This line can cause NPE, fileExtension NULL check need to be performed.

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -180,6 +193,27 @@
                 if (contentType == null)
                     contentType = MimeConstants.CT_APPLICATION_OCTET_STREAM;
             }
+
+            Account acct = Provisioning.getInstance().getAccount(acctId);
+            String [] blockedFileTypes = null;
```
thefoolwhocodes, line 198: Would recommend moving this variable creation inside if statement.

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -719,34 +770,20 @@ public void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOEx
         Upload up = new Upload(acct.getId(), fi, filename);
 
         if (filename.endsWith(".har")) {
-            File file = ((DiskFileItem) fi).getStoreLocation();
             try {
-                String mimeType = MimeDetect.getMimeDetect().detect(file);
+                TikaInputStream stream = TikaInputStream.get(fi.getInputStream());
```
thefoolwhocodes, line 774: Reuse the code from Upload class after making getExtension as static.

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -117,12 +126,15 @@
...
         final String   accountId;
         String         contentType;
+        String         extension;
         final String   uuid;
         final String   name;
         final FileItem file;
         long time;
         boolean deleted = false;
         BlobInputStream blobInputStream;
+        static TikaConfig tikaConfig = TikaConfig.getDefaultConfig();
```
thefoolwhocodes, line 136: final in both places.

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -188,6 +224,29 @@
...
+            } catch (IOException exp) {
+                mLog.warn("Cannot get content for upload", exp);
+            }
+            return mimeType;
+        }
+
+        public static String getExtension(FileItem fileItem) {
+            String fileExtension = null;
+            MimeType mimeType = getMimeType(fileItem);
+            fileExtension = StringUtils.strip(mimeType.getExtension(), ".");
```
thefoolwhocodes, line 246: This line can cause NPE.

thefoolwhocodes: mimeType.getExtension()

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -719,33 +778,11 @@ public void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOEx
...
-            return m.find();
-        })) {
-            mLog.debug("handlePlainUpload(): deleting %s", fi);
-            fi.delete();
-            mLog.info("File content type is blacklisted : %s" + finalMimeType);
-            drainRequestStream(req);
-            sendResponse(resp, HttpServletResponse.SC_FORBIDDEN, fmt, null, null, null);
-            return Collections.emptyList();
+            MimeType mimeType = Upload.getMimeType(fi);
+            up.contentType = mimeType.toString();
```
thefoolwhocodes, line 782: This line can cause NPE mimeType.toString()

thefoolwhocodes (approved):

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -180,6 +193,43 @@
...
+                blockedFileTypes = acct.getFileUploadBlockedFileTypes();
+                List<String> blockedExtensionList = new ArrayList<>(Arrays.asList(blockedFileTypes));
+                if (blockedExtensionList.size() > 0) {
+                    if (blockedExtensionList.stream().anyMatch(extension::equalsIgnoreCase)) {
+                        throw ServiceException.BLOCKED_FILE_TYPE_UPLOAD(
+                                String.format("Blocked attachment during uploading %s filetype ", extension), null);
+                    }
+
+                    MimeType mimeType = getMimeType(file);
+                    if (blockedExtensionList.stream().anyMatch((blacklistedContentType) -> {
```
Prashantsurana, line 209: avoid terms variable  like `blacklisted` or `whitelisted`

Prashantsurana (changes requested):

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9822,4 +9822,38 @@ TODO: delete them permanently from here
...
+  <defaultCOSValue>shb</defaultCOSValue>
+  <defaultCOSValue>shm</defaultCOSValue>
+  <defaultCOSValue>shs</defaultCOSValue>
+  <defaultCOSValue>vbe</defaultCOSValue>
+  <defaultCOSValue>vbs</defaultCOSValue>
+  <defaultCOSValue>vbx</defaultCOSValue>
+  <defaultCOSValue>vxd</defaultCOSValue>
+  <defaultCOSValue>wsf</defaultCOSValue>
+  <defaultCOSValue>wsh</defaultCOSValue>
+  <defaultCOSValue>xl</defaultCOSValue>
```
Prashantsurana, line 9852: add `application/x-ms*` as default.

Prashantsurana: remove LC

Prashantsurana (changes requested):

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -719,33 +804,10 @@ public void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOEx
         Upload up = new Upload(acct.getId(), fi, filename);
 
         if (filename.endsWith(".har")) {
```
Prashantsurana, line 806: Why do we need separate for .har? 
I think we have discuss to handle everything at one place.

Prashantsurana: .har should be added as default config.

Prashantsurana (changes requested):

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
log2akshat: Removed the unnecessary code for handling the .har file as it was based upon only the extension. As, if we remove the extension of the .har file ZBUG-711 still exist and making the web-client unresponsive and also in CPU spike. While it is correctly handled by the tika with it's content-type detected as application/pdf contrary to earlier where it is getting detected as application/octet-stream but the file actually contains html content and the convertd is trying to convert the document based on application/octet-stream and failing.

madhavdhuppe (approved):

PR #1138, opened 2021-04-20
ZCS-10443: File upload size restrictions per user

log2akshat:
**Problem:** 
File upload size restrictions per user.

**Description:**
- Added new `LDAP` attribute `zimbraMailAttachmentMaxSize` at `account` and `cos` level for restricting the maximum size in bytes for e-mail attachment.
- Set the `defaultCOSValue` for the `LDAP` attribute `zimbraMailAttachmentMaxSize` as `0`.
- Modified `zimbraFileUploadMaxSizePerFile` `LDAP` attribute to be used for file uploads restrictions other than e-mail. Now, the old `defaultCOSValue` is removed as its value should be less than the `LDAP` attribute `zimbraFileUploadMaxSize`. If the default value is not set then upload size is limited by `zimbraFileUploadMaxSize` which is global. Also, added accountInfo flag to `zimbraFileUploadMaxSizePerFile`.
- Added new ldap attributes for restricting  the maximum size in bytes for e-mail attachment `zimbraMailAttachmentMaxSize`
and the for file uploads restrictions `zimbraFileUploadMaxSizePerFile` other than e-mail. The default value for both the new ldap attributes is set to `0`. So, the provided file upload size limit to `account` and `COS` levels. As we are using the `zimbraMailAttachmentMaxSize` ldap attribute for e-mail attachments file size restriction and  `zimbraFileUploadMaxSizePerFile` for the briefcase and other file size restrictions. And if these new ldap attributes are not set then it falls back to the original ldap values `zimbraMtaMaxMessageSize` in case of e-mail attachment file size restriction else `zimbraFileUploadMaxSize`. The values of these new attributes are lower than the older global level ldap attributes (`zimbraMtaMaxMessageSize`, `zimbraFileUploadMaxSize`). If the value of (`zimbraMailAttachmentMaxSize`, `zimbraFileUploadMaxSizePerFile`) is set higher than the (`zimbraMtaMaxMessageSize`, `zimbraFileUploadMaxSize`) then the global value is overridden.

**Updates:**
- Bumped the version of newly introduced  `LDAP` attribute `zimbraMailAttachmentMaxSize` from `9.0.0` to `9.1.0`
- Added code for handling the error in the  catch block of the method `getFileUploadMaxSize`.
- 05May2021
-- Fixed the scenario when the user was able to attach more than 1 files after crossing the size restrictions imposed by the new ldap attributes but with single attachment it was working fine.
-- Also fixed when the new ldap attributes on the account level was set to `0`  files were getting uploaded as zero bytes now when the `zimbraMailAttachmentMaxSize` and `zimbraFileUploadMaxSizePerFile` is set to the default values then the values of `zimbraMtaMaxMessageSize` and `zimbraFileUploadMaxSize` will be used.

**Note:**
UI performs a check of base64 encoding on the size of the upload
```
$ zmprov -l gcf zimbraMtaMaxMessageSize
zimbraMtaMaxMessageSize: 10240000

$ zmprov -l gcf zimbraFileUploadMaxSize
zimbraFileUploadMaxSize: 10485760
```
uses `attSizeLimit` and `docSizeLimit` returned in `GetInfo` call:
```
<GetInfoResponse docSizeLimit="10485760" requestId="418fb075-dec3-49a6-9b75-6ad96a484a6b" attSizeLimit="10240000" xmlns="urn:zimbraAccount">
=> attSizeLimit :: zimbraMtaMaxMessageSize :: 10240000 (9.7656MB)
=> docSizeLimit :: zimbraFileUploadMaxSize :: 10485760 (10MB)
```
But UI is restricting to `7MB` limit, details are in the following bug.
```
Attachment file size check should be before encoding
The admin sets the limit based on the base64 encoded size. (The actual size sent over the network). And since 1.3333 * originalSize = base64Size, this means that:
If the admin sets the limit (for either attSizeLimit or docSizeLimit) to some value (m), then as far as the client (and end-user) is concerned the limit is actually m / 1.34. (1.34 is approximation of 1.333333....)
```
[Bug 76919](https://bugzilla.zimbra.com/show_bug.cgi?id=76919) - [PR](https://github.com/Zimbra/zm-web-client/commit/78f82dedfd70da53ac87556712787230d4bd89e3)

**Testing Done:**
- Verified the new ldap attributes are present.
- Verified that if zimbraMailAttachmentMaxSize is set to a lower value than `zimbraMtaMaxMessageSize` for e-mail attachment then `zimbraMailAttachmentMaxSize` value is used and if `zimbraMailAttachmentMaxSize` is set to a higher value than `zimbraMtaMaxMessageSize` for e-mail attachment then the value of `zimbraMtaMaxMessageSize` is used.
- Verified that if `zimbraFileUploadMaxSizePerFile` is set to a lower value than `zimbraFileUploadMaxSize` for e-mail attachment then `zimbraFileUploadMaxSizePerFile` value is used and if `zimbraFileUploadMaxSizePerFile` is set to a higher value than `zimbraFileUploadMaxSize` for e-mail attachment then the value of `zimbraFileUploadMaxSize` is used.
- Verified now the whole attachment upload is also now limited by the `zimbraMailAttachmentMaxSize` and `zimbraFileUploadMaxSizePerFile` ldap attributes.
- Verified that when the new ldap attributes on the account level was set to `0`  files are not getting uploaded as zero bytes.

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -1349,7 +1349,7 @@ TODO - add support for multi-line values in globalConfigValue and defaultCOSValu
 
 <attr id="227" name="zimbraFileUploadMaxSize" type="long" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited,accountInfo">
   <globalConfigValue>10485760</globalConfigValue>
-  <desc>Maximum size in bytes for file uploads</desc>
+  <desc>Maximum size in bytes for file uploads.</desc>
 </attr>
 
 <attr id="229" name="zimbraTimeZoneStandardDtStart" type="string" max="256" cardinality="multi" requiredIn="timeZone" deprecatedSince="5.0">
@@ -6786,7 +6786,10 @@ TODO: delete them permanently from here
 <attr id="1350" name="zimbraFileUploadMaxSizePerFile" type="long" cardinality="single" optionalIn="account,cos,domain,globalConfig" flags="accountCosDomainInherited,domainInherited,accountInfo" since="8.0.0">
   <globalConfigValue>2147483648</globalConfigValue>
   <defaultCOSValue>0</defaultCOSValue>
-  <desc>Maximum size in bytes for each attachment.</desc>
+  <desc>
+    Maximum size in bytes for each attachment for the file uploads.
+    If the default value is not set then upload size is limited by zimbraFileUploadMaxSize.
+  </desc>
 </attr>
 
 <attr id="1351" name="zimbraPublicSharingEnabled" type="boolean" cardinality="single" optionalIn="account,cos,domain" flags="accountInfo,accountCosDomainInherited,domainAdminModifiable" since="8.0.0">
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9825,6 +9825,7 @@ TODO: delete them permanently from here
   <defaultCOSValue>TRUE</defaultCOSValue>
   <desc>whether to use Web Client feature</desc>
 </attr>
+
 <attr id="3091" name="zimbraFileUploadBlockedFileTypes" type="string" cardinality="multi" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" since="9.1.0">
   <defaultCOSValue>asd</defaultCOSValue>
   <defaultCOSValue>bat</defaultCOSValue>
@@ -9858,7 +9859,7 @@ TODO: delete them permanently from here
   <desc>Commonly blocked file types for uploading</desc>
  </attr>
 
-<attr id="3092" name="zimbraFeatureFileTypeUploadRestrictionsEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" since="9.1.0">
+ <attr id="3092" name="zimbraFeatureFileTypeUploadRestrictionsEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" since="9.1.0">
   <defaultCOSValue>TRUE</defaultCOSValue>
   <desc>Enable/Disable blocked file types set in zimbraFileUploadBlockedFileTypes for uploading</desc>
  </attr>
@@ -9902,8 +9903,8 @@ TODO: delete them permanently from here
   <defaultCOSValue>TRUE</defaultCOSValue>
   <desc>Whether or not to log document accessed time</desc>
  
-<attr id="4000" name="zimbraMailAttachmentMaxSize" type="long" cardinality="single" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" since="9.1.0">
+ <attr id="4000" name="zimbraMailAttachmentMaxSize" type="long" cardinality="single" optionalIn="account,cos" flags="accountInherited,domainAdminModifiable" since="9.1.0">
   <defaultCOSValue>0</defaultCOSValue>
   <desc>Maximum size in bytes for e-mail attachment.</desc>
-</attr>
+ </attr>
 </attrs>
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -6783,9 +6783,9 @@ TODO: delete them permanently from here
   </desc>
 </attr>
 
-<attr id="1350" name="zimbraFileUploadMaxSizePerFile" type="long" cardinality="single" optionalIn="account,cos,domain,globalConfig" flags="accountCosDomainInherited,domainInherited" since="8.0.0">
+<attr id="1350" name="zimbraFileUploadMaxSizePerFile" type="long" cardinality="single" optionalIn="account,cos,domain,globalConfig" flags="accountCosDomainInherited,domainInherited,accountInfo" since="8.0.0">
   <globalConfigValue>2147483648</globalConfigValue>
-  <defaultCOSValue>2147483648</defaultCOSValue>
+  <defaultCOSValue>0</defaultCOSValue>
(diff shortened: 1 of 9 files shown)


store/src/java/com/zimbra/cs/account/ZAttrAccount.java
```diff
@@ -25913,6 +25923,78 @@ public void unsetMailAllowReceiveButNotSendWhenOverQuota() throws com.zimbra.com
         return attrs;
     }
 
+    /**
+     * Maximum size in bytes for e-mail attachment.
+     *
+     * @return zimbraMailAttachmentMaxSize, or 0 if unset
+     *
+     * @since ZCS 9.0.0
```
silentsakky, line 25931: 9.1.0?

chinmay15: yes it should be 9.1.0

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -881,10 +881,50 @@ private static long getFileUploadMaxSize(boolean limitByFileUploadMaxSize) {
...
+                        maxSize = mtaMaxMsgSize;
+                    } else {
+                        maxSize = mailAttachmentMaxSize;
+                    }
+                    if (maxSize == 0) {
+                        maxSize = -1;
+                    }
+                }
+            } catch (ServiceException e) {
+                // TODO Auto-generated catch block
```
chinmay15, line 911: do we want to log and re-throw the exception? i think we are not supposed to swallow exception here.

chinmay15 (changes requested):

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -6783,10 +6783,13 @@ TODO: delete them permanently from here
   </desc>
 </attr>
 
-<attr id="1350" name="zimbraFileUploadMaxSizePerFile" type="long" cardinality="single" optionalIn="account,cos,domain,globalConfig" flags="accountCosDomainInherited,domainInherited" since="8.0.0">
+<attr id="1350" name="zimbraFileUploadMaxSizePerFile" type="long" cardinality="single" optionalIn="account,cos,domain,globalConfig" flags="accountCosDomainInherited,domainInherited,accountInfo" since="8.0.0">
   <globalConfigValue>2147483648</globalConfigValue>
-  <defaultCOSValue>2147483648</defaultCOSValue>
-  <desc>Maximum size in bytes for each attachment.</desc>
+  <defaultCOSValue>0</defaultCOSValue>
```
chinmay15, line 6788: LGTM. just make sure 2 default values are not breaking anything.

chinmay15 (approved):

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -881,10 +881,52 @@ private static long getFileUploadMaxSize(boolean limitByFileUploadMaxSize) {
         return maxSize;
     }
 
+    private static long getFileUploadMaxSize(boolean limitByFileUploadMaxSize, Account acct) {
+        long maxSize = DEFAULT_MAX_SIZE;
+        String attributeUsed = null;
+        try {
+            if (limitByFileUploadMaxSize && acct.getFileUploadMaxSizePerFile() > 0) {
+                long fileUploadSizePerFile = acct.getFileUploadMaxSizePerFile();
```
thefoolwhocodes, line 889: Move the variable creation inside respective if and else condition, so that at least one variable creation step can be saved in each execution.

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -881,10 +881,52 @@ private static long getFileUploadMaxSize(boolean limitByFileUploadMaxSize) {
...
+                        Provisioning.A_zimbraFileUploadMaxSize, DEFAULT_MAX_SIZE);
+                if (fileUploadSizePerFile >= fileUploadSize) {
+                    maxSize = fileUploadSize;
+                    attributeUsed = Provisioning.A_zimbraFileUploadMaxSize;
+                } else {
+                    maxSize = fileUploadSizePerFile;
+                    attributeUsed = Provisioning.A_zimbraFileUploadMaxSizePerFile;
+                }
+            } else if (!limitByFileUploadMaxSize && acct.getMailAttachmentMaxSize() > 0) {
+                long mailAttachmentMaxSize = acct.getMailAttachmentMaxSize();
```
thefoolwhocodes, line 900: Move the variable creation inside respective if and else condition, so that at least one variable creation step can be saved in each execution.

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -881,10 +881,52 @@ private static long getFileUploadMaxSize(boolean limitByFileUploadMaxSize) {
...
+                long mtaMaxMsgSize = Provisioning.getInstance().getConfig().getLongAttr(
+                        Provisioning.A_zimbraMtaMaxMessageSize, DEFAULT_MAX_SIZE);
+                if (mailAttachmentMaxSize > mtaMaxMsgSize) {
+                    maxSize = mtaMaxMsgSize;
+                    attributeUsed = Provisioning.A_zimbraMtaMaxMessageSize;
+                } else {
+                    maxSize = mailAttachmentMaxSize;
+                    attributeUsed = Provisioning.A_zimbraMailAttachmentMaxSize;
+                }
+                if (maxSize == 0) {
```
thefoolwhocodes, line 910: In what condition maxSize can be zero, and what does -1 means, one liner comment please.

log2akshat: `0` means `no limit`.  The return value from this function gets used by `FileUploadBase` `sizeMax` where `-1` means `no limit`

thefoolwhocodes (approved):

RBDurgin (approved):

PR #1139, opened 2021-04-21
TSS-18300: Fix for emails displaying as blank

log2akshat:
**Problem:**
Some of the emails displaying as blank.

**Analysis and Fix:**
It was observed that certain e-mails were displaying as blank. On analysing the input mime it was found that there was issue with the input html that there was no closing `style` tag. Here, there are 2 issues:

1. In the `JTidy` configuration `setPrintBodyOnly()` was earlier set to true so it was processing only the body part and skipping the style tags.

2. Earlier in the code skipped the trimming of the parsed string which was getting null value as it was not handling the head of the html because of `setPrintBodyOnly()` was set to true but the code  was written to handle these situation earlier and fallback to the `OWASP` if the output from the `JTidy` was `null`. In this particular case it was not able to handle because
the parsed string was having a whitespace character and it was not null so it skipped the string trimming which is causing not getting fallback to `OWASP` directly.

**Testing Done:**
- Imported few e-mails from the account mentioned in the [TSS-18300](https://jira.corp.synacor.com/browse/TSS-18300) and verified they are displaying properly.
- Verified the older malformed html mime, attached in the [TSS-18004](https://jira.corp.synacor.com/browse/TSS-18004) and verified it is displaying properly.

### DIFF
--- a/store/src/java-test/com/zimbra/cs/html/owasp/OwaspHtmlSanitizerTest.java
+++ b/store/src/java-test/com/zimbra/cs/html/owasp/OwaspHtmlSanitizerTest.java
@@ -705,4 +705,21 @@ public class OwaspHtmlSanitizerTest {
         // check that the id and class attributes are not removed
         Assert.assertTrue(result.equals(html));
     }
+
+    @Test
+    public void testBugTSS18004() throws Exception {
+        String malformedHtml = "<HTML><BODY><p dir=\"ltr\" style=\"-webkit-text-size-adjust:auto;line-height:1.38;margin-top:0pt;margin-bottom:0pt;\">"
+                + "<span style=\"font-family:Arial;font-size:14.666666984558105px-webkit-text-size-adjust:auto;\"\">Hello h</span>"
+                + "<span style=\"font-size:11pt;font-family:Arial;font-variant-ligatures:normal;font-variant-east-asian:normal;font-variant-position:normal;vertical-align:baseline\"\">ave a nice day.</span>"
+                + "</p></BODY></HTML>";
+        String result = new OwaspHtmlSanitizer(malformedHtml, true, null).cleanMalformedHtml(malformedHtml, true);
+        String output = "<p dir=\"ltr\"\n"
+                + "style=\"-webkit-text-size-adjust:auto;line-height:1.38;margin-top:0pt;margin-bottom:0pt;\"><span\n"
+                + " style=\"font-family:Arial;font-size:14.666666984558105px-webkit-text-size-adjust:auto;\">Hello\n"
+                + "h</span><span\n"
+                + "style=\"font-size:11pt;font-family:Arial;font-variant-ligatures:normal;font-variant-east-asian:normal;font-variant-position:normal;vertical-align:baseline\">ave\n"
+                + "a nice day.</span></p>";
+        // check that the extra double quotes are removed
+        Assert.assertTrue(output.equals(result.trim()));
+    }
 }
--- a/store/src/java-test/com/zimbra/cs/html/owasp/OwaspHtmlSanitizerTest.java
+++ b/store/src/java-test/com/zimbra/cs/html/owasp/OwaspHtmlSanitizerTest.java
@@ -34,6 +34,7 @@ import org.junit.rules.TestName;
 
 import com.google.common.base.Charsets;
 import com.google.common.io.CharStreams;
+import com.zimbra.common.localconfig.DebugConfig;
 import com.zimbra.common.localconfig.LC;
 import com.zimbra.common.util.ByteUtil;
 import com.zimbra.cs.mailbox.MailboxTestUtil;
@@ -90,7 +91,7 @@ public class OwaspHtmlSanitizerTest {
         + "            )";
 
     private void defangHtmlString(String html, String expected) throws IOException {
-        String result = new OwaspHtmlSanitizer(html, true, null).sanitize();
+        String result = new OwaspHtmlSanitizer(html, true, null).sanitize(false);
         Assert.assertEquals("Defanged HTML result", expected, result);
     }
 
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
```diff
@@ -98,12 +98,12 @@ private String cleanMalformedHtml(String str) throws UnsupportedEncodingExceptio
             Tidy tidy = new Tidy();
             tidy.setInputEncoding("UTF-8");
             tidy.setOutputEncoding("UTF-8");
-            tidy.setPrintBodyOnly(true);
+            tidy.setPrintBodyOnly(false);
```
Prashantsurana, line 101: can you please explain the option?

store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
```diff
@@ -98,12 +98,12 @@ private String cleanMalformedHtml(String str) throws UnsupportedEncodingExceptio
...
             tidy.setXHTML(true);
             ByteArrayInputStream inputStream = new ByteArrayInputStream(str.getBytes("UTF-8"));
             ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
             tidy.parseDOM(inputStream, outputStream);
-            String outStream = outputStream.toString("UTF-8");
+            String outStream = outputStream.toString("UTF-8").trim();
```
Prashantsurana, line 106: this should not trim() if outputStream.toString("UTF-8")=null then it will be NPE. 

you want to trim() to check blank String which can be handled  if condition next line.

store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
```diff
@@ -98,12 +98,12 @@ private String cleanMalformedHtml(String str) throws UnsupportedEncodingExceptio
...
             if (outStream.isEmpty() || outStream == null) {
```
Prashantsurana, line 107: 1. `if (outStream.isEmpty() || outStream == null) {`  this is wrong check.  Null check does not make any sense.
if outStream is null then outStream.isEmpty() will throw NullPointerException. 

See below code snippet. it will always throw NPE. 
`        String b = null; 
        if(b.isEmpty() || b == null) {
            System.out.println("blank");
        }` 

The correct code first null check and then operations. 

`if (outStream == null || outStream.isEmpty()) {` 

2. you can use trim() while validating. do not unnecessary remove  leading and trailing spaces and set back to the string. 

`if (outStream == null || outStream.trim().isEmpty()) {`

Prashantsurana (changes requested):
Please correct.

chinmay15 (approved):
LGTM

madhavdhuppe (approved):

store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
```diff
@@ -63,6 +63,14 @@ private void instantiatePolicy() {
      * @throws UnsupportedEncodingException
      */
     public String sanitize() throws UnsupportedEncodingException {
+        return processSanitization(true).toString();
+    }
+
+    public String sanitize(boolean cleanData) throws UnsupportedEncodingException {
```
RBDurgin, line 69: Sanitizing is cleaning up data, can we rename **cleanData** to something a little more verbose, like **cleanMalformedHtml**?

store/src/java-test/com/zimbra/cs/html/owasp/OwaspHtmlSanitizerTest.java
```diff
@@ -158,39 +159,39 @@ public void testBug46948() throws Exception {
     @Test
     public void testBug98215() throws Exception {
         String html = "<a href=\"vbscript:alert(parent.csrfToken)\">CLICK</a>";
-        String result = new OwaspHtmlSanitizer(html, true, null).sanitize();
+        String result = new OwaspHtmlSanitizer(html, true, null).sanitize(false);
```
RBDurgin, line 162: You should add test cases to make sure all these conditions pass with jtidy on as well.

store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
```diff
@@ -73,38 +81,46 @@ public String sanitize() throws UnsupportedEncodingException {
...
+    private void runSanitizer(String str, Policy policy, boolean cleanData) throws UnsupportedEncodingException {
+        if (cleanData) {
+            HtmlSanitizer.sanitize(cleanMalformedHtml(str, false), policy);
+        } else {
+            HtmlSanitizer.sanitize(str, policy);
+        }
     }
 
-    private String cleanMalformedHtml(String str) throws UnsupportedEncodingException {
+    public String cleanMalformedHtml(String str, boolean printBodyOnly) throws UnsupportedEncodingException {
```
RBDurgin, line 110: Does this need to be made public, since it's conditionally being called by the **cleanData** param in the sanitize() method?

store/src/java-test/com/zimbra/cs/html/owasp/OwaspHtmlSanitizerTest.java
```diff
@@ -689,23 +686,40 @@ public void testBug102910() throws Exception {
...
         Assert.assertTrue(result.equals(html));
     }
+
+    @Test
+    public void testBugTSS18004() throws Exception {
+        String malformedHtml = "<HTML><BODY><p dir=\"ltr\" style=\"-webkit-text-size-adjust:auto;line-height:1.38;margin-top:0pt;margin-bottom:0pt;\">"
+                + "<span style=\"font-family:Arial;font-size:14.666666984558105px-webkit-text-size-adjust:auto;\"\">Hello h</span>"
+                + "<span style=\"font-size:11pt;font-family:Arial;font-variant-ligatures:normal;font-variant-east-asian:normal;font-variant-position:normal;vertical-align:baseline\"\">ave a nice day.</span>"
+                + "</p></BODY></HTML>";
+        String result = new OwaspHtmlSanitizer(malformedHtml, true, null).cleanMalformedHtml(malformedHtml, true);
```
RBDurgin, line 715: Do we have a test to check this is returning the expected result when running through full owasp sanitization?

RBDurgin (changes requested):
Added some comments. Also make sure we enable this test when the full suite is run https://github.com/ZimbraOS/zm-mailbox/blob/develop/build-common.xml#L322

store/src/java-test/com/zimbra/cs/html/owasp/OwaspHtmlSanitizerTest.java
```diff
@@ -689,23 +686,40 @@ public void testBug102910() throws Exception {
...
+        String output = "<p dir=\"ltr\"\n"
+                + "style=\"-webkit-text-size-adjust:auto;line-height:1.38;margin-top:0pt;margin-bottom:0pt;\"><span\n"
+                + " style=\"font-family:Arial;font-size:14.666666984558105px-webkit-text-size-adjust:auto;\">Hello\n"
+                + "h</span><span\n"
+                + "style=\"font-size:11pt;font-family:Arial;font-variant-ligatures:normal;font-variant-east-asian:normal;font-variant-position:normal;vertical-align:baseline\">ave\n"
+                + "a nice day.</span></p>";
+        // check that the extra double quotes are removed
+        Assert.assertTrue(output.equals(result.trim()));
```
RBDurgin, line 723: A brief error message when the test fails would be helpful as well, so the next developer doesn't have to keep jumping back and forth between Jira tickets to understand what's expected.

store/src/java-test/com/zimbra/cs/html/owasp/OwaspHtmlSanitizerTest.java
log2akshat: All these test cases are passing with `JTidy` enabled.

store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
log2akshat: It is made public as it's being called from `OwaspHtmlSanitizerTest.testBugTSS18004` for verifying extra double quotes are removed.

store/src/java-test/com/zimbra/cs/html/owasp/OwaspHtmlSanitizerTest.java
log2akshat: Used `cleanMalformedHtml(malformedHtml, true)` to avoid the insertion of html tags by the `JTidy`.

PR #1141, opened 2021-04-29
TSS-18349/ZCS-10544: Fix for emails displaying as blank [2]

log2akshat:
**Problem:**
Some of the emails displaying as blank.

**Analysis and Fix:**
It was observed that certain e-mails were displaying as blank. On analysing the input mime it was found that there was issue
with the input html after getting parsed div tag was containing a new line between `<` and `/div>`. ~~To fix this issue a new
condition is introduced to handle if any of the end tag contains this situation then newline is removed from that tag and it
should constitute a well-formed end tag.~~
**Update:** Handled the newline condition using the `JTidy` library option `setWraplen(0)` as earlier `JTidy` was wrapping the html at 68th character. So, setting this option to set the wrapping to `0` which in removes any chances of having the unnecessary newline in the html.

Also, added `localconfig` `hide_jtidy_warnings` to disable the printing of warnings and messages in the `zmmailboxd.out` log.
Also, added the following configs:
- Set to omit the doctype to not have unnecessary jtidy generated doctype in the parsed html.
- Set the tidy mark to false so that tidy should not add meta element indicating tidied doc.
- Set the hide end tags to true so that it suppress optional end tags.

**Testing Done:**
- Imported few e-mails from the account mentioned in the [TSS-18349](https://jira.corp.synacor.com/browse/TSS-18349) and verified they are displaying properly.
- Imported few e-mails from the account mentioned in the [TSS-18300](https://jira.corp.synacor.com/browse/TSS-18300) and verified they are displaying properly.
- Verified the older malformed html mime, attached in the [TSS-18004](https://jira.corp.synacor.com/browse/TSS-18004) and verified it is displaying properly.

**Related PRs:**
[TSS-18004](https://github.com/Zimbra/zm-mailbox/pull/1128)
[TSS-18300](https://github.com/ZimbraOS/zm-mailbox/pull/1139)

### DIFF
--- a/common/src/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/java/com/zimbra/common/localconfig/LC.java
@@ -1455,6 +1455,9 @@ public final class LC {
     // imap folder pagination enabled
     public static final KnownKey zimbra_imap_folder_pagination_enabled =  KnownKey.newKey(false);
 
+    // For disabling the printing of jtidy warnings in zmmailboxd.out
+    public static final KnownKey hide_jtidy_warnings = KnownKey.newKey(true);
+
     static {
         // Automatically set the key name with the variable name.
         for (Field field : LC.class.getFields()) {
--- a/common/src/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/java/com/zimbra/common/localconfig/LC.java
@@ -1455,9 +1455,6 @@ public final class LC {
     // imap folder pagination enabled
     public static final KnownKey zimbra_imap_folder_pagination_enabled =  KnownKey.newKey(false);
 
-    // For disabling the printing of jtidy warnings in zmmailboxd.out
-    public static final KnownKey hide_jtidy_warnings = KnownKey.newKey(true);
-
     static {
         // Automatically set the key name with the variable name.
         for (Field field : LC.class.getFields()) {
--- a/store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
+++ b/store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
@@ -123,6 +123,9 @@ public class OwaspHtmlSanitizer implements Callable<String> {
             if (outStream == null || outStream.trim().isEmpty()) {
                 return str;
             }
+            if (outStream.contains("<\n/")) {
+                outStream = outStream.replaceAll("<\n/","</");
+            }
             ZimbraLog.mailbox.debug("End - Using JTidy library for cleaning the markup. Taken %d milliseconds.",
                     (System.currentTimeMillis() - startTime));
             return outStream;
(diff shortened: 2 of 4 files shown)


store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
```diff
@@ -123,6 +131,9 @@ public String cleanMalformedHtml(String str, boolean printBodyOnly) throws Unsup
             if (outStream == null || outStream.trim().isEmpty()) {
                 return str;
             }
+            if (outStream.contains("<\n/")) {
+                outStream = outStream.replaceAll("<\n/","</");
```
chinmay15, line 135: String.replaceAll() takes regex as 1st argument, did you check behaviour with `\` in text? I suggest to use String.replace(), if there is no need of regex matching.
Also it would be great, if you can add unit test for this specific case.

chinmay15 (changes requested):

common/src/java/com/zimbra/common/localconfig/LC.java
```diff
@@ -1444,6 +1444,9 @@
     // imap folder pagination enabled
     public static final KnownKey zimbra_imap_folder_pagination_enabled =  KnownKey.newKey(false);
 
+    // For disabling the printing of jtidy warnings in zmmailboxd.out
+    public static final KnownKey hide_jtidy_warnings = KnownKey.newKey(true);
```
Prashantsurana, line 1448: Please move to DebugConfig, it is more for debugging.

RBDurgin:
Can you add some unit tests with some of the mime examples you used to test?

common/src/java/com/zimbra/common/localconfig/LC.java
log2akshat: Initially I tried that one only, but it is not getting effective until I explicitly set the parameter.

Prashantsurana: why?  did you debug?

Prashantsurana: Why is it not working same as `jtidy_parser` ? which is also boolean?

log2akshat: Done

store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
```diff
@@ -29,6 +29,7 @@
 import org.w3c.tidy.Tidy;
 
 import com.zimbra.common.localconfig.DebugConfig;
+import com.zimbra.common.localconfig.LC;
```
thefoolwhocodes, line 32: Remove the import.

thefoolwhocodes (approved):

sreejan-chowdhury (approved):

chinmay15 (approved):
LGTM

PR #1151, opened 2021-06-08
ZCS-10642: SearchGal | add support for query in request

itsmeprashant:
Added query param in searchGal request to support searching based on queries. This is needed for the organisation chart zimlet.

### DIFF
--- a/store/src/java/com/zimbra/cs/gal/GalSearchControl.java
+++ b/store/src/java/com/zimbra/cs/gal/GalSearchControl.java
@@ -346,12 +346,15 @@ public class GalSearchControl {
     private String getSearchQuery(Account galAcct, boolean addInId) throws ServiceException {
         String query = mParams.getQuery();
         String searchByDn = mParams.getSearchEntryByDn();
+        String galSearchQuery = mParams.getGalSearchQuery();
 
         GalSearchType type = mParams.getType();
         StringBuilder searchQuery = new StringBuilder();
 
         if (searchByDn != null) {
             searchQuery.append("#dn:(" + searchByDn + ")");
+        } else if (!Strings.isNullOrEmpty(galSearchQuery)) {
+            searchQuery.append(galSearchQuery.replace("\"", "\\\"")); // escape quotes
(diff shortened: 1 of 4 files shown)


store/src/java/com/zimbra/cs/gal/GalSearchControl.java
```diff
@@ -346,12 +346,15 @@ private void generateSearchQuery(Account galAcct) throws ServiceException {
     private String getSearchQuery(Account galAcct, boolean addInId) throws ServiceException {
         String query = mParams.getQuery();
         String searchByDn = mParams.getSearchEntryByDn();
+        String galSearchQuery = mParams.getGalSearchQuery();
```
dasiyogesh, line 349: what's the difference between mParams.getQuery() & mParams.getGalSearchQuery(), is there an existing query param?

itsmeprashant: So, SearchGal searches data on `name` request param, and mParams.getQuery(), holds the value for the same. Which searches inside the contact field.
And we want to search data inside any field, so we introduce a new parameter named `query`, and mParams.getGalSearchQuery() holds value for the same.

sreejan-chowdhury (approved):

store/src/java/com/zimbra/cs/service/account/SearchGal.java
```diff
@@ -71,14 +71,27 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
         
-        // otherwise require a query
+        String galSearchQuery = null;
+        
         String name = null;
-        if (ref == null) {
-            name = request.getAttribute(AccountConstants.E_NAME);
-        }
+        
+		if (ref == null) {
```
dasiyogesh, line 81: indentation looks off for this if block, please correct it.

dasiyogesh (approved):

store/src/java/com/zimbra/cs/service/account/SearchGal.java
RahulMehraSynacor: done

PR #1153, opened 2021-06-10
Feature/ZCS 10618

sreejan-chowdhury:
**Problem**
1. Generate the onlyoffice nginx config to redirect the onlyoffice requests ( url which starts with /zdocument) to the particular onlyoffice server for : 
 1.1 Single Deployment of onlyoffice (Onlyoffice on a single mailbox server) or One onlyoffice on each mailbox server.
 1.2 Single Onlyoffice Deployment on a different server. It is defined by setting `zimbraDocumentEditingHost`

2. Generate the nginx config while using `zmproxyctl` for both type of deployments as mentioned in 1
3. Generate the hash for the zimbraId of the server where onlyoffice is installed ( based on conditions in problem 1)

Solution : 
1. Added 2 new keys or placeholders ( web.upstream.onlyoffice.docservice , web.upstream.onlyoffice.spellchecker). Added new classes to generate the upstream blocks for docservice and spellchecker.
2. URL Safe encoding for the zimbraID

Related PR : https://github.com/Zimbra/zm-nginx-conf/pull/28
https://github.com/Zimbra/zm-doc-server-ext/pull/4

### DIFF
--- a/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
+++ b/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
@@ -5697,14 +5697,6 @@ public class ZAttrProvisioning {
     public static final String A_zimbraDocumentEditingCallbackPort = "zimbraDocumentEditingCallbackPort";
 
     /**
-     * Host server where the onlyoffice is hosted
-     *
-     * @since ZCS 9.1.0
-     */
-    @ZAttr(id=3098)
-    public static final String A_zimbraDocumentEditingHost = "zimbraDocumentEditingHost";
-
-    /**
      * Document editing JWT secret
      *
      * @since ZCS 9.1.0
@@ -5719,6 +5711,14 @@ public class ZAttrProvisioning {
      */
     @ZAttr(id=3101)
     public static final String A_zimbraDocumentRecentlyViewedEnabled = "zimbraDocumentRecentlyViewedEnabled";
+
+    /**
+     * Host server where the onlyoffice is hosted
+     *
+     * @since ZCS 9.1.0
+     */
+    @ZAttr(id=3098)
+    public static final String A_zimbraDocumentServerHost = "zimbraDocumentServerHost";
 
     /**
      * maximum amount of mail quota a domain admin can set on a user
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9893,7 +9893,6 @@ TODO: delete them permanently from here
   <desc>Whether or not document editing feature is enabled within briefcase</desc>
 </attr>
 <attr id="3098" name="zimbraDocumentServerHost" type="string" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="9.1.0">
-  <globalConfigValue>zimbra.com</globalConfigValue>
   <desc>Host server where the onlyoffice is hosted</desc>
 </attr>
 <attr id="3099" name="zimbraDocumentEditingCallbackPort" type="port" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" callback="CheckPortConflict" since="9.1.0">
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9892,7 +9892,7 @@ TODO: delete them permanently from here
   <defaultCOSValue>TRUE</defaultCOSValue>
   <desc>Whether or not document editing feature is enabled within briefcase</desc>
 </attr>
-<attr id="3098" name="zimbraDocumentEditingHost" type="string" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="9.1.0">
+<attr id="3098" name="zimbraDocumentServerHost" type="string" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="9.1.0">
   <globalConfigValue>zimbra.com</globalConfigValue>
   <desc>Host server where the onlyoffice is hosted</desc>
 </attr>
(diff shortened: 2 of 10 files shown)


store/src/java/com/zimbra/cs/util/ProxyConfGen.java
```diff
@@ -1106,6 +1107,107 @@ public void update() throws ServiceException {
...
+
+        super("web.upstream.onlyoffice.docservice", null, null, ProxyConfValueType.STRING, ProxyConfOverride.CUSTOM,
+                "List of upstream HTTPS servers towards docservice port used by Web Proxy ");
+    }
+
+    @Override
+    public void update() throws ServiceException {
+        StringBuffer sb = new StringBuffer();
+        // check if zimbraDocumentEditingHost has been set
+        // if it is set, it means onlyoffice is to be deployed on a single
```
dasiyogesh, line 1124: comment here is this a assumptions if zimbraDocumentEditingHost is set there is single onlyoffice host?
zimbraDocumentEditingHost is server config so can be different for each server

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
```diff
@@ -1106,6 +1107,107 @@ public void update() throws ServiceException {
...
+        // server
+        String onlyofficeServer = mProv.getConfig().getDocumentEditingHost();
+        if (onlyofficeServer != null && onlyofficeServer.trim().length() > 0 && !defaultHost.equals(onlyofficeServer)) {
+            Server server = mProv.get(Key.ServerBy.name, onlyofficeServer);
+            sb.append(generateUpstreamBlock(server.getId(), onlyofficeServer));
+            mLog.debug("Added docservice upstream for single onlyoffice server : " + onlyofficeServer);
+        } else {
+            List<Server> mailclientservers = mProv.getAllMailClientServers();
```
thefoolwhocodes, line 1132: Broke use case:
1.Scenario where only few of the mailbox servers have running onlyoffice.

thefoolwhocodes (changes requested):

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
sreejan-chowdhury: The config is as such : 
1. We set the onlyoffice server in `zmprov mcf zimbraDocumentEditingHost <host>` on the box where nginx is running.
2. For Onlyoffice on each mailbox : we do not set it.

Is there any other way of setting it other than `zmprov mcf zimbraDocumentEditingHost <host>`?
If we set it in this way, won't it be reflected on all servers.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
sreejan-chowdhury: This isn't handled over here. The 2 scenarios that are handled here are : 
1. onlyoffice on a single server.
2. onlyoffice on each mailbox server

thefoolwhocodes: @sreejan-chowdhury we need to handle this as per discussion with Product,

sreejan-chowdhury: Changes have been done to handle the scenario too.

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
sreejan-chowdhury: It was a global, which has been changed to server level now.

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9892,7 +9892,7 @@ TODO: delete them permanently from here
   <defaultCOSValue>TRUE</defaultCOSValue>
   <desc>Whether or not document editing feature is enabled within briefcase</desc>
 </attr>
-<attr id="3098" name="zimbraDocumentEditingHost" type="string" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="9.1.0">
+<attr id="3098" name="zimbraDocumentServerHost" type="string" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="9.1.0">
   <globalConfigValue>zimbra.com</globalConfigValue>
```
Prashantsurana, line 9896: Why **zimbra.com** as default?

Prashantsurana (changes requested):

store/conf/attrs/zimbra-attrs.xml
sreejan-chowdhury: Not sure. The attribute has been renamed. I think it is set to zimbra.com to keep it similar to zimbraZulipBaseHost

sreejan-chowdhury: set to null by default

store/src/java/com/zimbra/cs/service/SharedFileServletContext.java
```diff
@@ -174,4 +174,5 @@ public static EncodedId decode(String encoded) throws ServiceException {
             }
         }
     }
+
```
dasiyogesh, line 177: adding empty line here not needed

store/src/java/com/zimbra/cs/util/ProxyConfGen.java
```diff
@@ -1106,6 +1108,114 @@ public void update() throws ServiceException {
     }
 }
 
+class OnlyOfficeDocServiceServersVar extends ProxyConfVar {
+
+    public OnlyOfficeDocServiceServersVar() {
+
```
dasiyogesh, line 1114: empty line not needed

common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
```diff
@@ -5720,6 +5712,14 @@ public static TwoFactorAuthSecretEncoding fromString(String s) throws ServiceExc
     @ZAttr(id=3101)
     public static final String A_zimbraDocumentRecentlyViewedEnabled = "zimbraDocumentRecentlyViewedEnabled";
 
+    /**
```
dasiyogesh, line 5715: changing attrs position here seems to have caused unnecessary changes in ZAttrConfig.java  when ran `ant generate-getters`

dasiyogesh (approved):

common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
sreejan-chowdhury: This was auto generated.

log2akshat (approved):

PR #1165, opened 2021-07-13
ZBUG-2334:- zmprov is broken when zimbraFeatureTwoFactorAuthRequired is TRUE after Z9P16, Z8.8.15P23

madhavdhuppe:
**Problem:** 

- ZMPROV is broken when zimbraFeatureTwoFactorAuthRequired is TRUE
- Throwing nullPointerException when zmprov is called after zimbraFeatureTwoFactorAuthRequired is TRUE

**Fix:** 

- Added check while doing admin auth for account.getServer() null because getServer will be null for zmprov account as it is not a admin created account.
- we don't need to check 2fa for zmprov account.

**Testing to be done by QA:** 

1. Should not throw null pointer exception when the user does any zmprov operation after zimbraFeatureTwoFactorAuthRequired is TRUE
2. Check if auth is working as before.

**NOTE**
View PR diff by selecting "hide whitespace changes"

### DIFF
--- a/store/src/java/com/zimbra/cs/service/admin/Auth.java
+++ b/store/src/java/com/zimbra/cs/service/admin/Auth.java
@@ -148,11 +148,9 @@ public class Auth extends AdminDocumentHandler {
                 String byStr = acctEl.getAttribute(AccountConstants.A_BY, AccountBy.name.name());
                 by = AccountBy.fromString(byStr);
             }
-            value = valuePassedIn;
-
-            try {
-
-                if (by == AccountBy.name && value.indexOf("@") == -1) {
+			value = valuePassedIn;
+			try {
+				if (by == AccountBy.name && value.indexOf("@") == -1) {
                     // first try to get by adminName, which resolves the account under cn=admins,cn=zimbra
                     // and does not need a domain
                     acct = prov.get(AccountBy.adminName, value, zsc.getAuthToken());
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/account/Account.java
```diff
@@ -245,10 +245,10 @@ public boolean isAccountExternal() throws ServiceException {
         return !server.mailTransportMatches(getAttr(Provisioning.A_zimbraMailTransport));
     }
 
-    public Server getServer() throws ServiceException {
-        String serverName = getAttr(Provisioning.A_zimbraMailHost);
-        return (serverName == null ? null : getProvisioning().get(Key.ServerBy.name, serverName));
-    }
+	public Server getServer() throws ServiceException {
```
madhavdhuppe, line 248: code formatting autofix

log2akshat (changes requested):
Please split your changes into 2 separate commits and push formatting changes in a separate commit.

madhavdhuppe:
> Please split your changes into 2 separate commits and push formatting changes in a separate commit.

I think we follow to maintain 1 commit for 1 ticket.
and pushing separate commit for formatting changes is not that good to break this.

Prashantsurana (changes requested):
Agreed with @log2akshat  , 
It is difficult to review this pull request. 
Please split formatting changes and code fix in separate commit.

Prashantsurana:
> > Please split your changes into 2 separate commits and push formatting changes in a separate commit.
> 
> I think we follow to maintain 1 commit for 1 ticket. & Add more commits if the changes are separate e.g 1 commit for modifying account & 1 for cos etc.
> 
> and pushing separate commit for formatting changes is not that good to break this coding standards

**We should have one merge commit for bug. Inside a merge commit you can have multiple changes,  this is standard git practice.**

PR #1167, opened 2021-07-13
ZBUG-2334:- Added code to skip two factor auth check when auth request is coming from ZMPROV

madhavdhuppe:
**Problem:** 

- ZMPROV is broken when zimbraFeatureTwoFactorAuthRequired is TRUE
- Throwing nullPointerException when zmprov is called after zimbraFeatureTwoFactorAuthRequired is TRUE

**Fix:** 

- Added check while doing admin auth for account.getServer() null because getServer will be null for zmprov account as it is not a admin created account.
- we don't need to check 2fa for zmprov account.

**Testing to be done by QA:** 

1. Should not throw null pointer exception when the user does any zmprov operation after zimbraFeatureTwoFactorAuthRequired is TRUE
2. Check if auth is working as before.

**Note:**
The whitespaces are unavoidable as they occurred due to the tabSpace added when we create wrapper if loop.
Please use hide whitespace setting from GitHub while reviewing PR.

### DIFF
--- a/common/src/java/com/zimbra/common/soap/AccountConstants.java
+++ b/common/src/java/com/zimbra/common/soap/AccountConstants.java
@@ -640,5 +640,4 @@ public class AccountConstants {
     public static final String E_SKIN_SELECTION_COLOR = "zimbraSkinSelectionColor";
     public static final String E_SKIN_FAVICON = "zimbraSkinFavicon";
     public static final String E_HOSTNAME = "hostname";
-    public static final String E_IS_SYSTEM_RESOURCE = "zimbraIsSystemResource";
 }
--- a/common/src/java/com/zimbra/common/soap/AccountConstants.java
+++ b/common/src/java/com/zimbra/common/soap/AccountConstants.java
@@ -640,4 +640,5 @@ public class AccountConstants {
     public static final String E_SKIN_SELECTION_COLOR = "zimbraSkinSelectionColor";
     public static final String E_SKIN_FAVICON = "zimbraSkinFavicon";
     public static final String E_HOSTNAME = "hostname";
+    public static final String E_IS_SYSTEM_RESOURCE = "zimbraIsSystemResource";
 }
--- a/store/src/java/com/zimbra/cs/service/admin/Auth.java
+++ b/store/src/java/com/zimbra/cs/service/admin/Auth.java
@@ -76,7 +76,6 @@ public class Auth extends AdminDocumentHandler {
         ZimbraSoapContext zsc = getZimbraSoapContext(context);
         AuthToken at = null;
         Account acct = null;
-        String acctValuePassedIn = null;
 
         Provisioning prov = Provisioning.getInstance();
         boolean csrfSupport = request.getAttributeBool(AccountConstants.A_CSRF_SUPPORT, false);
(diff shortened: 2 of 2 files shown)


sreejan-chowdhury (approved):

store/src/java/com/zimbra/cs/service/admin/Auth.java
```diff
@@ -185,135 +185,139 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
-                if (registerTrustedDevice && (trustedToken == null || trustedToken.isExpired())) {
-                    //generate a new trusted device token if there is no existing one or if the current one is no longer valid
-                    Map<String, Object> attrs = getTrustedDeviceAttrs(zsc, newDeviceId == null? deviceId: newDeviceId);
-                    TrustedDevices trustedDeviceManager = TwoFactorAuth.getFactory().getTrustedDevices(acct);
-                    if (trustedDeviceManager != null) {
-                        trustedToken = trustedDeviceManager.registerTrustedDevice(attrs);
-                    }
-                }
+                // Only perform 2fa authentication on admin account not on the resource accounts
+                if(acct != null && !Boolean.parseBoolean(acct.getAttr(AccountConstants.E_IS_SYSTEM_RESOURCE))) {
```
Prashantsurana, line 189: !acct.isIsSystemAccount()   -- should work

store/src/java/com/zimbra/cs/service/admin/Auth.java
```diff
@@ -185,135 +185,139 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
+	                String deviceId = request.getAttribute(AccountConstants.E_DEVICE_ID, null);
```
Prashantsurana, line 190: I prefer to put entire block into private method

common/src/java/com/zimbra/common/soap/AccountConstants.java
```diff
@@ -640,4 +640,5 @@
     public static final String E_SKIN_SELECTION_COLOR = "zimbraSkinSelectionColor";
     public static final String E_SKIN_FAVICON = "zimbraSkinFavicon";
     public static final String E_HOSTNAME = "hostname";
+    public static final String E_IS_SYSTEM_RESOURCE = "zimbraIsSystemResource";
```
Prashantsurana, line 643: if you use acct.isIsSystemAccount() this is not required.

Prashantsurana (changes requested):

store/src/java/com/zimbra/cs/service/admin/Auth.java
madhavdhuppe: We don't have any method named isIsSystemAccount() in account class are you suggesting create one ?

store/src/java/com/zimbra/cs/service/admin/Auth.java
madhavdhuppe: doing this

Prashantsurana (approved):
Thanks

madhavdhuppe:
commit a37774079845adbdf1ad68d8e3c95ae4c3822b1d (HEAD -> develop, origin/develop, origin/HEAD)
Author: Madhav Dhuppe <44395134+madhavdhuppe@users.noreply.github.com>
Date:   Fri Jul 16 13:21:09 2021 +0530

    ZBUG-2334:- Added code to skip two factor auth check when auth request is coming from ZMPROV (#1167)

    * ZBUG-2334:- Added code to skip two factor auth check when auth request is coming from ZMPROV

    * ZBUG-2334: Moved 2FA entire code into private method.

PR #1169, opened 2021-07-15
ZCS-10761: remove default value of zimbra.com from zimbraZulipBaseHost

dasiyogesh:
**Issue**
There is default value of zimbra.com set for zimbraZulipBaseHost LDAP attr. Once the extension loads on every domain creation there is zulip realm creation request gets fired.

**Fix**
Removed the default value of zimbra.com from zimbraZulipBaseHost. Zulip extension will check for this value first before initializing the extension, if its set domain listeners will register. If not set extension won't initialize and listeners won't get called for realm creation.

https://github.com/Zimbra/zm-zulip-proxy-ext/pull/2

### DIFF
--- a/store/src/java/com/zimbra/cs/account/ZAttrConfig.java
+++ b/store/src/java/com/zimbra/cs/account/ZAttrConfig.java
@@ -31972,6 +31972,21 @@ public abstract class ZAttrConfig extends Entry {
     }
 
     /**
+     * Blocked mobile device list for ActiveSync/ABQ
+     *
+     * @param attrs existing map to populate, or null to create a new map
+     * @return populated map to pass into Provisioning.modifyAttrs
+     *
+     * @since ZCS 8.9.0
+     */
+    @ZAttr(id=3090)
+    public Map<String,Object> unsetMobileBlockedDevices(Map<String,Object> attrs) {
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/account/ZAttrConfig.java
```diff
@@ -31971,6 +31971,21 @@ public void unsetMobileBlockedDevices() throws com.zimbra.common.service.Service
...
+    /**
+     * Blocked mobile device list for ActiveSync/ABQ
+     *
+     * @param attrs existing map to populate, or null to create a new map
+     * @return populated map to pass into Provisioning.modifyAttrs
+     *
+     * @since ZCS 8.9.0
+     */
+    @ZAttr(id=3090)
+    public Map<String,Object> unsetMobileBlockedDevices(Map<String,Object> attrs) {
```
prashantajari, line 31983: unsetMobileBlockedDevices attribute is not related to current fix. Do we need this change?

dasiyogesh: not required for the current change but this is auto generated when we run `ant generate-getters` after making change to zimbra-attrs.xml

itsmeprashant (approved):

prashantajari (approved):

PR #1186, opened 2021-08-20
ZBUG-2026: display cos id from domain if unset on account

dasiyogesh:
**Issue**
**zimbraCosId** if not set on account but set on domain as **zimbraDomainDefaultCosId** does not shows for account in `zmprov ga <account> zimbraCosId`

**Fix**
Add the cos Id from domain to list of account defaults in display of the zmprov get account request.

### DIFF
--- a/store/src/java/com/zimbra/cs/account/Account.java
+++ b/store/src/java/com/zimbra/cs/account/Account.java
@@ -409,8 +409,12 @@ public class Account extends ZAttrAccount implements GroupedEntry, AliasedEntry 
         Cos cos = getProvisioning().getCOS(this); // will set cos if not set yet
 
         Map<String, Object> defaults = null;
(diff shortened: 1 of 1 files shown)


sreejan-chowdhury (approved):

store/src/java/com/zimbra/cs/account/Account.java
```diff
@@ -412,6 +412,10 @@ public void setAccountDefaults(boolean setSecondaryDefaults) throws ServiceExcep
         if (cos != null)
             defaults = cos.getAccountDefaults();
 
+        if (cos != null && cos.getId() != null && this.getCOSId() == null) {
```
thefoolwhocodes, line 415: If possible merge to previous condition.

thefoolwhocodes (approved):

PR #1191, opened 2021-09-08
ZBUG-2316 : Changes to set desc html when missing

sreejan-chowdhury:
Problem : When sending meeting invites from MS teams Desktop, ALT-DESC param is not being sent. This causes the DescHTML not to be set while creating the invites. When we use EWS to add an account in Outlook, the meeting urls are removed.

Solution: Check if descHtml is missing and desc is present in the request. Convert the desc as html and add it is descHtml.

Test : 
Sent recurring and normal meetings from MS Teams to verify if the URLs are present or not.

### DIFF
--- a/store/src/java/com/zimbra/cs/mailbox/calendar/Invite.java
+++ b/store/src/java/com/zimbra/cs/mailbox/calendar/Invite.java
@@ -511,6 +511,11 @@ public class Invite {
         String desc = descInMeta ? meta.get(FN_DESC, null) : null;
         String descHtml = descInMeta ? meta.get(FN_DESC_HTML, null) : null;
 
+        // if desc html is missing but desc is present
+        if (desc != null && descHtml == null) {
+            descHtml = Util.textToHtml(desc);
+        }
+
         long completed = meta.getLong(FN_COMPLETED, 0);
 
         ParsedDateTime dtStart = null;
--- a/store/src/java/com/zimbra/cs/mailbox/calendar/Util.java
+++ b/store/src/java/com/zimbra/cs/mailbox/calendar/Util.java
@@ -23,6 +23,8 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import com.zimbra.common.calendar.Attach;
 import com.zimbra.common.calendar.Geo;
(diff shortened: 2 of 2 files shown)


store/src/java/com/zimbra/cs/mailbox/calendar/Util.java
```diff
@@ -351,4 +353,56 @@ public static Metadata encodeMetadata(Geo geo) {
...
+                    builder.append("&amp;");
+                    break;
+                case '"':
+                    builder.append("&quot;");
+                    break;
+                case '\n':
+                    builder.append("<br>");
+                    break;
+                case '\t':
+                    builder.append("&nbsp; &nbsp; &nbsp;");
```
dasiyogesh, line 394: for tab should be 4 spaces

store/src/java/com/zimbra/cs/mailbox/calendar/Util.java
```diff
@@ -351,4 +353,56 @@ public static Metadata encodeMetadata(Geo geo) {
...
+                if (previousWasASpace) {
+                    builder.append("&nbsp;");
+                    previousWasASpace = false;
+                    continue;
+                }
+                previousWasASpace = true;
+            } else {
+                previousWasASpace = false;
+            }
+            switch (c) {
```
dasiyogesh, line 377: is the text limited to only below characters matched in the switch, wouldn't there be any more characters in the description?

sreejan-chowdhury: This is to follow the same functionality as in : https://github.com/Zimbra/zm-sync-store/blob/739373e9f6aa1c7e62a8251036a74e9f1a847[…]384/src/java/com/zimbra/zimbrasync/commands/EmailSyncUtils.java 
No new changes have been done other than code porting.

store/src/java/com/zimbra/cs/mailbox/calendar/Util.java
sreejan-chowdhury: This is to follow the same functionality as in : https://github.com/Zimbra/zm-sync-store/blob/739373e9f6aa1c7e62a8251036a74e9f1a847[…]384/src/java/com/zimbra/zimbrasync/commands/EmailSyncUtils.java
No new changes have been done other than code porting.

store/src/java/com/zimbra/cs/mailbox/calendar/Invite.java
```diff
@@ -511,6 +511,11 @@ public static Invite decodeMetadata(int mailboxId, Metadata meta, CalendarItem c
         String desc = descInMeta ? meta.get(FN_DESC, null) : null;
         String descHtml = descInMeta ? meta.get(FN_DESC_HTML, null) : null;
 
+        // if desc html is missing but desc is present
+        if (desc != null && descHtml == null) {
```
thefoolwhocodes, line 515: Check descHtml first then go for desc..Else it could cause override of descHtml from desc.

thefoolwhocodes (approved):

dasiyogesh (approved):

store/src/java/com/zimbra/cs/mailbox/calendar/Util.java
```diff
@@ -351,4 +353,56 @@ public static Metadata encodeMetadata(Geo geo) {
...
         return meta;
     }
+
+    /**
+     * Converts text to html.
+     *
+     * @param String containing plain text
+     * @return String Html converted string
+     */
+    public static String textToHtml(String s) {
```
ronstra-synacor, line 363: This doesn't seem to belong here. Maybe com.zimbra.common.util.StringUtil?

PR #1208, opened 2021-11-17
ZCS-11066: update java identifier check for LDAP enum values to use the first character too for uniqueness

dasiyogesh:
**Issue**
LDAP attribute added for type enum with values starting with numeric is leading the compilation errors. For e.g. if the enum is set with values 10, 15, 30 here when generate getters is ran there first character from enum gets removed because for Character.isJavaIdentifierStart first character cannot be numeric and this leads duplicate enum values for '0' as from 10 & 30, 1 & 3 gets removed. This leads to below compilation error.

**Fix**
Java identifier check for enum values requires first character to be non numeric and using numeric enum as list for LDAP attrs enums causes duplicate method errors during compliation. As a fix made identifier check to include the numeric character too for maintaining uniqueness.

```
compile:
 [javac] Compiling 1 source file to /Users/psurana/git/zm-mailbox/common/build/classes
 [javac] /Users/psurana/git/zm-mailbox/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java:1670: error: variable _0 is already defined in enum PrefCalenderScaling
 [javac] _0("30");
 [javac] ^
 [javac] /Users/psurana/git/zm-mailbox/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java:1682: error: method is_0() is already defined in enum PrefCalenderScaling
 [javac] public boolean is_0() { return this == _0;}
 [javac] ^
 [javac] 2 errors

```

### DIFF
--- a/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
+++ b/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
@@ -956,7 +956,7 @@ public class ZAttrProvisioning {
 
     public static enum MtaNotifyClasses {
         bounce("bounce"),
-        _bounce("2bounce"),
+        _2bounce("2bounce"),
         data("data"),
         delay("delay"),
         policy("policy"),
@@ -973,7 +973,7 @@ public class ZAttrProvisioning {
              throw ServiceException.INVALID_REQUEST("invalid value: "+s+", valid values: "+ Arrays.asList(values()), null);
         }
         public boolean isBounce() { return this == bounce;}
-        public boolean is_bounce() { return this == _bounce;}
+        public boolean is_2bounce() { return this == _2bounce;}
         public boolean isData() { return this == data;}
         public boolean isDelay() { return this == delay;}
         public boolean isPolicy() { return this == policy;}
@@ -1665,9 +1665,9 @@ public class ZAttrProvisioning {
     }
 
     public static enum PrefCalenderScaling {
-        _0("10"),
-        _5("15"),
-        _10("30");
+        _10("10"),
+        _15("15"),
+        _30("30");
         private String mValue;
         private PrefCalenderScaling(String value) { mValue = value; }
         public String toString() { return mValue; }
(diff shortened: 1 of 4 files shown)


common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
```diff
@@ -973,7 +973,7 @@ public static MtaNotifyClasses fromString(String s) throws ServiceException {
              throw ServiceException.INVALID_REQUEST("invalid value: "+s+", valid values: "+ Arrays.asList(values()), null);
         }
         public boolean isBounce() { return this == bounce;}
-        public boolean is_bounce() { return this == _bounce;}
+        public boolean is_2bounce() { return this == _2bounce;}
```
Prashantsurana, line 976: can you please check references of this method. it should not break anything,

common/src/java/com/zimbra/common/util/StringUtil.java
```diff
@@ -889,7 +889,7 @@ public static String escapeJavaIdentifier(String s) {
         for (int i = 0; i < s.length(); i++) {
             char ch = s.charAt(i);
             if (i == 0) {
-                result.append(Character.isJavaIdentifierStart(ch) ? ch : "_");
+                result.append(Character.isJavaIdentifierStart(ch) ? ch : "_" + ch);
```
Prashantsurana, line 892: escapeJavaIdentifier can you please check references and see if there are any other places this utility function used.  look from github in zimbra org.

dasiyogesh: had checked it, its being only used here.

common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
dasiyogesh: didn't find any code references for `is_bounce` method

Prashantsurana (approved):
LGTM

PR #1209, opened 2021-11-17
ZBUG-2368: ProxyServlet domain check: substring match vulnerability

thefoolwhocodes:
**Problem**
- Vulnerability while domain match

**Testing**
- Test for the use cases mentioned in ticket.

### DIFF
--- a/store/src/java/com/zimbra/cs/zimlet/ProxyServlet.java
+++ b/store/src/java/com/zimbra/cs/zimlet/ProxyServlet.java
@@ -99,12 +99,12 @@ public class ProxyServlet extends ZimbraServlet {
 
     protected boolean checkPermissionOnTarget(URL target, AuthToken auth) {
         String host = target.getHost().toLowerCase();
-        ZimbraLog.zimlet.debug("checking allowedDomains permission on target host: "+host);
+        ZimbraLog.zimlet.debug("checking allowedDomains permission on target host: " + host);
         Set<String> domains;
         try {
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/zimlet/ProxyServlet.java
```diff
@@ -99,22 +99,22 @@
...
+            ZimbraLog.zimlet.info("error getting allowedDomains: " + se.getMessage());
             return false;
         }
         for (String domain : domains) {
-            if (domain.equals("*")) {
-                return true;
-            }
             if (domain.charAt(0) == '*') {
                 domain = domain.substring(1);
+                if (host.endsWith(domain)) {
```
dasiyogesh, line 113: what if there is vulnerable domain for proxy with wildcard, should we handle it? In that case we will still allow proxy for allowed domain list.

thefoolwhocodes: Can be controlled via allowed domain configuration.

girishbhamare (approved):
LGTM

abhishekdeshmukh2209 (approved):
LGTM

ronstra-synacor (approved):
LGTM as long as the administrators know they should use _both_ "zimbra.com" and "*.zimbra.com" if they want zimbra.com and its legitimate subdomains, and not just "*zimbra.com" which will match "foozimbra.com" etc. as the bug suggests.

PR #1210, opened 2021-11-18
ZBUG-2316: store text/html from message to calendar invite metadata

dasiyogesh:
**Issue**
Invite sent from MS Teams does not shows up the links from the invite since it uses x-alt-desc or html description from invite.

**Fix**
Since MS Team sent the invites text part with links enclosed in <> , this content is sent in the EWS response is sent as text type and EWS ignores this link. As a fix the email invite which is received contains html body within the multipart we can send this html content in EWS response so the content type is sent as HTML and the links display fine in the invite, for this storing the text/html description into invite metadata as `xDescH`.
If `xDescH` is missing as fallback sending text/plain description in EWS response with converting the text links to anchor tags.

**Testing done**
Verfied by injecting the test MIME can see the links fine in the invite after accepting the invite and message is moved the trash.

### DIFF
--- a/store/src/java/com/zimbra/cs/mailbox/Mailbox.java
+++ b/store/src/java/com/zimbra/cs/mailbox/Mailbox.java
@@ -1,7 +1,7 @@
 /*
  * ***** BEGIN LICENSE BLOCK *****
  * Zimbra Collaboration Suite Server
- * Copyright (C) 2004, 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016 Synacor, Inc.
+ * Copyright (C) 2004, 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2021 Synacor, Inc.
  *
  * This program is free software: you can redistribute it and/or modify it under
  * the terms of the GNU General Public License as published by the Free Software Foundation,
@@ -18,6 +18,7 @@ package com.zimbra.cs.mailbox;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.Reader;
 import java.io.StringWriter;
 import java.io.UnsupportedEncodingException;
 import java.io.Writer;
@@ -46,7 +47,9 @@ import java.util.concurrent.locks.ReentrantLock;
 import java.util.regex.Pattern;
 
 import javax.mail.Address;
+import javax.mail.MessagingException;
 import javax.mail.internet.MimeMessage;
+import javax.mail.internet.MimePart;
 
 import com.google.common.base.CharMatcher;
 import com.google.common.base.MoreObjects;
@@ -56,6 +59,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import com.google.common.io.Closeables;
 import com.googlecode.concurrentlinkedhashmap.ConcurrentLinkedHashMap;
 import com.zimbra.client.ZFolder;
 import com.zimbra.client.ZMailbox;
@@ -86,6 +90,7 @@ import com.zimbra.common.mailbox.ZimbraMailItem;
 import com.zimbra.common.mailbox.ZimbraQueryHitResults;
 import com.zimbra.common.mailbox.ZimbraSearchParams;
 import com.zimbra.common.mime.InternetAddress;
+import com.zimbra.common.mime.MimeConstants;
 import com.zimbra.common.mime.Rfc822ValidationInputStream;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.soap.SoapProtocol;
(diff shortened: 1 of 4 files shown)


abhishekdeshmukh2209 (approved):
LGTM

shrutig0510 (approved):

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -5851,6 +5863,48 @@ public AddInviteData addInvite(OperationContext octxt, Invite inv, int folderId,
         }
     }
 
+    public String getMsgHtmlDesc(ParsedMessage pm) throws ServiceException {
+        String htmlDesc = null;
+        if (pm == null)
+            return htmlDesc;
```
thefoolwhocodes, line 5869: nit: try enclosing in curly braces.

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -5851,6 +5863,48 @@ public AddInviteData addInvite(OperationContext octxt, Invite inv, int folderId,
...
+
+        Account acct = getAccount();
+        MimeMessage mm = pm.getMimeMessage();
+        List<MPartInfo> parts;
```
thefoolwhocodes, line 5873: nit: null assignment.

dasiyogesh: null assignment shouldn't be needed as in this case its empty and its being checked for empty list in the code below.

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
dasiyogesh: its a single line statement in if block, no need for curly braces

thefoolwhocodes: Technically it looks good. That's why mentioned as nit.

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
thefoolwhocodes: Technically it looks good. That's why mentioned as nit. As discussed let's do it.

store/src/java/com/zimbra/cs/mailbox/calendar/Invite.java
```diff
@@ -759,6 +763,12 @@ public synchronized String getDescription() throws ServiceException {
         return mDescription;
     }
 
+    public synchronized String getXZimbraDescriptionHtml() throws ServiceException {
+        return xZimbraDescHtml;
+    }
+    public synchronized void setXZimbraDescriptionHtml(String html) throws ServiceException {
+        xZimbraDescHtml = html;
+    }
```
thefoolwhocodes, line 771: nit: add space after method body.

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -5851,6 +5863,48 @@ public AddInviteData addInvite(OperationContext octxt, Invite inv, int folderId,
         }
     }
 
+    public String getMsgHtmlDesc(ParsedMessage pm) throws ServiceException {
```
thefoolwhocodes, line 5866: Technically this function can return null but seems like the callers are never checking for null return and going ahead with value settings.
Example:
```
if (firstInvite.getXZimbraDescriptionHtml() == null) {
	firstInvite.setXZimbraDescriptionHtml(mbox.getMsgHtmlDesc(pm));
}
```
There are many ways to better handle it.

thefoolwhocodes: Good to add some descriptive comment.

store/src/java/com/zimbra/cs/mailbox/calendar/Invite.java
```diff
@@ -759,6 +763,12 @@ public synchronized String getDescription() throws ServiceException {
         return mDescription;
     }
 
+    public synchronized String getXZimbraDescriptionHtml() throws ServiceException {
+        return xZimbraDescHtml;
+    }
+    public synchronized void setXZimbraDescriptionHtml(String html) throws ServiceException {
+        xZimbraDescHtml = html;
```
thefoolwhocodes, line 770: Null check can be added here also.

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
dasiyogesh: done

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
dasiyogesh: done

store/src/java/com/zimbra/cs/mailbox/calendar/Invite.java
dasiyogesh: done

thefoolwhocodes (approved):
LGTM

PR #1213, opened 2021-12-17
Zbug-1280 added isSpellCheckEnabled attribute to the GetInfoResponse

rcyarrapothu:
Jira ticket: https://jira.corp.synacor.com/browse/ZBUG-1280
Frontend PR: https://github.com/Zimbra/zm-web-client/pull/659

Retrieving list of urls using zimbraSpellCheckURL attribute and checking if it's empty on the primary server and adding it as a boolean value to the GetInfo response. This value will be used by frontend to display/hide the spell check button on compose page and mandatory spell check.

### DIFF
--- a/soap/src/java/com/zimbra/soap/account/message/GetInfoResponse.java
+++ b/soap/src/java/com/zimbra/soap/account/message/GetInfoResponse.java
@@ -77,8 +77,8 @@ import io.leangen.graphql.annotations.types.GraphQLType;
 @XmlType(propOrder = {"version", "accountId", "profileImageId", "accountName", "crumb", "lifetime", "adminDelegated", "restUrl",
         "quotaUsed", "isTrackingIMAP", "previousSessionTime", "lastWriteAccessTime", "recentMessageCount", "cos", "prefs", "attrs",
         "zimlets", "props", "identities", "signatures", "dataSources", "childAccounts", "discoveredRights",
-        "soapURL", "publicURL", "changePasswordURL", "license", "adminURL", "boshURL"})
-@JsonPropertyOrder({"version", "id", "profileImageId", "name", "crumb", "lifetime", "adminDelegated", "docSizeLimit", "attSizeLimit",
+        "soapURL", "publicURL", "changePasswordURL", "license", "adminURL", "boshURL", "spellCheckEnabled"})
+@JsonPropertyOrder({"version", "id", "profileImageId", "name", "crumb", "lifetime", "adminDelegated", "docSizeLimit", "spellCheckEnabled", "attSizeLimit",
         "rest", "used", "isTrackingIMAP", "prevSession", "accessed", "recent", "cos", "prefs", "attrs", "zimlets", "props", "identities",
         "signatures", "dataSources", "childAccounts", "discoveredRights", "soapURL", "publicURL", "license", "adminURL", "boshURL"})
 public final class GetInfoResponse {
@@ -96,6 +96,14 @@ public final class GetInfoResponse {
      */
     @XmlAttribute(name=AccountConstants.A_DOCUMENT_SIZE_LIMIT /* docSizeLimit */, required=false)
     private Long documentSizeLimit;
+
+    /**
+     * @zm-api-field-tag spell-check-enabled
+     * @zm-api-field-description returns true if the spell check is enabled on the server
+     */
+    @XmlElement(name=AccountConstants.A_IS_SPELL_CHECK_ENABLED /* isSpellCheckEnabled */, required=false)
+    @ZimbraJsonAttribute
+    private ZmBoolean spellCheckEnabled;
 
     /**
      * @zm-api-field-description Server version:
@@ -580,6 +588,14 @@ public final class GetInfoResponse {
         this.isTrackingIMAP = ZmBoolean.fromBool(trackingEnabled);
     }
 
+    @GraphQLQuery(name=GqlConstants.IS_SPELL_CHECK_ENABLED, description="Boolean value denoting if spell check is enabled on a server")
+    public Boolean getSpellCheckEnabled() {
+        return ZmBoolean.toBool(spellCheckEnabled, Boolean.FALSE);
+    }
+    public void setSpellCheckEnabled(Boolean spellCheckEnabled) {
+        this.spellCheckEnabled = ZmBoolean.fromBool(spellCheckEnabled);
+    }
+
     public MoreObjects.ToStringHelper addToStringInfo(
                 MoreObjects.ToStringHelper helper) {
         return helper
@@ -610,6 +626,7 @@ public final class GetInfoResponse {
             .add("boshURL", boshURL)
             .add("changePasswordURL", changePasswordURL)
             .add("license", license)
+            .add("isSpellCheckEnabled", spellCheckEnabled)
             .add("isTrackingIMAP", ZmBoolean.toBool(isTrackingIMAP) ? "1": "0");
 
     }
--- a/soap/src/java/com/zimbra/soap/account/message/GetInfoResponse.java
+++ b/soap/src/java/com/zimbra/soap/account/message/GetInfoResponse.java
@@ -98,8 +98,8 @@ public final class GetInfoResponse {
     private Long documentSizeLimit;
 
     /**
-     * @zm-api-field-tag spell-check-enabled
-     * @zm-api-field-description returns true if the spell check is enabled on the server
+     * @zm-api-field-tag spell-check-available
+     * @zm-api-field-description returns true if the spell check is available on the server
      */
     @XmlElement(name=AccountConstants.A_IS_SPELL_CHECK_AVAILABLE /* isSpellCheckAvailable */, required=false)
     @ZimbraJsonAttribute
--- a/soap/src/java/com/zimbra/soap/account/message/GetInfoResponse.java
+++ b/soap/src/java/com/zimbra/soap/account/message/GetInfoResponse.java
@@ -77,8 +77,8 @@ import io.leangen.graphql.annotations.types.GraphQLType;
 @XmlType(propOrder = {"version", "accountId", "profileImageId", "accountName", "crumb", "lifetime", "adminDelegated", "restUrl",
         "quotaUsed", "isTrackingIMAP", "previousSessionTime", "lastWriteAccessTime", "recentMessageCount", "cos", "prefs", "attrs",
         "zimlets", "props", "identities", "signatures", "dataSources", "childAccounts", "discoveredRights",
-        "soapURL", "publicURL", "changePasswordURL", "license", "adminURL", "boshURL", "spellCheckEnabled"})
-@JsonPropertyOrder({"version", "id", "profileImageId", "name", "crumb", "lifetime", "adminDelegated", "docSizeLimit", "spellCheckEnabled", "attSizeLimit",
+        "soapURL", "publicURL", "changePasswordURL", "license", "adminURL", "boshURL", "spellCheckAvailable"})
+@JsonPropertyOrder({"version", "id", "profileImageId", "name", "crumb", "lifetime", "adminDelegated", "docSizeLimit", "spellCheckAvailable", "attSizeLimit",
         "rest", "used", "isTrackingIMAP", "prevSession", "accessed", "recent", "cos", "prefs", "attrs", "zimlets", "props", "identities",
         "signatures", "dataSources", "childAccounts", "discoveredRights", "soapURL", "publicURL", "license", "adminURL", "boshURL"})
 public final class GetInfoResponse {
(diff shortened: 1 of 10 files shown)


store/src/java/com/zimbra/cs/service/account/GetInfo.java
```diff
@@ -477,6 +480,12 @@ private void doDiscoverRights(Element eRights, Account account, Set<Right> right
         DiscoverRights.discoverRights(account, rights, eRights, false);
     }
 
+    private boolean checkIfSpellCheckServiceIsAvailable(Server server) {
+        String[] installedServices = server.getMultiAttr(Provisioning.A_zimbraServiceInstalled);
+        String[] enabledServices = server.getMultiAttr(Provisioning.A_zimbraServiceEnabled);
+        return (Arrays.asList(installedServices).contains("spell") && Arrays.asList(enabledServices).contains("spell"));
+    }
+
```
k-kato, line 488: Spell check server does not always run on the mailstore server where an account exists.
For example, there are 2 mailstore server on a system. mailstore1 runs mailboxd and spell. mailstore2 runs mailboxd only and uses spell service which runs on mailstore1.
```
<attr id="267" name="zimbraSpellCheckURL" type="astring" max="256" cardinality="multi" optionalIn="globalConfig,server" flags="serverInherited">
  <desc>URL of the server running the spell checking service.  Multi-valued attribute that allows multiple spell check servers to be specified.  If the request to the first server fails, a request to the second server is sent and so on.</desc>
</attr>

$ zmprov gs `zmhostname` zimbraSpellCheckURL
# name zcs.local
zimbraSpellCheckURL: http://zcs.local:7780/aspell.php
```

soap/src/java/com/zimbra/soap/account/message/GetInfoResponse.java
```diff
@@ -610,6 +626,7 @@ public void setIsTrackingIMAP(Boolean trackingEnabled) {
             .add("boshURL", boshURL)
             .add("changePasswordURL", changePasswordURL)
             .add("license", license)
+                .add("isSpellCheckEnabled", spellCheckEnabled)
```
k-kato, line 629: fix indent

store/src/java/com/zimbra/cs/service/account/GetInfo.java
dasiyogesh: @rcyarrapothu with this change isn't it still checking for spell service on the local server? I think we need to fetch the servers and check the service in them.

k-kato: If my understanding is correct, `zimbraSpellCheckURL` in the server setting should be checked in the same way as CheckSpellingRequest handling.
```
com.zimbra.cs.service.mail.CheckSpelling
        // Make sure a spell check server is specified
        Provisioning prov = Provisioning.getInstance();
        Server localServer = prov.getLocalServer();
        String[] urls = localServer.getMultiAttr(Provisioning.A_zimbraSpellCheckURL);
        if (ArrayUtil.isEmpty(urls)) {
            log.warn("Unable to check spelling.  No value specified for %s", Provisioning.A_zimbraSpellCheckURL);
            return unavailable(response);
        }
```
`zimbraServiceEnabled spell` just shows whether Spell Service runs on the server or not. If `zimbraServiceEnabled spell` is set on a server, httpdis up on 7780 port at `zmcontrol start`. It does not mean that whether or not "Spell check function is available for the account who belongs to the server".

rcyarrapothu: @k-kato that's correct. Have modified the logic to consider "zimbraSpellCheckURL" instead of zimbraServiceEnabled/zimbraServiceInstalled. So now, setting the boolean value to false whenever zimbraSpellCheckURL returns empty on  the primary server. Thanks for your suggestion!

rcyarrapothu: @dasiyogesh that's was the commit while we were still discussing the approach. The latest commit checks for the zimbraSpellCheckURL.

store/src/java-test/com/zimbra/cs/service/account/GetInfoTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class GetInfoTest {
+
+    private static Server server = null;
+    private String[] servicesEnabled = null;
+    private String[] servicesInstalled = null;
+    private String[] spellCheckUrls = null;
```
k-kato, line 19: The three variables should not be member variables but be defined in each test case to keep dependency among tests.
Server settings should be set using `setAttributes(String[] servicesEnabled, String[] servicesInstalled, String[] urls)` in each test case for the same reason.

store/src/java-test/com/zimbra/cs/service/account/GetInfoTest.java
```diff
@@ -0,0 +1,94 @@
...
+        map.put("zimbraServiceEnabled", servicesEnabled);
+        map.put("zimbraServiceInstalled", servicesInstalled);
+        server.setAttrs(map);
+    }
+
+    /**
+     * spellcheck should return true when zimbraSpellCheckURL on the current server returns non-empty list of urls.
+     */
+    @Test
+    public void spellCheckIsTrueWhenItsEnabled() {
```
k-kato, line 47: I prefer not to use "Its" but describe a proper object name or such. Other functions as well.

store/src/java-test/com/zimbra/cs/service/account/GetInfoTest.java
```diff
@@ -0,0 +1,94 @@
...
+        boolean value = getInfo.isSpellCheckServiceAvailable(server);
+        Assert.assertEquals(false, value);
+    }
+
+    /**
+     * spellcheck should return true when zimbraSpellCheckURL on the current server returns non-empty list of urls.
+     * zimbraServiceInstalled and zimbraServiceEnabled shouldn't impact the return value of this method even "spell" service is not available on current server.
+     */
+    @Test
+    public void spellCheckIsFalseWhenItsNotEnabled() {
```
k-kato, line 71: Isn't it "spellChecIsTrue..." because `true` is an expected result?

store/src/java-test/com/zimbra/cs/service/account/GetInfoTest.java
```diff
@@ -0,0 +1,94 @@
...
+        GetInfo getInfo = new GetInfo();
+        boolean value = getInfo.isSpellCheckServiceAvailable(server);
+        Assert.assertEquals(true, value);
+    }
+
+    /**
+     * spellcheck should return false when zimbraSpellCheckURL on the current server returns an empty list of urls.
+     */
+    @Test
+    public void spellCheckIsTrueWhenItsEnabledFalse() {
```
k-kato, line 59: It's not clear what "ItsEnabledFalse" means.

store/src/java-test/com/zimbra/cs/service/account/GetInfoTest.java
```diff
@@ -0,0 +1,94 @@
...
+        boolean value = getInfo.isSpellCheckServiceAvailable(server);
+        Assert.assertEquals(true, value);
+    }
+
+    /**
+     * spellcheck should return false when zimbraSpellCheckURL on the current server returns with empty list of urls.
+     * zimbraServiceInstalled and zimbraServiceEnabled shouldn't impact the return value of this method even "spell" service is available on current server.
+     */
+    @Test
+    public void spellCheckIsTrueEvenWhenItsNotInstalled() {
```
k-kato, line 86: Isn't it "spellCheckIsFalseWhen..."?
True -> False?
Is "Even" necessary?
I don't know how "Enabled" or "Installed" is chosen for names of the functions.

store/src/java/com/zimbra/cs/service/account/GetInfo.java
```diff
@@ -176,6 +177,8 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
             Server server = prov.getLocalServer();
             if (server != null) {
                 response.addAttribute(AccountConstants.A_DOCUMENT_SIZE_LIMIT, server.getFileUploadMaxSize());
+                //For ZBUG-1280: checks if "spell" service is available from the list of enabled services on the server.
```
k-kato, line 180: Fix a comment. "spell" service is no longer checked.

k-kato (commented):
**Discussion:**
Is the word "isSpellCheck**Enabled**" appropriate? It looks like "Spell service is running" because we often see `zimbraServiceEnabled spell`.
Another option: "isSpellCheck**Available**"

store/src/java/com/zimbra/cs/service/account/GetInfo.java
rcyarrapothu: updated.

store/src/java-test/com/zimbra/cs/service/account/GetInfoTest.java
rcyarrapothu: updated.

store/src/java-test/com/zimbra/cs/service/account/GetInfoTest.java
rcyarrapothu: updated

store/src/java-test/com/zimbra/cs/service/account/GetInfoTest.java
rcyarrapothu: updated

store/src/java-test/com/zimbra/cs/service/account/GetInfoTest.java
rcyarrapothu: updated

store/src/java-test/com/zimbra/cs/service/account/GetInfoTest.java
rcyarrapothu: updated

rcyarrapothu:
> **Discussion:** Is the word "isSpellCheck**Enabled**" appropriate? It looks like "Spell service is running" because we often see `zimbraServiceEnabled spell`. Another option: "isSpellCheck**Available**"

Yeah isSpellCheckAvailable seems to be more appropriate here. Update the field name to **isSpellCheckAvailable**.

soap/src/java/com/zimbra/soap/account/message/GetInfoResponse.java
```diff
@@ -97,6 +97,14 @@
     @XmlAttribute(name=AccountConstants.A_DOCUMENT_SIZE_LIMIT /* docSizeLimit */, required=false)
     private Long documentSizeLimit;
 
+    /**
+     * @zm-api-field-tag spell-check-enabled
+     * @zm-api-field-description returns true if the spell check is enabled on the server
```
k-kato, line 102: For both annotations, enabled -> available

store/src/java-test/com/zimbra/cs/service/account/GetInfoTest.java
```diff
@@ -0,0 +1,104 @@
...
+        Map<String, Object> map = new HashMap<>();
+        map.put("zimbraSpellCheckURL", spellCheckUrls);
+        server.setAttrs(map);
+        GetInfo getInfo = new GetInfo();
+        boolean value = getInfo.isSpellCheckServiceAvailable(server);
+        Assert.assertEquals(false, value);
+    }
+
+    /**
+     * spellcheck should return false when zimbraSpellCheckURL on the current server returns an empty list of urls.
```
k-kato, line 55: "an empty list of urls" -> null

store/src/java/com/zimbra/cs/service/account/GetInfo.java
```diff
@@ -477,6 +480,16 @@ private void doDiscoverRights(Element eRights, Account account, Set<Right> right
         DiscoverRights.discoverRights(account, rights, eRights, false);
     }
 
+    /**
+     * This check is done for zbug-1280.
+     * checks and returns true when zimbraSpellCheckURL is empty on a server
```
k-kato, line 485: empty -> "not empty"

store/src/java-test/com/zimbra/cs/service/account/GetInfoTest.java
```diff
@@ -0,0 +1,104 @@
...
+        boolean value = getInfo.isSpellCheckServiceAvailable(server);
+        Assert.assertEquals(false, value);
+    }
+
+    /**
+     * spellcheck should return true when zimbraSpellCheckURL on the current server returns non-empty list of urls.
+     * zimbraServiceInstalled and zimbraServiceEnabled shouldn't impact the return value of this method when "spell" service is not available on current server.
+     */
+    @Test
+    public void trueWhenZimbraSpellCheckURLReturnsNonEmptyListWhenEnabled() {
```
k-kato, line 72: Isn't it "...WhenDisabled" since "spell" is not included in `servicesEnabled`?

k-kato (approved):
LGTM.
Please check my comment about function name and fix it as necessary. It does not affect any functionality, though.

dasiyogesh (approved):

PR #1216, opened 2022-01-11
ZBUG-1765

rupeshtelus:
Problem- While we are clicking next page on task view nothing is coming . That is only happening when we are checking for past dates.

Solution- In a query we had taken date validation and we were comparing with current date when we are clicking next page on task view but that is not required because it should show us the result for every dates .  

FixMode- So in Cursor object we are taking date validation . I am making null on that particular because in this case date validation is not required .   

Test Done - Yes

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/Search.java
+++ b/store/src/java/com/zimbra/cs/service/mail/Search.java
@@ -80,6 +80,13 @@ public class Search extends MailDocumentHandler  {
         OperationContext octxt = getOperationContext(zsc, context);
         fixBooleanRecipients(request);
         SearchRequest req = zsc.elementToJaxb(request);
+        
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/service/mail/Search.java
```diff
@@ -80,6 +80,13 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
         OperationContext octxt = getOperationContext(zsc, context);
         fixBooleanRecipients(request);
         SearchRequest req = zsc.elementToJaxb(request);
+        
+        // ZBUG-1765 ... task selection time date validation is not required
+        if( req.getSearchTypes().equalsIgnoreCase("task") && req.getCursor() != null) {
```
dasiyogesh, line 85: add space between if & (

PR #1217, opened 2022-01-11
Bugfix/zbug 651

swaatiTelus:
Problem: zmprov gal -v command was not showing correct member details.
Solution: 
1. Send false value for hideMembers attribute in a function.
2. Created a constructor which is not call addDlm() as this method is not allowing to get member details and added one parameter to distinguish the constructor.
Fix Mode: Modified code
Test Done: Unit test done(tried steps mentioned in tickets and able to get members information.)

### DIFF
--- a/store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
+++ b/store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
@@ -81,6 +81,13 @@ class SoapDistributionList extends DistributionList implements SoapEntry {
                 SoapProvisioning.getAttrs(e), prov);
         addDlm(e, getRawAttrs());
     }
+    
+    //ZBUG-651: This bug is about to get correct member details for gadl command;
+  	//ZBUG-651: Below constructor is added, which does not call addDlm() as this method is not allowing to get member details
+  	//ZBUG-651: Added one more boolean parameter in here, to differentiate from other constructor
+  	SoapDistributionList(DistributionListInfo dlInfo, Provisioning prov, boolean hideMembers) throws ServiceException {
+  		super(dlInfo.getName(), dlInfo.getId(), Attr.collectionToMap(dlInfo.getAttrList()), prov);
+  	}
 
     private void addDlm(List <String> members, Map<String, Object> attrs) {
         attrs.put(Provisioning.A_zimbraMailForwardingAddress,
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
```diff
@@ -81,6 +81,13 @@
                 SoapProvisioning.getAttrs(e), prov);
         addDlm(e, getRawAttrs());
     }
+    
+    //ZBUG-651: This bug is about to get correct member details for gadl command;
+  	//ZBUG-651: Below constructor is added, which does not call addDlm() as this method is not allowing to get member details
+  	//ZBUG-651: Added one more boolean parameter in here, to differentiate from other constructor
+  	SoapDistributionList(DistributionListInfo dlInfo, Provisioning prov, boolean hideMembers) throws ServiceException {
```
silentsakky, line 88: from where exactly we are calling this constructor form? most of the callers are using two params when creating instance, there is no caller which is using three params in object instantiation

swaatiTelus: Added the missed out changes into the PR. The changes were missed while squashing process.

silentsakky: on line no 62, we have a constructor which should ideally set member data properly, isn't working or not called at all?

swaatiTelus: Before calling addDlm(dlInfo.getMembers(), getRawAttrs()) method, we are getting members value. And in this method, we further calling getMembers() in which member value is not setting from anywhere hence it is zero. And inside addDlm(), we are putting member value in map and that's where member value gets override to zero.

rcyarrapothu:
Is PR#1220 solving the same bug? Is there a reason why you had 2 separate PRs for these changes? I would recommend to merge these changes into one or close the PR that you don't want to merge.

silentsakky:
> Is PR#1220 solving the same bug? Is there a reason why you had 2 separate PRs for these changes? I would recommend to merge these changes into one or close the PR that you don't want to merge.

Yes, this PR was created by swati first but I didn't like the fix so I have asked her to go with different approach so she created another PR, so we should compare both PR and pick the best suited solution, both PRs are fixing same issue so functionally both are same

swaatiTelus:
Another PR gas been created for this issue:
https://github.com/Zimbra/zm-mailbox/pull/1220

PR #1218, opened 2022-01-11
Bugfix/zbug 1765

rupeshtelus:
Problem- While we are clicking next page on task view nothing is coming . That is only happening when we are checking for past dates.

Solution- In a query we had taken date validation and we were comparing with current date when we are clicking next page on task view but that is not required because it should show us the result for every dates .  

FixMode- So in Cursor object we are taking date validation . I am making null on that particular because in this case date validation is not required .   

Test Done - Yes

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/Search.java
+++ b/store/src/java/com/zimbra/cs/service/mail/Search.java
@@ -80,6 +80,13 @@ public class Search extends MailDocumentHandler  {
         OperationContext octxt = getOperationContext(zsc, context);
         fixBooleanRecipients(request);
         SearchRequest req = zsc.elementToJaxb(request);
+        
+        // ZBUG-1765 ... task selection time date validation is not required
+        if( req.getSearchTypes().equalsIgnoreCase("task") && req.getCursor() != null) {
+        	req.setCursor(null);
+        }
+        
+        
         if (MoreObjects.firstNonNull(req.getWarmup(), false)) {
             mbox.index.getIndexStore().warmup();
             return zsc.createElement(MailConstants.SEARCH_RESPONSE);
--- a/store/src/java/com/zimbra/cs/service/mail/Search.java
+++ b/store/src/java/com/zimbra/cs/service/mail/Search.java
@@ -82,7 +82,7 @@ public class Search extends MailDocumentHandler  {
         SearchRequest req = zsc.elementToJaxb(request);
         
         // ZBUG-1765 ... task selection time date validation is not required
-        if( req.getSearchTypes().equalsIgnoreCase("task") && req.getCursor() != null) {
+        if ( req.getSearchTypes().equalsIgnoreCase("task") && req.getCursor() != null) {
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/service/mail/Search.java
```diff
@@ -80,6 +80,13 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
         OperationContext octxt = getOperationContext(zsc, context);
         fixBooleanRecipients(request);
         SearchRequest req = zsc.elementToJaxb(request);
+        
+        // ZBUG-1765 ... task selection time date validation is not required
+        if ( req.getSearchTypes().equalsIgnoreCase("task") && req.getCursor() != null) {
+        	req.setCursor(null);
+        }
+
```
dasiyogesh, line 88: remove extra empty line

rupeshtelus: removed empty lines

store/src/java/com/zimbra/cs/service/mail/Search.java
```diff
@@ -80,6 +80,11 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
         OperationContext octxt = getOperationContext(zsc, context);
         fixBooleanRecipients(request);
         SearchRequest req = zsc.elementToJaxb(request);
+        // ZBUG-1765 ... task selection time date validation is not required
+        if ( req.getSearchTypes().equalsIgnoreCase("task") && req.getCursor() != null) {
```
silentsakky, line 84: if the functionality is working fine in modern ui then why we need to put a fix in the API layer? did we analyse if modern ui and classic ui are using different request parameters?

rupeshtelus: Yes we are using different request parameters like offset  and limit . We are taking different parameters like in modern ui we are taking limit as 1000 so at a time 1000 tasks will come with a scrollbar
and for classic ui we are using pagination so for each page limit we are taking 50.. 1st time offset will be 0 and limit 50 then in next page offset will be 50 limit 50 like this we are retrieving 50 tasks at a time  … So limit and offset parameter is different in both ui

silentsakky: ok thanks for info
so as per changes, we don't support use of cursor in tasks Search request, any reason for this deviation? do we support cursor in mails Search request?
What would be implication of not supporting it and can we get documentation updated for this change?

rupeshtelus: I don’t have idea why that validation was there but I can tell we were getting current date in cursor object and we were comparing it in a query to get data. So we were not facing issue on future date but facing issue on past dates . As my view I have checked there is no use of dates on task labels .

Wherever we have date or time validation we need cursor object with particular date value that we can compare it in queries .

Ya I can provide but in which document I will update ? Or should I create a new document and provide?

PR #1219, opened 2022-01-11
ZBUG-2397: Added cookie SameSite=Strict 

zimsuchitgupta:
Supported SameSite=Strict cookie in ZM_AUTH_TOKEN cookie.
For this added a localconfig varible in LC.java named **zimbra_same_site_cookie**
By default value will be '**Strict**' but user can change it to **Lax**, **None** and ""(**empty**) for existing behavior.

### DIFF
--- a/common/src/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/java/com/zimbra/common/localconfig/LC.java
@@ -863,6 +863,7 @@ public final class LC {
     public static final KnownKey zimbra_deregistered_authtoken_queue_size = KnownKey.newKey(5000);
     public static final KnownKey zimbra_jwt_cookie_size_limit = KnownKey.newKey(4096);
     public static final KnownKey zimbra_authtoken_cookie_domain = KnownKey.newKey("");
+    public static final KnownKey zimbra_same_site_cookie = KnownKey.newKey("Strict");
     public static final KnownKey zimbra_zmjava_options = KnownKey.newKey("-Xmx256m" +
             " -Dhttps.protocols=TLSv1.2,TLSv1.3" +
             " -Djdk.tls.client.protocols=TLSv1.2,TLSv1.3");
--- a/common/src/java/com/zimbra/common/util/ZimbraCookie.java
+++ b/common/src/java/com/zimbra/common/util/ZimbraCookie.java
@@ -47,6 +47,12 @@ public class ZimbraCookie {
             cookie.setDomain(LC.zimbra_authtoken_cookie_domain.value());
         }
 
+        String cookieVal = LC.zimbra_same_site_cookie.value();
+        if (!StringUtil.isNullOrEmpty(cookieVal)) {
+            // setting cookie value like "SameSite=Strict;", value can be Strict, Lax, None
+            path = new StringBuilder(path).append(";SameSite=").append(cookieVal).append(";").toString();
+        }
+
         cookie.setPath(path);
     }
 
--- a/common/src/java/com/zimbra/common/util/ZimbraCookie.java
+++ b/common/src/java/com/zimbra/common/util/ZimbraCookie.java
@@ -76,7 +76,12 @@ public class ZimbraCookie {
         }
         ZimbraCookie.setAuthTokenCookieDomainPath(cookie, ZimbraCookie.PATH_ROOT);
 
-        cookie.setSecure(secure);
+        String cookieVal = LC.zimbra_same_site_cookie.value();
+        if (!StringUtil.isNullOrEmpty(cookieVal)) {
+            cookie.setSecure(true);
+        } else {
+            cookie.setSecure(secure);
+        }
 
         if (httpOnly) {
             cookie.setHttpOnly(httpOnly);
--- a/common/src/java/com/zimbra/common/util/ZimbraCookie.java
+++ b/common/src/java/com/zimbra/common/util/ZimbraCookie.java
@@ -47,12 +47,6 @@ public class ZimbraCookie {
             cookie.setDomain(LC.zimbra_authtoken_cookie_domain.value());
         }
 
-        String cookieVal = LC.zimbra_same_site_cookie.value();
-        if (!StringUtil.isNullOrEmpty(cookieVal)) {
-            // setting cookie value like "SameSite=Strict;", value can be Strict, Lax, None
-            path = new StringBuilder(path).append(";SameSite=").append(cookieVal).append(";").toString();
-        }
-
         cookie.setPath(path);
     }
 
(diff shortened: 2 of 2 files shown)


shrutig0510 (approved):

thefoolwhocodes (changes requested):
Please update testing details.

common/src/java/com/zimbra/common/localconfig/LC.java
```diff
@@ -863,6 +863,7 @@
     public static final KnownKey zimbra_deregistered_authtoken_queue_size = KnownKey.newKey(5000);
     public static final KnownKey zimbra_jwt_cookie_size_limit = KnownKey.newKey(4096);
     public static final KnownKey zimbra_authtoken_cookie_domain = KnownKey.newKey("");
+    public static final KnownKey zimbra_same_site_cookie = KnownKey.newKey("Strict");
```
dasiyogesh, line 866: by default should the value be Strict or None? 
Since this fix is addressing this header not be used in zimbraReverseProxyResponseHeaders and there its not default, so should we keep it as Strict here

silentsakky: It's better to go with Strict as default value, we don't want to compromise on security aspect
Just need to be sure that we are not regressing anything on the ui clients
we should document this if customer wants to go to earlier behaviour then they need to change value to None, although it's not recommended by Zimbra

barrydegraaff:
We probably should document exactly what we are changing and where and create a test build so that Q&A can verify this change.

While the code looks good, a fail here means people can potentially no longer log in.

common/src/java/com/zimbra/common/util/ZimbraCookie.java
```diff
@@ -47,6 +47,12 @@ public static void setAuthTokenCookieDomainPath(Cookie cookie, String path) {
             cookie.setDomain(LC.zimbra_authtoken_cookie_domain.value());
         }
 
+        String cookieVal = LC.zimbra_same_site_cookie.value();
+        if (!StringUtil.isNullOrEmpty(cookieVal)) {
+            // setting cookie value like "SameSite=Strict;", value can be Strict, Lax, None
+            path = new StringBuilder(path).append(";SameSite=").append(cookieVal).append(";").toString();
```
jeastman, line 53: If the `cookieVal` is set to `None`, you will need to ensure that the `Secure` flag is also set. If `Secure` is not set when the `SameSite` value is `None`, the cookie will be blocked. It would be possible to accidentally block all cookies by simply setting the config value to `None`.

May also want to consider the case when cookies are being cleared (value of cooke set to `""`). At a minimum, existing cookies may not be cleared once this change is deployed (and cookie updates restricted).

common/src/java/com/zimbra/common/localconfig/LC.java
jeastman: I agree the default value should be `Strict`.

barrydegraaff:
see also: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite

common/src/java/com/zimbra/common/util/ZimbraCookie.java
```diff
@@ -76,7 +76,12 @@ private static void addCookie(HttpServletResponse response, String name, String
         }
         ZimbraCookie.setAuthTokenCookieDomainPath(cookie, ZimbraCookie.PATH_ROOT);
 
-        cookie.setSecure(secure);
+        String cookieVal = LC.zimbra_same_site_cookie.value();
+        if (!StringUtil.isNullOrEmpty(cookieVal)) {
+            cookie.setSecure(true);
+        } else {
+            cookie.setSecure(secure);
+        }
```
jeastman, line 84: If we are going to maintain this logic here, we should ensure that all of the code is actually using this method to add cookies. I was able to find at least one example where `setAuthTokenCookieDomainPath` is called directly and not followed by a call to `addCookie`. (see https://github.com/Zimbra/zm-saml-consumer-store/blob/b37b0409787a490e5824b78c2e6807c1320313fa/src/java/com/zimbra/cs/security/saml/SamlLogoutReceiverHandler.java#L95)

zimsuchitgupta:
Testing Instruction:

**1** - On login in webclient, by default user can see the SameSite=Strict cookie for ZM_AUTH_TOKEN.
http://synscreen.buf.synacor.com/view/a5d6a2.png
**2** - User can control behavior using zmlocalconfig

zmlocalconfig -e zimbra_same_site_cookie="Strict" 
User can provide value like Strict, Lax, None or "" (Empty)

**3** - To disable this behavior we can set value of SameSite=""
http://synscreen.buf.synacor.com/view/a7d8d2.png

common/src/java/com/zimbra/common/util/ZimbraCookie.java
zimsuchitgupta: @jeastman  moved implementation in addCookie method also every module mentor have to backport these changes to support SameSite=Strict.

thefoolwhocodes (approved):

barrydegraaff (approved):

dasiyogesh (approved):

log2akshat (approved):

ronstra-synacor (approved):

common/src/java/com/zimbra/common/util/ZimbraCookie.java
```diff
@@ -70,7 +70,14 @@ private static void addCookie(HttpServletResponse response, String name, String
         }
         ZimbraCookie.setAuthTokenCookieDomainPath(cookie, ZimbraCookie.PATH_ROOT);
 
-        cookie.setSecure(secure);
+        String cookieVal = LC.zimbra_same_site_cookie.value();
```
silentsakky, line 73: Hope this code is used in JWT based auth also, as it uses ZM_JWT cookie

barrydegraaff:
@uttamtakalkar can you verify that this change does not break NG Modules?

NG Modules make requests to ng-activation.com (and other domains). In case it breaks, Zextras has a 2,5 month grace period after which the NG Modules de-activate. So even if it may work now, it may not work 2,5 months from now. Probably this can be either seen in the logs. Or we could ask Zextras.

ghen2:
FYI, SameSite=Strict breaks "preauth" authentication (https://wiki.zimbra.com/wiki/Preauth) when coming from a portal on another hostname.  We need SameSite=Lax in this scenario.  

(tested by playing with proxy_cookie_flags on nginx)

zimsuchitgupta:
This work has been taken care in new PR : https://github.com/Zimbra/zm-mailbox/pull/1225 so closing this.

PR #1220, opened 2022-01-12
Bugfix/zbug 651 refix

swaatiTelus:
Problem: zmprov gal -v command was not showing correct member details.
Solution:

1. addDlm() is modified as it is overriding the existing member/s value with zero against zimbraMailForwardingAddress attribute. Now checking if members has some data then only it will allow to override else not.
2. Since other constructors are also calling this method which might need empty list of members so created a new resetDlm() which is adding blank list of members against the attribute.
3. Send false value for hideMembers attribute in a function.

Fix Mode: Modified code

Test Done: Unit test done(tried steps mentioned in tickets and able to get members information.)

### DIFF
--- a/store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
+++ b/store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
@@ -18,6 +18,7 @@
 package com.zimbra.cs.account.soap;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -52,7 +53,7 @@ class SoapDistributionList extends DistributionList implements SoapEntry {
         super(dlInfo.getName(), dlInfo.getId(),
                 new HashMap<String,Object>(), prov);
         // DistributionListMembershipInfo does not supply membership info
-        addDlm(new ArrayList<String>(), getRawAttrs());
+        resetDlm(getRawAttrs());
     }
 
     SoapDistributionList(DistributionListInfo dlInfo, Provisioning prov)
@@ -72,7 +73,7 @@ class SoapDistributionList extends DistributionList implements SoapEntry {
         attrs.put(Provisioning.A_zimbraId, dlInfo.getId());
         
         // DLInfo does not supply membership info
-        addDlm(new ArrayList<String>(), getRawAttrs());
+        resetDlm(getRawAttrs());
     }
 
     SoapDistributionList(Element e, Provisioning prov) throws ServiceException {
@@ -83,15 +84,17 @@ class SoapDistributionList extends DistributionList implements SoapEntry {
     }
     
     //ZBUG-651: This bug is about to get correct member details for gadl command;
-  	//ZBUG-651: Below constructor is added, which does not call addDlm() as this method is not allowing to get member details
-  	//ZBUG-651: Added one more boolean parameter in here, to differentiate from other constructor
-  	SoapDistributionList(DistributionListInfo dlInfo, Provisioning prov, boolean hideMembers) throws ServiceException {
-  		super(dlInfo.getName(), dlInfo.getId(), Attr.collectionToMap(dlInfo.getAttrList()), prov);
-  	}
-
+    //ZBUG-651: Below method is modified as it is overridding the existing member/s value with zero against zimbraMailForwardingAddress attribute. Now checking if members has some data then only it will allow to override else not
     private void addDlm(List <String> members, Map<String, Object> attrs) {
-        attrs.put(Provisioning.A_zimbraMailForwardingAddress,
+    	if(members.size() > 0)
+    		attrs.put(Provisioning.A_zimbraMailForwardingAddress,
                 members.toArray(new String[members.size()]));
+    }
+    
+  //ZBUG-651: Added one more method in here against the change in addDlm(), so if some method need blank list of members below method can provide.
+    private void resetDlm(Map<String, Object> attrs) {
+    	attrs.put(Provisioning.A_zimbraMailForwardingAddress,
+    			new ArrayList<String>().toArray(new String[0]));
     }
 
     private void addDlm(Element e, Map<String, Object> attrs) {
--- a/store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
+++ b/store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
@@ -18,7 +18,6 @@
 package com.zimbra.cs.account.soap;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
```diff
@@ -81,11 +81,20 @@
                 SoapProvisioning.getAttrs(e), prov);
         addDlm(e, getRawAttrs());
     }
-
+    
+    //ZBUG-651: This bug is about to get correct member details for gadl command;
```
dasiyogesh, line 85: add the comments to method documentation here and below

store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
```diff
@@ -81,11 +81,20 @@
...
+    //ZBUG-651: Below method is modified as it is overridding the existing member/s value with zero against zimbraMailForwardingAddress attribute. Now checking if members has some data then only it will allow to override else not
     private void addDlm(List <String> members, Map<String, Object> attrs) {
-        attrs.put(Provisioning.A_zimbraMailForwardingAddress,
+    	if(members.size() > 0)
```
dasiyogesh, line 88: add space between if & (

store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
dasiyogesh, line 85: this comment `//ZBUG-651: This bug is about to get correct member details for gadl command;` is not required here can move it to method body where the change is made

swaatiTelus: Method documentation is added.

store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
```diff
@@ -81,11 +81,31 @@
...
+    /**
+     * This method is modified as it is overriding the existing member/s value with zero against zimbraMailForwardingAddress attribute.
+     * So now checking if members list is zero than it should not override. (ZBUG-651)
+     * @param members This is list of members
+     * @param attrs This is list of attribute
+     * @return nothing
+     */
     private void addDlm(List <String> members, Map<String, Object> attrs) {
-        attrs.put(Provisioning.A_zimbraMailForwardingAddress,
+    	if (members.size() > 0)
```
log2akshat, line 94: Use `{}` style for the `if` condition.

store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
```diff
@@ -81,11 +81,31 @@
...
+    		attrs.put(Provisioning.A_zimbraMailForwardingAddress,
                 members.toArray(new String[members.size()]));
     }
+    
+    /**
+     * This method is added against the change in addDlm(), which provides blank list of members.
+     * @param attrs to put/override attribute with blank member list.
+     * @return nothing
+     * ZBUG-651
+     */
```
log2akshat, line 104: Please draft the comment starting with the bug number:
`ZBUG-651: This method is added against the change in addDlm(), which...`

log2akshat (changes requested):

store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
swaatiTelus: Done.

store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
```diff
@@ -81,10 +81,29 @@
                 SoapProvisioning.getAttrs(e), prov);
         addDlm(e, getRawAttrs());
     }
-
+    
+    /**
+     * ZBUG-651: This method is modified as it is overriding the existing member/s value with zero against zimbraMailForwardingAddress attribute.
```
log2akshat, line 86: Avoid lines longer than 120 characters.

store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
swaatiTelus: Done.

store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
swaatiTelus: Done

store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
swaatiTelus: Done.

desouzas:
A few notes:
* Test this with some dynamic distribution lists, one that is defined by a manually added `memberURL`, and one that is not. The counts/list probably will not track with those.
  * `zmprov cddl test-ddl@zimbra.com`
  * `zmprov adlm test-ddl@zimbra.com test-user@zimbra.com`
  * `zmprov ca dadmin@zimbra.com passhere zimbraIsDelegatedAdminAccount TRUE`
  * `zmprov cddl da@zimbra.com memberURL 'ldap:///??sub?(&(objectClass=zimbraAccount)(zimbraIsDelegatedAdminAccount=TRUE))' zimbraIsACLGroup FALSE`
* The PR comment mentions unit tests, but I do not see any added in this PR? Can we add a few to easily test code changes locally?

store/src/java/com/zimbra/cs/account/soap/SoapDistributionList.java
```diff
@@ -81,10 +81,30 @@
...
+    }
+    
+    /**
+     * ZBUG-651: This method is added against the change in addDlm(), which provides blank list of members.
+     * @param attrs to put/override attribute with blank member list.
+     * @return nothing
+     */
+    private void resetDlm(Map<String, Object> attrs) {
+    	attrs.put(Provisioning.A_zimbraMailForwardingAddress,
+    			new ArrayList<String>().toArray(new String[0]));
```
ronstra-synacor, line 107:
```suggestion
    			new String[0]);
```
Unless there's some reason to create an empty list just to get an empty array that I'm missing.

silentsakky (approved):

swaatiTelus:
@desouzas  I added case tested on my local server.

log2akshat (approved):

thefoolwhocodes (approved):

ronstra-synacor (approved):

rcyarrapothu (approved):

desouzas:
For consideration:
* The members list (at the bottom of `zmprov gadl -v` output) for DDLs prints in ldap path format instead of email address format
  * This is a different output than what the ldap flag version shows, and may eventually come back to us in a new bug ticket.
* No unit tests added with this.
* I wonder if the change in `SoapDynamicGroup` has performance considerations with a large number of complicated dynamic groups. Any reason to bypass the `memberList` class property entirely instead of filling it out?

swaatiTelus:
> For consideration:
> * The members list (at the bottom of `zmprov gadl -v` output) for DDLs prints in ldap path format instead of email address format.
>   * This is a different output than what the ldap flag version shows, and may eventually come back to us in a new bug ticket.==> Yes, Stephan we have created a new ticket(ZBUG-2619) for the same, since we did not consider this while grooming this bug hence did not assign the story points for it.
> * No unit tests added with this. ==> We have done the manual testing on this whose steps are mentioned in the PR. And regarding Unit tests, I was facing some difficulties that I conveyed you over Slack also.
> * I wonder if the change in `SoapDynamicGroup` has performance considerations with a large number of complicated dynamic groups. Any reason to bypass the `memberList` class property entirely instead of filling it out?==> We were already getting data in attribute 'member' and found out other part of code is getting members list against attribute value.
, so used the same.

PR #1225, opened 2022-02-01
ZBUG-2397: Added cookie SameSite=Strict and zmlocalconfig attribute

zimsuchitgupta:
Supported SameSite=Strict cookie in ZM_AUTH_TOKEN cookie.
For this added a localconfig varible in LC.java named zimbra_same_site_cookie
By default value will be 'Strict' but user can change it to Lax, None and ""(empty) for existing behavior.

**Testing Instruction:**

**1** - On login in webclient, by default user can see the SameSite=Strict cookie for ZM_AUTH_TOKEN.
http://synscreen.buf.synacor.com/view/a5d6a2.png
**2** - User can control behavior using zmlocalconfig

zmlocalconfig -e zimbra_same_site_cookie="Strict"
User can provide value like Strict, Lax, None or "" (Empty)

**3** - To disable this behavior we can set value of SameSite=""
http://synscreen.buf.synacor.com/view/a7d8d2.png

### DIFF
--- a/common/src/java/com/zimbra/common/util/ZimbraCookie.java
+++ b/common/src/java/com/zimbra/common/util/ZimbraCookie.java
@@ -72,8 +72,10 @@ public class ZimbraCookie {
 
         String cookieVal = LC.zimbra_same_site_cookie.value();
         if (!StringUtil.isNullOrEmpty(cookieVal)) {
+            String pathStr = cookie.getPath();
             // setting cookie value like "SameSite=Strict;", value can be Strict, Lax, None
-            path = new StringBuilder(path).append(";SameSite=").append(cookieVal).append(";").toString();
+            pathStr = new StringBuilder(pathStr).append(";SameSite=").append(cookieVal).append(";").toString();
+            cookie.setPath(pathStr);
             cookie.setSecure(true);
         } else {
             cookie.setSecure(secure);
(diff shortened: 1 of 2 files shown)


common/src/java/com/zimbra/common/util/ZimbraCookie.java
```diff
@@ -70,7 +70,14 @@ private static void addCookie(HttpServletResponse response, String name, String
         }
         ZimbraCookie.setAuthTokenCookieDomainPath(cookie, ZimbraCookie.PATH_ROOT);
 
-        cookie.setSecure(secure);
+        String cookieVal = LC.zimbra_same_site_cookie.value();
```
silentsakky, line 73: hope this is used in JWT authentication also, can we check it?

silentsakky: can we also add unittests for this change?

zimsuchitgupta: @silentsakky Above method is also called for JWT token also can see here https://github.com/Zimbra/zm-mailbox/blob/develop/store/src/java/com/zimbra/cs/account/ZimbraJWToken.java#L301 but need to verify by doing testing.

log2akshat (approved):

common/src/java/com/zimbra/common/util/ZimbraCookie.java
zimsuchitgupta: Thanks @silentsakky  for you help verifying feature for JWT token, as you shared and I also verified that this is working for JWT token too. 
JWT Strict

PR #1229, opened 2022-02-15
Zbug 265: first pass at zmprov glps

jholder85638:
Feature to get the LDAP pool status via zmprov. This is useful for knowing the failure stats.


desouzas:
@jholder85638 something is wrong with this branch. Maybe git config line ending issues or wrong base branch?

PR #1231, opened 2022-02-15
ZBUG-1869: Setting effective quota in zimbraMailQuota

zimsuchitgupta:
**Problem**
* The value of `zimbraMailQuota` that we return in GetInfo is not the actual value that the server respects when limiting the user if `zimbraMailDomainQuota` contains a lower value than it.
  * This behavior is documented.
  * This behavior exists in order to support various configuration options on shared environments by admins/domain-admins.

**Solution**
* We can adjust the value of `zimbraMailQuota` that we return in `GetInfo` to respect the effective value (minimum between the relevant properties described above).
* We can also get the value of `zimbraMailQuota` from Admin `GetAccountRequest` so for consistency setting the effective quota value in `zimbraMailQuota` attribute.

**Pros**
* Keeps logic of choosing minimum between the two options in one place (the API).
* No UI or CLI tool changes needed to reflect the change (don't need to replicate the logic across the various clients we support).

**Cons**
* The actual value on the user is no longer exposed - only the effective value - so if any other clients need the actual value for some reason, that may be an issue.

### DIFF
--- a/common/src/java/com/zimbra/common/soap/AdminConstants.java
+++ b/common/src/java/com/zimbra/common/soap/AdminConstants.java
@@ -1342,7 +1342,8 @@ public final class AdminConstants {
 
     public static final String A_QUOTA_USED = "used";
     public static final String A_QUOTA_LIMIT = "limit";
-
+    public static final String A_EFFECTIVE_QUOTA = "effectiveQuota";
+    
     public static final String E_TEMPLATE = "template";
     public static final String E_TEST = "test";
     public static final String A_DEST = "dest";
--- a/soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java
+++ b/soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java
@@ -54,6 +54,17 @@ public class GetAccountRequest extends AttributeSelectorImpl {
     private ZmBoolean applyCos = ZmBoolean.ONE /* true */;
 
     /**
+     * @zm-api-field-tag effectiveQuota
+     * @zm-api-field-description Flag whether or not to get effective value (minimum of zimbraMailQuota and zimbraMailDomainQuota)
+     * <table>
+     * <tr> <td> <b>1 (true)</b> </td> <td> zimbraMailQuota attribute will contain effective value
+     * <tr> <td> <b>0 (false)[default]</b> </td> <td> zimbraMailQuota attribute will contain actual ldap value set </td> </tr>
+     * </table>
+     */
+    @XmlAttribute(name=AdminConstants.A_EFFECTIVE_QUOTA, required=false)
+    private ZmBoolean effectiveQuota;
+    
+    /**
      * @zm-api-field-description Account
      */
     @XmlElement(name=AdminConstants.E_ACCOUNT)
@@ -72,9 +83,15 @@ public class GetAccountRequest extends AttributeSelectorImpl {
 
     public GetAccountRequest(AccountSelector account, Boolean applyCos,
             Iterable<String> attrs) {
+        this(account, applyCos, false, attrs);
+    }
+
+    public GetAccountRequest(AccountSelector account, Boolean applyCos,
+            Boolean effectiveQuota, Iterable<String> attrs) {
         super(attrs);
         this.account = account;
         this.applyCos = ZmBoolean.fromBool(applyCos);
+        this.effectiveQuota = ZmBoolean.fromBool(effectiveQuota);
     }
 
     public void setAccount(AccountSelector account) {
@@ -85,6 +102,11 @@ public class GetAccountRequest extends AttributeSelectorImpl {
         this.applyCos = ZmBoolean.fromBool(applyCos);
     }
 
+    public void setEffectiveQuota(Boolean effectiveQuota) {
+        this.effectiveQuota = ZmBoolean.fromBool(effectiveQuota);
+    }
+
     public AccountSelector getAccount() { return account; }
     public Boolean isApplyCos() { return ZmBoolean.toBool(applyCos); }
+    public Boolean isEffectiveQuota() { return ZmBoolean.toBool(effectiveQuota); }
 }
--- a/store/src/java/com/zimbra/cs/service/admin/ToXML.java
+++ b/store/src/java/com/zimbra/cs/service/admin/ToXML.java
@@ -54,6 +54,13 @@ public class ToXML {
     public static Element encodeAccount(Element parent, Account account,
             boolean applyCos, boolean needsExternalIndicator,
             Set<String> reqAttrs, AttrRightChecker attrRightChecker) {
+        return encodeAccount(parent, account,
+                applyCos, needsExternalIndicator, reqAttrs,  attrRightChecker, false);
+    }
+    
+    public static Element encodeAccount(Element parent, Account account,
+            boolean applyCos, boolean needsExternalIndicator,
+            Set<String> reqAttrs, AttrRightChecker attrRightChecker, boolean isEffectiveQuota) {
         Element acctElem = parent.addNonUniqueElement(AccountConstants.E_ACCOUNT);
         acctElem.addAttribute(AccountConstants.A_NAME, account.getUnicodeName());
         acctElem.addAttribute(AccountConstants.A_ID, account.getId());
@@ -67,11 +74,11 @@ public class ToXML {
             }
         }
         Map attrs = account.getUnicodeAttrs(applyCos);
-        if (null != reqAttrs && reqAttrs.contains(Provisioning.A_zimbraMailQuota)) {
+        if (isEffectiveQuota) {
             try {
                 // setting effective quota value refer ZBUG-1869
                 long effectiveQuota = AccountUtil.getEffectiveQuota(account);
-                attrs.put(Provisioning.A_zimbraMailQuota, Long.toString(effectiveQuota));
+                attrs.put(Provisioning.A_zimbraMailQuota, String.valueOf(effectiveQuota));
             } catch (ServiceException e) {
                 ZimbraLog.account.error("error in getting effective zimbra mail quota", e);
             }
--- a/store/src/java/com/zimbra/cs/service/admin/ToXML.java
+++ b/store/src/java/com/zimbra/cs/service/admin/ToXML.java
@@ -34,6 +34,7 @@ import com.zimbra.cs.account.AttributeManager.IDNType;
 import com.zimbra.cs.account.CalendarResource;
 import com.zimbra.cs.account.IDNUtil;
 import com.zimbra.cs.account.Provisioning;
+import com.zimbra.cs.util.AccountUtil;
 
 public class ToXML {
     public static Element encodeAccount(Element parent, Account account) {
@@ -66,6 +67,15 @@ public class ToXML {
             }
         }
         Map attrs = account.getUnicodeAttrs(applyCos);
+        if (null != reqAttrs && reqAttrs.contains(Provisioning.A_zimbraMailQuota)) {
+            try {
+                // setting effective quota value refer ZBUG-1869
+                long effectiveQuota = AccountUtil.getEffectiveQuota(account);
+                attrs.put(Provisioning.A_zimbraMailQuota, Long.toString(effectiveQuota));
+            } catch (ServiceException e) {
+                ZimbraLog.account.error("error in getting effective zimbra mail quota", e);
+            }
+        }
         encodeAttrs(acctElem, attrs, AdminConstants.A_N, reqAttrs, attrRightChecker);
         return acctElem;
     }
--- a/store/src/java/com/zimbra/cs/service/account/GetInfo.java
+++ b/store/src/java/com/zimbra/cs/service/account/GetInfo.java
@@ -337,6 +337,9 @@ public class GetInfo extends AccountDocumentHandler  {
             } else if (Provisioning.A_zimbraFeatureZulipChatEnabled.equals(key)) {
                 value = AccountUtil.isZulipChatEnabled(acct) ?
                         ProvisioningConstants.TRUE : ProvisioningConstants.FALSE;
+            } else if (Provisioning.A_zimbraMailQuota.equals(key)) {
+                // setting effective quota value refer ZBUG-1869 
+                value = String.valueOf(AccountUtil.getEffectiveQuota(acct));
             } else {
                 value = attrsMap.get(key);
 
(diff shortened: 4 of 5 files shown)


zimsuchitgupta:
We can also get the value of `zimbraMailQuota` from Admin `GetAccountRequest` so for consistency made the same changes here ie setting the effective quota value in `zimbraMailQuota` attribute.

zimsuchitgupta:
**Testing instructions**
**Case1:**
zimbraMailDomainQuota = 25MB
zimbraMailQuota = 30MB
Effective value  would be `25MB` and will be display in the Classic UI.
**Case2:**
zimbraMailDomainQuota = 30MB
zimbraMailQuota = 25MB
Effective value  would be `25MB` and will be display in the Classic UI.
**Case3:**
zimbraMailDomainQuota = 25MB
zimbraMailQuota = 25MB
Effective value  would be `25MB` and will be display in the Classic UI.
**Case4:**
zimbraMailDomainQuota = 30MB
zimbraMailQuota is not set
Effective value  would be `30MB` and will be display in the Classic UI.
**Case5:**
zimbraMailQuota = 30MB
zimbraMailDomainQuota is not set
Effective value  would be `30MB` and will be display in the Classic UI.

* Verified that there is no impact on Admin Level `GetAccountInfo` as well as User Level `GetAccountInfo` because we are returning limited attribute in response.

k-kato:
How does the behavior change on Admin Console?  When zimbraMailDomainQuota is 25MB and zimbraMailQuota is 30MB, which is shown in the following page/item?
* Manage -> Accounts -> (account) -> Advanced -> Account quota (MB)
* Configure -> Domains -> (domain) -> Mailbox Quota

zimsuchitgupta:
@k-kato When zimbraMailDomainQuota is 25MB and zimbraMailQuota is 30MB, Below is answer -

**Case-1**
Manage -> Accounts -> (account) -> Advanced -> Account quota (MB) -- `30MB `
**Case-2**
Configure -> Domains -> (domain) -> Mailbox Quota -- `25 MB`

For case 1, Admin `GetAccountRequest` gets called. Below is the format of request -
`<GetAccountRequest xmlns="urn:zimbraAdmin" applyCos="0">`
     `<account by="id">b07b5db1-9f64-4faa-82e4-fc8f14bd38e1</account>`
`</GetAccountRequest>`

As in first case admin can set any value to `zimbraMailQuota` so in this case we must display the actual value set on ldap.
So we can't touch the default behavior but in case admin wanna specifically ask for the `zimbraMailQuota` .Like below -
`<urn1:GetAccountRequest attrs="zimbraMailQuota">`
   `<urn1:account by="name">user1@test.com</urn1:account>`
`</urn1:GetAccountRequest>`

So in this case we will return effective value. Below is the implementation used -

`if (null != reqAttrs && reqAttrs.contains(Provisioning.A_zimbraMailQuota)) {`
    `try {`
    `// setting effective quota value refer ZBUG-1869`
    `long effectiveQuota = AccountUtil.getEffectiveQuota(account);`
    `attrs.put(Provisioning.A_zimbraMailQuota, Long.toString(effectiveQuota));`
    `} catch (ServiceException e) {`
         `ZimbraLog.account.error("error in getting effective zimbra mail quota", e);`
    `}`
 `}`

For Case 2, Admin `GetQuotaUsageRequest` get called which in turns already returning the effective value.

k-kato:
@zimsuchitgupta I understood. Thank you for the clear explanation.

k-kato:
zmprov command returns account's zimbraMailQuota. It is fine.
```
zimbra@zcs:~$ zmprov gd dev.zimbra.com zimbraMailDomainQuota
# name dev.zimbra.com
zimbraMailDomainQuota: 25000000

zimbra@zcs:~$ zmprov ga test1@dev.zimbra.com zimbraMailQuota
# name test1@dev.zimbra.com
zimbraMailQuota: 31457280
```

GetAccountResponse with `attrs="zimbraMailQuota"` returns the effective value of `zimbraMailQuota` as written in the comments above.
```
<GetAccountRequest applyCos="1" xmlns="urn:zimbraAdmin">
  <account by="name">test1</account>
</GetAccountRequest>

<GetAccountResponse xmlns="urn:zimbraAdmin">
  <account name="test1@dev.zimbra.com" id="464d6fba-6d78-4889-8365-29100ecf9eb8">
    ...
    <a n="zimbraMailQuota">31457280</a>
    ...
  </account>
</GetAccountResponse>


<GetAccountRequest applyCos="1" attrs="zimbraMailQuota" xmlns="urn:zimbraAdmin">
  <account by="name">test1</account>
</GetAccountRequest>

<GetAccountResponse xmlns="urn:zimbraAdmin">
  <account name="test1@dev.zimbra.com" id="464d6fba-6d78-4889-8365-29100ecf9eb8">
    <a n="zimbraMailQuota">25000000</a>
  </account>
</GetAccountResponse>
```
Can you modify API document so that the admin can understand the difference of the behavior?
https://files.zimbra.com/docs/soap_api/9.0.0/api-reference/index.html  => GetAccount

soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java
https://github.com/Zimbra/zm-mailbox/blob/develop/soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java

around here
```
Get attributes related to an account
{request-attrs} - comma-seperated list of attrs to return

Note: this request is by default proxied to the account's home server 
```
I think the change affects GetAccountRequest only, but please update documents of all affected APIs.

k-kato:
I rethought about the design of GetAccountRequest. It's confusing that the result is not the same with/without `attrs="zimbraMailQuota"`. In addition, if a customer has created a script which gets actual zimbraMailQuota using GetAccountRequest, the change will break the script as described in Cons.
Another idea is, adding a flag like `<GetAccountRequest applyCos="1" getEffectiveQuota="[0|1]" xmlns="urn:zimbraAdmin">`. If `getEffectiveQuota` is true, return effective quota.
What do you think?

zimsuchitgupta:
> I rethought about the design of GetAccountRequest. It's confusing that the result is not the same with/without `attrs="zimbraMailQuota"`. In addition, if a customer has created a script which gets account's zimbraMailQuota using GetAccountRequest, the change will break the script as described in Cons. Another idea is, adding a flag like `<GetAccountRequest applyCos="1" getEffectiveQuota="[0|1]" xmlns="urn:zimbraAdmin">`. If `getEffectiveQuota` is true, return effective quota. What do you think?

It also seems good to me, default behavior will be untouched and to get effective value admins only need to send a true flag.
Thanks @k-kato , Let wait for others comments too.

zimsuchitgupta:
@k-kato added a flag `effectiveQuota` if its value set to 1 then `zimbraMailQuota` will be having an effective value else it will have actual value of `zimbraMailQuota` set on ldap.

By default this flag would be false. Below is example-

`       <urn1:GetAccountRequest attrs="zimbraMailQuota" effectiveQuota="1">
         <urn1:account by="name">user1@test.com</urn1:account>
      </urn1:GetAccountRequest>
`

soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java
```diff
@@ -53,6 +53,17 @@
     @XmlAttribute(name=AdminConstants.A_APPLY_COS, required=false)
     private ZmBoolean applyCos = ZmBoolean.ONE /* true */;
 
+    /**
+     * @zm-api-field-tag effectiveQuota
+     * @zm-api-field-description Flag whether want to return effective value (minimum of zimbraMailQuota & zimbraMailDomainQuota)
```
k-kato, line 58: Idea, referring the description of applyCos
```suggestion
     * @zm-api-field-description Flag whether or not to get effective value (minimum of zimbraMailQuota and zimbraMailDomainQuota)
```

soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java
```diff
@@ -53,6 +53,17 @@
...
+     * <table>
+     * <tr> <td> <b>1 (true)</b> </td> <td> zimbraMailQuota attribute will contain effective value.
```
k-kato, line 60: remove the period (dot).

soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java
```diff
@@ -77,6 +88,14 @@ public GetAccountRequest(AccountSelector account, Boolean applyCos,
         this.applyCos = ZmBoolean.fromBool(applyCos);
```
k-kato, line 88: change to call new `GetAccountRequest` defined at line 91

soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java
```diff
@@ -53,6 +53,17 @@
...
+    /**
+     * @zm-api-field-tag effectiveQuota
+     * @zm-api-field-description Flag whether want to return effective value (minimum of zimbraMailQuota & zimbraMailDomainQuota)
+     * <table>
+     * <tr> <td> <b>1 (true)</b> </td> <td> zimbraMailQuota attribute will contain effective value.
+     * <tr> <td> <b>0 (false)[default]</b> </td> <td> zimbraMailQuota attribute will contain actual ldap value set </td> </tr>
+     * </table>
+     */
+    @XmlAttribute(name=AdminConstants.A_EFFECTIVE_QUOTA, required=false)
+    private ZmBoolean effectiveQuota = ZmBoolean.ZERO /* false */;
```
k-kato, line 65: Does `ZmBoolean.ZERO` need to be set?  I couldn't find any statement which sets `ZmBoolean.ZERO` in other files.

store/src/java/com/zimbra/cs/service/admin/ToXML.java
```diff
@@ -66,6 +73,15 @@ public static Element encodeAccount(Element parent, Account account,
             }
         }
         Map attrs = account.getUnicodeAttrs(applyCos);
+        if (isEffectiveQuota) {
+            try {
+                // setting effective quota value refer ZBUG-1869
+                long effectiveQuota = AccountUtil.getEffectiveQuota(account);
+                attrs.put(Provisioning.A_zimbraMailQuota, Long.toString(effectiveQuota));
```
k-kato, line 80: In GetInfo.java, `String.valueOf` is used. Is there any reason why `Long.toString` is used here?
I think `String.valueOf` is better to avoid NullPointerException.

soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java
```diff
@@ -53,6 +53,17 @@
     @XmlAttribute(name=AdminConstants.A_APPLY_COS, required=false)
     private ZmBoolean applyCos = ZmBoolean.ONE /* true */;
 
+    /**
```
silentsakky, line 56: do we need extra changes to update soap api documentation with this flag? let's create a ticket to release updated soap documentation with this ticket

common/src/java/com/zimbra/common/soap/AdminConstants.java
```diff
@@ -1342,7 +1342,7 @@
 
     public static final String A_QUOTA_USED = "used";
     public static final String A_QUOTA_LIMIT = "limit";
-
+    public static final String A_EFFECTIVE_QUOTA = "effectiveQuota";
```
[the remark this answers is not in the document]
zimsuchitgupta: Done

soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java
zimsuchitgupta: Done

soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java
zimsuchitgupta: Removed

soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java
zimsuchitgupta: Done

store/src/java/com/zimbra/cs/service/admin/ToXML.java
zimsuchitgupta: Purpose of both is same, yes agreed, String.valueOf avoid NullPointer . Now using String.valueOf()

soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java
zimsuchitgupta: Done

k-kato: Lines 84-96:
move statements from GetAccountRequest defined at line 84 to another at line 91, and call the new GetAccountRequest with `effectiveQuota false`.
```
    public GetAccountRequest(AccountSelector account, Boolean applyCos,
            Iterable<String> attrs) {
        this(account, applyCos, false, (Iterable<String>) null);
    }

    public GetAccountRequest(AccountSelector account, Boolean applyCos,
            Boolean effectiveQuota, Iterable<String> attrs) {
        super(attrs);
        this.account = account;
        this.applyCos = ZmBoolean.fromBool(applyCos);
        this.effectiveQuota = ZmBoolean.fromBool(effectiveQuota);
    }
```

soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java
k-kato: I mean, just defining a variable might work.
```
private ZmBoolean effectiveQuota;
```

soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java
zimsuchitgupta: Done

soap/src/java/com/zimbra/soap/admin/message/GetAccountRequest.java
zimsuchitgupta: Yes.. we can... but in this case effectiveQuota will gave null as default. Need to taken care of this.
Declared variable as you said

k-kato (approved):
LGTM

store/src/java/com/zimbra/cs/service/admin/ToXML.java
```diff
@@ -50,9 +51,22 @@ public static Element encodeAccount(Element parent, Account account, boolean app
                 applyCos, false, reqAttrs,  attrRightChecker);
     }
 
+    public static Element encodeAccount(Element parent, Account account, boolean applyCos,
```
dasiyogesh, line 54: Do we need this overloaded method?

dasiyogesh (approved):

log2akshat (approved):

ronstra-synacor (approved):

PR #1235, opened 2022-03-01
ZBUG-2165: fixes mail filtering not working with forwarding.

wilfredo-alfaro:
### Solution

Implement additional rules within the `SMFIC_Rcpt()` method of `MilterHandler` class to reject incoming messages where one or more recipients have set their `zimbraPrefMailForwardingAddress` property value to a distribution list address that they are not permitted to send messages according to the distribution list ACL.

### Notes

- Messages will be rejected even if there are other compliant recipients on the message, which matches the existing behavior for mail filtering. So, it takes just one recipient that attempts to forward to a non-permitted distribution list for the whole message to be rejected.
- This rules will apply to both:
   - Messages that come from the front-end through the back-end and then reach the MTA, and
   - Messages that come directly to the MTA from the outside world.

### Testing Done

**Setting up server, domain and accounts.**

1. enable server-wide mail filter
2. Create domain
3. Create Account Random to which messages will be forwarded under some scenarios
4. Create Distribution List 1 and add two member Accounts to it
5. Create Distribution List 2 and add two member Accounts to it
6. Create Account Alpha and grant sendToDistList permission on Distribution List 1 only
7. Create Account Beta and grant sendToDistList permission on Distribution List 2 only
8. Create Account Delta from which messages will be originated in all scenarios

**Test internal and external messaging for every scenario.**

1. Set the forwarding address of Alpha to a given combination between regular accounts, permitted and non-permitted distribution lists, including:
- Forwarding to regular account.
- Forwarding to permitted distribution list.
- Forwarding to non-permitted distribution list.
- Forwarding to regular account with permitted distribution list.
- Forwarding to regular account with non-permitted distribution list.
- Forwarding to regular account, with permitted and non-permitted distribution list.

2. For each of the above scenarios:
- Send a message from Delta to Alpha account through Zimbra Web Client (internal messaging), and,
- Inject a message directly to the MTA with SMTP commands (external messaging).
```
telnet localhost 25
ehlo your.zimbra.fqdn.tld
mail from: random@external.tld
rcpt to: alpha@zbug2165.tld
data
Subject: TEST
 
This is a test.
.
quit
```

For any scenarios were Alpha has included a non-permitted distribution list as a forwarding address the message should be _rejected_, otherwise the message should be _accepted_.

Finally, you can test all these scenarios by sending messages to account Beta.

### DIFF
--- a/store/src/java/com/zimbra/cs/milter/MilterHandler.java
+++ b/store/src/java/com/zimbra/cs/milter/MilterHandler.java
@@ -38,6 +38,7 @@ import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.util.StringUtil;
 import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.cs.account.AccessManager;
+import com.zimbra.cs.account.Account;
 import com.zimbra.cs.account.Group;
 import com.zimbra.cs.account.Provisioning;
 import com.zimbra.cs.account.accesscontrol.Rights.User;
@@ -353,6 +354,23 @@ public final class MilterHandler implements NioHandler {
         if (sender == null || rcpt == null) {
             connection.send(new MilterPacket(SMFIR_TEMPFAIL));
             return;
+        }
+        // ZBUG-2165: prevent forwarding to non-permitted distribution list
+        final Account account = prov.getAccountByName(rcpt);
+        if (account != null && account.getPrefMailForwardingAddress() != null) {
+            ZimbraLog.milter.trace("account: %s", account.getName());
+            for (final String a : account.getPrefMailForwardingAddress().split(",")) {
+                final String address = a.trim();
+                ZimbraLog.milter.trace("address: %s", address);
+                if (!address.isEmpty() && prov.isDistributionList(address)) {
+                    final Group group = prov.getGroupBasic(Key.DistributionListBy.name, address);
+                    if (group != null && !accessMgr.canDo(account, group, User.R_sendToDistList, false)) {
+                        ZimbraLog.milter.debug("Recipient %s is not allowed to forward to this distribution list: %s", account.getName(), group.getName());
+                        SMFIR_ReplyCode("571", "571 Recipient " + account.getName() + " is not allowed to forward to this distribution list: " + group.getName());
+                        return;
+                    }
+                }
+            }
         }
         if (prov.isDistributionList(rcpt)) {
             Group group = prov.getGroupBasic(Key.DistributionListBy.name, rcpt);


store/src/java/com/zimbra/cs/milter/MilterHandler.java
```diff
@@ -354,6 +355,30 @@ private void SMFIC_Rcpt() throws ServiceException {
             connection.send(new MilterPacket(SMFIR_TEMPFAIL));
             return;
         }
+        try { // ZBUG-2165: prevent forwarding to non-permitted distribution list
+            ZimbraLog.milter.trace("evaluating forwarding to non-permitted distribution list");
+            final Account account = prov.getAccountByName(rcpt);
+            ZimbraLog.milter.trace("account: %s", account.getName());
```
shrutig0510, line 361: On line 360 if account is null then at line 361 getName() method will give exception.
Can you move this line inside if block

store/src/java/com/zimbra/cs/milter/MilterHandler.java
```diff
@@ -354,6 +355,30 @@ private void SMFIC_Rcpt() throws ServiceException {
...
+            if (account != null) {
+                final String prefMailForwardingAddress = account.getPrefMailForwardingAddress();
+                ZimbraLog.milter.trace("prefMailForwardingAddress: %s", prefMailForwardingAddress);
+                if (prefMailForwardingAddress != null) {
+                    if (prov.isDistributionList(prefMailForwardingAddress)) {
```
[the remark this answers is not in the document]
wilfredo-alfaro: Done

store/src/java/com/zimbra/cs/milter/MilterHandler.java
```diff
@@ -354,6 +355,30 @@ private void SMFIC_Rcpt() throws ServiceException {
...
+                        final Group group = prov.getGroupBasic(Key.DistributionListBy.name, prefMailForwardingAddress);
+                        ZimbraLog.milter.trace("group: %s", group);
+                        if (group != null) {
+                            if (!accessMgr.canDo(account, group, User.R_sendToDistList, false)) {
```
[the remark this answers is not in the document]
wilfredo-alfaro: Done

store/src/java/com/zimbra/cs/milter/MilterHandler.java
wilfredo-alfaro: Done

shrutig0510 (approved):

dasiyogesh (approved):

rcyarrapothu (approved):

zimsuchitgupta (approved):
LGTM

store/src/java/com/zimbra/cs/milter/MilterHandler.java
```diff
@@ -354,6 +355,26 @@ private void SMFIC_Rcpt() throws ServiceException {
...
             return;
         }
+        try { // ZBUG-2165: prevent forwarding to non-permitted distribution list
+            ZimbraLog.milter.trace("evaluating forwarding to non-permitted distribution list");
+            final Account account = prov.getAccountByName(rcpt);
+            if (account != null) {
+                ZimbraLog.milter.trace("account: %s", account.getName());
+                final String prefMailForwardingAddress = account.getPrefMailForwardingAddress();
+                ZimbraLog.milter.trace("prefMailForwardingAddress: %s", prefMailForwardingAddress);
+                if (prefMailForwardingAddress != null && prov.isDistributionList(prefMailForwardingAddress)) {
```
desouzas, line 365: The `zimbraPrefMailForwardingAddress` is a comma-separated list of email addresses. Let's make sure to test this with multiple email addresses, and DLs set in that property and address handling it.

https://github.com/Zimbra/zm-mailbox/blob/develop/store/conf/attrs/zimbra-attrs.xml#L1868

desouzas (changes requested):
* Support multiple forwarding addresses.

store/src/java/com/zimbra/cs/milter/MilterHandler.java
wilfredo-alfaro: Done

store/src/java/com/zimbra/cs/milter/MilterHandler.java
```diff
@@ -354,6 +355,33 @@ private void SMFIC_Rcpt() throws ServiceException {
...
+                            ZimbraLog.milter.trace("group: %s", group);
+                            if (group != null && !accessMgr.canDo(account, group, User.R_sendToDistList, false)) {
+                                ZimbraLog.milter.debug("Recipient %s is not allowed to forward to this distribution list: %s", account.getName(), group.getName());
+                                SMFIR_ReplyCode("571", "571 Recipient " + account.getName() + " is not allowed to forward to this distribution list: " + group.getName());
+                                return;
+                            }
+                        }
+                    }
+                }
+            }
```
log2akshat, line 381: There are lots of nested if please refactor this code and try to make it simpler and more readable.
Also, make sure you put only risky code inside a `try` `catch` block.

store/src/java/com/zimbra/cs/milter/MilterHandler.java
```diff
@@ -354,6 +355,33 @@ private void SMFIC_Rcpt() throws ServiceException {
...
+        } catch (Exception e) {
+            ZimbraLog.milter.warn("failed to evaluate forwarding to non-permitted distribution list");
```
log2akshat, line 383: Please `catch` specific exception not the `Exception` class.

log2akshat (changes requested):

store/src/java/com/zimbra/cs/milter/MilterHandler.java
wilfredo-alfaro: Done.

Refactored code for simplicity and readability.

store/src/java/com/zimbra/cs/milter/MilterHandler.java
wilfredo-alfaro: Done.

I was going to catch the `ServiceException` class, which is the specific exception to be expected here, but then I realized that such class is already part of the `throw` clause on the parent `SMFIC_Rcpt()` method, so no point in catching it anywhere in this code. I completely removed the `try{}catch(){}` block instead.

desouzas: Note: this changes the behavior of this addition to fail, instead of continuing, if an exception is thrown while processing forwarding addresses.

log2akshat: @wilfredo-alfaro You have get rid of `try` `catch` blocks but can you please address @desouzas concern.

store/src/java/com/zimbra/cs/milter/MilterHandler.java
```diff
@@ -354,6 +355,23 @@ private void SMFIC_Rcpt() throws ServiceException {
...
+                ZimbraLog.milter.trace("address: %s", address);
+                if (!address.isEmpty() && prov.isDistributionList(address)) {
+                    final Group group = prov.getGroupBasic(Key.DistributionListBy.name, address);
+                    if (group != null && !accessMgr.canDo(account, group, User.R_sendToDistList, false)) {
+                        ZimbraLog.milter.debug("Recipient %s is not allowed to forward to this distribution list: %s", account.getName(), group.getName());
+                        SMFIR_ReplyCode("571", "571 Recipient " + account.getName() + " is not allowed to forward to this distribution list: " + group.getName());
+                        return;
+                    }
+                }
+            }
```
log2akshat, line 373: @wilfredo-alfaro You have get rid of `try` `catch` blocks but can you please address @desouzas concern.

store/src/java/com/zimbra/cs/milter/MilterHandler.java
desouzas: This is fine. It behaves the same as any ServiceException thrown in the section below.

desouzas (approved):

log2akshat (approved):

PR #1237, opened 2022-03-08
ZBUG-2608: Contacts not importing references from ContactGroups

ronstra-synacor:
Problem:
When importing an exported CSV or TGZ previously exported by Zimbra that has a ContactGroup, `CONTACT_REF` references are being turned into `INLINE`, losing any attributes defined for that contact in the imported file.

Fix:
For CSV, inside the `ContactGroup` class, for each member of the "dlist" field, convert all bare inline addresses that have also been imported in the same file with a reference to that imported contact.
For TGZ, the approach is similar, but it was very difficult to find just where to implement it. Ended up in `ArchiveFormatter::postProcessContacts`.

Progress:
CSV fix appears to be good. For TGZ, while the contact does go through `postProcessContacts`, I can't figure out how to save it back to the Zimbra store after converting inline addresses to contact references.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1,5 +1,4 @@
 <project name="zm-mailbox" default="all">
-
     <target name="all" depends="publish-local-all">
       <ant dir="./native" target="generate-native-headers" inheritAll="true"/>
       <exec dir="./native" executable="make"          failonerror="true"/>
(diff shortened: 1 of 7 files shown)


shrutig0510:
Formatting issues... Please fix it to make it more readable

PR #1240, opened 2022-03-14
ZBUG-2361: Modified Draft not synced to external imap account

gabhinavTI:
Issue:
Modified Drafts are not synced to an external IMAP account

Fix:
- The issue was basically the logic was not in place to identify the draft content change when changed locally
- In order to identify the draft change content we have added a logic where we figure out the draft content has been updated
- Next step to query and find the modified items for Draft since previous sync state. This returns the required and expected modified items based on lastChangeID
- lastChangeID remains 0 when the server is restarted and we modify and sync the messages. In this case syncState remains null hence we have to check for all drafts messages and sync them up
- once we have Message changes identified we delete the corresponding message and push the updated new message

Repro:
There are two Accounts:

a Zimbra account
any IMAP account
Steps to reproduce:
1. Create a draft in the imap account

Open the IMAP account with any Client
create a draft and save it so that a mail is saved in the draft folder
2. Add the IMAP account as external account to zimbra

Go to "Preferences > Accounts"
"Add External Account"
Enter the data of the imap account
3. Try to edit the Draft

Go to "Mail" and open the folder structure of the added IMAP account
Open the "Drafts" folder. There should be the saved mail.
Click in "Edit"
Edit the draft
Click on "Save Draft

Testing:

Thorough Manual testing through different scenarios  in Draft message and also paralley modifying and syncing the Inbox , Sent, Drafts etc folder message all at a time

### DIFF
--- a/store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
+++ b/store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
@@ -100,6 +100,8 @@ class ImapFolderSync {
     // Max number of total per-item failures before we abort sync
     private static final int MAX_TOTAL_ERRORS = 10;
 
+    private static final String DRAFTS = "Drafts"; 
+
     private static class Statistics {
         int flagsUpdatedLocally;
         int flagsUpdatedRemotely;
@@ -277,6 +279,10 @@ class ImapFolderSync {
 
         // Restore previously cached sync state or create new state
         syncState = imapSync.removeSyncState(localFolder.getId());
+        int lastChangeId = 0;
+        if (syncState != null && DRAFTS.equals(remoteFolder.getPath())) {
+            lastChangeId = syncState.getLastChangeId();
+        }
         if (syncState == null || fullSync) {
             syncState = newSyncState();
             fullSync = true;
@@ -348,6 +354,14 @@ class ImapFolderSync {
                 pushChanges(changes);
                 syncState.setLastChangeId(changes.getLastChangeId());
             }
+        }
+
+        // Push local draft changes into remote
+        if (DRAFTS.equals(remoteFolder.getPath())) {
+            changes = MessageChanges.getChangesForDrafts(ds, localFolder.getFolder(), lastChangeId,
+                    localFolder.getId());
+            pushChangesForDraft(changes, remoteFolder.getPath());
+            syncState.setLastChangeId(changes.getLastChangeId());
         }
 
         // Fetch new messages
@@ -431,10 +445,34 @@ class ImapFolderSync {
      * let it be handled when the originating folder is processed.
      */
     private void pushChanges(MessageChanges changes) throws ServiceException, IOException {
+        pushChangesForDraft(changes, null);
+    }
+
+    /**
+     * Message has been modified in either of below ways
+     * 
+     * 1. Message flags changed (UPDATED)
+     * 2. New message added to this folder (ADDED)
+     * 3. Message moved from this folder to another (MOVED)
+     * 4. Message moved to this folder from another
+     * 5. Local Draft message modified or it's Attachment changed
+     * 
+     * @param changes
+     * @param folderName
+     * @throws ServiceException
+     * @throws IOException
+     */
+    private void pushChangesForDraft(MessageChanges changes, String folderName) throws ServiceException, IOException {
         localFolder.debug("Pushing changes: %s", changes);
         for (MessageChange change : changes.getChanges()) {
             clearError(change.getItemId());
-            if (change.isAdded()) {
+            if ((change.isModified() || change.isUpdated()) && DRAFTS.equalsIgnoreCase(folderName)
+                    && deleteMessage(change.getTracker().getUid())) {
+                List<Integer> pushId = new ArrayList<>();
+                pushId.add(change.getItemId());
+                appendMsgs(pushId);
+            }
+            else if (change.isAdded()) {
                 newMsgIds.add(change.getItemId());
             } else if (change.isDeleted()) {
                 deleteMessage(change.getTracker().getUid());
--- a/store/src/java/com/zimbra/cs/datasource/imap/MessageChanges.java
+++ b/store/src/java/com/zimbra/cs/datasource/imap/MessageChanges.java
@@ -20,9 +20,11 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.cs.account.DataSource;
+import com.zimbra.cs.mailbox.Flag;
 import com.zimbra.cs.mailbox.Folder;
 import com.zimbra.cs.mailbox.MailItem;
 import com.zimbra.cs.mailbox.MailServiceException;
@@ -39,6 +41,11 @@ class MessageChanges {
     public static MessageChanges getChanges(DataSource ds, Folder folder, int changeId)
         throws ServiceException {
         return new MessageChanges(ds, folder).findChanges(changeId);
+    }
+
+    public static MessageChanges getChangesForDrafts(DataSource ds, Folder folder, int changeId, int folderId)
+            throws ServiceException {
+        return new MessageChanges(ds, folder).findChangesForDraft(changeId, folderId);
     }
 
     private MessageChanges(DataSource ds, Folder folder) {
@@ -90,6 +97,28 @@ class MessageChanges {
         return this;
     }
 
+    private MessageChanges findChangesForDraft(int changeId, int draftFolderId) throws ServiceException {
+        // find modified items since specified change id for draft folder
+        List<Integer> modifiedItems;
+        mbox.lock.lock(false);
+        try {
+            Set<Integer> folderIds = new HashSet<>();
+            folderIds.add(draftFolderId);
+            modifiedItems = mbox.getModifiedItems(null, changeId, MailItem.Type.MESSAGE, folderIds).getFirst();
+        } finally {
+            mbox.lock.release();
+        }
+        changes = new ArrayList<MessageChange>();
+        // Find modified messages for this folder
+        for (int id : modifiedItems) {
+            MessageChange change = getChange(id);
+            if (change != null) {
+                changes.add(change);
+            }
+        }
+        return this;
+    }
+
     private MessageChange getChange(int msgId) throws ServiceException {
         Message msg = getMessage(msgId);
         if (msg != null) {
@@ -103,6 +132,12 @@ class MessageChanges {
                     if (tracker.getFlags() != msg.getFlagBitmask()) {
                         // Message flags updated
                         return MessageChange.updated(msg, tracker);
+                    }
+                    final int flagMessageBitMask = msg.getFlagBitmask();
+                    if ((flagMessageBitMask & Flag.BITMASK_DRAFT) == flagMessageBitMask) {
+                        // Message is a draft and none of the above changes are relevant, assume it has
+                        // been modified
+                        return MessageChange.modifiedDraft(msg, tracker);
                     }
                 } else if (tracker.getFolderId() == folder.getId()) {
                     // Message moved from this folder to another
@@ -162,18 +197,18 @@ class MessageChanges {
 
     @Override
     public String toString() {
-        int added = 0, updated = 0, moved = 0, deleted = 0;
+        int added = 0, updated = 0, moved = 0, deleted = 0, modified=0;
         for (MessageChange change : getChanges()) {
             switch (change.getType()) {
             case ADDED:   added++; break;
             case UPDATED: updated++; break;
             case MOVED:   moved++; break;
             case DELETED: deleted++; break;
+            case MODIFIED: modified++; break;
             }
         }
         return String.format(
-            "{changeId=%d,added=%d,updated=%d,moved=%d,deleted=%d}",
-            lastChangeId, added, updated, moved, deleted);
+            "{changeId=%d,added=%d,updated=%d,moved=%d,deleted=%d,modified=%d}",
+            lastChangeId, added, updated, moved, deleted, modified);
     }
-
 }
--- a/store/src/java/com/zimbra/cs/datasource/imap/MessageChange.java
+++ b/store/src/java/com/zimbra/cs/datasource/imap/MessageChange.java
@@ -20,7 +20,7 @@ import com.zimbra.cs.mailbox.Message;
 
 final class MessageChange {
     public enum Type {
-        ADDED, UPDATED, MOVED, DELETED
+        ADDED, UPDATED, MOVED, DELETED, MODIFIED
     }
 
     private final Type type;
@@ -43,7 +43,11 @@ final class MessageChange {
     public static MessageChange deleted(int itemId, ImapMessage tracker) {
         return new MessageChange(Type.DELETED, itemId, null, tracker);
     }
-    
+
+    public static MessageChange modifiedDraft(Message msg, ImapMessage tracker) {
+        return new MessageChange(Type.MODIFIED, msg.getId(), msg, tracker);
+    }
+
     MessageChange(Type type, int itemId, Message msg, ImapMessage tracker) {
         this.type = type;
         this.itemId = itemId;
@@ -82,4 +86,9 @@ final class MessageChange {
     public boolean isDeleted() {
         return type == Type.DELETED;
     }
+
+    public boolean isModified() {
+        return type == Type.MODIFIED;
+    }
+
 }


store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
```diff
@@ -277,6 +279,10 @@ public void syncMessages() throws ServiceException, IOException {
 
         // Restore previously cached sync state or create new state
         syncState = imapSync.removeSyncState(localFolder.getId());
+        int lastChangeId = 0;
+        if (remoteFolder.getPath().equals(DRAFTS) && syncState != null) {
+            lastChangeId = syncState.getLastChangeId();
+        }
```
k-kato, line 285: Can we move the 4 lines to around line 359? `lastChangeId` is used there. It is more readable.

gabhinavTI: We require older lastChangeId from older syncstate, and the value of syncState changes at line 287 
hence cannot move this code around 359 as we won't get required value then

k-kato: I got it.

store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
```diff
@@ -431,10 +444,37 @@ private void syncFlags(long lastUid) throws ServiceException, IOException {
      * let it be handled when the originating folder is processed.
      */
     private void pushChanges(MessageChanges changes) throws ServiceException, IOException {
+        localFolder.debug("Pushing changes: %s", changes);
```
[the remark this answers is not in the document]
gabhinavTI: Done

store/src/java/com/zimbra/cs/datasource/imap/MessageChanges.java
```diff
@@ -41,6 +43,11 @@ public static MessageChanges getChanges(DataSource ds, Folder folder, int change
         return new MessageChanges(ds, folder).findChanges(changeId);
     }
 
+    public static MessageChanges getChangesForDrafts(DataSource ds, Folder folder, int changeId, int folderId)
+            throws ServiceException {
+            return new MessageChanges(ds, folder).findChangesForDraft(changeId, folderId);
+        }
```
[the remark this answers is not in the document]
gabhinavTI: done

store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
```diff
@@ -431,10 +444,37 @@ private void syncFlags(long lastUid) throws ServiceException, IOException {
...
+     * @throws ServiceException
+     * @throws IOException
+     */
+    private void pushChangesForDraft(MessageChanges changes, String folderName) throws ServiceException, IOException {
         localFolder.debug("Pushing changes: %s", changes);
         for (MessageChange change : changes.getChanges()) {
             clearError(change.getItemId());
-            if (change.isAdded()) {
+            if ((change.isModified() || change.isUpdated()) && DRAFTS.equalsIgnoreCase(folderName)) {
+                if (deleteMessage(change.getTracker().getUid())) {
```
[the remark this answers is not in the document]
gabhinavTI: combined the if condition, thanks

k-kato (approved):
LGTM

store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
```diff
@@ -277,6 +279,10 @@ public void syncMessages() throws ServiceException, IOException {
 
         // Restore previously cached sync state or create new state
         syncState = imapSync.removeSyncState(localFolder.getId());
+        int lastChangeId = 0;
+        if (remoteFolder.getPath().equals(DRAFTS) && syncState != null) {
```
shrutig0510, line 283: Can we keep it like this if (syncState != null && remoteFolder.getPath().equals(DRAFTS)),
so that null check fails, there will no need to call getPath() method

shrutig0510, line 283: Can we keep it like this if (syncState != null && remoteFolder.getPath().equals(DRAFTS)),
so that null check fails, there will no need to call getPath() method

store/src/java/com/zimbra/cs/datasource/imap/MessageChanges.java
```diff
@@ -90,6 +97,28 @@ private MessageChanges findChanges(int changeId) throws ServiceException {
         return this;
     }
 
+    private MessageChanges findChangesForDraft(int changeId, int draftFolderId) throws ServiceException {
+        // Find modified items since specified change id for draft folder
```
dasiyogesh, line 101: use small case for comments to start with here and other places

store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
gabhinavTI: Done , thanks

store/src/java/com/zimbra/cs/datasource/imap/MessageChanges.java
gabhinavTI: made all single-line comments as small case in all my changes

store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
```diff
@@ -431,10 +444,35 @@ private void syncFlags(long lastUid) throws ServiceException, IOException {
...
+     * @throws ServiceException
+     * @throws IOException
+     */
+    private void pushChangesForDraft(MessageChanges changes, String folderName) throws ServiceException, IOException {
         localFolder.debug("Pushing changes: %s", changes);
         for (MessageChange change : changes.getChanges()) {
             clearError(change.getItemId());
-            if (change.isAdded()) {
+            if ((change.isModified() || change.isUpdated()) && DRAFTS.equalsIgnoreCase(folderName)
+                    && deleteMessage(change.getTracker().getUid())) {
```
dasiyogesh, line 470: whats the purpose of `deleteMessage` in the if check, is it actually deleting message?

gabhinavTI: yes this actually deletes the message from Remote IMAP and then pushes the new modified draft to remote in next step

store/src/java/com/zimbra/cs/datasource/imap/MessageChanges.java
```diff
@@ -90,6 +97,28 @@ private MessageChanges findChanges(int changeId) throws ServiceException {
...
     }
 
+    private MessageChanges findChangesForDraft(int changeId, int draftFolderId) throws ServiceException {
+        // find modified items since specified change id for draft folder
+        List<Integer> modifiedItems;
+        mbox.lock.lock(false);
+        try {
+            Set<Integer> folderIds = new HashSet<>();
+            folderIds.add(draftFolderId);
+            modifiedItems= mbox.getModifiedItems(null, changeId, MailItem.Type.MESSAGE, folderIds).getFirst();
```
dasiyogesh, line 107: add space before =

gabhinavTI: Thanks , done

store/src/java/com/zimbra/cs/datasource/imap/MessageChanges.java
```diff
@@ -90,6 +97,28 @@ private MessageChanges findChanges(int changeId) throws ServiceException {
         return this;
     }
 
+    private MessageChanges findChangesForDraft(int changeId, int draftFolderId) throws ServiceException {
```
desouzas, line 100: Did we separate this out from findChanges because of the fullSync thing?
* It looks like this method additionally runs after findChanges and the conditional push of partial changes? If not then we need to capture tombstones and add a delete change since Drafts can be deleted.
* It looks like we don't care about comparing changeId in here now, so this always runs?

store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
```diff
@@ -350,6 +356,13 @@ public void syncMessages() throws ServiceException, IOException {
             }
         }
 
+        // ZBUG-2361: this is to push local draft changes into remote
```
desouzas, line 359: You don't need to list the ticket here.
```suggestion
        // Push local draft changes into remote 
```

store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
```diff
@@ -431,10 +444,35 @@ private void syncFlags(long lastUid) throws ServiceException, IOException {
      * let it be handled when the originating folder is processed.
      */
     private void pushChanges(MessageChanges changes) throws ServiceException, IOException {
+        pushChangesForDraft(changes, null);
+    }
+
+    /**
+     * Message has been modified in either of below ways
+     * ZBUG-2361: Add condition to identify if draft content has been modified
```
desouzas, line 452:
```suggestion
```

desouzas (commented):
* Try to use the modern client to modify a draft that was moved to the trash - curious what happens.
* Make sure deleting drafts is still captured.
* Sign the contributor agreement: https://github.com/Zimbra/zm-mailbox/pull/1240#issuecomment-1066479836

store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
```diff
@@ -350,6 +356,13 @@ public void syncMessages() throws ServiceException, IOException {
             }
         }
 
+        // ZBUG-2361: this is to push local draft changes into remote 
+        if (DRAFTS.equals(remoteFolder.getPath())) {
+            changes = MessageChanges.getChangesForDrafts(ds, localFolder.getFolder(), lastChangeId,
+                    localFolder.getId());
+            pushChangesForDraft(changes, remoteFolder.getPath());
```
desouzas, line 363: This will probably need to [handle updating the changeId](https://github.com/Zimbra/zm-mailbox/blob/9ece3daf78fd27bed7db95d27e12b9896855109a/store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java#L310) again if you use two separate calls to find changes on the draft folder since we could capture a later change from the second call, findChangesForDrafts.

store/src/java/com/zimbra/cs/datasource/imap/MessageChanges.java
desouzas: You don't need to change anything, but I usually aim for consistency with the rest of the file.

store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
```diff
@@ -100,6 +100,8 @@
     // Max number of total per-item failures before we abort sync
     private static final int MAX_TOTAL_ERRORS = 10;
 
+    private static final String DRAFTS = "Drafts";
```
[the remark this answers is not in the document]
gabhinavTI: I don't see the draft folder loaded up when configured Yahoo as an External Imap account
The same can be seen in attached screenshot
When debugging code this is the list of folders passed as an input to syncFolders/syncMessages

```
inbox=Folder{n=Inbox, id=941, type=1, name=Inbox, unread=1, flags=uy~, folder_id=940, size=1, meta_version=3, version=1, parent_id=940, color=0, subject=Inbox, date=1647409817, rev=8702, deleted=0, del_unread=0, attributes=0}
trash=Folder{n=Trash, id=942, type=1, name=Trash, unread=0, flags=y~, folder_id=940, size=0, meta_version=3, version=1, parent_id=940, color=0, subject=Trash, date=1647409818, rev=8705, deleted=0, del_unread=0, attributes=0}
sent=Folder{n=Sent, id=943, type=1, name=Sent, unread=0, flags=y~, folder_id=940, size=0, meta_version=3, version=1, parent_id=940, color=0, subject=Sent, date=1647409818, rev=8708, deleted=0, del_unread=0, attributes=0}
junk=Folder{n=Junk, id=944, type=1, name=Junk, unread=0, flags=y~, folder_id=940, size=0, meta_version=3, version=1, parent_id=940, color=0, subject=Junk, date=1647409819, rev=8711, deleted=0, del_unread=0, attributes=0}
archive=Folder{n=Archive, id=945, type=1, name=Archive, unread=0, flags=y~, folder_id=940, size=0, meta_version=3, version=1, parent_id=940, color=0, subject=Archive, date=1647409820, rev=8714, deleted=0, del_unread=0, attributes=0}
```

Attching screenshot for same

Screenshot 2022-03-16 at 11 47 44 AM

When same Yahoo account configured in Outlook Draft Folder is ther and it is named as "Draft"
this code change wouldn't be much if we have Draft folder passed as an input as we can check for "contains" instead of "equals"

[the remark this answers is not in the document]
desouzas: Relying on `contains` is not ideal either. Yahoo is one of many providers. The special folder can be mapped to different names.

store/src/java/com/zimbra/cs/datasource/imap/MessageChanges.java
gabhinavTI: The code is a reference from findChanges() only and removed the tombstones check as it's not required here to check for that

The deletion of the draft if any are taken care of before steps, hence removing the tombstones code and check from this as it would be repeatability of same. 
In this, we are only bothered about the draft that is locally modified.
It runs in all scenarios to check for drafts that have been modified if there are any modified it return with modified item details else returns empty

desouzas: You're still querying twice, in separate locations? Will you receive _more_ results the second time if things are happening from other threads (e.g. other clients deleting things, modifying things, etc).

store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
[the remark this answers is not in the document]
desouzas: Alternatively, see if using the localfolder's id and MailConstants drafts folderId works.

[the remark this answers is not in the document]
gabhinavTI: The thing is we don't have a draft folder available as an input in the code , and even in UI we don't see draft folder

store/src/java/com/zimbra/cs/datasource/imap/MessageChanges.java
gabhinavTI: In ideal scenario or when I was doing testing it never entered findChanges() in syncMesage as `fullSync` remains true in all possible scenarios . but for this we can check for `!fullsync`
If there is any modification happened by other thread still it would sync up those messages which has been modified locally

store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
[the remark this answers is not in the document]
gabhinavTI: I see that for Yahoo draft folder is ignored 
` <folder remotePath="Draft" ignore="true"/>`
I think, for this reason, the Draft folder for Yahoo is not visible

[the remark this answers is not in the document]
desouzas: Either way, using the localfolder id will allow it work if the draft folder is being synced (since the mapping would have been done elsewhere) while using the folder path will only work if the path matches?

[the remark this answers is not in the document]
gabhinavTI: Using Folder path will only allow if the patch matches as of now,
it can be changed to contains if you think it would be appropriate to do it

store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
gabhinavTI: The tomstone wouldn't be required here as delete operation is taken care in previous scenarios, hence removed from here
I'll quickly do the test by updating the changeId , I assume we have to handle it
Thanks for this. I'll update test and let know

store/src/java/com/zimbra/cs/datasource/imap/ImapFolderSync.java
gabhinavTI: done

store/src/java/com/zimbra/cs/datasource/imap/MessageChanges.java
desouzas: Try non force sync - polling interval.

store/src/java/com/zimbra/cs/datasource/imap/MessageChanges.java
gabhinavTI: Yes they were capital, changing it back to capital

gabhinavTI:
> * Try to use the modern client to modify a draft that was moved to the trash - curious what happens.
> * Make sure deleting drafts is still captured.

When the draft is moved to the trash and modified, it won't sync up. The reason is we have restricted the draft sync up only for the drafts folder. Is it even required to consider a draft message in Trash folder? I yes then we can add conditions to check for modified message in Trash folder as well

Deleting of drafts is captured and working as expected

store/src/java/com/zimbra/cs/datasource/imap/MessageChanges.java
gabhinavTI: if we happen to call findChanges() which also looks for tombstones , in case of drafts this errors out . hence we need to have a separate method which won't look for tombstone and looks for only modified items findChangesForDraft()

PR #1196, opened 2021-09-22
ZCS-10914: Introduce calendar resolution preference backed by LDAP attribute

log2akshat:
**Problem:**
Introduce calendar resolution preference backed by the LDAP attribute.

**Approach:**
- Introduced a new LDAP attribute `zimbraPrefCalenderScaling` on the `account` and `COS` level to have the calendar resolution preference for `10, 15, 30` minutes keeping the default value as `30 minutes`.
- Reverted the [ZCS-10883](https://jira.corp.synacor.com/browse/ZCS-10883) changes for the temporary boolean `local-config` for `15 minutes` calendar resolution.

**Testing Done:**
- Created a new build with these changes and verified that the newly introduced LDAP attribute `zimbraPrefCalenderScaling`  is available at the `account` and `COS` levels.
- Verified with the following `GetPrefsRequest` that the newly introduced LDAP attribute `zimbraPrefCalenderScaling` is now available as `GetPrefsResponse`.
```                                                                                                                                                               
<GetPrefsRequest xmlns="urn:zimbraAccount">
</GetPrefsRequest>
```

```
<GetPrefsResponse xmlns="urn:zimbraAccount">
  <pref name="zimbraPrefCalendarReminderMobile">FALSE</pref>
....
....
   <pref name="zimbraPrefOpenMailInNewWindow">FALSE</pref>
  <pref name="zimbraPrefMailSignatureStyle">outlook</pref>
  <pref name="zimbraPrefCalenderScaling">15</pref>
  <pref name="zimbraPrefAdminConsoleWarnOnExit">TRUE</pref>
....
....
</GetPrefsResponse>
```
- Verified that the newly introduced LDAP attribute `zimbraPrefCalenderScaling` default value is `30 minutes`.
- Verified that the newly introduced LDAP attribute `zimbraPrefCalenderScaling` accepts the ENUM (10 mins, 15 mins, 30 mins) by changing them.
 ```
zimbra@dev-akshat:~$ zmprov -l mc default zimbraPrefCalenderScaling 15
zimbra@dev-akshat:~$ zmprov -l gc default zimbraPrefCalenderScaling
# name default
zimbraPrefCalenderScaling: 15
```
- Verified that the newly introduced LDAP attribute `zimbraPrefCalenderScaling` accepts only the valid ENUM values.
```
zimbra@dev-akshat:~$ zmprov -l mc default zimbraPrefCalenderScaling 100
[...]

PR #1192, opened 2021-09-08
ZCS-10883: Create a boolean local-config for fifteen minutes calendar resolution

log2akshat:
**Problem:**
Create a boolean local-config for fifteen minutes calendar resolution.

**Approach and Fix:**
- Added local config `zimbraPrefCalenderScaling` for fifteen minutes calendar resolution.
- Made the LC value available as pref on the front-end side based on the boolean condition of the value of `zimbraPrefCalenderScaling`.

**Testing Done:**
- Verified with the following `GetPrefsRequest` that the local-config value `zimbraPrefCalenderScaling` is now available as `GetPrefsResponse`.
```                                                                                                                                                               
<GetPrefsRequest xmlns="urn:zimbraAccount">
</GetPrefsRequest>
```

```
<GetPrefsResponse xmlns="urn:zimbraAccount">
  <pref name="zimbraPrefCalendarReminderMobile">FALSE</pref>
....
....
  <pref name="zimbraPrefCalendarReminderSendEmail">FALSE</pref>
  <pref name="zimbraPrefCalendarSendInviteDeniedAutoReply">FALSE</pref>
  <pref name="zimbraPrefIMIdleTimeout">10</pref>
  <pref name="zimbraPrefCalenderScaling">15</pref>
</GetPrefsResponse>
```

PR #1241, opened 2022-03-25
ZCS-10678: Server Side work to Force users not to use username in the password

gabhinavTI:
https://jira.corp.synacor.com/browse/ZCS-10678
Feature: Server Side work to Force users not to use username in the password

Problem:
One of the customer (The Open university of Tanzania) is looking for an option in password policy - users should not be able to use their username in the password.

Code fix:
As discussed on the approach, we would be making using of Local config to enable the feature is password can contain the username or not. If the value is set to true then we check in checkPasswordStrength() to check for password containing username and throw a relevant exception

Test:
Through UI and soap UI API calls, the relevant exception is thrown

### DIFF
--- a/common/src/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/java/com/zimbra/common/localconfig/LC.java
@@ -1466,6 +1466,8 @@ public final class LC {
 
     public static final KnownKey invite_ignore_x_alt_description = KnownKey.newKey(true);
 
+    public static final KnownKey allow_username_within_password = KnownKey.newKey(false);
+
     static {
         // Automatically set the key name with the variable name.
         for (Field field : LC.class.getFields()) {
--- a/store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
+++ b/store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
@@ -6147,6 +6147,12 @@ public class LdapProvisioning extends LdapProv implements CacheAwareProvisioning
                     new Argument(Provisioning.A_zimbraPasswordMaxLength, maxLength, Argument.Type.NUM));
         }
 
+        if (LC.allow_username_within_password.booleanValue()
+                && StringUtils.containsIgnoreCase(password, acct.getUCUsername())) {
+            throw AccountServiceException.INVALID_PASSWORD("password contains username",
+                    new Argument("zimbraPasswordAllowUsername", "", Argument.Type.STR));
+        }
+
         if (getBoolean(acct, cos, entry, Provisioning.A_zimbraPasswordBlockCommonEnabled, false)
                 && commonPasswordFilter.mightContain(password)) {
             throw AccountServiceException.INVALID_PASSWORD("too common",
--- a/store/src/java/com/zimbra/cs/service/account/GetInfo.java
+++ b/store/src/java/com/zimbra/cs/service/account/GetInfo.java
@@ -34,6 +34,7 @@ import com.google.common.collect.Sets;
 import com.zimbra.common.account.Key;
 import com.zimbra.common.account.Key.AccountBy;
 import com.zimbra.common.account.ProvisioningConstants;
+import com.zimbra.common.localconfig.LC;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.soap.AccountConstants;
 import com.zimbra.common.soap.Element;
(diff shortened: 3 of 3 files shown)


shrutig0510 (approved):

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
```diff
@@ -6147,6 +6147,11 @@ private void checkPasswordStrength(String password, Account acct, Cos cos, ZMuta
                     new Argument(Provisioning.A_zimbraPasswordMaxLength, maxLength, Argument.Type.NUM));
         }
 
+        if (LC.password_contains_username.booleanValue()
+                && StringUtils.containsIgnoreCase(password, acct.getUCUsername())) {
+            throw AccountServiceException.INVALID_PASSWORD("password contains username");
```
silentsakky, line 6152: on UI side, are we just going to handle this exception or we will be checking this logic on UI side as well?
recently we added these password related logics to UI screens also so user experience can be improved

gabhinavTI: I think on UI side we only need to handle exceptions, as the check is based on a local config value for which check is happening on the backend side. I am not sure the same check can be implemented on the UI side or not
@dawoodshaikh can better answer the same

common/src/java/com/zimbra/common/localconfig/LC.java
```diff
@@ -1466,6 +1466,8 @@
 
     public static final KnownKey invite_ignore_x_alt_description = KnownKey.newKey(true);
 
+    public static final KnownKey password_contains_username = KnownKey.newKey(false);
```
dasiyogesh, line 1469: config name doesn't sound right about its conditional usage, should we change it to `password_contains_username_allowed` or `allow_username_within_password`

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
gabhinavTI: @silentsakky @dawoodshaikh , I have changed the error code to the existing error code being used in password scenarios and this is the error message being displayed, I think we would not require UI work on this

Could you please confirm same

image

preshitaagrawal (approved):

zimsuchitgupta (approved):

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
gabhinavTI: @silentsakky , As a discussed updated the GetInfo call to include `zimbraPasswordAllowUsername` in the response
Also updated the Exception statement in checkPasswordStrength() for username case to pass on the argument which can be used on UI side `zimbraPasswordAllowUsername`

GetInfo response:
image

Password change response which contains keyword which can be used to show exact error on UI side
`{"Header":{"context":{"session":{"id":"12","_content":"12"},"change":{"token":1499},"_jsns":"urn:zimbra"}},"Body":{"Fault":{"Code":{"Value":"soap:Sender"},"Reason":{"Text":"invalid password: password contains username"},"Detail":{"Error":{"Code":"account.INVALID_PASSWORD","Trace":"qtp510109769-144:1648491353457:ffcdb33d43d3cabf","a":[{"n":"zimbraPasswordAllowUsername","t":"STR","_content":""}],"_jsns":"urn:zimbra"}}}},"_jsns":"urn:zimbraSoap"}`

gabhinavTI: @silentsakky , updated the Getinfo response `zimbraPasswordAllowUsername` to return in capaitalize form

k-kato (commented):
Regarding the new localconfig key `allow_username_within_password`, it is okay to add a LC key for a patch of 9.0 or 8.8.15. However, it should be not LC key but LDAP attribute in `develop` branch.

Suggestion:
* change the name of key from `allow_username_within_password` to `zimbraPasswordAllowUsername` (according to LdapPrivisioning line 6153 and GetInfo line 367).
    * note: this PR will be able to ported to patches.
* create another PR to change the LC key to a ldap attribute with the same name. All the three java files needs to be update as well.
 
Reference:
* https://github.com/Zimbra/zm-mailbox/pull/1192  => It has been ported into 9.0.0.p19 (commit id: e69fd65584dedc355414312606cd2310a5f657a9)
* https://github.com/Zimbra/zm-mailbox/pull/1196  => It has not been ported into the patch.

dasiyogesh (approved):

k-kato:
> change the name of key from allow_username_within_password to zimbraPasswordAllowUsername (according to LdapPrivisioning line 6153 and GetInfo line 367). 

the first item is optional.

k-kato (approved):

silentsakky (approved):

PR #1243, opened 2022-03-25
ZBUG-2640:Able to add member in DL and DDL without having that accoun…

swaatiTelus:
Bug: https://jira.corp.synacor.com/browse/ZBUG-2640

Problem: Able to add member in DL and DDL without having that account in that domain.

Solution: Not allowing user to add an invalid account by throwing exception if member account does not exist.

Fix Mode: After getting account details if account is null than throwing exception of no such account exists.

Test Done: Manual test done.
Test case: zmprov adlm 2640@zimbra.test invalidUser@zimbra.test
ERROR: account.NO_SUCH_ACCOUNT (no such account: invalidUser@zimbra.test)


store/src/java/com/zimbra/cs/service/admin/AddDistributionListMember.java
```diff
@@ -63,6 +65,13 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
             throw ServiceException.INVALID_REQUEST("members to add not specified", null);
         }
 
+        for (String member : memberList) {
```
zimsuchitgupta, line 68: Rather than adding this logic here we should check the method prov.addGroupMembers(group, members); as this is responsible for adding DLs. So we should check what's breaking or need to improve logic.

swaatiTelus: Yes, I understand the logic is written inside the method prov.addGroupMembers(group, members);
It is breaking for not validating account before adding it to the DL.
My motive is to write this here is that we will be checking whether the account is valid or not beforehand so that we could save method calling when it is not required at all and throw appropriate exception to user. Also, inside addGroupMembers() we need to write this throw exception code at 2 places, one for addDynamicGroupMembers() and another for addDistributionListMembers().
Please let me know if, I should do this inside addGroupMembers(). Thanks!

abhishekdeshmukh2209: How often is this condition being true where we get account as null?
If it is often happening then it makes sense to do as below but if it is not happening often then we can add it inside both the methods as then we will not need to process list of members everytime.

swaatiTelus: It is unpredictable, could be frequent or could not happen for a long time. As input is taken from user so can't determine its occurrence.
So now writing it inside both of the methods as suggested.

zimsuchitgupta (approved):
LGTM

shrutig0510 (approved):

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
```diff
@@ -10570,6 +10575,11 @@ private void addDynamicGroupMembers(LdapDynamicGroup group, String[] members)
             String memberName = member.toLowerCase();
             memberName = IDNUtil.toAsciiEmail(memberName);
 
+            Account account = getAccountByName(memberName);
```
dasiyogesh, line 10578: `getAccountByName()` calls `get(AccountBy.name, name)` to fetch the account, so getting account null on line **10579** is invalid account and on line **10583** getting account null could is being checked for **group/group unit address** isn't it contradicting each other ?

PR #1250, opened 2022-04-08
ZBUG-1455 : zimbraFeatureMailForwardingInFiltersEnabled, Attribute funtionality is wrong

rupeshtelus:
[ZBUG-1455](https://jira.corp.synacor.com/browse/ZBUG-1455)

Problem:
AccDescription of the zimbraFeatureMailForwardingInFiltersEnabled need to change . While we are hitting to get descriptions of zimbraFeatureMailForwardingInFiltersEnabled we are getting result as ( enable end-user mail forwarding defined in mail filters features ) but it should be ( enable end-user mail redirecting defined in mail filters features )

Solution:
I have replaced redirecting in the place of forwarding as we are doing operations of redirect on this attribute .

Testing:
Tested on a new build for above changes as for this bug we have changed on zimbra-attrs.xml file . Description is working as expected .

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -3720,7 +3720,7 @@ TODO - add support for multi-line values in globalConfigValue and defaultCOSValu
 
 <attr id="704" name="zimbraFeatureMailForwardingInFiltersEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInfo,accountInherited,domainAdminModifiable" since="5.0.10">
   <defaultCOSValue>TRUE</defaultCOSValue>
-  <desc>enable end-user mail forwarding defined in mail filters features</desc>
+  <desc>enable end-user mail redirecting defined in mail filters features</desc>
 </attr>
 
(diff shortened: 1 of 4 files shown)


lavegupta (approved):

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -3720,7 +3720,7 @@ TODO - add support for multi-line values in globalConfigValue and defaultCOSValu
 
 <attr id="704" name="zimbraFeatureMailForwardingInFiltersEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInfo,accountInherited,domainAdminModifiable" since="5.0.10">
   <defaultCOSValue>TRUE</defaultCOSValue>
-  <desc>enable end-user mail forwarding defined in mail filters features</desc>
+  <desc>enable end-user mail redirecting defined in mail filters features</desc>
```
dasiyogesh, line 3723: @rupeshtelus this being LDAP change and the bug affected version is 8.8.x, this change won't make update for older version as in the patch release LDAP schema won't be updated. Considering this do we need this change?

rupeshtelus: @dasiyogesh as I had a discussion with @dawoodshaikh , he is ok to validate this LDAP changes in 9.1.0 Release

dasiyogesh (approved):

PR #1252, opened 2022-04-19
ZBUG-1403: Not able to set zimbraDistributionListSendShareMessageToNewMembers on Dynamic Distribution lists

swaatiTelus:
[ZBUG-1403](https://jira.corp.synacor.com/browse/ZBUG-1403)

Problem: User is Not able to set zimbraDistributionListSendShareMessageToNewMembers on Dynamic Distribution lists promoting error “ERROR: service.FAILURE (system failure: unable to modify attrs: object class violation - unable to modify attributes”, hence also it is not working through admin panel. But for dynamic list it works fine.

Solution: Any changes in LDAP attribute takes place in zimbra-attrs.xml file. Added group in the scope of zimbraDistributionListSendShareMessageToNewMembers and created getters.

Testing:
Did testing on my local system by modifying attribute from CLI as well as from admin panel. Now we are able to modify this attribute's value.

Additionally:

During above fix, 2 more attributes were found to be prompting error,
- zimbraDistributionListSendShareMessageFromAddress and 
- zimbraChatAllowDlMemberAddAsFriend
So, fix has been done for them also.

Note:
ZBUG-1403 affected version is 8.8.x. this change won't update for older version as in the patch release LDAP schema won't be updated. So this fix will get reflected in 9.1 version onwards.

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -4297,14 +4297,14 @@ TODO - add support for multi-line values in globalConfigValue and defaultCOSValu
   <desc>Maximum percent of conflicting instances allowed before declining schedule request for a recurring appointment; default 0 (means decline on any conflict)</desc>
 </attr>
 
-<attr id="810" name="zimbraDistributionListSendShareMessageToNewMembers" type="boolean" cardinality="single" optionalIn="distributionList" since="6.0.0_BETA1">
+<attr id="810" name="zimbraDistributionListSendShareMessageToNewMembers" type="boolean" cardinality="single" optionalIn="distributionList,group" since="6.0.0_BETA1">
   <desc>
     Whether to send an email with all the shares of the group when a new member is added to the group.
     If not set, default is to send the email.
   </desc>
 </attr>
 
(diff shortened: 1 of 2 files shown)


store/conf/attrs/zimbra-attrs.xml
```diff
@@ -4297,14 +4297,14 @@ TODO - add support for multi-line values in globalConfigValue and defaultCOSValu
   <desc>Maximum percent of conflicting instances allowed before declining schedule request for a recurring appointment; default 0 (means decline on any conflict)</desc>
 </attr>
 
-<attr id="810" name="zimbraDistributionListSendShareMessageToNewMembers" type="boolean" cardinality="single" optionalIn="distributionList" since="6.0.0_BETA1">
+<attr id="810" name="zimbraDistributionListSendShareMessageToNewMembers" type="boolean" cardinality="single" optionalIn="distributionList,group" since="6.0.0_BETA1">
```
silentsakky, line 4300: these changes will go in 9.1.0 only and can not be backported to 9.0.0 or 8.8.15, I hope the expectations are already cleared with product management

swaatiTelus: Yes Saket!

dasiyogesh (approved):
LGTM for 9.1.0

silentsakky (approved):

PR #1254, opened 2022-04-21
PREAPPS-6695: Added ignoreSameSite param, if true will not add Same SiteStrict cookie

zimsuchitgupta:
**Problem**
User was not able to login in Zimbra Desktop when SameSite=Strict

**Fix**
Added a new boolean param in AuthRequest named **ignoreSameSite**, By default its value is false.
If it is set to true then SameSite=Strict will not be added in ZM_AUTH_TOKEN cookie.

**Testing**
Verified on local, able to login in zimbra desktop.

### DIFF
--- a/common/src/java/com/zimbra/common/util/ZimbraCookie.java
+++ b/common/src/java/com/zimbra/common/util/ZimbraCookie.java
@@ -56,11 +56,16 @@ public class ZimbraCookie {
 
     public static void addHttpOnlyCookie(HttpServletResponse response, String name, String value,
             String path, Integer maxAge, boolean secure) {
-        addCookie(response, name, value, path, maxAge, true, secure);
+        addCookie(response, name, value, path, maxAge, true, secure, false);
+    }
+
+    public static void addHttpOnlyCookie(HttpServletResponse response, String name, String value,
+            String path, Integer maxAge, boolean secure, boolean ignoreSameSite) {
+        addCookie(response, name, value, path, maxAge, true, secure, ignoreSameSite);
     }
 
     private static void addCookie(HttpServletResponse response, String name, String value,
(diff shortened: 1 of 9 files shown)


silentsakky (approved):

common/src/java/com/zimbra/common/util/ZimbraCookie.java
```diff
@@ -56,11 +56,16 @@ public static boolean secureCookie(HttpServletRequest request) {
 
     public static void addHttpOnlyCookie(HttpServletResponse response, String name, String value,
             String path, Integer maxAge, boolean secure) {
-        addCookie(response, name, value, path, maxAge, true, secure);
+        addCookie(response, name, value, path, maxAge, true, secure, false);
```
dasiyogesh, line 59: here we should call overloaded method addHttpOnlyCookie() with default value false for new param instead of addCookie().
addCookie will get called by overloaded method addHttpOnlyCookie()

store/src/java/com/zimbra/cs/service/account/ChangePassword.java
```diff
@@ -108,6 +108,7 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
         Element response = zsc.createElement(AccountConstants.CHANGE_PASSWORD_RESPONSE);
         if (!dryRun) {
            AuthToken at = AuthProvider.getAuthToken(acct);
+           at.setIgnoreSameSite(request.getAttributeBool(AccountConstants.A_IGNORE_SAME_SITE_COOKIE, false));
```
silentsakky, line 111: if I remember correctly then changepassword just changes the password and doesn't update auth token, so why do we need this?

zimsuchitgupta:
Create a new PR for same https://github.com/Zimbra/zm-mailbox/pull/1256

PR #1257, opened 2022-04-28
ZCS-11344: set default value of SameSite cookie to empty

zimsuchitgupta:
**Problem**

From regression testing, we found some login issues in Desktop and Web.

**Solution**
We have decided that we will set default value zimbra_same_site_cookie to Empty.


shrutig0510 (approved):

deepakzimb (approved):
LGTM

PR #1258, opened 2022-04-29
ZBUG-2662:When using preauth, CSRF tokens are not checked on some pos…

javierflores05telus:
Solution
ZBUG-2662: When using preauth, CSRF tokens are not checked on some post endpoints
In a normal flow authentication with user and password the validation of csrfToken works properly but when you are using 
preauth authentication those validations are missing.
So in PreAuthServlet we have multiples conditions to create the authToken where is the property csrfTokenEnabled,
now it is required set this property as TRUE to make sure csrfToken validation will be present.

Notes
preauth authentication is a particular flow where is not required your password to go to mailbox client and use the
preauth key generated as authentication

Tests
Please see ticket for further details.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/PreAuthServlet.java
+++ b/store/src/java/com/zimbra/cs/service/PreAuthServlet.java
@@ -248,6 +248,7 @@ public class PreAuthServlet extends ZimbraServlet {
                     else
                         at = (expires ==  0) ? AuthProvider.getAuthToken(acct) : AuthProvider.getAuthToken(acct, expires);
 
+                    at.setCsrfTokenEnabled(true);
                     setCookieAndRedirect(req, resp, at);
 
                 } else {


shrutig0510 (approved):

store/src/java/com/zimbra/cs/service/PreAuthServlet.java
```diff
@@ -248,6 +248,7 @@ public void doGet(HttpServletRequest req, HttpServletResponse resp)
                     else
                         at = (expires ==  0) ? AuthProvider.getAuthToken(acct) : AuthProvider.getAuthToken(acct, expires);
 
+                    at.setCsrfTokenEnabled(true);
```
silentsakky, line 251: we are generating auth token on line no 145 as well, don't we need this fix there as well?

zimsuchitgupta (approved):

store/src/java/com/zimbra/cs/service/PreAuthServlet.java
zimsuchitgupta: @silentsakky That authtoken is generated by auth token passed in request while doing Preauth so will keep all the original properties. So no need to handle here.

silentsakky (approved):

dasiyogesh (approved):

RakeshAMore (approved):

PR #1261, opened 2022-05-04
ZCS-11318 Add zimbraWebClientSkipLogoff to ClientInfo. LC implementation for patch support.

dasiyogesh:
Webclient needs the list of SAML extension logout urls to be compared with **zimbraWebClientLogoutURL** during ClientInfoRequest and to decide whether to skip logoff of token de-registration. 

During ClientInfoRequest, backend will read this list from LC **zimbra_web_client_logoff_urls** which will be a space separated list of URLs and compare to **zimbraWebClientLogoutURL** for any match and if present will respond with zimbraWebClientSkipLogoff TRUE else FALSE

* Webclient can use this to end session, then send the user to the logout url to end tokens.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/account/ClientInfo.java
+++ b/store/src/java/com/zimbra/cs/service/account/ClientInfo.java
@@ -16,8 +16,11 @@
  */
 package com.zimbra.cs.service.account;
 
+import java.util.Arrays;
 import java.util.Map;
+import org.apache.commons.lang.StringUtils;
 
+import com.zimbra.common.localconfig.LC;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.soap.AccountConstants;
 import com.zimbra.common.soap.Element;
@@ -42,6 +45,7 @@ public class ClientInfo extends AccountDocumentHandler {
 
         Element parent =  zsc.createElement(AccountConstants.CLIENT_INFO_RESPONSE);
         if (domain != null) {
+            String webClientLogoutURL = domain.getWebClientLogoutURL();
             ToXML.encodeAttr(parent, Provisioning.A_zimbraFeatureResetPasswordStatus, domain.getFeatureResetPasswordStatusAsString());
             ToXML.encodeAttr(parent, Provisioning.A_zimbraSkinBackgroundColor, domain.getSkinBackgroundColor());
             ToXML.encodeAttr(parent, Provisioning.A_zimbraSkinFavicon, domain.getSkinFavicon());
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/service/account/ClientInfo.java
```diff
@@ -52,12 +56,36 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
             ToXML.encodeAttr(parent, Provisioning.A_zimbraSkinSecondaryColor, domain.getSkinSecondaryColor());
             ToXML.encodeAttr(parent, Provisioning.A_zimbraSkinSelectionColor, domain.getSkinSelectionColor());
             ToXML.encodeAttr(parent, Provisioning.A_zimbraWebClientLoginURL, domain.getWebClientLoginURL());
-            ToXML.encodeAttr(parent, Provisioning.A_zimbraWebClientLogoutURL, domain.getWebClientLogoutURL());
+            ToXML.encodeAttr(parent, Provisioning.A_zimbraWebClientLogoutURL, webClientLogoutURL);
```
shrutig0510, line 59: Do we need this change...Do we need variable declaration at line 48.

store/src/java/com/zimbra/cs/service/account/ClientInfo.java
```diff
@@ -52,12 +56,36 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
+     * when sending EndSession request, if the configured zimbraWebClientLogoutURL is
+     * known to handle token de-registration.
+     *
+     * @param parent The element to add the attr to
+     * @param webClientLogoutURL The configured webclient logout url
+     * @param logoffURLs URLs to skip logoff for - we expect them to handle it when we send the user there
+     */
+    protected void encodeAttrSkipLogoff(Element parent, String webClientLogoutURL, String[] logoffURLs) {
+        // always include the attr if webclient logout url is non-empty
+        if (StringUtils.isNotEmpty(webClientLogoutURL)) {
```
shrutig0510, line 79: We have one more method to check both bull and empty on String in com.zimbra.common.util.StringUtil class

abhishekdeshmukh2209 (approved):
LGTM

RakeshAMore (approved):

store/src/java/com/zimbra/cs/service/account/ClientInfo.java
dasiyogesh: I think we should update to null & empty check, as the default value for attribute is null if not set

dasiyogesh: StringUtils.isNotEmpty does the same, so its fine.

shrutig0510: OK

shrutig0510 (approved):

deepakzimb (approved):

zimsuchitgupta (approved):

PR #1274, opened 2022-05-25
ZCS-11191: [Backend] Integrate postsrsd

carlosromero03:


common/src/java/com/zimbra/common/localconfig/LC.java
```diff
@@ -1527,7 +1527,9 @@ public enum PUBLIC_SHARE_VISIBILITY { samePrimaryDomain, all, none };
     public static final KnownKey zimbra_strict_unclosed_comment_tag = KnownKey.newKey(true);
     public static final KnownKey zimbra_skip_tags_with_unclosed_cdata = KnownKey.newKey("style");
 
-    
+    //ZCS-11191 : Allow to enable/disable SRS feature
+    public static final KnownKey zimbra_srs_enabled = KnownKey.newKey(false);
```
silentsakky, line 1531: any reason to put this into LC instead of LDAP? are we planning to release this feature in a patch?

store/src/java/com/zimbra/cs/service/admin/ModifyServer.java
```diff
@@ -58,6 +67,8 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
         }
         checkRight(zsc, context, server, attrs);
 
+        startOrStopPostSRSd(server, attrs);
```
silentsakky, line 70: is this going to be called with every modifyServer call?

swaatiTelus: @silentsakky, I think the same this code should not be here, as zmcontrol and all other service's stop/start functionality is written in zm-core-utils repo, which is in Perl script. This logic should also be written in that repo, also would like to confirm the same with @desouzas . 
@desouzas : Could you please provide your inputs here? Thanks..!!

PR #1276, opened 2022-06-01
ZBUG-1375 : IMAP Clients can not delete messages when account is close to quota

gabhinavTI:
Issue:
When an account is close to quota, users can not delete messages from IMAP clients.
Cause:  IMAP does a copy and then purge so when a user tries to move a message from INBOX to trash, the message will be copied to the trash folder and then purge from the inbox.  If the message exceeds the quota, the copy will fail.  

Code fix:
The issue was due to when messages were tried to delete from Inbox/Sent etc folders in the code, the messages were being copied and during the process, total email quota check is in place and once 90% quota is crossed the copy item fails as a result deletion operation fails.
To resolve this fix was to remove the quota check when the message is being deleted i.e. the message is being moved/deleted to the Trash folder(folder id = 3) as a result deletion operation is successful, and people will be able to delete the message and once emptied from Trash and able to send emails again

Test:
Able to delete emails from an IMAP account, checked from ZWC and no exception in mailbox logs


store/src/java/com/zimbra/cs/mailbox/MailItem.java
```diff
@@ -1914,7 +1918,11 @@ protected void finishCreation(MailItem parent) throws ServiceException {
         if (isLeafNode()) {
             boolean isDeleted = isTagged(Flag.FlagInfo.DELETED);
 
-            mMailbox.updateSize(mData.size, isQuotaCheckRequired());
+            if (trashFolderId == 3) {
```
dasiyogesh, line 1921: use folder constants for comparison refer https://github.com/Zimbra/zm-mailbox/blob/develop/common/src/java/com/zimbra/common/mailbox/FolderConstants.java#L27

store/src/java/com/zimbra/cs/mailbox/MailItem.java
```diff
@@ -2703,6 +2712,10 @@ MailItem copy(Folder folder, int copyId, String copyUuid, MailItem parent) throw
      *    <li><tt>service.PERM_DENIED</tt> - if you don't have sufficient
      *        permissions</ul> */
     MailItem icopy(Folder target, int copyId, String copyUuid) throws IOException, ServiceException {
+        return icopy(target, copyId, copyUuid, 0);
+    }
+
+    MailItem icopy(Folder target, int copyId, String copyUuid, int trashFolderId) throws IOException, ServiceException {
```
k-kato, line 2718: same

dasiyogesh, line 2718: @gabhinavTI do we need trashFolderId param,  as there is already target Folder object can we not get the Id from it for comparison? If yes we won't need overloading these methods

store/src/java/com/zimbra/cs/mailbox/MailItem.java
```diff
@@ -2753,7 +2766,11 @@ MailItem icopy(Folder target, int copyId, String copyUuid) throws IOException, S
         DbMailItem.icopy(this, data, shareIndex);
 
         MailItem copy = constructItem(mMailbox, data);
-        copy.finishCreation(null);
+        if (trashFolderId == FolderConstants.ID_FOLDER_TRASH) {
+            copy.finishCreation(null, trashFolderId);
+        } else {
+            copy.finishCreation(null);
+        }
```
k-kato, line 2773: same

store/src/java/com/zimbra/cs/mailbox/MailItem.java
gabhinavTI: @dasiyogesh `target` param is available only in icopy() , hence removed overloading for this method but other methods  `finishCreation()`  and `updateSize() `overloading is still required hence using boolean values as for isTrashFolder in paramter

store/src/java/com/zimbra/cs/mailbox/MailItem.java
```diff
@@ -1897,6 +1898,10 @@ void markBlobForDeletion(MailboxBlob mblob) {
      * @param parent  The created item's parent.  The parent's addChild()
      *                method will be called during the function. */
     protected void finishCreation(MailItem parent) throws ServiceException {
+        finishCreation(parent, 0);
+    }
+
+    protected void finishCreation(MailItem parent, int trashFolderId) throws ServiceException {
```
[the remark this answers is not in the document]
gabhinavTI: @k-kato replaced the trashFolderId with the boolean values 
done for all overloaded method

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -1456,12 +1456,16 @@ void updateSize(long delta) throws ServiceException {
     }
 
     void updateSize(long delta, boolean checkQuota) throws ServiceException {
+        updateSize(delta, checkQuota, 0);
+    }
+
+    void updateSize(long delta, boolean checkQuota, int trashFolderID) throws ServiceException {
```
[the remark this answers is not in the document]
gabhinavTI: replaced the trashFolderId with the boolean values
done for all overloaded method

store/src/java/com/zimbra/cs/mailbox/MailItem.java
gabhinavTI: replaced the trashFolderId with the boolean values
done for all overloaded method

store/src/java/com/zimbra/cs/mailbox/MailItem.java
```diff
@@ -1914,7 +1919,11 @@ protected void finishCreation(MailItem parent) throws ServiceException {
         if (isLeafNode()) {
             boolean isDeleted = isTagged(Flag.FlagInfo.DELETED);
 
-            mMailbox.updateSize(mData.size, isQuotaCheckRequired());
+            if (isTrashFolder) {
+                mMailbox.updateSize(mData.size, isQuotaCheckRequired(), isTrashFolder);
+            } else {
+                mMailbox.updateSize(mData.size, isQuotaCheckRequired());
+            }
```
k-kato, line 1926: It can be simplified without if-else.
`mMailbox.updateSize(mData.size, isQuotaCheckRequired(), isTrashFolder);`

note: please don't remove `void updateSize(long delta, boolean checkQuota) throws ServiceException` since it is not a private function.

store/src/java/com/zimbra/cs/mailbox/MailItem.java
```diff
@@ -2753,7 +2762,11 @@ MailItem icopy(Folder target, int copyId, String copyUuid) throws IOException, S
         DbMailItem.icopy(this, data, shareIndex);
 
         MailItem copy = constructItem(mMailbox, data);
-        copy.finishCreation(null);
+        if (target.getId() == FolderConstants.ID_FOLDER_TRASH) {
+            copy.finishCreation(null, true);
+        } else {
+            copy.finishCreation(null);
+        }
```
k-kato, line 2769: Can we simplify it without if-else as well? Is the following code less readable? What do you think?
`copy.finishCreation(null, target.getId() == FolderConstants.ID_FOLDER_TRASH);`

store/src/java/com/zimbra/cs/mailbox/MailItem.java
gabhinavTI: implemented as per suggestion, the updated method looks like below

```
    void updateSize(long delta, boolean checkQuota) throws ServiceException {
        updateSize(delta, checkQuota, false);
    }
```

store/src/java/com/zimbra/cs/mailbox/MailItem.java
gabhinavTI: This is much readable and also removed unwanted code , updated as per suggestion

k-kato (approved):
LGTM

dasiyogesh (approved):

shrutig0510 (approved):

PR #1282, opened 2022-06-08
ZCS-11425 Configure LDAP attributes for storing S3 configurations

inpati:
**Problem Statement**
Add LDAP attributes to store S3 configurations

**Solution**
Added New Attributes:
1. zimbraGlobalExternalStoreConfig -> global level
2. zimbraServerExternalStoreConfig -> server level

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9929,4 +9929,10 @@ TODO: delete them permanently from here
   <defaultCOSValue>30</defaultCOSValue>
   <desc>Default calendar resolution preference</desc>
 </attr>
+<attr id="5000" name="globalExternalStoreConfig" type="string" cardinality="single" optionalIn="globalConfig" since="9.1.0">
+  <desc>Global level external store config</desc>
+</attr>
+<attr id="5001" name="serverExternalStoreConfig" type="string" cardinality="single" optionalIn="server" since="9.1.0">
+  <desc>Server level external store config</desc>
+</attr>
 </attrs>
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -9929,10 +9929,10 @@ TODO: delete them permanently from here
   <defaultCOSValue>30</defaultCOSValue>
   <desc>Default calendar resolution preference</desc>
 </attr>
-<attr id="5000" name="zimbraGlobalExternalStoreConfig" type="string" cardinality="multi" optionalIn="globalConfig" since="9.1.0">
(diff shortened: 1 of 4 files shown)


store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9929,4 +9929,10 @@ TODO: delete them permanently from here
   <defaultCOSValue>30</defaultCOSValue>
   <desc>Default calendar resolution preference</desc>
 </attr>
+<attr id="5000" name="zimbraGlobalExternalStoreConfig" type="string" cardinality="single" optionalIn="globalConfig" since="9.1.0">
```
dasiyogesh, line 9932: since there can be multiple configs, I think we  should make multi value attribute

inpati: Updated cardinality="multi"

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9929,4 +9929,10 @@ TODO: delete them permanently from here
   <defaultCOSValue>30</defaultCOSValue>
   <desc>Default calendar resolution preference</desc>
 </attr>
+<attr id="5000" name="zimbraGlobalExternalStoreConfig" type="string" cardinality="multi" optionalIn="globalConfig" since="9.1.0">
```
deepakzimb, line 9932: why we are using multi-value attributes?  we will store info in json object and it would be a single json object - right?

deepakzimb, line 9932: Also, last added attr id was 404, not you start with 5000, why so?

store/conf/attrs/zimbra-attrs.xml
deepakzimb: What is here "multiple config" - we are storing info in one json object

@RakeshAMore - do you want to store multiple json object and if yes please tell one test case?

store/conf/attrs/zimbra-attrs.xml
inpati: The id is not sequential, please go through the file. Some where it starts from 3xxx and for some attributes it starts for 5xxxx.
I started totally new series for storage management attributes.

inpati: Done !

store/conf/attrs/zimbra-attrs.xml
inpati: This atrr should be single value, reverted to siungle cardinality

RakeshAMore (approved):

dasiyogesh (approved):

zimsuchitgupta (approved):

PR #1283, opened 2022-06-10
ZCS-11499: Regression - Message written in CP932 is displayed incorrectly (garbled)

wilfredo-alfaro:
### Solution

The issue described in ZCS-11499 was introduced by the ZCS-11096 implementation for External Email Warning (EEW), so that it would prevent for messages from being saved and rendered correctly under certain character encoding scenarios, even when the EEW feature was turned off. As a solution, further changes have been done on the `ZimbraLmtpBackend `class in order to make it compatible with all charsets described in the `MimeConstants `class rather than expecting a default charset, which was the root cause for the issue.

### Notes

This issue was first identified from incompatibility the CP932 charset, also known as Windows-31J, and the solution fixes that. However, it also works for other charsets, including:

- us-ascii
- utf-8
- iso-8859-1
- windows-1252
- euc_cn
- gb2312
- gbk
- Windows-31J
- shift_jis

### Testing Done

### DIFF
--- a/store/src/java/com/zimbra/cs/lmtpserver/ZimbraLmtpBackend.java
+++ b/store/src/java/com/zimbra/cs/lmtpserver/ZimbraLmtpBackend.java
@@ -48,6 +48,7 @@ import com.zimbra.common.lmtp.LmtpClient;
 import com.zimbra.common.lmtp.LmtpProtocolException;
 import com.zimbra.common.localconfig.DebugConfig;
 import com.zimbra.common.localconfig.LC;
+import com.zimbra.common.mime.MimeConstants;
 import com.zimbra.common.mime.Rfc822ValidationInputStream;
 import com.zimbra.common.service.DeliveryServiceException;
 import com.zimbra.common.service.ServiceException;
@@ -330,6 +331,13 @@ public class ZimbraLmtpBackend implements LmtpBackend {
         }
     }
 
+    // ZCS-1149/ZCS-11096
+    private static final String[] CHARSET_NAMES = { MimeConstants.P_CHARSET_ASCII, MimeConstants.P_CHARSET_UTF8,
+            MimeConstants.P_CHARSET_LATIN1, MimeConstants.P_CHARSET_WINDOWS_1252, MimeConstants.P_CHARSET_EUC_CN,
+            MimeConstants.P_CHARSET_GB2312, MimeConstants.P_CHARSET_GBK, MimeConstants.P_CHARSET_WINDOWS_31J,
+            MimeConstants.P_CHARSET_SHIFT_JIS, MimeConstants.P_CHARSET_DEFAULT };
+    private static final String GARBLED_CHARACTER = String.valueOf((char) 0xfffD);
+
     @Override
     public void deliver(LmtpEnvelope env, InputStream in, int sizeHint) throws UnrecoverableLmtpException {
         CopyInputStream cis = null;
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/lmtpserver/ZimbraLmtpBackend.java
```diff
@@ -339,11 +347,22 @@ public void deliver(LmtpEnvelope env, InputStream in, int sizeHint) throws Unrec
...
-            inEEW = IOUtils.toInputStream(contentEEW);
+            for (int i = 0; i < CHARSET_NAMES.length; i++) {
+                final String charsetName = CHARSET_NAMES[i];
+                // copy input stream and evaluate content encoding
+                final byte[] ba = IOUtils.toByteArray(in);
+                IOUtils.closeQuietly(in);
+                in = new ByteArrayInputStream(ba);
+                final String content = new String(ba, charsetName);
+                // if current encoding is suitable, use it; otherwise, continue or use default
+                if (content.contains(GARBLED_CHARACTER) && i < CHARSET_NAMES.length - 1) {
```
shrutig0510, line 358: Why do we need this "i < CHARSET_NAMES.length - 1" check here?

swaatiTelus: As per the understanding of logic, for the last element of array i.e. `MimeConstants.P_CHARSET_DEFAULT`, Code has to be always executed in else block, that is the reason `i < CHARSET_NAMES.length - 1` is implemented.

zimsuchitgupta: We may write with single if block .
`if (!content.contains(GARBLED_CHARACTER) || i == CHARSET_NAMES.length - 1) {
}`

preshitaagrawal (approved):

store/src/java/com/zimbra/cs/lmtpserver/ZimbraLmtpBackend.java
gabhinavTI: @zimsuchitgupta @shrutig0510 modified as suggested

store/src/java/com/zimbra/cs/lmtpserver/ZimbraLmtpBackend.java
```diff
@@ -339,11 +347,21 @@ public void deliver(LmtpEnvelope env, InputStream in, int sizeHint) throws Unrec
...
-            inEEW = IOUtils.toInputStream(contentEEW);
+            for (int i = 0; i < CHARSET_NAMES.length; i++) {
+                final String charsetName = CHARSET_NAMES[i];
+                // copy input stream and evaluate content encoding
+                final byte[] ba = IOUtils.toByteArray(in);
+                IOUtils.closeQuietly(in);
+                in = new ByteArrayInputStream(ba);
+                final String content = new String(ba, charsetName);
+                // if current encoding is suitable, use it; otherwise, continue or use default
+                if (!content.contains(GARBLED_CHARACTER) || i == CHARSET_NAMES.length - 1) { }
```
zimsuchitgupta, line 358: I doesn't mean this. You should write this only.
`if (!content.contains(GARBLED_CHARACTER) || i == CHARSET_NAMES.length - 1) { 
       final String contentEEW = ExternalEmailWarning.getInstance().getUpdatedContent(content);
       inEEW = IOUtils.toInputStream(contentEEW, charsetName);
       break;
}`

zimsuchitgupta (changes requested):

store/src/java/com/zimbra/cs/lmtpserver/ZimbraLmtpBackend.java
shrutig0510: Yes....expecting this without else bock

RakeshAMore, line 358: why is this empty block?
do you mean if the variable content contains `GARBLED_CHARACTER ` and `i` value lesser than CHARSET names? then you want to execute the below code and break the loop?
```
final String contentEEW = ExternalEmailWarning.getInstance().getUpdatedContent(content);
inEEW = IOUtils.toInputStream(contentEEW, charsetName);
break;
```
this seems to be very confusing to me.

gabhinavTI: @zimsuchitgupta @shrutig0510 , Could you please check now

zimsuchitgupta (approved):

store/src/java/com/zimbra/cs/lmtpserver/ZimbraLmtpBackend.java
gabhinavTI: @RakeshAMore , Updated as per Suchit's comment

shrutig0510 (approved):

deepakzimb (approved):

PR #1286, opened 2022-06-13
ZCS-11535: Briefcase content accessible without 2FA

wilfredo-alfaro:
### Solution
This is a revision for previous work done for the ZBUG-2390, which is about certain resources (e.g. Briefcase) being accessible with username and password alone even when two factor authentication was enabled for the account. The previous work passed every test scenario on Patch 23 at the moment, but failed some of them when tested on Patch 25.

### Testing Done

### DIFF
--- a/store/src/java/com/zimbra/cs/service/UserServlet.java
+++ b/store/src/java/com/zimbra/cs/service/UserServlet.java
@@ -357,9 +357,12 @@ public class UserServlet extends ZimbraServlet {
         } else if (ctxt != null && ctxt.getAuthAccount() instanceof GuestAccount && ctxt.basicAuthAllowed()) {
             resp.addHeader(AuthUtil.WWW_AUTHENTICATE_HEADER, getRealmHeader(req, null));
             resp.sendError(HttpServletResponse.SC_UNAUTHORIZED, L10nUtil.getMessage(MsgKey.errMustAuthenticate, req));
-        } else {
+        } else if (!checkTwoFactorAuthentication(ctxt)) {
+            resp.addHeader(AuthUtil.WWW_AUTHENTICATE_HEADER, getRealmHeader(req, null));
+            resp.sendError(HttpServletResponse.SC_UNAUTHORIZED, L10nUtil.getMessage(MsgKey.errMustAuthenticate, req));
+         } else {
             resp.sendError(HttpServletResponse.SC_NOT_FOUND, message);
-        }
+         }
     }
 
     protected UserServletContext createContext(HttpServletRequest req, HttpServletResponse resp, UserServlet servlet)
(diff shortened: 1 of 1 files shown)


shrutig0510 (approved):

zimsuchitgupta (approved):

dasiyogesh (approved):

store/src/java/com/zimbra/cs/service/UserServlet.java
```diff
@@ -382,8 +382,7 @@ public void doGet(HttpServletRequest req, HttpServletResponse resp) throws Servl
             checkTargetAccountStatus(context);
 
             if (!checkTwoFactorAuthentication(context)) {
-                resp.addHeader(AuthUtil.WWW_AUTHENTICATE_HEADER, getRealmHeader(req, null));
-                resp.sendError(HttpServletResponse.SC_UNAUTHORIZED, L10nUtil.getMessage(MsgKey.errMustAuthenticate, req));
+                throw ServiceException.AUTH_REQUIRED(String.format("two factor auth is required"));
```
RakeshAMore, line 385: I think you can pass a plain string instead of `String.format()` since there is nothing to replace right? or am I missing something?

store/src/java/com/zimbra/cs/service/UserServlet.java
```diff
@@ -405,6 +404,8 @@ public void doGet(HttpServletRequest req, HttpServletResponse resp) throws Servl
         } catch (ServiceException se) {
             if (se.getCode() == ServiceException.PERM_DENIED || se instanceof NoSuchItemException)
                 sendError(context, req, resp, L10nUtil.getMessage(MsgKey.errNoSuchItem, req));
+            else if (se.getCode() == ServiceException.AUTH_REQUIRED)
```
RakeshAMore, line 407: Personally, I like to wrap if statements in `{}` it makes code more readable.

store/src/java/com/zimbra/cs/service/UserServlet.java
```diff
@@ -382,8 +382,7 @@ public void doGet(HttpServletRequest req, HttpServletResponse resp) throws Servl
             checkTargetAccountStatus(context);
 
             if (!checkTwoFactorAuthentication(context)) {
-                resp.addHeader(AuthUtil.WWW_AUTHENTICATE_HEADER, getRealmHeader(req, null));
```
RakeshAMore, line 385: I am not too familiar with the Zimbra codebase yet, but want to confirm if two-factor auth is mandatory? or is enabled based on flag/attribute?

RakeshAMore (changes requested):

store/src/java/com/zimbra/cs/service/UserServlet.java
wilfredo-alfaro: Agreed. Done.

store/src/java/com/zimbra/cs/service/UserServlet.java
wilfredo-alfaro: Agreed. I just didn't at first to follow the style of the existing block around it, but just changed it to `{}` wrapped statements as per your suggestion.

Done.

store/src/java/com/zimbra/cs/service/UserServlet.java
wilfredo-alfaro: Two-Factor Auth can be either mandatory or optional (or even unavailable as a feature) to a given a account depending on the settings for its Class of Service. The `checkTwoFactorAuthentication` method on line 514 finds whether the account should be using 2FA while trying to access a URL with Basic authentication alone, for the purpose of throwing the ServiceException from line 385 in such scenario, as required in the user story.

deepakzimb (approved):

PR #1291, opened 2022-06-16
ZBUG-2835 : Changing debug level to info from debug for syslog

abhishekdeshmukh2209:


silentsakky (approved):

deepakzimb (approved):

umagmrit (approved):

dasiyogesh (approved):

zimsuchitgupta (approved):

desouzas (approved):
Sign CLA

ghen2 (commented):
There's an error on the line below: %%uncomment instead of %%comment, it breaks the config if you're not using syslog:

system failure: org.apache.logging.log4j.core.config.ConfigurationException: No name attribute provided for Logger slogger

no logs are written at all due to this.

ghen2:
> this will happen only if we haven't set zimbraLogHostname. which by default is set to hostname.

`zimbraLogHostname` is not set (empty) by default here.  With this, it generates the following log4j.properties:

```
#logger.slogger.name = zimbra.slogger
#logger.slogger.level=INFO
logger.slogger.level=ERROR
#logger.slogger.additivity = false
#logger.slogger.appenderRef.SLOGGER.ref = sloggerAppender
```
Which is broken because the third line is not commented, but `slogger` is not defined, resulting in no logs at all:

system failure: org.apache.logging.log4j.core.config.ConfigurationException: No name attribute provided for Logger slogger

Either this block should be commented completely, or uncommented completely, not partially like the current code.

Another side effect is that, after fixing this, we now get these "slogger" statistics (mailboxd.csv, ldap,csv etc) in our standard mailbox.log.

abhishekdeshmukh2209:
I agree @ghen2 
the only thing that needs to done for now is that we will have to remove line 211 from log4j.properties.in and do a
zmmailboxdctl restart
we will fix this in next PR

ghen2:
Ah I missed that duplicate, indeed just remove it then.

ghen2:
Perhaps the goal of this construction to log only ERROR level when "slogger" is NOT going to syslog?

ghen2:
Yes indeed, the [old config](https://github.com/Zimbra/zm-mailbox/blob/71c02c95f45c421cbb365d457988c922c8c85dfe/store-conf/conf/log4j.properties.production) (before log4j 2.17 upgrade) looked like:

```
```

abhishekdeshmukh2209:
@ghen2  
https://github.com/Zimbra/zm-mailbox/pull/1293

PR #1293, opened 2022-06-17
ZBUG-2841 : Disable Syslog log submission by default in log4j V2

abhishekdeshmukh2209:
Issue :- Bydefault syslogs where getting enabled as there was no flag to check weather to enable or disable syslogs.

Fix:- Added pre existing flag zimbraLogToSyslog to check either to enable or disable, by default the flag is FALSE.

### DIFF
--- a/store-conf/conf/log4j.properties.production
+++ b/store-conf/conf/log4j.properties.production
@@ -126,14 +126,14 @@ appender.EWS.policies.time.modulate = true
 appender.EWS.strategy.type = DefaultRolloverStrategy
 
 # Syslog appender
-appender.SYSLOG.type = Syslog
-appender.SYSLOG.name = syslogAppender
-appender.SYSLOG.host = localhost
-appender.SYSLOG.port = 514
-appender.SYSLOG.protocol = UDP
-appender.SYSLOG.facility = LOCAL0
-appender.SYSLOG.layout.type = PatternLayout
-appender.SYSLOG.layout.pattern = mailboxd: %-5p [%t] [%z] %c{1} - %m
+%%uncomment VAR:zimbraLogToSyslog%%appender.SYSLOG.type = Syslog
+%%uncomment VAR:zimbraLogToSyslog%%appender.SYSLOG.name = syslogAppender
+%%uncomment VAR:zimbraLogToSyslog%%appender.SYSLOG.host = localhost
+%%uncomment VAR:zimbraLogToSyslog%%appender.SYSLOG.port = 514
+%%uncomment VAR:zimbraLogToSyslog%%appender.SYSLOG.protocol = UDP
+%%uncomment VAR:zimbraLogToSyslog%%appender.SYSLOG.facility = LOCAL0
+%%uncomment VAR:zimbraLogToSyslog%%appender.SYSLOG.layout.type = PatternLayout
+%%uncomment VAR:zimbraLogToSyslog%%appender.SYSLOG.layout.pattern = mailboxd: %-5p [%t] [%z] %c{1} - %m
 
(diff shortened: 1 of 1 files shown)


preshitaagrawal (approved):

dasiyogesh (approved):

zimsuchitgupta (approved):

silentsakky (approved):

deepakzimb (approved):

preshitaagrawal (approved):

zimsuchitgupta (approved):

store-conf/conf/log4j.properties.production
```diff
@@ -207,8 +207,7 @@ logger.memcached.name = net.spy.memcached
 logger.memcached.level = WARN
 
 %%comment VAR:!zimbraLogHostname%%logger.slogger.name = zimbra.slogger
-%%comment VAR:!zimbraLogHostname%%logger.slogger.level=INFO
-%%uncomment VAR:!zimbraLogHostname%%logger.slogger.level=ERROR
+%%comment VAR:!zimbraLogHostname%%logger.slogger.level = info
```
ghen2, line 210: I would still suggest the following config to restore previous (log4j v1) behaviour:
```
logger.slogger.name = zimbra.slogger
%%comment VAR:!zimbraLogHostname%%logger.slogger.level = info
%%uncomment VAR:!zimbraLogHostname%%logger.slogger.level = ERROR
logger.slogger.additivity = false
%%comment VAR:!zimbraLogHostname%%logger.slogger.appenderRef.SLOGGER.ref = sloggerAppender
```
Otherwise it keeps spamming mailbox.log with statistics logging when zimbraLogHostname is not set (the default).

abhishekdeshmukh2209: zimbraLogHostname is setup in most system as it copies hostname by default.

ghen2: Looking at zmsetup.pl, that's only the case when zimbra-logger is installed.

abhishekdeshmukh2209: I guess both are same

ghen2: This was the old config:
```
```

Hence slogger level was set to ERROR when zimbraLogHostname wasn't set (and these logs go to mailbox.log).

In the current v2 config, the slogger block gets all commented when zimbraLogHostname isn't set, so everything goes to mailbox.log at default loglevel, spamming the log with unwanted statistics.

(In the old v2 config, only the level=ERROR line got uncommented, but this broke the config since "slogger" definition was still commented.)

abhishekdeshmukh2209: @ghen2  check now

store-conf/conf/log4j.properties.production
```diff
@@ -207,8 +207,8 @@ logger.memcached.name = net.spy.memcached
...
+%%uncomment VAR:!zimbraLogHostname%%logger.slogger.level = error
```
ghen2, line 211: Can you please actually test this with `zimbraLogHostname` unset ?  Now it results in:
```
#logger.slogger.name = zimbra.slogger
#logger.slogger.level = info
logger.slogger.level = error
#logger.slogger.additivity = false
#logger.slogger.appenderRef.SLOGGER.ref = sloggerAppender
```
So again no logs (like before) because slogger is undefined:

`ERROR: service.FAILURE (system failure: org.apache.logging.log4j.core.config.ConfigurationException: No name attribute provided for Logger slogger)`

shrutig0510 (approved):

PR #1303, opened 2022-06-29
ZCS-11491 - Changing system-defaults for Backup and Restore

rcyarrapothu:
Feature:
Change system-defaults for the Zimbra Backup.

- Make Auto-Grouped backup as default.
- Auto-grouped backup divides the total number of mailboxes into groups by using this existing attribute zimbraBackupAutoGroupedNumGroups.
- The scheduler runs as-is but now takes a full-backup of 1/7 of all the accounts and incremental backups for the rest of the 6/7 accounts for each day. This process continues for the cycle of the default scheduled 7 days. This will allow us to backup 1/x the number of accounts each day. This prevents backups from running into business hours when backups are slow and can not be completed within the allotted time.

Implementation:
Changing the default backup type to Auto-Grouped Mode

Related PRs:
https://github.com/Zimbra/zm-backup-utilities/pull/6
https://github.com/Zimbra/zm-backup-store/pull/26


store/conf/attrs/zimbra-attrs.xml
```diff
@@ -2690,7 +2690,7 @@ TODO - add support for multi-line values in globalConfigValue and defaultCOSValu
 </attr>
 
 <attr id="512" name="zimbraBackupMode" type="enum" value="Standard,Auto-Grouped" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited">
-  <globalConfigValue>Standard</globalConfigValue>
+  <globalConfigValue>Auto-Grouped</globalConfigValue>
```
silentsakky, line 2693: for upgrade scenario, will this change applied automatically or do we need to write some migration script?

RakeshAMore: So by default, it will be `Auto-Grouped`, and what for the existing setups?

deepakzimb: whenever we upgrade to the old setup, local config, and LDAP values should not change. 

If we want to apply new value to old setup, then we need a script.

PR #1305, opened 2022-06-29
ZCS-11436: Enhance zmvolume CLI to support S3 volume type

vishnukantjangid:
**Feature**
Enhance zmvolume CLI to support S3 volume type

**Solution**
Modified zmvolume command for **add** and **edit** for external and internal volume storetype. 

**Added validations as per requirement like:**

-vp, -sp, -bid options should only be executed only when -st has the external option. In case it's internal, give an error.
Validation should be done if all the options are specified when -st = external.
Validation for each and every parameter should be done:
vp should follow the same validation for the string as -name
-stp should only allow the permitted s3 providers - AWS, Ceph, OpenIO, StorageGrid. Any option other than this should give an error.
bid should do the validation of the UID format. Error should be given if the bid is not specified in the expected format. For e.g. InvalidBucket, Bucket123_123.
If admin tries to edit an external volume store and changes any other parameter apart from name, error should be given.
The command and its option should be case sensitive.

**Additional Options for zmvolume**

-st, --storeType <arg> Store type: internal or external)
-vp, --volumePrefix Volume Preifx
-stp, --storageType Name of the store provider (S3, ObjectStore)
-bid, --bucketId S3 Bucket ID.


store/src/java/com/zimbra/cs/volume/VolumeCLI.java
```diff
@@ -252,12 +287,103 @@ private void addVolume() throws ParseException, SoapFaultException, IOException,
...
+                volumeInfo.setRootPath(path);
+            }
+            if (!Strings.isNullOrEmpty(compress)) {
+                volumeInfo.setCompressBlobs(Boolean.parseBoolean(compress));
+            }
+            if (!Strings.isNullOrEmpty(compressThreshold)) {
+                volumeInfo.setCompressionThreshold(Long.parseLong(compressThreshold));
+            }
+        } else if (volStoreType.equals(Volume.StoreType.EXTERNAL)) {
+            if (!Strings.isNullOrEmpty(type)) {
```
RakeshAMore, line 320: This is the mandatory parameter I think, regardless of store_type

store/src/java/com/zimbra/cs/volume/VolumeCLI.java
```diff
@@ -252,12 +287,103 @@ private void addVolume() throws ParseException, SoapFaultException, IOException,
...
+                throw new ParseException(A_TYPE + NOT_ALLOWED_EXTERNAL);
+            }
+            if (!Strings.isNullOrEmpty(path)) {
```
RakeshAMore, line 323: see this is how it works 
```
**Store Type -> 1/Internal**

1. type -> required -> Primary | Secodary | Index
2. name -> required ->
3. rootPath -> required
4. compressBlobs -> not required -> by default false
5. compressionThreadshold -> not reuired -> default value 4096
6. storeType -> not required -> default value 1

**Store Type -> 2/External** 

1. type -> required -> Primary | Secodary | Index
2. name -> required ->
3. rootPath -> required -> default (/S3-{globalBucketId})
4. compressBlobs -> not required -> by default false
5. compressionThreadshold -> not reuired -> default value 4096
6. storeType -> not required -> default value 1 (Internal/External)
7. volumePrefix -> not required -> 
8. storageType -> not required. -> default "S3" ( other value can be ObjectStore)
9. globalBucketConfigurationId -> required 
10. useInfrequentAccess -> not required -> default false
11. useIntelligentTiering -> not required -> default false
12. infrequentAccessThreshold -> not required -> default 65536

store/src/java/com/zimbra/cs/volume/VolumeCLI.java
vishnukantjangid: This is the edit command operation and this is mandatory only in add command. If we have store type external then we have to edit the name only.

zimsuchitgupta (approved):

store/src/java/com/zimbra/cs/volume/VolumeCLI.java
RakeshAMore: Yeah got it. thanks :)

RakeshAMore (approved):

PR #1316, opened 2022-07-08
ZCS-11491 - Changing default backup mode for Backup and Restore

rcyarrapothu:
Feature:
Change system-defaults for the Zimbra Backup.

Make Auto-Grouped backup as default.

- Auto-grouped backup divides the total number of mailboxes into groups by using this existing attribute zimbraBackupAutoGroupedNumGroups.
- The scheduler runs as-is but now takes a full-backup of 1/7 of all the accounts and incremental backups for the rest of the 6/7 accounts for each day. This process continues for the cycle of the default scheduled 7 days. This will allow us to backup 1/x the number of accounts each day. This prevents backups from running into business hours when backups are slow and can not be completed within the allotted time.

Implementation:
- Changing the default backup type to Auto-Grouped Mode

Related PRs:
https://github.com/Zimbra/zm-backup-utilities/pull/6
https://github.com/Zimbra/zm-backup-store/pull/26


rcyarrapothu:
After discussing with @silentsakky - we need a migration script to change the default value as we will have 2 types of users. Users migrating from NG backup needs their backupmode set to "Auto-Grouped" and users who are already using classic backup and have their default set to "Standard" should remain as-is.

silentsakky (approved):

ronstra-synacor (approved):

deepakzimb (approved):

zimsuchitgupta (approved):

PR #1318, opened 2022-07-13
ZBUG-2849 : Disabling stats logger by default

abhishekdeshmukh2209:
Issue :- Bydefault slogger level was set to info which caused unnecessary logs getting filled in /var/log folder.

Fix:- Set  default log level to off

### DIFF
--- a/store-conf/conf/log4j.properties.production
+++ b/store-conf/conf/log4j.properties.production
@@ -207,7 +207,7 @@ logger.memcached.name = net.spy.memcached
 logger.memcached.level = WARN
 
 logger.slogger.name = zimbra.slogger
-logger.slogger.level = off
+logger.slogger.level = error
 logger.slogger.additivity = false
 logger.slogger.appenderRef.SLOGGER.ref = sloggerAppender
 
--- a/store-conf/conf/log4j.properties.production
+++ b/store-conf/conf/log4j.properties.production
@@ -207,7 +207,7 @@ logger.memcached.name = net.spy.memcached
 logger.memcached.level = WARN
 
 %%comment VAR:!zimbraLogHostname%%logger.slogger.name = zimbra.slogger
-%%comment VAR:!zimbraLogHostname%%logger.slogger.level = info
+%%comment VAR:!zimbraLogHostname%%logger.slogger.level = off
 %%uncomment VAR:!zimbraLogHostname%%logger.slogger.level = error
 %%comment VAR:!zimbraLogHostname%%logger.slogger.additivity = false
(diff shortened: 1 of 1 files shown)


zimsuchitgupta (approved):

abhishekdeshmukh2209:
> Hi
> 
> This still breaks when zimbraLogHostname is unset, as this scenario uncomments only one line: `logger.slogger.level = error`
> 
> Leading to: `system failure: org.apache.logging.log4j.core.config.ConfigurationException: No name attribute provided for Logger slogger` because referenced "slogger" is not defined (those lines are still commented).
> 
> Looking back at the original v1 config, I think you should go to something like:
> 
> ```
> logger.slogger.name = zimbra.slogger
> logger.slogger.level = off
> logger.slogger.additivity = false
> %%comment VAR:!zimbraLogHostname%%logger.slogger.appenderRef.SLOGGER.ref = sloggerAppender
> ```
> 
> So "slogger" is always defined, and only the destination (appender) defends on whether zimbraLogHostname is set or not.

Check now @ghen2

ghen2:
This is better (no more fatal error), but if zimbraLogHostname is not set, statistics logging are still enabled and go to mailbox.log by default, because the line setting loglevel=off still gets commented.

Can you still uncomment the above lines like I suggested?  (equivalent to previous log4j v1 config)

ghen2:
FYI the old log4j v1 config ([before 8.8.15 patch 32](https://github.com/Zimbra/zm-mailbox/blob/71c02c95f45c421cbb365d457988c922c8c85dfe/store-conf/conf/log4j.properties.production)) was:
```
log4j.logger.zimbra=INFO
log4j.logger.zimbra.op=WARN
log4j.logger.com.zimbra=INFO
```
Setting loglevel to INFO and destination to syslog (SLOGGER) when zimbraLogHostname was set, and loglevel to ERROR (ie. silent) when zimbraLogHostname was not set.

abhishekdeshmukh2209:
> Just test the latest change with zimbraLogHostname unset. It still generates statistics, and because no specific appender or loglevel is set (all commented), it goes to mailbox.log by default.

yes it does flood mailbox.log
I think we are going to remove dependency on zimbraLogHostname as stats logger isn't required by everybody and if someone wants they can change log level to info
At info it still floods /var/log location that isn't required by all

ghen2:
Thanks.  (btw "error" level should be sufficient to mute the statistics.)

ghen2 (approved):

yyoshiho (approved):
This change will work for this bug.

PR #1323, opened 2022-07-16
Feature/zcs 11436-new changes in validation of root path in case of external store type

vishnukantjangid:
Feature/ZCS 11436-new changes in the validation of root path in case of external store type

**Original PR:** https://github.com/Zimbra/zm-mailbox/pull/1305

### DIFF
--- a/store/src/java/com/zimbra/cs/volume/VolumeCLI.java
+++ b/store/src/java/com/zimbra/cs/volume/VolumeCLI.java
@@ -62,7 +62,7 @@ public final class VolumeCLI extends SoapCLI {
     private static final String O_P = "p";
     private static final String O_C = "c";
     private static final String O_CT = "ct";
-
+    
     /** attributes for external storetype **/
     private static final String O_ST = "st";
     private static final String O_VP = "vp";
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/volume/VolumeCLI.java
```diff
@@ -276,14 +276,10 @@ private void addVolume() throws ParseException, SoapFaultException, IOException,
         if (Strings.isNullOrEmpty(name)) {
             throw new ParseException(A_NAME + MISSING_ATTRS);
         }
-        if (Strings.isNullOrEmpty(path)) {
```
RakeshAMore, line 279: This path is optional only in the case of EXTERNAL volumes, it should be mandatory param for INTERNAL right?

RakeshAMore (approved):

store/src/java/com/zimbra/cs/volume/VolumeCLI.java
vishnukantjangid: Yes, this is mandatory in INTERNAL block validation only. In the case of EXTERNAL, Path will be optional and we are not using it at all. The default path will work in an EXTERNAL case.

zimsuchitgupta (approved):

deepakzimb (approved):

PR #1324, opened 2022-07-16
ZCS-11428  Modify Volumes SOAP API to support S3 volume type

inpati:
**Problem Statement**
Modify SOAP Below SOAP API -> 1. CreateVolumeRequest 2. ModifyVolumeRequest 3. DeleteVolumeRequest 4. GetAllVolumeRequest 5. GetVolumeRequest

**Solution**
Modified these SOAP API's and added supporting files and code.

**Use-Cases tested**
1. Create Volume Request can be used to create FILE_STORE volume or EXTERNAL volume depending on store_type. if store_type is 2, external volumes are created if store_type is 1, internal volumes are created. Tested compulsory and optional arguments while creating Volumes
2. Modify Volume Request can be used to modify already created volumes. The parameters available for modification depends on the store type. If store_type is 2, only volume name can be modified and if store_type is 1 parameters CompressBlobs, CompressionThreshold, RootPath, Name and Type can be modified
3. Get Volume Request returns volume information identified by unique parameter volume id
4. Get All Volume Requests returns all volume information present at the moment
5. Delete Volume Request deleted volume based on unique parameter volume id

### DIFF
--- a/store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
+++ b/store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
@@ -17,6 +17,7 @@
 
 package com.zimbra.cs.service.util;
 
+import com.zimbra.client.ZMailbox;
 import org.json.JSONException;
 import org.json.JSONObject;
 
@@ -40,6 +41,8 @@ import com.zimbra.soap.admin.type.VolumeInfo;
 import com.zimbra.util.ExternalVolumeInfoHandler;
 
 public class VolumeConfigUtil {
+
+    private static final String  ROOT_PATH_ELE_SEPARATOR = "-";
 
     /**
      * Validate the create volume request parameters
@@ -307,11 +310,18 @@ public class VolumeConfigUtil {
                     storageType = properties.getString(AdminConstants.A_VOLUME_STORAGE_TYPE);
                     globalBucketConfigId = properties.getString(AdminConstants.A_VOLUME_GLB_BUCKET_CONFIG_ID);
                 } catch (JSONException e) {
-                    String errMsg = "Error while reading json data for external volume: " + req.getId();
-                    throw ServiceException.FAILURE(errMsg, e);
+                    throw ServiceException.FAILURE("Error while reading json data for external volume: " + req.getId(), e);
+                }
+                // storageType should not be null
+                if (StringUtil.isNullOrEmpty(storageType)) {
+                    throw VolumeServiceException.INVALID_REQUEST("StorageType Empty for external volume " + req.getId(), VolumeServiceException.INVALID_REQUEST);
                 }
                 builder.setName(volInfo.getName());
-                String extRootPath = "/" + storageType + "-" + volInfo.getName() + "-" + globalBucketConfigId;
+                String extRootPath = ZMailbox.PATH_SEPARATOR + storageType + ROOT_PATH_ELE_SEPARATOR + volInfo.getName();
+                // append global bucket id as well in case it is available
+                if (!StringUtil.isNullOrEmpty(globalBucketConfigId)) {
+                    extRootPath = extRootPath + ROOT_PATH_ELE_SEPARATOR + globalBucketConfigId;
+                }
                 builder.setPath(extRootPath, false);
             }
         }
--- a/store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
+++ b/store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
@@ -18,14 +18,24 @@
 package com.zimbra.cs.service.util;
 
 import org.json.JSONException;
+import org.json.JSONObject;
 
 import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.soap.AdminConstants;
 import com.zimbra.common.util.StringUtil;
-import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.cs.account.Provisioning;
+import com.zimbra.cs.store.StoreManager;
 import com.zimbra.cs.volume.Volume;
+import com.zimbra.cs.volume.VolumeManager;
 import com.zimbra.cs.volume.VolumeServiceException;
 import com.zimbra.soap.admin.message.CreateVolumeRequest;
+import com.zimbra.soap.admin.message.DeleteVolumeRequest;
+import com.zimbra.soap.admin.message.ModifyVolumeRequest;
+import com.zimbra.soap.admin.message.GetAllVolumesRequest;
+import com.zimbra.soap.admin.message.GetAllVolumesResponse;
+import com.zimbra.soap.admin.message.GetVolumeRequest;
+import com.zimbra.soap.admin.message.GetVolumeResponse;
+import com.zimbra.soap.admin.type.VolumeExternalInfo;
 import com.zimbra.soap.admin.type.VolumeInfo;
 import com.zimbra.util.ExternalVolumeInfoHandler;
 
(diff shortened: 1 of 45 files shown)


store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
```diff
@@ -300,8 +300,50 @@ public static void parseModifyVolumeRequest(ModifyVolumeRequest req) throws Serv
...
             if (volInfo.getName() != null) {
                 builder.setName(volInfo.getName());
+                String extRootPath = "/" + readStorageType(req.getId()) + "-" + volInfo.getName() + "-" + readGlobalBucketConfigId(req.getId());
+                builder.setPath(extRootPath, false);
             }
         }
         mgr.update(builder.build());
     }
+
+    /**
```
RakeshAMore, line 310: I think you don't need to create the below methods, you can change this code as shown below(instead of creating an object of `ExternalVolumeInfoHandler` multiple times )
```
ExternalVolumeInfoHandler extVolInfoHandler = new ExternalVolumeInfoHandler(Provisioning.getInstance());
JSONObject jsonProperties = extVolInfoHandler.readServerProperties(id);
String  storageType = properties.getString(AdminConstants.A_VOLUME_STORAGE_TYPE);
String globalBucketConfigId = properties.getString(AdminConstants.A_VOLUME_GLB_BUCKET_CONFIG_ID);
```

RakeshAMore (changes requested):

store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
inpati: Fixed

RakeshAMore (approved):

zimsuchitgupta (approved):

vishnukantjangid (approved):

store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
```diff
@@ -299,7 +299,20 @@ public static void parseModifyVolumeRequest(ModifyVolumeRequest req) throws Serv
...
+                try {
+                    JSONObject properties = extVolInfoHandler.readServerProperties(req.getId());
+                    storageType = properties.getString(AdminConstants.A_VOLUME_STORAGE_TYPE);
+                    globalBucketConfigId = properties.getString(AdminConstants.A_VOLUME_GLB_BUCKET_CONFIG_ID);
+                } catch (JSONException e) {
+                    String errMsg = "Error while reading json data for external volume: " + req.getId();
+                    throw ServiceException.FAILURE(errMsg, e);
+                }
                 builder.setName(volInfo.getName());
+                String extRootPath = "/" + storageType + "-" + volInfo.getName() + "-" + globalBucketConfigId;
```
dasiyogesh, line 314: Would be better to use constants for the delimiters( '/' & '-' ) used in constructing root path

store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
```diff
@@ -299,7 +299,20 @@ public static void parseModifyVolumeRequest(ModifyVolumeRequest req) throws Serv
...
             if (volInfo.getName() != null) {
+                String storageType = "";
+                String globalBucketConfigId = "";
+                ExternalVolumeInfoHandler extVolInfoHandler = new ExternalVolumeInfoHandler(Provisioning.getInstance());
+                try {
+                    JSONObject properties = extVolInfoHandler.readServerProperties(req.getId());
+                    storageType = properties.getString(AdminConstants.A_VOLUME_STORAGE_TYPE);
+                    globalBucketConfigId = properties.getString(AdminConstants.A_VOLUME_GLB_BUCKET_CONFIG_ID);
+                } catch (JSONException e) {
+                    String errMsg = "Error while reading json data for external volume: " + req.getId();
```
dasiyogesh, line 310: there is no need of declaring errMsg var as its not used anywhere further, just use the string in the exception call.

store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
dasiyogesh, line 314: what would happen if storageType Or globalBucketConfigId are empty/null?

RakeshAMore: "storageType" should not be null/empty, but globalBucketConfigId can be(in the case of OPEN_IO, will be adding this as separate ticket). But empty/null should be added here.

RakeshAMore: Done

store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
RakeshAMore: Done

store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
RakeshAMore: Done

store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
```diff
@@ -299,7 +302,27 @@ public static void parseModifyVolumeRequest(ModifyVolumeRequest req) throws Serv
...
+                try {
+                    JSONObject properties = extVolInfoHandler.readServerProperties(req.getId());
+                    storageType = properties.getString(AdminConstants.A_VOLUME_STORAGE_TYPE);
+                    globalBucketConfigId = properties.getString(AdminConstants.A_VOLUME_GLB_BUCKET_CONFIG_ID);
+                } catch (JSONException e) {
+                    throw ServiceException.FAILURE("Error while reading json data for external volume: " + req.getId(), e);
+                }
+                // storageType should not be null
+                if (StringUtil.isNullOrEmpty(storageType)) {
+                    throw VolumeServiceException.INVALID_REQUEST("StorageType Empty for external volume " +req.getId(), VolumeServiceException.INVALID_REQUEST);
```
dasiyogesh, line 317: add space after + 
change `StorageType Empty for external volume` to `Storage type empty for external volume`

store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
```diff
@@ -93,7 +96,7 @@ public static void validateCreateVolumeRequest(CreateVolumeRequest request, Volu
 
         // validate current
         if (false != volInfoRequest.isCurrent() &&
```
dasiyogesh, line 98: why are we checking `volInfoRequest.isCurrent()` twice in this condition?

RakeshAMore: @inpati this condition will never be satisfied. can you please check this again?

store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
inpati: Actually there was no need of adding null check because globalBucketConfigId was initialized to empty string when object gets created

store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
inpati: Done, fixed

store/src/java/com/zimbra/cs/service/util/VolumeConfigUtil.java
inpati: Done. fixed. Basically we are checking whether if current is not true and not false, this is corner case as anybody can pass string in this current variable.

dasiyogesh (approved):

PR #1326, opened 2022-07-18
ZCS-11694 : File Shared with me implementation

gabhinavTI:
Problem: As of current when the file is shared i.e. User A shares any file with User B then the file is shared as an accessible link on the mail. So assume when I have to access the file again after a month I receive I'll have to scroll down to check the mail where the link was shared. By any chance, if that mail is deleted from the grantee's side, they won't be able to access it again and have to ask for resharing of the file

Solution: To overcome the above issue we came up with a solution i.e. to create a system folder `File Shared with` Me under the briefcase tab. Now when user A shares the file with User B, the file is accessible via the link, and also file is available under the File Shared with Me folder. This file can be previewed upon clicking, and based on the rights provided file can be viewed and edited accordingly. There is an option to download the file as well.

Implementation Approach: When user A shares a file with user B, then in SendShareNotification the grantee ID is used to fetch the mailbox details of user B. Based on the mailbox details retrieved DB actions are made. Now the mailbox can either be on the local server or on another server , hence added a check for that, if it's on the Local server it processes it away directly else we have created a new API `FileSharedWithMeRequest` which gets' triggered, and then the mailbox is retrieved from context and the DB action happens in new API
There are basically 3 scenarios that can happen:
When the File is shared -> The file details are captured, i.e. fileName, content type, permissions to the file, size, etc. Based on the grantee ID grantee's mailbox is retrieved and using mailbox detail and input available a new DB entry is made directly on the grantee's side on the File Shared with Me .This file can be accessed based on an ID that has a value like "AccountID: FieldId"

When Acess is revoked --> In this case the file is deleted from the grantee side using the delete query and file will no longer be accessible on the File Shared with Me folder

When File details/rights are changed --> In this case firstly older DB entry is deleted and then newer entry is created on UserB mailbox

Testing:
Manual testing through SOAP API to check what files are available when input is "File Shared With me Folder"
Testing through UI on a single Node server, when the file which has been shared can be accessed via a link. ALso same file is available under the File Shared With me Folder. These files can be previewed, and there is an option to open the file in a new tab and edit it according to the rights provided. The users could also download single file at a time . The scenario us verified on Multi-node server as well

### DIFF
--- a/common/src/java/com/zimbra/common/soap/MailConstants.java
+++ b/common/src/java/com/zimbra/common/soap/MailConstants.java
@@ -302,6 +302,8 @@ public final class MailConstants {
     public static final String E_GET_IMAP_RECENT_CUTOFF_RESPONSE = "GetIMAPRecentCutoffResponse";
     public static final String E_IMAP_COPY_REQUEST = "IMAPCopyRequest";
     public static final String E_IMAP_COPY_RESPONSE = "IMAPCopyResponse";
+    public static final String E_FILE_SHARED_WITH_ME_REQUEST = "FileSharedWithMeRequest";
+    public static final String E_FILE_SHARED_WITH_ME_RESPONSE = "FileSharedWithMeResponse";
 
     // noop
     public static final QName NO_OP_REQUEST = QName.get(E_NO_OP_REQUEST, NAMESPACE);
@@ -604,6 +606,10 @@ public final class MailConstants {
     public static final QName RECORD_IMAP_SESSION_RESPONSE = QName.get(E_RECORD_IMAP_SESSION_RESPONSE, NAMESPACE);
     public static final QName IMAP_COPY_REQUEST = QName.get(E_IMAP_COPY_REQUEST, NAMESPACE);
     public static final QName IMAP_COPY_RESPONSE = QName.get(E_IMAP_COPY_RESPONSE, NAMESPACE);
+
+    //File Shared With Me
+    public static final QName FILE_SHARED_WITH_ME_REQUEST = QName.get(E_FILE_SHARED_WITH_ME_REQUEST, NAMESPACE);
+    public static final QName FILE_SHARED_WITH_ME_RESPONSE = QName.get(E_FILE_SHARED_WITH_ME_RESPONSE, NAMESPACE);
 
     public static final String E_MAILBOX = "mbx";
     public static final String E_ITEM = "item";
--- a/soap/src/java/com/zimbra/soap/mail/message/FileSharedWithMeRequest.java
+++ b/soap/src/java/com/zimbra/soap/mail/message/FileSharedWithMeRequest.java
@@ -1,0 +1,174 @@
+/* 
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2022 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ * ***** END LICENSE BLOCK *****
+ */
+
+package com.zimbra.soap.mail.message;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+
+import com.zimbra.common.soap.MailConstants;
+
+/**
+ * @zm-api-command-auth-required true
+ * @zm-api-command-admin-auth-required false
+ * @zm-api-command-description File Share With Me
+ * <br />
+ * This is an internal API, cannot be invoked directly
+ */
+@XmlAccessorType(XmlAccessType.NONE)
+@XmlRootElement(name = MailConstants.E_FILE_SHARED_WITH_ME_REQUEST)
+public class FileSharedWithMeRequest {
+
+    /**
+     * @zm-api-field-tag action
+     * @zm-api-field-description Action - Create, Edit, Revoke
+     */
+    @XmlElement(name = MailConstants.E_ACTION, required = true)
+    private String action;
+
+    /**
+     * @zm-api-field-tag fileName
+     * @zm-api-field-description Name of the file which is to be shared
+     */
+    @XmlElement(name = MailConstants.A_CONTENT_FILENAME, required = true)
+    private String fileName;
+
+    /**
+     * @zm-api-field-tag ownerFileId
+     * @zm-api-field-description Owner File ID 
+     */
+    @XmlElement(name = MailConstants.A_ITEMID, required = true)
+    private int ownerFileId;
+
+    /**
+     * @zm-api-field-tag fileUUID
+     * @zm-api-field-description Owner File UUID
+     */
+    @XmlElement(name = MailConstants.A_REMOTE_UUID, required = true)
+    private String fileUUID;
+
+    /**
+     * @zm-api-field-tag fileOwnerName
+     * @zm-api-field-description File Owner Name
+     */
+    @XmlElement(name = MailConstants.A_OWNER_NAME, required = true)
+    private String fileOwnerName;
+
+    /**
+     * @zm-api-field-tag rights
+     * @zm-api-field-description Permission provided to the file
+     */
+    @XmlElement(name = MailConstants.A_RIGHTS, required = true)
+    private String rights;
+
+    /**
+     * @zm-api-field-tag contentType
+     * @zm-api-field-description Content type of the file
+     */
+    @XmlElement(name = MailConstants.A_CONTENT_TYPE, required = true)
+    private String contentType;
+
+    /**
+     * @zm-api-field-tag size
+     * @zm-api-field-description Actual file size
+     */
+    @XmlElement(name = MailConstants.A_SIZE, required = true)
+    private long size;
+
+    public FileSharedWithMeRequest() {
+    }
+
+    public FileSharedWithMeRequest(String action, String granteeId, String fileName, String ownerAccountId,
+            int ownerFileId, String fileUUID, String fileOwnerName, String rights, String contentType, long size) {
+        super();
+        this.action = action;
+        this.fileName = fileName;
+        this.ownerFileId = ownerFileId;
+        this.fileUUID = fileUUID;
+        this.fileOwnerName = fileOwnerName;
+        this.rights = rights;
+        this.contentType = contentType;
+        this.size = size;
+    }
+
+    public String getAction() {
+        return action;
+    }
+
+    public void setAction(String action) {
+        this.action = action;
+    }
+
+    public String getFileName() {
+        return fileName;
+    }
+
+    public void setFileName(String fileName) {
+        this.fileName = fileName;
+    }
+
+    public int getOwnerFileId() {
+        return ownerFileId;
+    }
+
+    public void setOwnerFileId(int ownerFileId) {
+        this.ownerFileId = ownerFileId;
+    }
+
+    public String getFileUUID() {
+        return fileUUID;
+    }
+
+    public void setFileUUID(String fileUUID) {
+        this.fileUUID = fileUUID;
+    }
+
+    public String getFileOwnerName() {
+        return fileOwnerName;
+    }
+
+    public void setFileOwnerName(String fileOwnerName) {
+        this.fileOwnerName = fileOwnerName;
+    }
+
+    public String getRights() {
+        return rights;
+    }
+
+    public void setRights(String rights) {
+        this.rights = rights;
+    }
+
+    public String getContentType() {
+        return contentType;
+    }
+
+    public void setContentType(String contentType) {
+        this.contentType = contentType;
+    }
+
+    public long getSize() {
+        return size;
+    }
+
+    public void setSize(long size) {
+        this.size = size;
+    }
+
+}
--- a/soap/src/java/com/zimbra/soap/mail/message/FileSharedWithMeResponse.java
+++ b/soap/src/java/com/zimbra/soap/mail/message/FileSharedWithMeResponse.java
@@ -1,0 +1,47 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2022 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ * ***** END LICENSE BLOCK *****
+ */
+
+package com.zimbra.soap.mail.message;
+
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+
+import com.google.common.base.MoreObjects;
+import com.zimbra.common.soap.MailConstants;
+
+@XmlRootElement(name = MailConstants.E_FILE_SHARED_WITH_ME_RESPONSE)
+public class FileSharedWithMeResponse {
+
+    @XmlElement(name = MailConstants.A_STATUS, required = true)
+    private String status;
+
+    public FileSharedWithMeResponse() {
+    }
+
+    public FileSharedWithMeResponse(String status) {
+        this.status = status;
+    }
+
+    public MoreObjects.ToStringHelper addToStringInfo(MoreObjects.ToStringHelper helper) {
+        return helper.add(MailConstants.A_STATUS, status);
+    }
+
+    @Override
+    public String toString() {
+        return addToStringInfo(MoreObjects.toStringHelper(this)).toString();
+    }
+}
--- a/store/src/java/com/zimbra/cs/mailbox/Mailbox.java
+++ b/store/src/java/com/zimbra/cs/mailbox/Mailbox.java
@@ -142,10 +142,13 @@ import com.zimbra.cs.ldap.LdapConstants;
 import com.zimbra.cs.mailbox.CalendarItem.AlarmData;
 import com.zimbra.cs.mailbox.CalendarItem.Callback;
 import com.zimbra.cs.mailbox.CalendarItem.ReplyInfo;
+import com.zimbra.cs.mailbox.Flag.FlagInfo;
 import com.zimbra.cs.mailbox.FoldersTagsCache.FoldersTags;
 import com.zimbra.cs.mailbox.MailItem.CustomMetadata;
+import com.zimbra.cs.mailbox.MailItem.CustomMetadata.CustomMetadataList;
 import com.zimbra.cs.mailbox.MailItem.PendingDelete;
 import com.zimbra.cs.mailbox.MailItem.TargetConstraint;
+import com.zimbra.cs.mailbox.MailItem.Type;
 import com.zimbra.cs.mailbox.MailItem.UnderlyingData;
 import com.zimbra.cs.mailbox.MailServiceException.NoSuchItemException;
 import com.zimbra.cs.mailbox.MailboxListener.ChangeNotification;
@@ -180,6 +183,7 @@ import com.zimbra.cs.redolog.op.CreateCalendarItemRecorder;
 import com.zimbra.cs.redolog.op.CreateChat;
 import com.zimbra.cs.redolog.op.CreateComment;
 import com.zimbra.cs.redolog.op.CreateContact;
+import com.zimbra.cs.redolog.op.CreateFileSharedWithMe;
 import com.zimbra.cs.redolog.op.CreateFolder;
 import com.zimbra.cs.redolog.op.CreateFolderPath;
 import com.zimbra.cs.redolog.op.CreateInvite;
@@ -308,8 +312,9 @@ public class Mailbox implements MailboxStore {
     public static final int ID_FOLDER_PROFILE = FolderConstants.ID_FOLDER_PROFILE; // 18;
 
     public static final int ID_FOLDER_UNSUBSCRIBE = FolderConstants.ID_FOLDER_UNSUBSCRIBE; // 19;
+    public static final int ID_FOLDER_FILE_SHARED_WITH_ME = FolderConstants.ID_FOLDER_FILE_SHARED_WITH_ME; // 20;
     //This id should be incremented if any new ID_FOLDER_* is added.
-    public static final int HIGHEST_SYSTEM_ID = FolderConstants.HIGHEST_SYSTEM_ID; // 19;
+    public static final int HIGHEST_SYSTEM_ID = FolderConstants.HIGHEST_SYSTEM_ID; // 20;
 
     public static final int EAS_BLOCKED_EMAIL_ID = 255;
     public static final int FIRST_USER_ID = 256;
@@ -2266,6 +2271,8 @@ public class Mailbox implements MailboxStore {
                             MailItem.Type.MESSAGE, 0, MailItem.DEFAULT_COLOR_RGB, null, null, null);
             Folder.create(ID_FOLDER_BRIEFCASE, UUIDUtil.generateUUID(), this, userRoot, "Briefcase", system,
                             MailItem.Type.DOCUMENT, 0, MailItem.DEFAULT_COLOR_RGB, null, null, null);
+            Folder.create(ID_FOLDER_FILE_SHARED_WITH_ME, UUIDUtil.generateUUID(), this, userRoot, "Files shared with me", system,
+                    MailItem.Type.DOCUMENT, 0, MailItem.DEFAULT_COLOR_RGB, null, null, null);
 
             if (this.getAccount().isFeatureSafeUnsubscribeFolderEnabled()) {
                 Folder.create(ID_FOLDER_UNSUBSCRIBE, UUIDUtil.generateUUID(), this, userRoot, "Unsubscribe", system,
@@ -9062,6 +9069,97 @@ public class Mailbox implements MailboxStore {
         }
 
         return getMountpointById(octxt, mountpointId);
+    }
+
+    public void createFileSharedWithMe(Mailbox mbox, OperationContext octxt, int folderId, String fileName,
+            String ownerAccountId, int ownerFileId, String remoteUuid, String fileOwnerName,
+            String contentType, long fileSize, String rights) throws ServiceException {
+
+        CreateFileSharedWithMe redoRecorder = new CreateFileSharedWithMe(mId, folderId, fileName, ownerAccountId,
+                ownerFileId, remoteUuid, fileOwnerName, contentType, rights);
+
+        boolean success = false;
+        try {
+            beginTransaction("createFileSharedWithMe", octxt, redoRecorder);
+            CreateFileSharedWithMe redoPlayer = (CreateFileSharedWithMe) currentChange().getRedoPlayer();
+
+            int mptId = getNextItemId(redoPlayer == null ? ID_AUTO_INCREMENT : redoPlayer.getId());
+            String uuid = redoPlayer == null ? UUIDUtil.generateUUID() : redoPlayer.getUuid();
+            UnderlyingData data = new UnderlyingData();
+            data.uuid = remoteUuid + ":" + ownerFileId;
+            data.id = mptId;
+            data.type = Type.DOCUMENT.toByte();
+            data.folderId = folderId;
+            data.date = mbox.getOperationTimestamp();
+            data.name = fileName;
+            data.setSubject(fileName);
+            data.size = fileSize;
+            data.metadata = encodeFileSharedMetadata(new Metadata(), fileOwnerName, ownerAccountId,
+                    ownerFileId, contentType, rights).toString();
+            data.contentChanged(mbox);
+            new DbMailItem(mbox).create(data);
+            Document doc = new Document(mbox, data);
+            doc.finishCreation(null);
+            success = true;
+        } finally {
+            endTransaction(success);
+        }
+    }
+
+    public void deleteFileSharedWithMe(Mailbox mbox, OperationContext octxt, String remoteUuid, int ownerFileId,
+            String ownerAccountId, String granteeAccountID) throws ServiceException {
+        CreateFileSharedWithMe redoRecorder = new CreateFileSharedWithMe(mId, remoteUuid);
+        boolean success = false;
+        try {
+            beginTransaction("deleteFileSharedWithMe", octxt, redoRecorder);
+            Document item = (Document) mbox.getItemByUuid(remoteUuid + ":" + ownerFileId, Type.DOCUMENT);
+            if (item.getOwnerAccountId().equals(ownerAccountId) && item.getAccountId().equals(granteeAccountID)) {
+                DbMailItem.delete(mbox, Collections.singletonList(item.getId()), false);
+                // update the folder item count and size accordingly
+                mbox.getFolderById(ID_FOLDER_FILE_SHARED_WITH_ME).updateSize(-1, -1, -item.getSize());
+                success = true;
+            } else {
+                throw ServiceException.PERM_DENIED("Revoke operation failed, Invalid request source");
+            }
+        } finally {
+            endTransaction(success);
+        }
+    }
+
+    public void updateFileSharedWithMe(Mailbox mbox, OperationContext octxt, String ownerAccountID, int ownerFileId,
+            String remoteUuid, String fileOwnerName, String contentType, String rights, String granteeAccountID)
+            throws ServiceException {
+        CreateFileSharedWithMe redoRecorder = new CreateFileSharedWithMe(mId, remoteUuid);
+        boolean success = false;
+        try {
+            beginTransaction("updateFileSharedWithMe", octxt, redoRecorder);
+            Document item = (Document) mbox.getItemByUuid(remoteUuid + ":" + ownerFileId, Type.DOCUMENT);
+            if (item.getOwnerAccountId().equals(ownerAccountID) && item.getAccountId().equals(granteeAccountID)) {
+                Metadata meta = new Metadata();
+                meta.put(Metadata.FN_CREATOR, fileOwnerName);
+                meta.put(Metadata.FN_MIME_TYPE, contentType);
+                meta.put(Metadata.FN_FILE_PERMISSION, rights);
+                meta.put(Metadata.FN_OWNER_ACCOUNT_ID, ownerAccountID);
+                meta.put(Metadata.FN_OWNER_FILE_ID, ownerFileId);
+                mbox.getItemCache().remove(item.getId());
+                new DbMailItem(mbox).update(item, meta);
+                success = true;
+            } else {
+                throw ServiceException.PERM_DENIED("Update operation failed, Invalid request source");
+            }
+        } finally {
+            endTransaction(success);
+        }
+    }
+
+    Metadata encodeFileSharedMetadata(Metadata meta, String creator, String ownerAccountId,
+            int ownerFileId, String contentType, String rights) {
+        meta.put(Metadata.FN_CREATOR, creator);
+        meta.put(Metadata.FN_MIME_TYPE, contentType);
+        meta.put(Metadata.FN_FILE_PERMISSION, rights);
+        meta.put(Metadata.FN_OWNER_ACCOUNT_ID, ownerAccountId);
+        meta.put(Metadata.FN_OWNER_FILE_ID, ownerFileId);
+        return meta;
     }
 
     public void enableSharedReminder(OperationContext octxt, int mountpointId, boolean enable) throws ServiceException {
--- a/store/src/java/com/zimbra/cs/redolog/op/CreateFileSharedWithMe.java
+++ b/store/src/java/com/zimbra/cs/redolog/op/CreateFileSharedWithMe.java
@@ -1,0 +1,120 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2022 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ * ***** END LICENSE BLOCK *****
+ */
+
+package com.zimbra.cs.redolog.op;
+
+import java.io.IOException;
+
+import com.zimbra.cs.mailbox.MailboxOperation;
+import com.zimbra.cs.redolog.RedoLogInput;
+import com.zimbra.cs.redolog.RedoLogOutput;
+
+public class CreateFileSharedWithMe extends RedoableOp {
+
+    private int mId;
+    private String mUuid;
+    private int mFolderId;
+    private String mName;
+    private String mOwnerId;
+    private int mRemoteId;
+    private String mRemoteUuid;
+    private String mOwnerName;
+    private String mContentType;
+    private String mRights;
+
+    public CreateFileSharedWithMe() {
+        super(MailboxOperation.CreateMountpoint);
+        mId = UNKNOWN_ID;
+    }
+
+    public CreateFileSharedWithMe(int mailboxId, int folderId, String name, String ownerId, int remoteId,
+            String remoteUuid, String fileOwnerName, String contetnType, String rights) {
+        this();
+        setMailboxId(mailboxId);
+        mId = UNKNOWN_ID;
+        mFolderId = folderId;
+        mName = name != null ? name : "";
+        mOwnerId = ownerId;
+        mRemoteId = remoteId;
+        mRemoteUuid = remoteUuid;
+        mOwnerName = fileOwnerName;
+        mContentType = contetnType;
+        mRights = rights;
+    }
+
+    public CreateFileSharedWithMe(int mailboxId, String remoteUuid) {
+        this();
+        setMailboxId(mailboxId);
+        mRemoteUuid = remoteUuid;
+    }
+
+    public int getId() {
+        return mId;
+    }
+
+    public String getUuid() {
+        return mUuid;
+    }
+
+    public void setIdAndUuid(int id, String uuid) {
+        mId = id;
+        mUuid = uuid;
+    }
+
+    @Override
+    protected String getPrintableData() {
+        StringBuilder sb = new StringBuilder("id=").append(mId);
+        sb.append(", uuid=").append(mUuid);
+        sb.append(", name=").append(mName).append(", folder=").append(mFolderId);
+        sb.append(", owner=").append(mOwnerId).append(", remoteId=").append(mRemoteId).append(", remoteUuid=")
+                .append(mRemoteUuid);
+        return sb.toString();
+    }
+
+    @Override
+    protected void serializeData(RedoLogOutput out) throws IOException {
+        out.writeInt(mId);
+        out.writeUTF(mUuid);
+        out.writeUTF(mName);
+        out.writeUTF(mOwnerId);
+        out.writeInt(mRemoteId);
+        out.writeUTF(mRemoteUuid);
+        out.writeInt(mFolderId);
+        out.writeUTF(mOwnerName);
+        out.writeUTF(mContentType);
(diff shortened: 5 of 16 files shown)


store/src/java/com/zimbra/cs/db/DbMailItem.java
```diff
@@ -1697,6 +1697,30 @@ public static void delete(Mailbox mbox, Collection<Integer> ids, boolean fromDum
...
+     * @throws ServiceException 
+     * The below delete method is used to delete files that  were shared with me
+     */
+    public void deleteSharedFile(Mailbox mbox, String uuid, boolean fromDumpster) throws ServiceException {
+        DbConnection conn = mbox.getOperationConnection();
+        PreparedStatement stmt = null;
+        try {
+            stmt = conn
+                    .prepareStatement("DELETE FROM " + getMailItemTableName(mbox, fromDumpster) + " WHERE " + "uuid=?");
+            int pos = 1;
```
dasiyogesh, line 1714: no need to declare pos here as there is only one field, can use directly 1 in the setstring call

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
```diff
@@ -745,5 +754,31 @@ public void run() {
...
+                    docreqItem.getSize(), rights);
+        } // if file access is revoked at source delete it from receiver end as well
+        else if (reqItem.getType().equals(Type.DOCUMENT) && action != null && action.name().equals(REVOKE)) {
+            String fileUUIDToDelete = fileUUID + ":" + ownerFileId;
+            mbox.deleteFileSharedWithMe(mbox, octxt, fileUUIDToDelete);
+        } // if file information is updated, then update at receivers end as well
+          // the reason for delete and create is, upon executing update query the latest
+          // information wasn't returned
+          // on Search response as items were returned from cached data
+        else if (reqItem.getType().equals(Type.DOCUMENT) && action != null && action.name().equalsIgnoreCase("edit")) {
```
dasiyogesh, line 775: can we use constant for edit here

store/src/java/com/zimbra/cs/service/mail/SearchResponse.java
```diff
@@ -149,13 +149,17 @@ int size() {
...
      */
-    void add(ZimbraHit zimbraHit) throws ServiceException{
-		add(zimbraHit,false);
+    void add(ZimbraHit zimbraHit) throws ServiceException {
+        add(zimbraHit, false);
+    }
+
+    void add(ZimbraHit zimbraHit, boolean expandMsg) throws ServiceException {
+        add(zimbraHit, expandMsg, false);
+    }
```
k-kato, line 158: both look incorrect; a wrong function seems to be called on each.

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
```diff
@@ -745,5 +754,31 @@ public void run() {
...
+                    ownerAccountId, ownerFileId, fileUUID, fileOwnerName, docreqItem.getContentType(),
+                    docreqItem.getSize(), rights);
+        } // if file access is revoked at source delete it from receiver end as well
+        else if (reqItem.getType().equals(Type.DOCUMENT) && action != null && action.name().equals(REVOKE)) {
+            String fileUUIDToDelete = fileUUID + ":" + ownerFileId;
+            mbox.deleteFileSharedWithMe(mbox, octxt, fileUUIDToDelete);
+        } // if file information is updated, then update at receivers end as well
+          // the reason for delete and create is, upon executing update query the latest
+          // information wasn't returned
+          // on Search response as items were returned from cached data
```
k-kato, line 774: cache should be removed rather than recreating an item if possible. DB write process is higher cost on performance.

store/src/java/com/zimbra/cs/service/mail/Search.java
```diff
@@ -165,6 +167,7 @@ private void putHits(ZimbraSoapContext zsc, OperationContext octxt, Element el,
         while (pager.hasNext() && resp.size() < params.getLimit()) {
             hitNum ++;
             ZimbraHit hit = pager.getNextHit();
+            params.getQueryString();
```
k-kato, line 170: is it necessary?

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -9064,6 +9071,63 @@ public Mountpoint refreshMountpoint(OperationContext octxt, int mountpointId, St
...
+        }
+    }
+
+    public void deleteFileSharedWithMe(Mailbox mbox, OperationContext octxt, String remoteUuid)
+            throws ServiceException {
+        CreateFileSharedWithMe redoRecorder = new CreateFileSharedWithMe(mId, remoteUuid);
+        boolean success = false;
+        try {
+            beginTransaction("deleteFileSharedWithMe", octxt, redoRecorder);
+            CreateFileSharedWithMe redoPlayer = (CreateFileSharedWithMe) currentChange().getRedoPlayer();
```
k-kato, line 9113: * Is it okay to use **Create**FileSharedWithMe here? Don't we need to to define DeleteFileSharedWithMe or like that?
* this `redoPlayer` is not used.

store/src/java/com/zimbra/cs/service/mail/SearchResponse.java
gabhinavTI: we have 3 add methods,  

> add(ZimbraHit zimbraHit)

> add(ZimbraHit zimbraHit, boolean expandMsg)

> add(ZimbraHit hit, boolean expandMsg, boolean isSharedFileWithMe)

if ` add(ZimbraHit zimbraHit)` is invoked then we assume `expandMsg` is false and similarly `isSharedFileWithMe` also becomes false.
similary when `add(ZimbraHit zimbraHit, boolean expandMsg) ` is called the, apart from respective values , `isSharedFileWithMe` becomes false.
And when `add(ZimbraHit hit, boolean expandMsg, boolean isSharedFileWithMe)` is called all the respective values are provided
May I know what exactly seems wrong here so I can fix it up ?

store/src/java/com/zimbra/cs/db/DbMailItem.java
gabhinavTI: done

gabhinavTI:
**Problem**: As of current when the file is shared i.e. User A shares any file with User B then the file is shared as an accessible link on the mail. So assume when I have to access the file again after a month I receive I'll have to scroll down to check the mail where the link was shared. By any chance, if that file is deleted from the grantee's side, they won't be able to access it again and have to ask for resharing of the file

**Solution**: To overcome the above issue we came up with a solution i.e. to create a system folder `File Shared with Me` under the briefcase tab. Now when user A shares the file with User B, the file is accessible via the link, and also file is available under the `File Shared with Me` folder. This file can be previewed upon clicking, and based on the rights provided file can be viewed and edited accordingly. There is an option to download the file as well.

**Implementation Approach**: When user A shares a file with user B, then in `SendShareNotification` the grantee ID is used to fetch the mailbox details of user B. Based on the mailbox details retrieved DB actions are made. There are basically 3 scenarios that can happen:
1) When the File is shared -> The file details are captured, i.e. fileName, content type, permissions to the file, size, etc. Based on the grantee ID grantee's mailbox is retrieved and using mailbox detail and input available a new DB entry is made directly on the grantee's side on the `File Shared with Me `.This file can be accessed based on an ID that has a value like "AccountID: FieldId"

2) When Acess is revoked --> In this case the file is deleted from the grantee side using the delete query and file will no longer be accessible on the `File Shared with Me` folder

3) When File details/rights are changed --> In this case firstly older DB entry is deleted and then newer entry is created on UserB mailbox

Testing:
Manual testing through SOAP API to check what files are available when input is "File Shared With me Folder"
Testing through UI on a single Node server, when the file which has been shared can be accessed via a link. ALso same file is available under the `File Shared With me` Folder. These files can be previewed, and there is an option to open the file in a new tab and edit it according to the rights provided. The users could also download single file at a time

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -2266,6 +2271,8 @@ protected void createDefaultFolders() throws ServiceException {
                             MailItem.Type.MESSAGE, 0, MailItem.DEFAULT_COLOR_RGB, null, null, null);
             Folder.create(ID_FOLDER_BRIEFCASE, UUIDUtil.generateUUID(), this, userRoot, "Briefcase", system,
                             MailItem.Type.DOCUMENT, 0, MailItem.DEFAULT_COLOR_RGB, null, null, null);
+            Folder.create(ID_FOLDER_FILE_SHARED_WITH_ME, UUIDUtil.generateUUID(), this, userRoot, "Files shared with me", system,
```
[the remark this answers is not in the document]
dasiyogesh: UI will check.

soap/src/java/com/zimbra/soap/mail/message/FileSharedWithMeRequest.java
```diff
@@ -0,0 +1,160 @@
+/* 
+* ***** BEGIN LICENSE BLOCK *****
```
k-kato, line 2: fix indent

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
```diff
@@ -0,0 +1,82 @@
...
+            String rights = request.getAttribute(MailConstants.A_RIGHTS);
+            String contentType = request.getAttribute(MailConstants.A_CONTENT_TYPE);
+            long size = request.getAttributeLong(MailConstants.A_SIZE);
+
+            if (Provisioning.onLocalServer(account)) {
+                // if file is shared, this file should be available at receivers end
+                if (action.equals(CREATE)) {
+                    mbox.createFileSharedWithMe(mbox, octxt, FolderConstants.ID_FOLDER_FILE_SHARED_WITH_ME, fileName,
+                            ownerAccountId, ownerFileId, fileUUID, fileOwnerName, contentType, size, rights);
+                } // if file access is revoked at source delete it from receiver end as well
```
k-kato, line 61:
```suggestion
                } // if file access is revoked at source, delete it from receiver end as well
```

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
```diff
@@ -0,0 +1,82 @@
...
+                else if (action != null && action.equals(REVOKE)) {
+                    String fileUUIDToDelete = fileUUID + ":" + ownerFileId;
+                    mbox.deleteFileSharedWithMe(mbox, octxt, fileUUIDToDelete);
+                } // if file rights is changed , update file permission accordingly
```
k-kato, line 65:
```suggestion
                } // if file rights is changed, update file permission accordingly
```

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
```diff
@@ -745,5 +765,83 @@ public void run() {
...
+        if (uri == null || uri.length() == 0) {
+            soapUri = AccountUtil.getSoapUri(acct);
+        } else {
+            soapUri = uri + AccountConstants.USER_SERVICE_URI;
+        }
+        SoapHttpTransport transport = new SoapHttpTransport(soapUri);
+        try {
+            transport.setUserAgent(ZIMBRA_SOAP_USER_AGENT, ZIMBRA_SOAP_VERSION);
+            transport.setTimeout(timeout);
+            transport.setRetryCount(1);
```
k-kato, line 828: Do we need to set them rather than use default values set in SoapHttpTransport?

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
```diff
@@ -745,5 +765,83 @@ public void run() {
...
+                mbox.deleteFileSharedWithMe(mbox, octxt, fileUUIDToDelete);
+            } // if file rights is changed , update file permission accordingly
+            else if (reqItem.getType().equals(Type.DOCUMENT) && action != null
+                    && action.name().equalsIgnoreCase(EDIT)) {
+                Document docReqItem = (Document) reqItem;
+                mbox.updateFileSharedWithMe(mbox, octxt, ownerAccountId, ownerFileId, fileUUID, fileOwnerName,
+                        docReqItem.getContentType(), rights);
+            }
+        } else {
+            Document docReqItem = (Document) reqItem;
```
k-kato, line 795: Cant't we use `proxyRequest`?

store/src/java/com/zimbra/cs/redolog/op/CreateFileSharedWithMe.java
```diff
@@ -0,0 +1,111 @@
+package com.zimbra.cs.redolog.op;
```
k-kato, line 1: add license block

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
k-kato: Please verify that playing redolog can restore the change(s) correctly at testing phase as well.

common/src/java/com/zimbra/common/soap/MailConstants.java
```diff
@@ -605,6 +607,10 @@ private MailConstants() {
     public static final QName IMAP_COPY_REQUEST = QName.get(E_IMAP_COPY_REQUEST, NAMESPACE);
     public static final QName IMAP_COPY_RESPONSE = QName.get(E_IMAP_COPY_RESPONSE, NAMESPACE);
 
+    //File Shared With Me
+    public static final QName FILE_SHARED_WITH_ME_REQUEST = QName.get( E_FILE_SHARED_WITH_ME_REQUEST, NAMESPACE);
```
dasiyogesh, line 611: remove space after (

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
```diff
@@ -80,6 +91,12 @@ public class SendShareNotification extends MailDocumentHandler {
     }
 
     private static final String REVOKE = "revoke";
+    private static final String EDIT = "edit";
+    private static final String CREATE = "edit";
+
+    public static final String ZIMBRA_SOAP_USER_AGENT = "ZimbraSync";
```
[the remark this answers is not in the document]
dasiyogesh: I think this could be `ZimbraWebClient` since the request being triggered from web client for sharing Or better capture the UA from current request and forward it in the further SOAP call.

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
dasiyogesh: here the request is getting changed from SendShareNotification to FileSharedWithMe
@gabhinavTI did you try proxyRequest here instead since the new request can be formulated here to pass on along with the current context from the handle of SendShareNotification?

store/src/java/com/zimbra/cs/service/mail/Search.java
```diff
@@ -72,6 +73,8 @@
  */
 public class Search extends MailDocumentHandler  {
 
+    public static final String FILE_SHARED = "in:\"Files shared with me\"";
```
silentsakky, line 76: why do we need this check? client can also send in:20 by using folder id

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
```diff
@@ -80,6 +91,12 @@ public class SendShareNotification extends MailDocumentHandler {
...
+    public static final String ZIMBRA_SOAP_VERSION = "1.0";
```
silentsakky, line 98: aren't these already defined, almost all Soap apis will be using this right?

store/src/java/com/zimbra/cs/db/DbMailItem.java
```diff
@@ -1697,6 +1697,29 @@ public static void delete(Mailbox mbox, Collection<Integer> ids, boolean fromDum
...
+    public void deleteSharedFile(Mailbox mbox, String uuid, boolean fromDumpster) throws ServiceException {
+        DbConnection conn = mbox.getOperationConnection();
+        PreparedStatement stmt = null;
+        try {
+            stmt = conn
+                    .prepareStatement("DELETE FROM " + getMailItemTableName(mbox, fromDumpster) + " WHERE " + "uuid=?");
+            stmt.setString(1, uuid);
+            stmt.executeUpdate();
+        } catch (SQLException e) {
+            throw ServiceException.FAILURE("deleting file shared with me failed " + uuid, e);
```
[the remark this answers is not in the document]
gabhinavTI: making use of existing delete methof , hence would not require to make change here

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
```diff
@@ -745,5 +765,83 @@ public void run() {
         senderThread.start();
     }
 
-}
+    private void processDBFilesSharedWithMe(MailItem reqItem, Action action, OperationContext octxt, String granteeId,
+            String fileName, String ownerAccountId, int ownerFileId, String fileUUID, String fileOwnerName,
+            String rights, Map<String, Object> context) throws ServiceException {
+        // Mailbox mbox = MailboxManager.getInstance().getMailboxByAccountId(granteeId);
+        // Account target = Provisioning.getInstance().get(Key.AccountBy.id,
+        // mRemoteOwnerId);
```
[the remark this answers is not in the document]
gabhinavTI: done

store/src/java/com/zimbra/cs/service/mail/ToXML.java
```diff
@@ -291,6 +292,10 @@ public static Element encodeFolder(Element parent, ItemIdFormatter ifmt, Operati
         }
         Element elem = parent.addNonUniqueElement(MailConstants.E_FOLDER);
         encodeFolderCommon(elem, ifmt, folder, fields);
+        if (folder.getId() == FolderConstants.ID_FOLDER_FILE_SHARED_WITH_ME) {
+            elem.addAttribute(MailConstants.A_RIGHTS, "r");
+            elem.addAttribute(MailConstants.A_ZIMBRA_ID, "0");
```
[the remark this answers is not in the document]
gabhinavTI: this field is required in UI implementation when we make folder read-only , not sure of the purpose but when this was discussed with UI people any value would work , hence "0"

soap/src/java/com/zimbra/soap/mail/message/FileSharedWithMeRequest.java
gabhinavTI: done

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
gabhinavTI: done

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
gabhinavTI: done

store/src/java/com/zimbra/cs/redolog/op/CreateFileSharedWithMe.java
gabhinavTI: done

common/src/java/com/zimbra/common/soap/MailConstants.java
gabhinavTI: done

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
```diff
@@ -80,6 +91,12 @@ public class SendShareNotification extends MailDocumentHandler {
     }
 
     private static final String REVOKE = "revoke";
+    private static final String EDIT = "edit";
+    private static final String CREATE = "edit";
```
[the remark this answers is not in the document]
gabhinavTI: modified , these were used directly at time of testing , changed it as constants before raising PR
updated the value accordingly

soap/src/java/com/zimbra/soap/mail/message/FileSharedWithMeRequest.java
k-kato: it has not been fixed.
```suggestion
 * ***** BEGIN LICENSE BLOCK *****
```

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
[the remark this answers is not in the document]
k-kato: it still exists

store/src/java/com/zimbra/cs/service/mail/ToXML.java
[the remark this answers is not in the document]
k-kato: * We should declare a variable with an appropriate name to match its meaning/purpose.
* It must be defined in/added to SOAP API reference in soap/src/java/... class(es) as well.

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
k-kato: it has not been fixed

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
k-kato: it has not been fixed

store/src/java/com/zimbra/cs/redolog/op/CreateFileSharedWithMe.java
k-kato: it has not been added

common/src/java/com/zimbra/common/soap/MailConstants.java
k-kato: it has not been fixed

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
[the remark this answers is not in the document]
gabhinavTI: removed the commented code

soap/src/java/com/zimbra/soap/mail/message/FileSharedWithMeRequest.java
gabhinavTI: done!

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
gabhinavTI: commited code!

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
gabhinavTI: updated it to get data from SoapHttpTransport class

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
gabhinavTI: tried using proxyRequest here, but I am coming across errors . hence proceeding with sendRequest call which works for both single node and multi node environment without any issue

store/src/java/com/zimbra/cs/redolog/op/CreateFileSharedWithMe.java
gabhinavTI: commited code for this

common/src/java/com/zimbra/common/soap/MailConstants.java
gabhinavTI: commited code!

store/src/java/com/zimbra/cs/service/mail/MailService.java
```diff
@@ -255,5 +255,8 @@ public void registerHandlers(DocumentDispatcher dispatcher) {
         dispatcher.registerHandler(OctopusXmlConstants.GET_NOTIFICATIONS_REQUEST, new GetNotifications());
         dispatcher.registerHandler(OctopusXmlConstants.GET_DOCUMENT_SHARE_URL_REQUEST, new GetDocumentShareURL());
         dispatcher.registerHandler(OctopusXmlConstants.GET_SHARE_DETAILS_REQUEST, new GetShareDetails());
+
+        //File Shared With Me
```
dasiyogesh, line 259: add space after //

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
gabhinavTI: I couldn't find out where exactly these are defined , check in SoapHTTPTransport class . couldn't find any similar constant . if you any reference it would be helpful

store/src/java/com/zimbra/cs/service/mail/MailService.java
gabhinavTI: done

store/src/java/com/zimbra/cs/service/mail/Search.java
gabhinavTI: since the folder name was fixed , and no other folder could  have this name or passed in similar format(for subfolder)
hence we had decided to go with this, Let me know your thought on this if requires modification from client can send folder ID , will convey same to Mitesh

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
gabhinavTI: done

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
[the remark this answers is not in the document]
gabhinavTI: since the method which was using this is  removed, remove these constants

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
gabhinavTI: the method is removed and used proxyRequest

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
gabhinavTI: used proxyRequest , some changes in `ZimbraSoapContext` class was required . Implemented with support of Stephan

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
gabhinavTI: the method using this was removed , hence this removed this constants as well

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
```diff
@@ -114,9 +130,12 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
                     sharesWithGroupGrantee.add(sid);
                 } else {
                     sendNotificationEmail(octxt, mbox, authAccount, account, sid, notes, action, null, null, reqItem);
+                    // create, delete, update
```
k-kato, line 133: We may be able to remove the comment. If it is necessary, please add more information so that it will be meaningful.

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
jiteshsojitra: @k-kato don't have much idea about this API but you meant when messages are permanently deleted and user recovers deleted mails, this emails should be recovered and shared links should work?

store/src/java/com/zimbra/cs/redolog/op/CreateFileSharedWithMe.java
```diff
@@ -0,0 +1,128 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2004, 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2018 Synacor, Inc.
```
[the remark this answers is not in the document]
jiteshsojitra: I think, these changes applicable to entire zimbra (classic) repos and should be fixed separately for all.

store/src/java/com/zimbra/cs/service/mail/ToXML.java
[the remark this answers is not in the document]
k-kato: Please ignore the first point. As we discussed in a different channel, `zid` is used in the same ways as others like https://github.com/Zimbra/zm-mailbox/blob/9.0.0.p26/store/src/java/com/zimbra/cs/service/mail/ToXML.java#L575

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
k-kato: @jiteshsojitra In `createFileSharedWithMe`, `deleteFileSharedWithMe` and `updateFileSharedWithMe` functions in Mailbox.java, DB entry is created/updated/deleted for an item which is shared with the account.
For example, after full back up is taken, a file is shared (or the sharing is updated or revoked). The process is recorded in redolog. When an account is restored from backup data and redolog, the target record should be restored (= created/updated/deleted) by replaying the redolog.
We need to check all cases; create, update and delete.

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
```diff
@@ -0,0 +1,84 @@
...
+                    mbox.createFileSharedWithMe(mbox, octxt, FolderConstants.ID_FOLDER_FILE_SHARED_WITH_ME, fileName,
+                            ownerAccountId, ownerFileId, fileUUID, fileOwnerName, contentType, size, rights);
+                } // if file access is revoked at source, delete it from receiver end as well
+                else if (action != null && action.equals(REVOKE)) {
+                    mbox.deleteFileSharedWithMe(mbox, octxt, fileUUID, ownerFileId, ownerAccountId, granteeAccountID);
+                } // if file rights is changed, update file permission accordingly
+                else if (action != null && action.equalsIgnoreCase(EDIT)) {
+                    mbox.updateFileSharedWithMe(mbox, octxt, ownerAccountId, ownerFileId, fileUUID, fileOwnerName,
+                            contentType, rights, granteeAccountID);
+                }
```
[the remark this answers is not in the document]
gabhinavTI: @k-kato in this API `ownerAccountId` and `granteeAccountID`  are retrieved from `zsc` . this is possible to only retrieved when the call is from UI. when the request is from SOAP UI the auth token would contain the same `ownerAccountId` and `granteeAccountID`, added an extra check if it is the same then don't share the file as sharing the file to self is not a valid request. for other validations like invalid Id, or invalid UUID it would be taken care of in DB operation. The reason of extracting `ownerAccountId` and `granteeAccountID` from zcs was to implement security concern so that no 3rd party could make delete/update operation

store/src/java/com/zimbra/cs/service/mail/Search.java
silentsakky: I think we need to write generic code on server side, client can pass either folder name or folder id in query or any other supported query parameter
we shouldn't hardcode that on server side, is it possible to remove this hardcoded check?
@dasiyogesh any thoughts?

store/src/java/com/zimbra/cs/redolog/op/CreateFileSharedWithMe.java
```diff
@@ -0,0 +1,111 @@
...
+        sb.append(", name=").append(mName).append(", folder=").append(mFolderId);
+        sb.append(", owner=").append(mOwnerId).append(", remoteId=").append(mRemoteId).append(", remoteUuid=")
+                .append(mRemoteUuid);
+        return sb.toString();
+    }
+
+    @Override
+    protected void serializeData(RedoLogOutput out) throws IOException {
+        out.writeInt(mId);
+        if (getVersion().atLeast(1, 37)) {
```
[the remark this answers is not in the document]
gabhinavTI: this was taken in reference from some other class , removed this check

store/src/java/com/zimbra/cs/redolog/op/CreateFileSharedWithMe.java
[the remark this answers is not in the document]
gabhinavTI: done for this file

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
```diff
@@ -0,0 +1,84 @@
...
+            String fileName = request.getAttribute(MailConstants.A_CONTENT_FILENAME);
+            int ownerFileId = request.getAttributeInt(MailConstants.A_ITEMID);
+            String fileUUID = request.getAttribute(MailConstants.A_REMOTE_UUID);
+            String fileOwnerName = request.getAttribute(MailConstants.A_OWNER_NAME);
+            String rights = request.getAttribute(MailConstants.A_RIGHTS);
+            String contentType = request.getAttribute(MailConstants.A_CONTENT_TYPE);
+            long size = request.getAttributeLong(MailConstants.A_SIZE);
+            if (Provisioning.onLocalServer(account)) {
+                // if file is shared, this file should be available at receivers end
+                if (action.equals(CREATE)) {
```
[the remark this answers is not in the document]
gabhinavTI: added `action != null &&`  check

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
```diff
@@ -80,6 +91,11 @@ public class SendShareNotification extends MailDocumentHandler {
     }
 
     private static final String REVOKE = "revoke";
+    private static final String EDIT = "edit";
+    private static final String CREATE = "create";
+
+    public static final String ZIMBRA_SOAP_USER_AGENT = "ZimbraWebClient";
+    public static final String ZIMBRA_SOAP_VERSION = "1.0";
```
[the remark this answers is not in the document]
gabhinavTI: removed

store/src/java/com/zimbra/cs/service/mail/SendShareNotification.java
gabhinavTI: removed comment as it's defined inside this method

store/src/java/com/zimbra/cs/service/mail/Search.java
gabhinavTI: @silentsakky we can remove this hardcoded check , but we would require folder ID/folder name in request to do validation in puthits method

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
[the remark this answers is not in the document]
k-kato: I understand about values retrieved from `zsc`. However, for other variables/parameters, is it okay not to check values?
What will happen when
* fileName, contentType or size does not match the actual file's one?
* ownerFileId or fileUUID is a non-existing ID or invalid format?
* fileOwnerName is a strange name? Does it need to match actual owner name?
* rights is an invalid value like "abc"?

I don't think they are checked in Mailbox.java#createFileSharedWithMe around line 9086...

> it would be taken care of in DB operation

`new DbMailItem(mbox).create(data);` does not check the above but just writes an entry using given values.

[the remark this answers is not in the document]
k-kato: As we discussed, FileSharedWithMeRequest is called only from zimbra server via proxyRequest in SendShareNotification class. It cannot be called directly from outside of zimbra. Then all parameters in a soap request are always valid.

store/src/java/com/zimbra/cs/service/mail/ToXML.java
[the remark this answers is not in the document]
gabhinavTI: As discussed 0 would be okay, marking this as resolved. and making notes in Jira for the response returned back

soap/src/java/com/zimbra/soap/mail/message/FileSharedWithMeRequest.java
```diff
@@ -0,0 +1,135 @@
...
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+
+import com.zimbra.common.soap.MailConstants;
+
+@XmlAccessorType(XmlAccessType.NONE)
+@XmlRootElement(name = MailConstants.E_FILE_SHARED_WITH_ME_REQUEST)
+public class FileSharedWithMeRequest {
```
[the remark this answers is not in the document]
gabhinavTI: verified , also added note like `This is an internal API, cannot be invoked directly`

store/src/java/com/zimbra/cs/service/mail/Search.java
gabhinavTI: @silentsakky @dasiyogesh as discussed removed this check and implement check based on folder ID , checking on SOAP API and UI does seem to work

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
gabhinavTI: marking this as resolved as this needs to be in testing phase

soap/src/java/com/zimbra/soap/mail/message/FileSharedWithMeResponse.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public FileSharedWithMeResponse() {
+    }
+
+    public FileSharedWithMeResponse(String status) {
+        this.status = status;
+    }
+
+    public MoreObjects.ToStringHelper addToStringInfo(MoreObjects.ToStringHelper helper) {
+        return helper.add("status", status);
```
dasiyogesh, line 40: use MailConstants.A_STATUS here

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -9064,6 +9071,97 @@ public Mountpoint refreshMountpoint(OperationContext octxt, int mountpointId, St
...
+    public void deleteFileSharedWithMe(Mailbox mbox, OperationContext octxt, String remoteUuid, int ownerFileId,
+            String ownerAccountId, String granteeAccountID) throws ServiceException {
+        CreateFileSharedWithMe redoRecorder = new CreateFileSharedWithMe(mId, remoteUuid);
+        boolean success = false;
+        try {
+            beginTransaction("deleteFileSharedWithMe", octxt, redoRecorder);
+            Document item = (Document) mbox.getItemByUuid(remoteUuid + ":" + ownerFileId, Type.DOCUMENT);
+            if (item.getOwnerAccountId().equals(ownerAccountId) && item.getAccountId().equals(granteeAccountID)) {
+                DbMailItem.delete(mbox, Collections.singletonList(item.getId()), false);
+                mbox.getFolderById(ID_FOLDER_FILE_SHARED_WITH_ME).updateSize(-1, -1, -item.getSize()); //update the folder item count and size accordingly
```
dasiyogesh, line 9118: for comments add space after `//` and move the comment above the line

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -9064,6 +9071,97 @@ public Mountpoint refreshMountpoint(OperationContext octxt, int mountpointId, St
...
+                meta.put(Metadata.FN_FILE_PERMISSION, rights);
+                meta.put(Metadata.FN_OWNER_ACCOUNT_ID, ownerAccountID);
+                meta.put(Metadata.FN_OWNER_FILE_ID, ownerFileId);
+                mbox.getItemCache().remove(item.getId());
+                new DbMailItem(mbox).update(item, meta);
+                success = true;
+            } else {
+                throw ServiceException.PERM_DENIED("Update operation failed, Invalid request source");
+            }
+
```
dasiyogesh, line 9149: remove empty line

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
```diff
@@ -0,0 +1,90 @@
...
+
+            String action = request.getAttribute(MailConstants.E_ACTION);
+            String fileName = request.getAttribute(MailConstants.A_CONTENT_FILENAME);
+            int ownerFileId = request.getAttributeInt(MailConstants.A_ITEMID);
+            String fileUUID = request.getAttribute(MailConstants.A_REMOTE_UUID);
+            String fileOwnerName = request.getAttribute(MailConstants.A_OWNER_NAME);
+            String rights = request.getAttribute(MailConstants.A_RIGHTS);
+            String contentType = request.getAttribute(MailConstants.A_CONTENT_TYPE);
+            long size = request.getAttributeLong(MailConstants.A_SIZE);
+            // boolean isDifferentAccountId = !ownerAccountId.equals(granteeAccountID);
```
dasiyogesh, line 55: remove commented code

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
```diff
@@ -0,0 +1,90 @@
...
+            if (ownerAccountId.equals(granteeAccountID)) {
```
dasiyogesh, line 56: move this check to line 43

store/src/java/com/zimbra/cs/service/mail/SearchResponse.java
```diff
@@ -150,9 +150,10 @@ int size() {
      * @throws ServiceException error
      */
     void add(ZimbraHit zimbraHit) throws ServiceException{
-		add(zimbraHit,false);
+        add(zimbraHit,false);
+
```
dasiyogesh, line 154: remove empty line, this change is not needed anymore

store/src/java/com/zimbra/soap/ZimbraSoapContext.java
```diff
@@ -524,6 +526,11 @@ private void validateDelegatedAccess(Account targetAccount, DocumentHandler hand
             return;
         }
 
+        if ((handler != null) && handler instanceof DelegatableRequest
```
dasiyogesh, line 529: is this required?

soap/src/java/com/zimbra/soap/mail/message/FileSharedWithMeResponse.java
gabhinavTI: done

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
gabhinavTI: moved the comment above line and formatted accordingly

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
gabhinavTI: done

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
gabhinavTI: removed

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
gabhinavTI: moved the code after getting ownerAccountId and ownerAccountId from zsc , also added logger statement  before throwing exception

store/src/java/com/zimbra/cs/service/mail/SearchResponse.java
gabhinavTI: done

store/src/java/com/zimbra/soap/ZimbraSoapContext.java
gabhinavTI: Yes, this is required, so the proxyRequest was failing in this method on the receiver end. hence Stephan asked me to include this check so that request doesn't fail here. Post this fix proxyRequest was working fine

dasiyogesh (approved):

store/src/java/com/zimbra/cs/service/mail/FileSharedWithMe.java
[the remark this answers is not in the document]
desouzas: > It cannot be called directly from outside of zimbra.

It's part of the API so it can be. https://github.com/Zimbra/zm-mailbox/pull/1326/files#diff-aab6f1809776ef672dc9da1243f8c875a1d496cdad83b518f0aa3d35dbbe59a5R260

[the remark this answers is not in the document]
k-kato: > It cannot be called directly from outside of zimbra.

It was not exact. 
FileSharedWithMeRequest can be called from anywhere. However, as per previous discussion, ownerAccountId and granteeAccountID will be the same when it is called directly (= without using proxyRequest). Then it is addressed as an error in the code below. 
https://github.com/Zimbra/zm-mailbox/pull/1326/files#diff-48c461d013fb1197ea91e69b6c526f9f48088974958f12df74dd9512d20dccceR44
```
            if (ownerAccountId.equals(granteeAccountID)) {
                ZimbraLog.mailbox.error("File sharer and grantee cannot be same", new Throwable("invalid request source"));
                throw ServiceException.FAILURE("invalid request", new Throwable("invalid"));
            }
```
It will be described in API reference.
https://github.com/Zimbra/zm-mailbox/pull/1326/files#diff-f9dde6e255750d299d6da695c0170bc0fc1537a8c011382b2e3f4f676749c23bR32
```
This is an internal API, cannot be invoked directly
```

k-kato:
Please add a description in `Description` of each element in SOAP API document even though it is an internal API.

(scope)
soap/src/java/com/zimbra/soap/mail/message/FileSharedWithMeRequest.java
soap/src/java/com/zimbra/soap/mail/message/FileSharedWithMeResponse.java

current FileSharedWithMe

FileSharedWithMe
--

comparison - SendShareNotification
SendShareNotification

k-kato (approved):

store/src/java/com/zimbra/cs/service/mail/SearchResponse.java
```diff
@@ -150,9 +150,9 @@ int size() {
      * @throws ServiceException error
      */
     void add(ZimbraHit zimbraHit) throws ServiceException{
-		add(zimbraHit,false);
+        add(zimbraHit,false);
```
silentsakky, line 153: revert this if there are no changes in this file

silentsakky (approved):

store/src/java/com/zimbra/cs/service/mail/SearchResponse.java
gabhinavTI: this code was not formatted previously, it's formatted now. Should I still change it back?

silentsakky: ok

PR #1333, opened 2022-07-22
ZCS-11440: ZimbraBlobMover: add support for blob movement from S3 to S3

wilfredo-alfaro:
### Solution
ZCS-11440 is about adding support for S3-to-S3 movement when it comes to executing the existing HSM policy (i.e. when both primary and secondary volumes are S3 Buckets). The solution works by adding a few new classes and updating existing classes in the `zm-hsm-project`, such as the `ZimbraBlobMoverAdaptor` and `BlobMoverThread` classes respectively.

### Notes
Now, regarding the `zm-mailbox` project, a new `zimbra_hybrid_blob_mover_enabled` local config property has been added, which default to `false`, for safety purposes (i.e. making easy to handover among the current `BlobMover` and new `ZimbraBlobMoverAdaptor`).

### Tests

### DIFF
--- a/store/src/java/com/zimbra/cs/store/file/FileBlobStore.java
+++ b/store/src/java/com/zimbra/cs/store/file/FileBlobStore.java
@@ -437,4 +437,41 @@ public final class FileBlobStore extends StoreManager {
     private static void ensureParentDirExists(File file) throws IOException {
         ensureDirExists(file.getParentFile());
     }
+
+    /**
+     * Store a blob from external S3 storage to internal store in volume/path
+     * specified by dest* parameters. Note this method is not part of the
+     * StoreManager interface It is only to be used for FileBlobStore specific code
+     * such as ExternalToInternal
+     * 
+     * @param in
+     * @param storeAsIs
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/store/file/FileBlobStore.java
```diff
@@ -437,4 +437,41 @@ private static void ensureDirExists(File dir) throws IOException {
...
+     * @param in
+     * @param storeAsIs
+     * @param mbox
+     * @param itemId
+     * @param revision
+     * @param destVolumeId mail_item.volume_id for message in dest Mbox
+     * @throws IOException
+     * @throws ServiceException
+     */
+    public Blob storeExternalToInternal(InputStream in, boolean storeAsIs, Mailbox mbox, int itemId, int revision,
```
RakeshAMore, line 456: make it overloaded method `store`

zimsuchitgupta (approved):

RakeshAMore (approved):

PR #1340, opened 2022-07-27
ZCS-11799: Removed unnecessary code which had endless loop and blocki…

zimsuchitgupta:
**Problem**
Redo log replaying process was blocked and redo logs were not replaying while running `zmplayredo` command

**Fix**
Found that there were a endless loop in code which was unnecessary as it was looking for console appender and then was deleting it. This was working fine in log4j-1.2 but in log4j-2.17.1 this while code block was running for infinite times.

Have removed this code block because we are initializing the /opt/zimbra/log4j.properties file which do not have any default console appender so searching for console appender and then deleting this after the initialization was seems pointless.

**Testing**
Tested in local environment and as able to run `zmplayredo`

### DIFF
--- a/store/src/java/com/zimbra/cs/redolog/util/PlaybackUtil.java
+++ b/store/src/java/com/zimbra/cs/redolog/util/PlaybackUtil.java
@@ -381,22 +381,6 @@ public class PlaybackUtil {
     private static void setup() throws ServiceException {
         // set up log4j
         ZimbraLog.toolSetupLog4j("INFO", LC.zimbra_log4j_properties.value());
-        // remove the console appender if any
-        org.apache.logging.log4j.Logger rootLogger = LogManager.getRootLogger();
-        Appender consoleAppender = null;
-        LoggerContext context = LoggerContext.getContext(false);
-        Configuration configuration = context.getConfiguration();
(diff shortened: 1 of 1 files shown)


silentsakky (approved):

k-kato:
Just to confirm. Is it okay to remove the code rather than fixing log4j.properties.in or any other log4j-related configuration? Will we see events in a log file and/or stdout/stderr  in console as same as before (with log4j-1.2)?

abhishekdeshmukh2209 (approved):

zimsuchitgupta:
> Just to confirm. Is it okay to remove the code rather than fixing log4j.properties.in or any other log4j-related configuration? Will we see events in a log file and/or stdout/stderr in console as same as before (with log4j-1.2)?

Our standard log4j.properties do not have any stdout/stderr defined, also that code was only considering the last console appneder and was deleting it.
Also our log4j.properties.in just have appenders and loggers defined don't think we have any configuration to fix that.

k-kato (approved):
I got it. Thanks @zimsuchitgupta

dasiyogesh (approved):

rcyarrapothu (approved):

PR #1357, opened 2022-08-05
ZCS-11658 RedoLog for Storage Management Relaunch changes

inpati:
**Problem Statement**
Create Redo operation for CreateVolumeRequest and ModifyVolumeRequest

**Solution**
Added functionality which shall perform redoLog operations for CreateVolumeRequest, ModifyVolumeRequest and DeleteVolumeRequest

### DIFF
--- a/store/src/java/com/zimbra/cs/redolog/op/CreateVolume.java
+++ b/store/src/java/com/zimbra/cs/redolog/op/CreateVolume.java
@@ -17,15 +17,25 @@
 
 package com.zimbra.cs.redolog.op;
 
-import java.io.IOException;
-
 import com.google.common.base.MoreObjects;
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.soap.AdminConstants;
+import com.zimbra.common.util.ZimbraLog;
+import com.zimbra.cs.account.Provisioning;
 import com.zimbra.cs.mailbox.MailboxOperation;
 import com.zimbra.cs.redolog.RedoLogInput;
 import com.zimbra.cs.redolog.RedoLogOutput;
 import com.zimbra.cs.volume.Volume;
 import com.zimbra.cs.volume.VolumeManager;
 import com.zimbra.cs.volume.VolumeServiceException;
+import com.zimbra.soap.admin.type.VolumeExternalInfo;
+import com.zimbra.soap.admin.type.VolumeInfo;
+import com.zimbra.util.ExternalVolumeInfoHandler;
+
+import java.io.IOException;
+
+import org.json.JSONException;
+import org.json.JSONObject;
 
 public final class CreateVolume extends RedoableOp {
 
@@ -124,6 +134,9 @@ public final class CreateVolume extends RedoableOp {
                     .setFileGroupBits(fileGroupBits).setFileBits(fileBits)
                     .setCompressBlobs(compressBlobs).setCompressionThreshold(compressionThreshold)
                     .setStoreType(enumStoreType).build();
+            VolumeInfo volInfo = ExternalVolumeInfoHandler.buildVolumeInfoFromBackupData(volume);
+            ExternalVolumeInfoHandler extVolInfoHandler = new ExternalVolumeInfoHandler(Provisioning.getInstance());
+            extVolInfoHandler.addServerProperties(volInfo);
             mgr.create(volume, getUnloggedReplay());
         } catch (VolumeServiceException e) {
             if (e.getCode() == VolumeServiceException.ALREADY_EXISTS) {
@@ -133,5 +146,4 @@ public final class CreateVolume extends RedoableOp {
             }
         }
     }
-
 }
--- a/store/src/java/com/zimbra/cs/redolog/op/ModifyVolume.java
+++ b/store/src/java/com/zimbra/cs/redolog/op/ModifyVolume.java
@@ -23,8 +23,11 @@ import com.google.common.base.MoreObjects;
 import com.zimbra.cs.mailbox.MailboxOperation;
 import com.zimbra.cs.redolog.RedoLogInput;
 import com.zimbra.cs.redolog.RedoLogOutput;
+import com.zimbra.cs.account.Provisioning;
 import com.zimbra.cs.volume.Volume;
 import com.zimbra.cs.volume.VolumeManager;
+import com.zimbra.soap.admin.type.VolumeInfo;
+import com.zimbra.util.ExternalVolumeInfoHandler;
 
 public final class ModifyVolume extends RedoableOp {
 
@@ -40,6 +43,8 @@ public final class ModifyVolume extends RedoableOp {
 
     private boolean compressBlobs;
     private long compressionThreshold;
+
+    private short storeType;
 
     public ModifyVolume() {
         super(MailboxOperation.ModifyVolume);
@@ -59,6 +64,8 @@ public final class ModifyVolume extends RedoableOp {
 
         compressBlobs = volume.isCompressBlobs();
         compressionThreshold = volume.getCompressionThreshold();
+
+        storeType = (short)(volume.getStoreType().getStoreType());
     }
 
     @Override
@@ -67,6 +74,7 @@ public final class ModifyVolume extends RedoableOp {
                 .add("mboxGroupBits", mboxGroupBits).add("mboxBits", mboxBits)
                 .add("fileGroupBits", fileGroupBits).add("fileBits", fileBits)
                 .add("compressBlobs", compressBlobs).add("compressionThrehold", compressionThreshold)
+                .add("storeType", storeType)
                 .toString();
     }
 
@@ -80,6 +88,7 @@ public final class ModifyVolume extends RedoableOp {
         out.writeShort(mboxBits);
         out.writeShort(fileGroupBits);
         out.writeShort(fileBits);
+        out.writeShort(storeType);
     }
 
     @Override
@@ -92,6 +101,7 @@ public final class ModifyVolume extends RedoableOp {
         mboxBits = in.readShort();
         fileGroupBits = in.readShort();
         fileBits = in.readShort();
+        storeType = in.readShort();
     }
 
     @Override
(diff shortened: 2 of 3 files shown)


store/src/java/com/zimbra/cs/redolog/op/ModifyVolume.java
```diff
@@ -41,6 +44,8 @@ public final class ModifyVolume extends RedoableOp {
     private boolean compressBlobs;
     private long compressionThreshold;
 
+    private short storeType;
```
RakeshAMore, line 47: I think in modify request we are not allowing users to modify its storeType, if we do so we need to fix that.

store/src/java/com/zimbra/cs/redolog/op/ModifyVolume.java
```diff
@@ -102,6 +112,9 @@ public void redo() throws Exception {
                 .setMboxGroupBits(mboxGroupBits).setMboxBit(mboxBits)
                 .setFileGroupBits(fileGroupBits).setFileBits(fileBits)
                 .setCompressBlobs(compressBlobs).setCompressionThreshold(compressionThreshold).build();
+        VolumeInfo volInfo = vol.toJAXB();;
+        ExternalVolumeInfoHandler extVolInfoHandler = new ExternalVolumeInfoHandler(Provisioning.getInstance());
```
RakeshAMore, line 116: Check for this as well, do we need to override it here?

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -24,13 +34,25 @@
...
+
     /**
      * LDAP Provision instance
      */
     private Provisioning provisioning;
 
     public ExternalVolumeInfoHandler(Provisioning provisioning) {
         this.provisioning = provisioning;
+        try {
+            writeServerJSONFile(provisioning.getLocalServer().getServerExternalStoreConfig(), DIR_PATH, PREV_FILE_PATH);
```
RakeshAMore, line 51: why we are overriding files every time? consider a case this class can be used from multiple places and contracture is getting called from multiple places.

store/src/java/com/zimbra/cs/redolog/op/ModifyVolume.java
inpati: We are calling modifyServerProperties() function line "117". This function will take care for modifying only name when storeType is 2. 
This seems okay.

store/src/java/com/zimbra/cs/redolog/op/ModifyVolume.java
inpati: We dont need to overwite JSON backup files (prev, curr versions) as the modifyServerProperties function is taking care for that,
 Infact the architecture is such that where-ever we are setting LDAP attributes we are taking backup before and after as prev and current version respectively

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
inpati: This is done in-order to handle one corner case --> Assume our build is fresh build and its doesnt have any directory like "/opt/zimbra/smconfig/*", so whenever we are doing this operation it creates new files strictly if not existing.
This will not harm any data if the files are already present

RakeshAMore: There are two different things.
1. preparing directories
2. saving files.
separate these things, don't add confusing names and over responsibilities to a single method.
and as discussed earlier directory path should be `/opt/zimbra/config/sm/`

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -238,4 +300,58 @@ public Boolean isVolumePresentInJson(int volumeId) throws ServiceException, JSON
...
+        String result = null;
+        try {
+            File serverJSONBackupFile = new File(filePath);
+            // check if file exists
+            if (serverJSONBackupFile.exists()) {
+                BufferedReader reader = new BufferedReader(new FileReader(filePath));
+                String line = null;
+                while ((line = reader.readLine()) != null) {
+                    if (line.length() > 0) {
+                        result = line;
```
RakeshAMore, line 346: Do you want to read a single line only ? or am I missing something?

inpati: Yes I want to read single line only, as we don't have **'\n'** in json when we read single line only, all content/data gets read.
I have tested this

RakeshAMore: Don't do this like, read the whole file. you never know

inpati: Okay changed this logic.

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
inpati: Directory path changes to /opt/zimbra/config/sm

Removed preparing directories logic from constructor this is not needed now as I removed setExternalParam function. External directories and params will get initialized only in redo operations

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -186,7 +246,9 @@ public void modifyServerProperties(VolumeInfo volInfo) throws ServiceException,
             currentJsonObject.put("server/stores", currentJsonArray);
 
             // step 6: Set ldap attribute
+            writeServerJSONFile(provisioning.getLocalServer().getServerExternalStoreConfig(), DIR_PATH, PREV_FILE_PATH);
             provisioning.getLocalServer().setServerExternalStoreConfig(currentJsonObject.toString());
+            writeServerJSONFile(provisioning.getLocalServer().getServerExternalStoreConfig(), DIR_PATH, CURR_FILE_PATH);
```
vishnukantjangid, line 251: We can write currentJsonObject directly instead of getting data from provisioning.getLocalServer().getServerExternalStoreConfig()

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -146,7 +204,9 @@ public void addServerProperties(VolumeInfo volInfo) throws ServiceException, JSO
...
```
vishnukantjangid, line 209: We can write currentJsonObject directly instead of getting data from provisioning.getLocalServer().getServerExternalStoreConfig()

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -24,13 +34,25 @@
  */
 public class ExternalVolumeInfoHandler {
 
+    public static final String DIR_PATH = File.separator + "opt" + File.separator + "zimbra" + File.separator + "smconfig";
```
vishnukantjangid, line 37: Reduce the scope of constants if these are used in the same class only.

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -67,6 +83,41 @@ public JSONObject readServerProperties(int volumeId) throws ServiceException, JS
...
+     * @param filePath
+     *
+     * @return JSONObject as a backup server level volume properties
+     * @throws JSONException, ServiceException
+     */
+    public JSONObject readBackupServerProperties(int volumeId, String filePath) throws ServiceException, JSONException {
+        JSONObject retJsonObj = new JSONObject();
+        try {
+            // step 1: Fetch current JSON state object and current JSON state array
+            String serverExternalStoreConfigJson = readServerJSONFile(filePath);
```
vishnukantjangid, line 98: serverExternalStoreConfigJson could be null

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
inpati: ok. Fixed

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
inpati: ok. Fixed

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -67,6 +89,41 @@ public JSONObject readServerProperties(int volumeId) throws ServiceException, JS
...
+            JSONObject currentJsonObject = new JSONObject(serverExternalStoreConfigJson);
+            JSONArray currentJsonArray = currentJsonObject.getJSONArray("server/stores");
```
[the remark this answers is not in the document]
inpati: Done

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
inpati: Initialized to empty in-order to avoid null pointer exception

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
inpati: Done

store/src/java/com/zimbra/cs/redolog/op/CreateVolume.java
```diff
@@ -134,4 +147,29 @@ public void redo() throws Exception {
         }
     }
 
+    private VolumeExternalInfo buildVolumeExternalInfo(Volume volume) {
+        VolumeExternalInfo volExtInfo = null;
```
RakeshAMore, line 151: This should not be part of this `...redo.CreateVolume` class, right? move this respective file like ExternalVolumeInfoHandler ..

inpati: Done.

RakeshAMore (approved):

zimsuchitgupta (approved):

PR #1365, opened 2022-08-11
ZCS-11861 SOAP API: Need soap API to get the list of Available StoreManagers for the volume

inpati:


inpati:
@RakeshAMore Lets have discussion on this before will raise final PR

PR #1368, opened 2022-08-16
ZCS-11780: Set default value false for zimbraFeatureFileTypeUploadRes…

zimsuchitgupta:
Set default value FALSE for zimbraFeatureFileTypeUploadRestrictionsEnabled

As well as reverted back the implementation done in ZBUG-1459 if value of zimbraFeatureFileTypeUploadRestrictionsEnabled is FALSE. 
PR for ZBUG-1459 was https://github.com/Zimbra/zm-mailbox/pull/1060

### DIFF
--- a/store/src/java/com/zimbra/cs/service/FileUploadServlet.java
+++ b/store/src/java/com/zimbra/cs/service/FileUploadServlet.java
@@ -58,6 +58,7 @@ import org.apache.http.HttpStatus;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.impl.client.HttpClientBuilder;
+import org.apache.tika.Tika;
 import org.apache.tika.config.TikaConfig;
 import org.apache.tika.detect.CompositeDetector;
 import org.apache.tika.detect.Detector;
@@ -77,6 +78,7 @@ import com.zimbra.common.localconfig.LC;
 import com.zimbra.common.mime.ContentDisposition;
 import com.zimbra.common.mime.ContentType;
 import com.zimbra.common.mime.MimeConstants;
+import com.zimbra.common.mime.MimeDetect;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.service.ServiceException.Argument;
 import com.zimbra.common.service.ServiceException.InternalArgument;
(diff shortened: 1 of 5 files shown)


vishnukantjangid (approved):

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -196,6 +195,44 @@ public static final class Upload {
...
+                }
+
+                // 2. detect by file extension
+                if (contentType == null) {
+                    contentType = MimeDetect.getMimeDetect().detect(name);
+                }
+
+                // 3. special-case text/xml to avoid detection
+                if (contentType == null && file.getContentType() != null) {
+                    if (file.getContentType().equals("text/xml"))
```
RakeshAMore, line 219: can use constant here.

RakeshAMore, line 219: Can use constant for `text/xml`

RakeshAMore (approved):

silentsakky (approved):

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
shrutig0510: We have a const defined in MimeConstants.java as MimeConstants.CT_TEXT_XML.
Plz check if it works for you

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -766,6 +803,38 @@ List<Upload> handlePlainUpload(HttpServletRequest req, HttpServletResponse resp,
...
+                blacklist.addAll(Arrays.asList(contentTypeBlacklist.trim().split(",")));
+            }
+            if (blacklist.stream().anyMatch((blacklistedContentType) -> {
+                Pattern p = Pattern.compile(blacklistedContentType);
+                Matcher m = p.matcher(finalMimeType);
+                return m.find();
+            })) {
+                mLog.debug("handlePlainUpload(): deleting %s", fi);
+                fi.delete();
+                mLog.info("File content type is blacklisted : %s" + finalMimeType);
```
shrutig0510, line 832: Replace + with ,

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
zimsuchitgupta: It's a revert and hence we should not fix review comments, this code is anyways going to be removed again when we implement this feature properly.

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
zimsuchitgupta: it's a revert and hence we should not fix review comments, this code is anyways going to be removed again when we implement this feature properly

shrutig0510 (approved):

dasiyogesh (approved):

PR #1371, opened 2022-08-17
ZBUG-2978:In latest Patch mails are showing with code in the mail con…

gabhinavTI:
Issue: [ZBUG-2978](https://jira.corp.synacor.com/browse/ZBUG-2978)
In latest Patch mails are showing with code in the mail content
Post code fix on unbalanced tags, there is a scenario where ending tag wasn't correctly present.
Hence this is fix for the particular scenario where the tag ends properly and the code is inside comment 
hence we don't see the particular code in emails now

Testing:
Inject mime using CLI and verify the email, the code is not seen post code change

### DIFF
--- a/store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
+++ b/store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
@@ -57,6 +57,7 @@ public class OwaspHtmlSanitizer implements Callable<String> {
         ArrayList<String> tags = new ArrayList<String>();
         // in tags array we can add multiple tags, so for each tag write start &
         // end with a & separator like(<!--&-->)
+        html = html.replaceAll("<!-->","-->");
         tags.add("<!--&-->");
         String formattedHtml = "";
         for (String str : tags) {


store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
```diff
@@ -57,6 +57,7 @@ private String checkUnbalancedTags(String html) {
         ArrayList<String> tags = new ArrayList<String>();
         // in tags array we can add multiple tags, so for each tag write start &
         // end with a & separator like(<!--&-->)
+        html = html.replaceAll("<!-->","-->");
```
silentsakky, line 60: we are doing replaceAll on entire html content of mail
- what would be performance implication of this change?
- if user has used <!--> inside mail as content, then what would happen with that data?

gabhinavTI: @silentsakky I do agree on the performance impact, but if we see below there're many string operations going on
Ideally, as I understood this issue should be fixed from the OWASP end, but since it's taking a long time for resolution from their end, hence this is just a workaround so that we don't get an issue until issue is fixed from OWASP

silentsakky: can you point me another place where we are doing string manipulation on html variable
Whole point of using OWASP was to make sure we don't do string/regex manipulations on html content in performance efficient way, so we are breaking that rule

gabhinavTI: OWASP library is supposed to do all this stuff, but for this particular scenario there seems to be a bug on their end which hasn't been resolved in some time and hence when Rupesh was working on 2547 he came up with this temporary solution
 This entire method is a temporary fix for unbalanced tags issue, which supposedly had to be handle by oWASP library
 @rupeshtelus can add further details of this method and ticket related details to OWASP if required

gabhinavTI:
Closing this PR ,
Fix below using PR
https://github.com/Zimbra/zm-mailbox/pull/1404 and
 https://github.com/Zimbra/java-html-sanitizer-release-20190610.1/pull/9

PR #1372, opened 2022-08-17
ZBUG-2914: Proxy servlet SSRF

ronstra-synacor:
**Problem**: https://jira.corp.synacor.com/browse/ZBUG-2914 describes this as a potential issue, but gives no indication of an actual problem in the real world.  So, going by the details and proposed fixes in that description, we want to prevent a proxied request from following any redirect that isn't also on our whitelist.

**Solution**: First, I wrote a custom redirect strategy that checked our whitelist. That should stop us from being redirected while the proxying request is running. Then I needed to re-check any ultimate "Location" header sent to report the proper client error  instead of allowing that header to fall through and (probably) fail due to existing browser CORS restrictions.

**Testing**: I grabbed a small URL-shortening go project to generate redirects and ensured our proxy did not follow them unless they matched a domain within our whitelist.

### DIFF
--- a/store/src/java/com/zimbra/cs/zimlet/ProxyServlet.java
+++ b/store/src/java/com/zimbra/cs/zimlet/ProxyServlet.java
@@ -88,7 +88,7 @@ public class ProxyServlet extends ZimbraServlet {
     private static final String AUTH_PARAM = "auth";
     private static final String AUTH_BASIC = "basic";
 
-    private static boolean isRedirectStatus(int status) {
+    protected static boolean isRedirectStatus(int status) {
         return status / 100 == 3; // 3xx Redirect
     }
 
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/zimlet/ProxyServlet.java
```diff
@@ -83,8 +88,11 @@ public class ProxyServlet extends ZimbraServlet {
     private static final String AUTH_PARAM = "auth";
     private static final String AUTH_BASIC = "basic";
 
+    private static boolean isRedirectStatus(int status) {
+        return status / 100 == 3; // 3xx Redirect
```
silentsakky, line 92: in case of status 301
301/100 = 3.01
so 3.01 is not equal to 3, so in that case are we going to return false?

ronstra-synacor: It'd be a problem if either was a float, but because they're both integers, `301/100 == 3`. Added some tests to prove the point.

silentsakky: aah good point, completely forgot I am reviewing java which is strict typed language unlike javascript :)

silentsakky (approved):

deepakzimb (approved):

dasiyogesh (approved):

PR #1378, opened 2022-08-20
ZCS-11687 Remove zulip feature

silentsakky:
- Remove ldap attributes also as in future this feature should be released in major/minor version
- Haven't adjusted attribute ids of other attributes as that is a big change


store/src/java/com/zimbra/cs/util/AccountUtil.java
```diff
@@ -906,7 +906,8 @@ public static Account getRequestedAccount(String identityId, Account authAcct) t
         return null;
     }
 
-    public static boolean isZulipChatEnabled(Account account) {
+    // @TODO this is bad way of doing this, we could have used domainInherited flag on ldap attribute zimbraFeatureZulipChatEnabled
```
RakeshAMore, line 909: Since we are not using this, we can remove commented code. And whenever we need it we can always get it from history :)

silentsakky: Sure removed code as well as ldap attributes, next minor version upgrade can reintroduce ldap attributes

deepakzimb (approved):

dasiyogesh (approved):

zimsuchitgupta (approved):

PR #1393, opened 2022-08-29
ZCS-12002:failed test case of testBug64974

rupeshtelus:
issue - https://jira.corp.synacor.com/browse/ZCS-12002
issue - failed test case of testBug64974

solution - if the html doesn't contain any start tag ( exp - '<!--' ) then it doesn't need to do operations of splitting & replacing tags , so added if conditions above it , if the length will be > 1 then only it will do all operations


store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
```diff
@@ -64,15 +64,19 @@ private String checkUnbalancedTags(String html) {
...
             String end = tagArr[1];
             String[] arrOfStr = html.split(start);
-            for (String strAfterSplit : arrOfStr) {
-                 if (!strAfterSplit.equals("") && strAfterSplit.contains(end)) {
-                     formattedHtml = formattedHtml + start + strAfterSplit;
-                 } else {
-                     formattedHtml = formattedHtml + strAfterSplit;
-                 }
+            //ZCS-12002 ( failed test case for testBug64974 )
+            //if the html doesn't containing start then skipping other operations
```
dasiyogesh, line 68: add space after //

store/src/java/com/zimbra/cs/html/owasp/OwaspHtmlSanitizer.java
```diff
@@ -64,15 +64,19 @@ private String checkUnbalancedTags(String html) {
...
+            if (arrOfStr.length > 1) {
+                for (String strAfterSplit : arrOfStr) {
+                    if (!strAfterSplit.equals("") && strAfterSplit.contains(end)) {
```
jeastman, line 71: For safety reasons (and avoiding a `NullPoitnerException`), you should compare the literal string to the variable.

`!strAfterSpit.equals("")`

should be:

`!"".equals(strAfterSpit)`

silentsakky:
The original fix of ZBUG-2547 has been reworked in https://github.com/Zimbra/zm-mailbox/commit/f82c5af7167f06c1aa8e3d247157a32124ebd496
We will have to do rework on this PR also

PR #1397, opened 2022-08-30
ZCS-12123: /opt/zimbra/bin/zmvolume --turnOffSecondary failing on 10.…

vishnukantjangid:
**Problem:**  /opt/zimbra/bin/zmvolume --turnOffSecondary failing on 10.0 build
**Solution:** Added condition for volume id none

### DIFF
--- a/store/src/java/com/zimbra/cs/service/admin/SetCurrentVolume.java
+++ b/store/src/java/com/zimbra/cs/service/admin/SetCurrentVolume.java
@@ -19,6 +19,7 @@ package com.zimbra.cs.service.admin;
 
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.soap.Element;
+import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.cs.account.Provisioning;
 import com.zimbra.cs.account.accesscontrol.AdminRight;
 import com.zimbra.cs.account.accesscontrol.Rights.Admin;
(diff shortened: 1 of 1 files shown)


zimsuchitgupta (approved):

store/src/java/com/zimbra/cs/service/admin/SetCurrentVolume.java
```diff
@@ -49,12 +50,16 @@ private SetCurrentVolumeResponse handle(SetCurrentVolumeRequest req, Map<String,
...
-        if (Volume.TYPE_MESSAGE == volume.getType()) {
-            // set current store manager
-            StoreManagerRuntimeSwitchResult runtimeSwitchResult = StoreManagerResetHelper.setNewStoreManager(volume.getStoreManagerClass());
-            response.setRuntimeSwitchResult(runtimeSwitchResult);
+        if (volId != Volume.ID_NONE) {
+            Volume volume = VolumeManager.getInstance().getVolume(volId);
+            // if its primary volume then
+            if (Volume.TYPE_MESSAGE == volume.getType()) {
+                // set current store manager
+                StoreManagerRuntimeSwitchResult runtimeSwitchResult = StoreManagerResetHelper.setNewStoreManager(volume.getStoreManagerClass());
```
dasiyogesh, line 58: as there is no use of runtimeSwitchResult further 
`response.setRuntimeSwitchResult(runtimeSwitchResult);` can be changed to 
`response.setRuntimeSwitchResult(StoreManagerResetHelper.setNewStoreManager(volume.getStoreManagerClass()));`

store/src/java/com/zimbra/cs/service/admin/SetCurrentVolume.java
```diff
@@ -49,12 +50,16 @@ private SetCurrentVolumeResponse handle(SetCurrentVolumeRequest req, Map<String,
...
-        Volume volume = VolumeManager.getInstance().getVolume(volId);
-        // if its primary volume then
-        if (Volume.TYPE_MESSAGE == volume.getType()) {
-            // set current store manager
-            StoreManagerRuntimeSwitchResult runtimeSwitchResult = StoreManagerResetHelper.setNewStoreManager(volume.getStoreManagerClass());
-            response.setRuntimeSwitchResult(runtimeSwitchResult);
+        if (volId != Volume.ID_NONE) {
+            Volume volume = VolumeManager.getInstance().getVolume(volId);
+            // if its primary volume then
+            if (Volume.TYPE_MESSAGE == volume.getType()) {
```
dasiyogesh, line 56: what happens if this is true?

store/src/java/com/zimbra/cs/service/admin/SetCurrentVolume.java
vishnukantjangid: Yes, this statement is too long so just to make it more readable we keep it this way.

store/src/java/com/zimbra/cs/service/admin/SetCurrentVolume.java
vishnukantjangid: There is no need to perform any action here for secondary volume after this condition. This is existing behavior so we didn't make any changes.

RakeshAMore (approved):

PR #1408, opened 2022-10-07
ZCS-12278: HSM, modifying compress threshold causing compress disable

vishnukantjangid:
**Issue:** HSM, modifying compress threshold causing compress disable
**Solution:** Added the validation for compress and compressThreshold. 

**Validations:** 

1. If compress is disabled we can't modify the value of compressThreshold.
2. If compress is enabled then we can modify the value of compressThreshold and it persists the old value of compress.


RakeshAMore (approved):
Looks good to me, just make sure we are covering all cases in testing.

zimsuchitgupta (approved):

deepakzimb (approved):

PR #1411, opened 2022-10-10
ZCS-11957 : Adding getter method for emailAddress and timeStamp

abhishekdeshmukh2209:
Issue :- All users configured in devices are not shown on ABQ MDM UI in admin console. Only 1 node users are shown on ABQ MDM UI in admin console. For other node configured users, we need to search that user and then we can see.

Solution:-
Making multi server call and getting data from each server.
Collating limited data from each server to display on UI.
Added getter missing getter methods in DeviceStatusInfo

### DIFF
--- a/common/src/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/java/com/zimbra/common/localconfig/LC.java
@@ -17,18 +17,17 @@
 
 package com.zimbra.common.localconfig;
 
+import com.google.common.base.Strings;
+import com.zimbra.common.util.Constants;
+import com.zimbra.common.util.ZimbraLog;
+import org.dom4j.DocumentException;
+
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
-
-import org.dom4j.DocumentException;
-
-import com.google.common.base.Strings;
-import com.zimbra.common.util.Constants;
-import com.zimbra.common.util.ZimbraLog;
 
 /**
  * Provides convenient means to get at local configuration - stuff that we do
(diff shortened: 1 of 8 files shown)


zimsuchitgupta (approved):

dasiyogesh:
@abhishekdeshmukh2209 ticket number seems to have type in commit message

common/src/java/com/zimbra/common/localconfig/LC.java
```diff
@@ -978,6 +977,8 @@ public final class LC {
     public static final KnownKey httpclient_soaphttptransport_retry_count = KnownKey.newKey(2);
     public static final KnownKey httpclient_soaphttptransport_so_timeout = KnownKey.newKey(300 * Constants.MILLIS_PER_SECOND);
     public static final KnownKey httpclient_soaphttptransport_keepalive_connections = KnownKey.newKey(true);
+    public static final KnownKey httpclient_internal_connmgr_mailbox_timeout = KnownKey.newKey(3 * Constants.MILLIS_PER_MINUTE);
```
dasiyogesh, line 980: should this be 3 secs?

abhishekdeshmukh2209: done

soap/src/java/com/zimbra/soap/admin/message/GetDeviceStatusRequest.java
```diff
@@ -54,6 +54,13 @@ public class GetDeviceStatusRequest {
     @XmlAttribute(name=SyncConstants.A_LIMIT /* limit */, required=false)
     private int limit = 0;
 
+    /**
+     * @zm-api-field-tag notificationEmail
+     * @zm-api-field-description Indicates if the request is coming from notification email that needs to be sent for all matching conditions across mailbox servers
+     */
+    @XmlAttribute(name=SyncConstants.A_NOTIFICATION_EMAIL /* notificationEmail */, required=false)
```
dasiyogesh, line 61: Attribute name `A_NOTIFICATION_EMAIL` is specific to email notification sent for quarantine devices and this API GetDeviceStatusRequest is only specific to device list.
Considering the need of the API to pull all records or with limit, may be can rename to GET_ALL with true Or USE_LIMIT  false to pull all records.

dasiyogesh (approved):

soap/src/java/com/zimbra/soap/admin/message/GetDeviceStatusRequest.java
abhishekdeshmukh2209: done

PR #1422, opened 2022-11-24
ZBUG-3129 - adding LC config for ClamAV timeout and data chunk size

RakeshAMore:
Adding LC config for ClamAV timeout and data chunk size

### DIFF
--- a/common/src/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/java/com/zimbra/common/localconfig/LC.java
@@ -408,6 +408,11 @@ public final class LC {
     @Supported
     public static final KnownKey ldap_starttls_required = KnownKey.newKey(true);
 
+    @Reloadable
+    public static final KnownKey clamav_socket_timeout = KnownKey.newKey(2000);
+    @Reloadable
+    public static final KnownKey clamav_scan_data_chunk_size = KnownKey.newKey(2048);
+
     @Supported
     public static final KnownKey zimbra_directory_max_search_result = KnownKey.newKey(5000);
 


common/src/java/com/zimbra/common/localconfig/LC.java
```diff
@@ -408,6 +408,11 @@ public final class LC {
     @Supported
     public static final KnownKey ldap_starttls_required = KnownKey.newKey(true);
 
+    @Reloadable
+    public static final KnownKey clamav_socket_timeout = KnownKey.newKey(2000);
```
dasiyogesh, line 412: the timeout & size used as default, why are these values selected?

RakeshAMore: I did not get your question

dasiyogesh: just trying to understand how these were selected as default values

deepakzimb (approved):

common/src/java/com/zimbra/common/localconfig/LC.java
dasiyogesh: why @Reloadable?

preshitaagrawal (approved):

common/src/java/com/zimbra/common/localconfig/LC.java
RakeshAMore: OK.got it. No specific reason but tried to keep it as ideal as possible. This can vary from setup to setup, for this reason, we introduced LC attributes so they can be controlled.

zimsuchitgupta: I think Redoable is OK if server restart is not required.
`If it's reloadable, changes are in effect
     * after LC reload. Otherwise, changes are in effect after server restart.`

abhishekdeshmukh2209 (approved):

zimsuchitgupta (approved):
LGTM

dasiyogesh (approved):

common/src/java/com/zimbra/common/localconfig/LC.java
```diff
@@ -408,6 +408,11 @@ public final class LC {
...
+    @Reloadable
+    public static final KnownKey clamav_scan_data_chunk_size = KnownKey.newKey(2048);
```
silentsakky, line 414: extensions can define their own LC so why these are added in mailbox and not https://github.com/Zimbra/zm-clam-scanner-store ?
I think we are increasing tight coupling between mailbox and extensions

RakeshAMore: I tried this in another extension, but it does not work and I ended up moving config into zm-mailbox. need to investigate this more.

PR #1425, opened 2022-12-02
ZBUG-3064: X-Zimbra-Forwarded is not added by SieveFilter rule due to…

vishnukantjangid:
**Issue:** X-Zimbra-Forwarded is not added by SieveFilter rule due to ParseException

**Solution:** If we are getting mail ParseException during saving changes and X-Zimbra-Forwarded header is not present in the mail header then we are adding X-Zimbra-Forwarded in the mail to detect the mail loop.

### DIFF
--- a/store/src/java/com/zimbra/cs/filter/FilterUtil.java
+++ b/store/src/java/com/zimbra/cs/filter/FilterUtil.java
@@ -19,6 +19,8 @@ package com.zimbra.cs.filter;
 import static com.zimbra.cs.filter.JsieveConfigMapHandler.CAPABILITY_VARIABLES;
 
 import java.io.IOException;
+import java.nio.charset.IllegalCharsetNameException;
+import java.nio.charset.UnsupportedCharsetException;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Date;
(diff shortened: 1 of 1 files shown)


zimsuchitgupta (approved):
LGTM

store/src/java/com/zimbra/cs/filter/FilterUtil.java
```diff
@@ -290,13 +290,23 @@ public static void redirect(OperationContext octxt, Mailbox sourceMbox, MimeMess
             try {
                 outgoingMsg = createRedirectMsgOnError(msg);
                 ZimbraLog.filter.info("Message format error detected.  Wrapper class in use.  %s", e.toString());
+                // ZBUG-3064: If MimeMessage.saveChanges fails, updating header in copy of the message
```
shrutig0510, line 293: looks like redundant code. Create separate bock and use it at both places

vishnukantjangid: done

yyoshiho (commented):
This fixes mail loop issue, but redirected message still has wrong headers because they are copied from original message. (This test case is wrong Content-Type with charset.) 
Should we follow to rfc2045? to avoid some mail problems with wrong headers for other messaging systems for associated addresses.

yyoshiho (approved):
This particular invalid Content-Type case is resolved by this fix.

PR #1431, opened 2022-12-23
ZCS-12704: adding mechanism to notify extension

k-kato:
**Enhancement:**
* Add an ability to insert a process into an existing code using extension framework to achieve more flexible customization.
* Add `ZimbraExtensionNotification.notifyExtension` to `Auth.java` (both account and admin) so that a custom validation can be added.

**Approach:**
* add a mechanism to call a extension handler with a notification ID and necessary parameters.
* an extension registers a handler for a notification.
  For example,
    * `handler1` is registered for `notif-id-A`
    * when a process (`process1`) calls `ZimbraExtensionNotification#notifyExtension("notif-ID-A", param1, param2, ..., paramN)`, `handler1#execute` is called. All parameters are passed to `handler1#execute` as `Object[]`.
    * In `handler1#execute` process, parameter(s) need to be casted to right class(es). After that, any process can be executed.
    * After the process of `handler1#execute` is completed, `process1` is resumed.
* If no handler has been registered for a specified notification ID, nothing happens.
* Only a single handler can be registered for each notification ID.
* It is responsible for an extension to cast parameters to right classes.
* It is responsible for an extension to make thread-safe process.
* `notifyExtension` is not Asynchronous; wait for completion of handler's process
* The approach is similar to `appCtxt.notifyZimlets` on Classic UI.

**Sample extension to handle a notification:**
* https://github.com/Zimbra/zm-notification-handler-ext-example/pull/1

### DIFF
--- a/store/src/java/com/zimbra/cs/extension/ZimbraExtensionNotification.java
+++ b/store/src/java/com/zimbra/cs/extension/ZimbraExtensionNotification.java
@@ -1,0 +1,78 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2022 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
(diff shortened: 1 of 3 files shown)


ronstra-synacor (approved):
Might it be useful to make the handler registry values a Set to allow more than one handler per hook?

k-kato:
I think allowing only a single handler is better for the following reasons:
* administrator cannot control the order of handlers (`notifyExtension` calls) per notification id. The order of loading extensions at mailboxd start is uncontrollable.
* a handler might affect aother handler for the same notification id since the same object(s) is passed as a parameter and there is a case that the object is not primitive but reference type. Chaining of calling handlers should be avoided for secure.
i.e. all necessary process should be implemented in a single handler rather than chaining handlers.

vishnukantjangid (approved):

deepakzimb (approved):

dasiyogesh (approved):

PR #1436, opened 2023-01-12
ZBUG-2722 : Install zimlet config if version updated in zimlet config

umagmrit:
**Problem:**

`zmzimletctl deploy zimlet.zip` overwrites all previous zimlet configuration stored in LDAP, back to their defaults.

For example, we had a config change in com_zimbar_url to disable youtubePreview and  this config change had been reverted by deploying the patch.

**Fix:**
Install zimlet config if version updated in latest zimlet config 

Compare latest zimlet config version with zimlet config version stored in LDAP. If latest zimlet config version is higher than the zimlet config version stored in LDAP then install zimlet config, which will override customer's changes and it will loose customer's changes. Customer will take updated zimlet config and apply their customizations and deploy using zmzimletctl.

### DIFF
--- a/store/src/java/com/zimbra/cs/zimlet/ZimletUtil.java
+++ b/store/src/java/com/zimbra/cs/zimlet/ZimletUtil.java
@@ -673,7 +673,20 @@ public class ZimletUtil {
 
         // install the config
         if (zf.hasZimletConfig()) {
-            installConfig(zf.getZimletConfig());
+            ZimletConfig zc = getZimletConfig(zimletName);
+            if (zc != null) {
+                Version zimletConfigVersion = new Version (zf.getZimletConfig().getVersion().toString());
+                Version zimletConfigVersionInLDAP = new Version (zc.getVersion().toString());
+                if (zimletConfigVersion.compareTo(zimletConfigVersionInLDAP)> 0) {
+                    installConfig(zf.getZimletConfig());
+                }
+                if (zimletConfigVersion.compareTo(zimletConfigVersionInLDAP) == 0) {
+                    return;
+                }
+            }
+            else {
+                installConfig(zf.getZimletConfig());
+            }
         }
 
         // activate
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/zimlet/ZimletUtil.java
```diff
@@ -673,7 +673,20 @@ public static void deployZimletLocally(ZimletFile zf) throws IOException, Zimlet
 
         // install the config
         if (zf.hasZimletConfig()) {
-            installConfig(zf.getZimletConfig());
+            ZimletConfig zc = getZimletConfig(zimletName);
+            if (zc != null) {
+                Version zimletConfigVersion = new Version (zf.getZimletConfig().getVersion().toString());
```
silentsakky, line 678: should we add try/catch to handle malformed versions?

store/src/java/com/zimbra/cs/zimlet/ZimletUtil.java
```diff
@@ -673,7 +673,20 @@ public static void deployZimletLocally(ZimletFile zf) throws IOException, Zimlet
...
+                Version zimletConfigVersionInLDAP = new Version (zc.getVersion().toString());
+                if (zimletConfigVersion.compareTo(zimletConfigVersionInLDAP)> 0) {
+                    installConfig(zf.getZimletConfig());
+                }
+                if (zimletConfigVersion.compareTo(zimletConfigVersionInLDAP) == 0) {
+                    return;
```
silentsakky, line 684: isn't this going to skip rest of the processing? don't we need to just skip calling installConfig?

umagmrit: yes this will skip rest of the processing, updated the code for same.

zimsuchitgupta (approved):

store/src/java/com/zimbra/cs/zimlet/ZimletUtil.java
zimsuchitgupta: Here we are using version data which is returned by ZimletConfig. Before creating ZimletConfig object all data gets validated. So if any exception comes then we'll get null object.

silentsakky (approved):

deepakzimb (approved):

PR #1440, opened 2023-01-18
ZCS-12826: added a logic to set custom X-Mailer header

k-kato:
Adding `ZimbraExtensionNotification.notifyExtension` to replace X-Mailer header using an extension.

An extension cannot return any value. Then a reference of string array `String[] customXMailer` is passed to an extension. The extension can set a preferred string for X-Mailer header.
If no extension is called or no value is set by an extension, default string is used.


ronstra-synacor (approved):

zimsuchitgupta (approved):

dasiyogesh (approved):

k-kato:
rebased

JayaG (approved):

PR #1443, opened 2023-02-01
ZBUG-733 SearchDirectoryRequest takes a long time on large directory (ISP size)

shrutig0510:
**Problem:**
SearchDirectoryRequest takes a long time on large directory (ISP size).

**Requirements:**

limit parameter is not getting passed to the SearchDirectoryRequest but it should as per it's API document..
If the limit is specified as 0 or if the limit is not specified then it should return everything as per the API document and there should not be any pagination.
If the limit is non-zero and it's specified then it should handle the pagination.

**Analysis and Approach:**
SearchDirectoryRequest takes a long time on large directory which only happens for a large directory sites such as ISP who has +1M accounts because it always sends with paged search options like -E pr=1000 whether limit is specified as limit="0" or not. And as per SearchDirectoryRequest API document maxResults is the maximum results that the backend will attempt to fetch from the directory before returning a account.TOO_MANY_SEARCH_RESULTS error, and limit is the number of accounts to return per page (0 is default which means all). But the limit is fixed value in the code for SearchLdapOptions class is set to 1000.

**Test-cases:**
Test performed added to https://jira.corp.synacor.com/browse/ZBUG-733 as attachment ZBUG-733_test_result.txt

### DIFF
--- a/soap/src/java/com/zimbra/soap/admin/message/SearchDirectoryRequest.java
+++ b/soap/src/java/com/zimbra/soap/admin/message/SearchDirectoryRequest.java
@@ -53,6 +53,7 @@ public class SearchDirectoryRequest extends AttributeSelectorImpl {
 
     /**
      * @zm-api-field-description The maximum number of accounts to return (0 is default and means all)
+     * and it must be greater than or equal to 0.
      */
     @XmlAttribute(name=AdminConstants.A_LIMIT, required=false)
     private Integer limit;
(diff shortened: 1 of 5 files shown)


zimsuchitgupta (approved):

vishnukantjangid (approved):

yyoshiho (approved):
Understood to approve the backward compatibility for UI side (limit=-1).

PR #1451, opened 2023-02-22
ZCS-12877: flush cache on all servers after grant or revoke global right

dasiyogesh:
Issue
Once the global rights are granted the global cache does not gets updated on multi server environment

Fix
Trigger FlushCacheRequest on all servers for global after grant/revoke rights.


store/src/java/com/zimbra/cs/service/admin/RevokeRight.java
```diff
@@ -43,6 +48,17 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
         RightCommand.revokeRight(Provisioning.getInstance(), getAuthenticatedAccount(zsc), rrReq.getTarget(),
                                 rrReq.getGrantee(), rrReq.getRight().getValue(), rightModifier);
+        EffectiveRightsTargetSelector erTargSel = rrReq.getTarget();
+        if (com.zimbra.soap.type.TargetType.global == erTargSel.getType()) {
+            CacheSelector cacheSel = new CacheSelector(true /* allServers */, CacheEntryType.globalgrant.toString());
+            FlushCacheRequest fcReq = new FlushCacheRequest(cacheSel);
+            try {
+                FlushCache.doFlushCache(this, context, fcReq);
+            } catch (ServiceException se) {
+                ZimbraLog.acl.info("Problem flushing acl cache for global %s/%s after revoking rights",
```
rcyarrapothu, line 58: warning log would be a better option here

shrutig0510 (approved):

deepakzimb (approved):

preshitaagrawal (approved):

store/src/java/com/zimbra/cs/service/admin/RevokeRight.java
ronstra-synacor: It's also "info" level in the change further above, so should probably change both to be consistent, if changing.

ronstra-synacor (approved):

zimsuchitgupta (approved):

store/src/java/com/zimbra/cs/service/admin/RevokeRight.java
dasiyogesh: done

yyoshiho (approved):
Agreed to use flushCache -a

PR #1456, opened 2023-03-03
ZCS-12471- adding LDAP attribute for mail recall

shubhamCS03:
Requirement of ticket [[ZCS-12471](https://jira.corp.synacor.com/browse/ZCS-12471)]:

- Two new LDAP attributes to control the recall feature in Zimbra to be introduced for three levels  i.e. account, COS and domain  - zimbraFeatureMailRecallEnabled(default value is FALSE), zimbraFeatureMailRecallTime(default value is 30 minutes and time limit is from 1 minute to 30 minutes ).

Solution:

- Two new attributes are added and their setters/getters are generated. 

Testing:

- Testing is done on remote machine by installing custom build for all the three levels and getting expected results.

PR #1458, opened 2023-03-06
ZCS-12966 Ldap attribute for CountObjectsRequest on admin console changes

shrutig0510:
**Problem:**
When millions of accounts exist on the system and CountObjectsRequest with userAccount type is sent from Admin Console to a server, it causes high load on a ldap server.

**Fix:**
Intoduced Ldap attr to enable/disable the counting of userAccount in CountObjectsRequest

**Test-cases:**
1. when zimbraCountObjectsEnabled is TRUE
send CountObjectsRequest with account or userAccount type.
show number of accounts in Home -> Summary -> Accounts and Manage -> left pane -> the number of accounts
2. when zimbraCountObjectsEnabled is FALSE
not to send CountObjectsRequest with account or userAccount type.
show N/A in Home -> Summary -> Accounts and Manage -> left pane -> the number of accounts

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -10352,4 +10352,9 @@ TODO: delete them permanently from here
     indexing         - Generate index
   </desc>
 </attr>
+
+<attr id="4096" name="zimbraCountAccountsEnabled" type="boolean" cardinality="single" optionalIn="globalConfig" since="11.0.0">
+  <globalConfigValue>TRUE</globalConfigValue>
+  <desc>whether to enable CountObjectsRequest with userAccount/account type on Zimbra Admin Console</desc>
+</attr>
 </attrs>
--- a/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
+++ b/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
@@ -5145,6 +5145,15 @@ public class ZAttrProvisioning {
     public static final String A_zimbraCOSInheritedAttr = "zimbraCOSInheritedAttr";
 
     /**
+     * whether to enable CountObjectsRequest with userAccount/account type on
+     * Zimbra Admin Console
+     *
+     * @since ZCS 11.0.0
+     */
+    @ZAttr(id=4096)
+    public static final String A_zimbraCountAccountsEnabled = "zimbraCountAccountsEnabled";
+
+    /**
      * time object was created
      *
      * @since ZCS 6.0.0_BETA1
--- a/store/src/java/com/zimbra/cs/account/ZAttrConfig.java
+++ b/store/src/java/com/zimbra/cs/account/ZAttrConfig.java
@@ -12697,6 +12697,83 @@ public abstract class ZAttrConfig extends Entry {
     }
 
     /**
+     * whether to enable CountObjectsRequest with userAccount/account type on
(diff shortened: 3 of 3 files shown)


store/conf/attrs/zimbra-attrs.xml
```diff
@@ -10104,4 +10104,9 @@ TODO: delete them permanently from here
     indexing         - Generate index
   </desc>
 </attr>
+
+<attr id="9030" name="zimbraCountObjectsEnabled" type="boolean" cardinality="single" optionalIn="globalConfig" since="11.0.0">
```
k-kato, line 10108: Is there a possibility to disable counting objects except accounts (cos, domain etc.) in the future?
If the scope is Account object only, `zimbraCountAccountsEnabled` may be better.

silentsakky, line 10108: can't we implement some type of caching mechanism on the server side instead of disabling the feature?

shrutig0510: Disabling through ldap attr was already discussed and part of ticket implementation. So, its implemented that way.

shrutig0510: I do not see it can be disabled for cos, domain etc.
Will make change for variale name as zimbraCountAccountsEnabled

k-kato: Please rename it to `zimbraCountAccountsEnabled`.
If we need to add ability to disables other object, we will add a separate attribute.

k-kato: The ability to disable the counting is necessary to avoid high load on ldap. Even if cache mechanism is introduced, the high load will happen at the first searching.
The cache mechanism should be implemented in a separate ticket as necessary. Having a cache might cause another issue like returning unexpected results due to old cache and/or increasing memory usage on a mailbox server.

shrutig0510: Done

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -10104,4 +10104,9 @@ TODO: delete them permanently from here
     indexing         - Generate index
   </desc>
 </attr>
+
+<attr id="9030" name="zimbraCountAccountsEnabled" type="boolean" cardinality="single" optionalIn="globalConfig" since="11.0.0">
```
k-kato, line 10108: @desouzas @dasiyogesh what id is good? 4095?
Related comment: https://github.com/Zimbra/zm-mailbox/pull/1456

store/conf/attrs/zimbra-attrs.xml
k-kato: thanks.

store/conf/attrs/zimbra-attrs.xml
shrutig0510: I have seen id upto 9029 in zimbrOS/zm-mailbox (https://github.com/ZimbraOS/zm-mailbox/blob/develop/store/conf/attrs/zimbra-attrs.xml) , So next i took.

shrutig0510: id changed to 4096. Thanks

k-kato (approved):

dasiyogesh (approved):

vishnukantjangid (approved):

PR #1468, opened 2023-03-23
ZCS-13176 : Add mail recall message verification header

shubhamCS03:
Requirement of ticket [[ZCS-13176](https://jira.corp.synacor.com/browse/ZCS-13176)]

Requirements:
-  To make sure we don't recall a message accidentally or spoofed, we should add a header to outgoing email.
-  We need to a header field, _Message-Verification_ , and verify the message-Id while recalling a mail.
-  Need to Hash data by using algorithm  SHA256 and further encrypt the data by using base64.

Solution:
- A new header field is added while sending mail,  _Message-Verification_ which contains hash(algorithm name) and guid(base64 of the guid's hash).
- The Message-Verification header field contains the following required tags, separated by semicolons (with no white space within):
hash=[string] --the algorithm used to produce the hash value in the "guid" tag. Current valid values are "SHA1" and "SHA256" [[SHA]].
guid=[base64] --the base64 encoding [[Base64]] of the hash of a globally unique ID (GUID) for the message, using the hash algorithm specified in the "hash" tag. The actual GUID (the pre-image of the hash) is kept secret by the sending side.
- We have created one LDAP attribute _secretKeyForMailRecall_ at global level to store secretKey.
- We have created one SOAP API to generate secretKey.
- We have generate the secretKey using this _SHA1PRNG_ and saved in LDAP attribute _secretKeyForMailRecall_ .
- GUID(globally unique ID) = Message-Id + Date + From + secretKey. 

Testing:
- Changes are tested  by sending an email and by checking its show original in the header field we are able to see _Message-Verification_ .
- Also tested by decrypting the hash by using base64 we get the same value.

[ZCS-13176]: https://synacor.atlassian.net/browse/ZCS-13176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
+++ b/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
@@ -7222,6 +7222,14 @@ public class ZAttrProvisioning {
     public static final String A_zimbraFeatureMailRecallEnabled = "zimbraFeatureMailRecallEnabled";
 
     /**
+     * Secret key used in Mail Recall to make it more secure from spoof.
+     *
+     * @since ZCS 10.1.5
+     */
+    @ZAttr(id=4136)
+    public static final String A_zimbraFeatureMailRecallSecretKey = "zimbraFeatureMailRecallSecretKey";
+
+    /**
      * Time(in minutes) within which a message can be recalled. The default
      * time is 30 minutes and accepts value from 1 to 30.
      *
--- a/common/src/java/com/zimbra/common/service/ServiceException.java
+++ b/common/src/java/com/zimbra/common/service/ServiceException.java
@@ -60,6 +60,7 @@ public class ServiceException extends Exception {
     public static final String UNSUPPORTED = "service.UNSUPPORTED";
     public static final String FORBIDDEN = "service.FORBIDDEN";
     public static final String SIEVE_SCRIPT_MAX_SIZE_EXCEPTION = "service.SIEVE_SCRIPT_MAX_SIZE_EXCEPTION";
+    public static final String MAIL_RECALL_ERROR = "service.MAIL_RECALL_ERROR";
     
     // generic "not found" error for objects other than mail items
     public static final String NOT_FOUND = "service.NOT_FOUND";
@@ -301,6 +302,10 @@ public class ServiceException extends Exception {
      */
     public static ServiceException LICENSE_ERROR(String message, Throwable cause) {
         return new ServiceException("license error: "+message, LICENSE_ERROR, RECEIVERS_FAULT, cause);
+    }
+
+    public static ServiceException ERROR_MESSAGE(String str, Throwable cause){
+        return new ServiceException(String.format("mailRecall error: %s", str), MAIL_RECALL_ERROR, SENDERS_FAULT, cause);
     }
 
     public static ServiceException ERROR_WHILE_PARSING_UPLOAD(String message, Throwable cause) {
--- a/soap/src/java/com/zimbra/soap/JaxbUtil.java
+++ b/soap/src/java/com/zimbra/soap/JaxbUtil.java
@@ -17,6 +17,8 @@
 
 package com.zimbra.soap;
 
+import com.zimbra.soap.mail.message.MailRecallRequest;
+import com.zimbra.soap.mail.message.MailRecallResponse;
 import java.io.ByteArrayInputStream;
 import java.util.HashMap;
 import java.util.Iterator;
@@ -1086,6 +1088,8 @@ public final class JaxbUtil {
             com.zimbra.soap.mail.message.VerifyCodeResponse.class,
             com.zimbra.soap.mail.message.WaitSetRequest.class,
             com.zimbra.soap.mail.message.WaitSetResponse.class,
+            com.zimbra.soap.mail.message.MailRecallRequest.class,
+            com.zimbra.soap.mail.message.MailRecallResponse.class,
             com.zimbra.soap.replication.message.BecomeMasterRequest.class,
             com.zimbra.soap.replication.message.BecomeMasterResponse.class,
             com.zimbra.soap.replication.message.BringDownServiceIPRequest.class,
@@ -1173,7 +1177,9 @@ public final class JaxbUtil {
             com.zimbra.soap.admin.message.ValidateS3BucketReachableRequest.class,
             com.zimbra.soap.admin.message.ValidateS3BucketReachableResponse.class,
             com.zimbra.soap.admin.message.EditS3BucketConfigRequest.class,
-            com.zimbra.soap.admin.message.EditS3BucketConfigResponse.class
+            com.zimbra.soap.admin.message.EditS3BucketConfigResponse.class,
+            com.zimbra.soap.admin.message.GenerateSecretKeyRequest.class,
+            com.zimbra.soap.admin.message.GenerateSecretKeyResponse.class
         };
 
         try {
@@ -1683,7 +1689,7 @@ public final class JaxbUtil {
         ImapMessageInfo messageInfo = new ImapMessageInfo(mod.getIdInMailbox(), mod.getImapUid(), mod.getMailItemType().toString(), mod.getFlagBitmask(), tags);
         return new ModifyNotification.ModifyItemNotification(messageInfo, reason);
     }
-    
+
     public static DeleteItemNotification getDeletedItemSOAP(int itemId, String type) throws ServiceException {
         return new DeleteItemNotification(itemId, type);
     }
--- a/soap/src/java/com/zimbra/soap/mail/message/MailRecallRequest.java
+++ b/soap/src/java/com/zimbra/soap/mail/message/MailRecallRequest.java
@@ -1,0 +1,39 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2025 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ * ***** END LICENSE BLOCK *****
+ */
+
+package com.zimbra.soap.mail.message;
+
+import com.zimbra.common.soap.MailConstants;
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlAccessorType(XmlAccessType.NONE)
+@XmlRootElement(name = MailConstants.E_MAIL_RECALL_REQUEST)
+public class MailRecallRequest {
+    @XmlAttribute(name = MailConstants.ITEM_ID , required = true)
+    private int itemId;
+
+    public int getItemId() {
+        return itemId;
+    }
+
+    public void setItemId(int itemId) {
+        this.itemId = itemId;
+    }
+}
--- a/soap/src/java/com/zimbra/soap/mail/message/MailRecallResponse.java
+++ b/soap/src/java/com/zimbra/soap/mail/message/MailRecallResponse.java
@@ -1,0 +1,64 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2025 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ * ***** END LICENSE BLOCK *****
+ */
+
+package com.zimbra.soap.mail.message;
+
+import com.zimbra.common.soap.MailConstants;
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlAccessorType(XmlAccessType.NONE)
+@XmlRootElement(name = MailConstants.E_MAIL_RECALL_RESPONSE)
+public class MailRecallResponse {
+
+    @XmlElement(name = "successfulRecall", required = true)
+    private int successfulRecall;
+
+    @XmlElement(name = "unsuccessfulRecall", required = true)
+    private int unsuccessfulRecall;
+
+    @XmlElement(name = "allMailRecalled", required = true)
+    private boolean allMailRecalled;
+
+    // Getters and Setters
+    public int getSuccessfulRecall() {
+        return successfulRecall;
+    }
+
+    public void setSuccessfulRecall(int successfulRecall) {
+        this.successfulRecall = successfulRecall;
+    }
+
+    public int getUnsuccessfulRecall() {
+        return unsuccessfulRecall;
+    }
+
+    public void setUnsuccessfulRecall(int unsuccessfulRecall) {
+        this.unsuccessfulRecall = unsuccessfulRecall;
+    }
+
+    public boolean isAllMailRecalled() {
+        return allMailRecalled;
+    }
+
+    public void setAllMailRecalled(boolean allMailRecalled) {
+        this.allMailRecalled = allMailRecalled;
+    }
+}
+
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -10339,7 +10339,7 @@ TODO: delete them permanently from here
 <attr id="4093" name="zimbraTrialConvertAtExpiration" type="boolean" cardinality="single" optionalIn="account,cos,domain" flags="accountCosDomainInherited,accountInfo" since="9.0.0">
   <desc>Can be used by an external service to indicate that a trial is set to convert to active at expiration.</desc>
 </attr>
-<attr id="4094" name="zimbraFeatureMailRecallEnabled" type="boolean" cardinality="single" optionalIn="globalConfig,account,cos,domain" flags="accountInfo,domainInherited,accountCosDomainInherited" since="10.1.0">
+<attr id="4094" name="zimbraFeatureMailRecallEnabled" type="boolean" cardinality="single" optionalIn="globalConfig,account,cos,domain" flags="accountInfo,domainInherited,accountCosDomainInherited" callback="GenerateSecretKeyCallback" since="10.1.0">
   <globalConfigValue>FALSE</globalConfigValue>
   <defaultCOSValue>FALSE</defaultCOSValue>
   <desc>Enables the mail recall functionality</desc>
@@ -10559,5 +10559,9 @@ TODO: delete them permanently from here
 <attr id="4135" name="zimbraFeatureDeliveryStatusNotificationEnabled" type="boolean" cardinality="single" optionalIn="cos,account" flags="accountInherited,accountInfo" since="10.1.4">
   <defaultCOSValue>FALSE</defaultCOSValue>
   <desc>Feature to enable delivery status notification</desc>
+</attr>
+
+<attr id="4136" name="zimbraFeatureMailRecallSecretKey" type="string" cardinality="single" optionalIn="globalConfig" since="10.1.5">
+  <desc>Secret key used in Mail Recall to make it more secure from spoof.</desc>
 </attr>
 </attrs>
--- a/store/src/java/com/zimbra/cs/account/ZAttrConfig.java
+++ b/store/src/java/com/zimbra/cs/account/ZAttrConfig.java
@@ -18824,6 +18824,78 @@ public abstract class ZAttrConfig extends Entry {
     }
 
     /**
+     * Secret key used in Mail Recall to make it more secure from spoof.
+     *
+     * @return zimbraFeatureMailRecallSecretKey, or null if unset
+     *
+     * @since ZCS 10.1.5
+     */
+    @ZAttr(id=4136)
+    public String getFeatureMailRecallSecretKey() {
+        return getAttr(Provisioning.A_zimbraFeatureMailRecallSecretKey, null, true);
+    }
+
+    /**
+     * Secret key used in Mail Recall to make it more secure from spoof.
+     *
+     * @param zimbraFeatureMailRecallSecretKey new value
+     * @throws com.zimbra.common.service.ServiceException if error during update
+     *
+     * @since ZCS 10.1.5
+     */
+    @ZAttr(id=4136)
+    public void setFeatureMailRecallSecretKey(String zimbraFeatureMailRecallSecretKey) throws com.zimbra.common.service.ServiceException {
+        HashMap<String,Object> attrs = new HashMap<String,Object>();
+        attrs.put(Provisioning.A_zimbraFeatureMailRecallSecretKey, zimbraFeatureMailRecallSecretKey);
+        getProvisioning().modifyAttrs(this, attrs);
+    }
+
+    /**
+     * Secret key used in Mail Recall to make it more secure from spoof.
+     *
+     * @param zimbraFeatureMailRecallSecretKey new value
+     * @param attrs existing map to populate, or null to create a new map
+     * @return populated map to pass into Provisioning.modifyAttrs
+     *
+     * @since ZCS 10.1.5
+     */
+    @ZAttr(id=4136)
+    public Map<String,Object> setFeatureMailRecallSecretKey(String zimbraFeatureMailRecallSecretKey, Map<String,Object> attrs) {
+        if (attrs == null) attrs = new HashMap<String,Object>();
+        attrs.put(Provisioning.A_zimbraFeatureMailRecallSecretKey, zimbraFeatureMailRecallSecretKey);
+        return attrs;
+    }
+
+    /**
+     * Secret key used in Mail Recall to make it more secure from spoof.
+     *
+     * @throws com.zimbra.common.service.ServiceException if error during update
+     *
+     * @since ZCS 10.1.5
+     */
+    @ZAttr(id=4136)
+    public void unsetFeatureMailRecallSecretKey() throws com.zimbra.common.service.ServiceException {
+        HashMap<String,Object> attrs = new HashMap<String,Object>();
+        attrs.put(Provisioning.A_zimbraFeatureMailRecallSecretKey, "");
+        getProvisioning().modifyAttrs(this, attrs);
+    }
+
+    /**
+     * Secret key used in Mail Recall to make it more secure from spoof.
+     *
+     * @param attrs existing map to populate, or null to create a new map
+     * @return populated map to pass into Provisioning.modifyAttrs
+     *
+     * @since ZCS 10.1.5
+     */
+    @ZAttr(id=4136)
+    public Map<String,Object> unsetFeatureMailRecallSecretKey(Map<String,Object> attrs) {
+        if (attrs == null) attrs = new HashMap<String,Object>();
+        attrs.put(Provisioning.A_zimbraFeatureMailRecallSecretKey, "");
+        return attrs;
+    }
+
+    /**
      * Time(in minutes) within which a message can be recalled. The default
      * time is 30 minutes and accepts value from 1 to 30.
      *
--- a/store/src/java/com/zimbra/cs/account/callback/GenerateSecretKeyCallback.java
+++ b/store/src/java/com/zimbra/cs/account/callback/GenerateSecretKeyCallback.java
@@ -1,0 +1,68 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2025 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ * ***** END LICENSE BLOCK *****
+ */
+
+package com.zimbra.cs.account.callback;
+
+import com.google.common.base.Strings;
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.util.ZimbraLog;
+import com.zimbra.cs.account.AttributeCallback;
+import com.zimbra.cs.account.Entry;
+import com.zimbra.cs.account.Provisioning;
+import com.zimbra.cs.account.soap.SoapProvisioning;
+import com.zimbra.soap.admin.message.GenerateSecretKeyRequest;
+
+import java.util.Map;
+import java.util.Optional;
+
+public class GenerateSecretKeyCallback extends AttributeCallback {
+
+    @Override
+    public void preModify(CallbackContext context, String attrName, Object value,
+            Map attrsToModify, Entry entry) throws ServiceException {
+        try {
+            Optional.ofNullable(value)
+                    .map(Object::toString)
+                    .map(String::trim)
+                    .filter("TRUE"::equals)
+                    .ifPresent(this::generateMailRecallSecretKey);
+        } catch (RuntimeException e) {
+            throw ServiceException.FAILURE("Unable to initialize SecureRandom for mail recall", e);
+        }
+    }
+
+    private void generateMailRecallSecretKey(String value) {
+        try {
+            String secretKey = Provisioning.getInstance()
+                    .getConfig()
+                    .getFeatureMailRecallSecretKey();
+
+            if (Strings.isNullOrEmpty(secretKey)) {
+                SoapProvisioning sp = SoapProvisioning.getAdminInstance();
+                sp.invokeJaxb(new GenerateSecretKeyRequest());
+            }
+        } catch (ServiceException e) {
+            ZimbraLog.misc.error("Error generating mail recall secret key", e);
+            throw new RuntimeException("Error generating mail recall secret key", e);
+        }
+    }
+
+    @Override
+    public void postModify(CallbackContext context, String attrName, Entry entry) {
+    }
+
+}
--- a/store/src/java/com/zimbra/cs/mailbox/MailSender.java
+++ b/store/src/java/com/zimbra/cs/mailbox/MailSender.java
@@ -33,6 +33,7 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 
 import javax.mail.Address;
@@ -41,6 +42,7 @@ import javax.mail.SendFailedException;
 import javax.mail.Session;
 import javax.mail.Transport;
 import javax.mail.internet.InternetAddress;
+import javax.mail.internet.MailDateFormat;
 import javax.mail.internet.MimeMessage;
 
 import com.google.common.base.Joiner;
@@ -81,6 +83,7 @@ import com.zimbra.cs.service.FileUploadServlet;
 import com.zimbra.cs.service.FileUploadServlet.Upload;
 import com.zimbra.cs.service.UserServlet;
 import com.zimbra.cs.service.util.ItemId;
+import com.zimbra.cs.service.util.SecretKey;
 import com.zimbra.cs.util.AccountUtil;
 import com.zimbra.cs.util.AccountUtil.AccountAddressMatcher;
 import com.zimbra.cs.util.BuildInfo;
@@ -91,6 +94,8 @@ public class MailSender {
     public static final String MSGTYPE_REPLY = String.valueOf(Flag.toChar(Flag.ID_REPLIED));
     public static final String MSGTYPE_FORWARD = String.valueOf(Flag.toChar(Flag.ID_FORWARDED));
     private static Map<String, PreSendMailListener> mPreSendMailListeners = new ConcurrentHashMap<String, PreSendMailListener>();
+    private static final MailDateFormat mailDateFormat = new MailDateFormat();
+    private static final String X_MESSAGE_VERIFICATION = "X-Zimbra-Message-Verification";
 
     private Boolean mSaveToSent;
     private Collection<Upload> mUploads;
@@ -1020,7 +1025,18 @@ public class MailSender {
         mm.setFrom(from);
         mm.setSender(sender);
 
-        mm.setSentDate(new Date());
+        Date date = new Date();
+        mm.setSentDate(date);
+
+        Optional.ofNullable(SecretKey.getMessageVerificationHeaderValue(mm.getMessageID(), mailDateFormat.format(date),
+                from.getAddress())).ifPresent(value -> {
+            try {
+                mm.addHeader(X_MESSAGE_VERIFICATION, value);
+            } catch (MessagingException e) {
+                throw new RuntimeException(String.format("Failed to add Message Verification header: %s", value), e);
+            }
+        });
+
         if (sender == null) {
             Address[] existingReplyTos = mm.getReplyTo();
             if (existingReplyTos == null || existingReplyTos.length == 0) {
@@ -1483,4 +1499,6 @@ public class MailSender {
             }
         }
     }
+
+
 }
--- a/store/src/java/com/zimbra/cs/service/admin/GenerateSecretKey.java
+++ b/store/src/java/com/zimbra/cs/service/admin/GenerateSecretKey.java
@@ -1,0 +1,56 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2025 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ * ***** END LICENSE BLOCK *****
+ */
+
+package com.zimbra.cs.service.admin;
+
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.soap.AdminConstants;
+import com.zimbra.common.soap.Element;
+import com.zimbra.common.util.ZimbraLog;
+import com.zimbra.cs.account.Provisioning;
+import com.zimbra.cs.account.soap.SoapProvisioning;
+import com.zimbra.cs.service.util.SecretKey;
+import com.zimbra.soap.ZimbraSoapContext;
+import com.zimbra.soap.admin.type.CacheEntryType;
+
+import java.util.Map;
+
+public class GenerateSecretKey extends AdminDocumentHandler {
+    public Element handle(Element request, Map<String, Object> context) throws ServiceException {
+        ZimbraSoapContext zsc = getZimbraSoapContext(context);
+
+        String randomString = SecretKey.generateRandomString();
+        Provisioning.getInstance().getConfig().setFeatureMailRecallSecretKey(randomString);
+        flushCache();
+
+        Element response = zsc.createElement(AdminConstants.GENERATE_SECRET_KEY_RESPONSE);
+        return response;
+    }
+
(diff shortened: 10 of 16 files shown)


store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -1478,4 +1479,31 @@ public static void unregisterPreSendMailListener(PreSendMailListener listener) {
             }
         }
     }
+
```
swaatiTelus, line 1502: Please add method documentation for both of the methods.

shubhamCS03: done

store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -90,6 +84,8 @@ public class MailSender {
 
     public static final String MSGTYPE_REPLY = String.valueOf(Flag.toChar(Flag.ID_REPLIED));
     public static final String MSGTYPE_FORWARD = String.valueOf(Flag.toChar(Flag.ID_FORWARDED));
+    public static final String HASH_ALGORITHM_INFO = "hash=SHA256;guid=";
+    public static final String ALGORITHM_NAME = "SHA-256";
```
ronstra-synacor:
```suggestion
    public static final String MSGVRFY_HEADER_PREFIX = "hash=SHA256;guid=";
    public static final String MSGVRFY_ALGORITHM_NAME = "SHA-256";
```

store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -1478,4 +1479,44 @@ public static void unregisterPreSendMailListener(PreSendMailListener listener) {
...
+            StringBuilder hexString = new StringBuilder(number.toString(16));
+
+            while (hexString.length() < 64) {
+                hexString.insert(0, '0');
+            }
+
+            String BasicBase64format
+                    = Base64.getEncoder()
+                    .encodeToString(hexString.toString().getBytes());
+            return BasicBase64format;
```
ronstra-synacor: Why is the `BasicBase64format` variable capitalized?

ronstra-synacor (changes requested):

store/src/java/com/zimbra/cs/mailbox/MailSender.java
shubhamCS03: Corrected , Thanks!

store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -1478,4 +1479,44 @@ public static void unregisterPreSendMailListener(PreSendMailListener listener) {
...
     }
+
+    /**
+     * provide the Hash for Message-Verification header field.
+     * @param id
+     * @param date
+     * @param from
+     * @throws ServiceException
+     */
+    public static String getHashForMessageVerification(String id, Date date, String from) throws MessagingException, ServiceException {
```
ronstra-synacor:
```suggestion
    private static String getHashForMessageVerification(String id, Date date, String from) throws MessagingException, ServiceException {
```

store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -1478,4 +1479,44 @@ public static void unregisterPreSendMailListener(PreSendMailListener listener) {
...
+        return hash;
+    }
+
+    /**
+     * provide the Encrypted value of the GUID.
+     * @param input
+     * @param algo
+     * @throws ServiceException
+     */
+    public static String getEncryptedValue(String input, String algo) throws ServiceException {
```
ronstra-synacor:
```suggestion
    private static String getEncryptedValue(String input, String algo) throws ServiceException {
```

store/src/java/com/zimbra/cs/mailbox/MailSender.java
ronstra-synacor:
```suggestion
    public static String getMessageVerificationHeaderValue(String id, Date date, String from) throws MessagingException, ServiceException {
```

store/src/java/com/zimbra/cs/mailbox/MailSender.java
ronstra-synacor:
```suggestion
    public static String getHashForMessageVerification(String input, String algo) throws ServiceException {
```

store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -1478,4 +1479,44 @@ public static void unregisterPreSendMailListener(PreSendMailListener listener) {
...
+     */
+    public static String getHashForMessageVerification(String id, Date date, String from) throws MessagingException, ServiceException {
+        String guid = (id + date + from);
+        String guidHash = getEncryptedValue(guid, MSGVRFY_ALGORITHM_NAME);
+        String hash = MSGVRFY_HEADER_PREFIX + guidHash;
+        return hash;
+    }
+
+    /**
+     * provide the Encrypted value of the GUID.
```
ronstra-synacor:
```suggestion
     * Create a digest of the given input with the given algorithm.
```

ronstra-synacor (changes requested):

store/src/java/com/zimbra/cs/mailbox/MailSender.java
shubhamCS03: Corrected ,Thank you!!

store/src/java/com/zimbra/cs/mailbox/MailSender.java
shubhamCS03: Done

store/src/java/com/zimbra/cs/mailbox/MailSender.java
shubhamCS03: Done

store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -1478,4 +1479,45 @@ public static void unregisterPreSendMailListener(PreSendMailListener listener) {
...
+     * Create a digest of the given input with the given algorithm.
+     * @param input
+     * @param algo
+     * @throws ServiceException
+     */
+    private static String getHashForMessageVerification(String input, String algo) throws ServiceException {
+        try {
+            MessageDigest md1 = MessageDigest.getInstance(algo);
+            byte[] messageDigest1 = md1.digest(input.getBytes(StandardCharsets.UTF_8));
+            BigInteger number = new BigInteger(1, messageDigest1);
```
k-kato: Is `1` necessary in each variable name?
```suggestion
            MessageDigest md = MessageDigest.getInstance(algo);
            byte[] messageDigest = md.digest(input.getBytes(StandardCharsets.UTF_8));
            BigInteger number = new BigInteger(1, messageDigest);
```

store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -1478,4 +1479,45 @@ public static void unregisterPreSendMailListener(PreSendMailListener listener) {
...
+    /**
+     * provide the Hash for Message-Verification header field.
+     * @param id
+     * @param date
+     * @param from
+     * @throws ServiceException
+     *
+     */
+    private static String getMessageVerificationHeaderValue(String id, Date date, String from) throws MessagingException, ServiceException {
+        String guid = (id + date + from);
```
k-kato: Is it secure? Don't we need to add any system-wide (= globalConfig) secret value?

https://www.ietf.org/archive/id/draft-leiba-morg-message-recall-00.html#anchor4
The actual GUID (the pre-image of the hash) is kept secret by the sending side.

ronstra-synacor (dismissed):

ghost (changes requested):
This implementation is incomplete.

The sha256 hash is one part of the solution, then when the recall action is started you have to match the account name with the from header.

So if the authenticated user account name (and or alias) matches the from header **_AND_** the sha256 is valid THEN you proceed with recall.

store/src/java/com/zimbra/cs/mailbox/MailSender.java
ghost: **Yes you are right, we need to add some secret to the sha256 hash.**

Otherwise a user could both spoof the from and sha256 header and still recall any message with certain message ID. Which is what we are trying to avoid.

In addition when we recall a message, the authenticated user account name and or alias should be validated against the from header. If that matches **_and_** the sha256 hash is correct we can proceed and recall.

store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -1015,7 +1013,12 @@ void updateHeaders(MimeMessage mm, Account acct, Account authuser, OperationCont
         mm.setFrom(from);
         mm.setSender(sender);
 
-        mm.setSentDate(new Date());
+        Date date = new Date();
+        mm.setSentDate(date);
+
+        String value = getMessageVerificationHeaderValue(mm.getMessageID(), mailDateFormat.format(date), from.getAddress());
+        mm.addHeader("Message-Verification", value);
```
silentsakky: I am not sure what's the standard we are following when adding custom headers but I have seen other products prefixing it with their name to identify it as custom header specific to that product like X-VADE-* etc
so I would suggest to do that here as well like X-ZIMBRA-*

shubhamCS03: Done, I used "X-Zimbra-Message-Verification" this as header name.

store/src/java/com/zimbra/cs/service/util/SecretKey.java
```diff
@@ -0,0 +1,81 @@
...
+     * @param id
+     * @param date
+     * @param from
+     * @throws ServiceException
+     */
+    public static String getMessageVerificationHeaderValue(String id, String date, String from) throws MessagingException, ServiceException {
+        String secretKey = Provisioning.getInstance().getConfig().getSecretKeyForMailRecall();
+        if(Strings.isNullOrEmpty(secretKey)) {
+            secretKey = SecretKey.generateRandomString();
+            Provisioning.getInstance().getConfig().setSecretKeyForMailRecall(secretKey);
```
k-kato: Multiple server environment need to be considered.
What occurs when the attribute is empty and 2 or more mailstore servers execute sending message process at the same time? A different value is generated on each mailstore, and then the different value is stored in each cache store until the cache is expired or mailboxd is restarted.
To make it more stable,
1. generate a key on one of mailstore servers
2. run flushcache config on all servers
3. enable recall function with the single key - without any conflict/mismatch among mailstore servers.

store/src/java/com/zimbra/cs/account/callback/GenerateSecretKeyCallback.java
```diff
@@ -0,0 +1,65 @@
...
+            ZimbraLog.misc.warn("Encountered exception during getting Secret Key", e);
+        }
+    }
+
+    private void flushCache() {
+        try {
+            SoapProvisioning sp = new SoapProvisioning();
+            sp.soapSetURI(SoapProvisioning.getLocalConfigURI());
+            sp.soapZimbraAdminAuthenticate();
+            sp.flushCache("config", null, true);
```
k-kato:
```suggestion
            sp.flushCache(CacheEntryType.config, null, true);
```

store/src/java/com/zimbra/cs/account/callback/GenerateSecretKeyCallback.java
```diff
@@ -0,0 +1,65 @@
...
+    @Override
+    public void postModify(CallbackContext context, String attrName, Entry entry) {
+        String secretKey = null;
+        try {
+            secretKey = Provisioning.getInstance().getConfig().getSecretKeyForMailRecall();
+
+            if (Strings.isNullOrEmpty(secretKey)) {
+                secretKey = SecretKey.generateRandomString();
+                Provisioning.getInstance().getConfig().setSecretKeyForMailRecall(secretKey);
+                flushCache();
```
k-kato: Can't GenerateSecretKeyRequest be called using SoapProvisioning instead? The three lines are the same as GenerateSecretKey.java

store/src/java/com/zimbra/cs/service/admin/GenerateSecretKey.java
```diff
@@ -0,0 +1,38 @@
...
+        Element response = zsc.createElement(AdminConstants.GENERATE_SECRET_KEY_RESPONSE);
+        return response;
+    }
+
+    private void flushCache() {
+        try {
+            SoapProvisioning sp = new SoapProvisioning();
+            sp.soapSetURI(SoapProvisioning.getLocalConfigURI());
+            sp.soapZimbraAdminAuthenticate();
+            sp.flushCache("config", null, true);
```
k-kato:
```suggestion
            sp.flushCache(CacheEntryType.config, null, true);
```

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -10369,4 +10369,7 @@ TODO: delete them permanently from here
   <globalConfigValue>TRUE</globalConfigValue>
   <desc>whether to enable CountObjectsRequest with userAccount/account type on Zimbra Admin Console</desc>
 </attr>
+<attr id="4097" name="secretKeyForMailRecall" type="string" max="128" cardinality="single" optionalIn="globalConfig" since="10.0.0">
+  <desc>Can be used in Mail Recall to make it more secure from spoof.</desc>
```
k-kato:
```suggestion
  <desc>Secret key used in Mail Recall to make it more secure from spoof.</desc>
```

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -10369,4 +10369,7 @@ TODO: delete them permanently from here
   <globalConfigValue>TRUE</globalConfigValue>
   <desc>whether to enable CountObjectsRequest with userAccount/account type on Zimbra Admin Console</desc>
 </attr>
+<attr id="4097" name="secretKeyForMailRecall" type="string" max="128" cardinality="single" optionalIn="globalConfig" since="10.0.0">
```
k-kato: Is the length `128` enough for a secure key? Removing it should be okay.
```suggestion
<attr id="4097" name="secretKeyForMailRecall" type="string" cardinality="single" optionalIn="globalConfig" since="10.0.0">
```

store/src/java/com/zimbra/cs/account/callback/GenerateSecretKeyCallback.java
```diff
@@ -0,0 +1,65 @@
...
+            }
+        } catch (ServiceException e) {
+            ZimbraLog.misc.warn("Encountered exception during getting Secret Key", e);
+        }
```
k-kato, line 61: Isn't it better to move the entire logic to `preModify`? 

If it can be moved to `preModify`, consider adding a logic to check if zimbraFeatureMailRecallEnabled is being set to TRUE. When zimbraFeatureMailRecallEnabled is set to FALSE, a secret key does not need to be generated/updated/removed.
* generate a secret key before zimbraFeatureMailRecallEnabled is set to TRUE
* If a secret key cannot be generated or set, throw an exception so that zimbraFeatureMailRecallEnabled cannot be set to TRUE

i.e. zimbraFeatureMailRecallEnabled can be set to TRUE only when a secret key is generated and set successfully or it has already been set.

store/src/java/com/zimbra/cs/account/callback/GenerateSecretKeyCallback.java
shubhamCS03: Done

store/src/java/com/zimbra/cs/mailbox/MailSender.java
shubhamCS03: Done

store/src/java/com/zimbra/cs/service/util/SecretKey.java
shubhamCS03: Done

store/src/java/com/zimbra/cs/mailbox/MailSender.java
shubhamCS03: Implemented the secret key at global level

store/src/java/com/zimbra/cs/account/callback/GenerateSecretKeyCallback.java
shubhamCS03: Yes, moved the entire logic to preModify and implemented the changes as per you suggestions

common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
```diff
@@ -2869,6 +2869,14 @@ public static TwoFactorAuthSecretEncoding fromString(String s) throws ServiceExc
     @ZAttr(id=-1)
     public static final String A_registeredAddress = "registeredAddress";
 
+    /**
+     * Can be used in Mail Recall to make it more secure from spoof.
+     *
+     * @since ZCS 10.0.0
+     */
+    @ZAttr(id=4097)
+    public static final String A_secretKeyForMailRecall = "secretKeyForMailRecall";
```
Prashantsurana: Please read the below lines at the start of the zimbra-attrs.xml file. 
The name of the attribute should start with **zimbra**

```
Zimbra devs - things to keep in mind when adding new attributes:

 1) Always add the new attribute at the end of the list, adding one to
    the highest attr id (the one that corresponds to the OID) number
    in use.  Do not sort by name - it will break OID allocation.

 2) When defining new attributes, try and use descriptive names. The
    name should *always* start with zimbra*.  When defining
    multiple-attributes use a standard prefix for all related
    attributes. For example, if you are defining multiple attrs all
    related to the blob store, then prefix them all with zimbraBlob*.
```

Prashantsurana (changes requested):

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -10369,4 +10369,7 @@ TODO: delete them permanently from here
   <globalConfigValue>TRUE</globalConfigValue>
   <desc>whether to enable CountObjectsRequest with userAccount/account type on Zimbra Admin Console</desc>
 </attr>
+<attr id="4097" name="secretKeyForMailRecall" type="string" cardinality="single" optionalIn="globalConfig" since="10.0.0">
```
Prashantsurana: Please read the below lines at the start of the zimbra-attrs.xml file. 
The name of the attribute should start with **zimbra**

```
Zimbra devs - things to keep in mind when adding new attributes:

 1) Always add the new attribute at the end of the list, adding one to
    the highest attr id (the one that corresponds to the OID) number
    in use.  Do not sort by name - it will break OID allocation.

 2) When defining new attributes, try and use descriptive names. The
    name should *always* start with zimbra*.  When defining
    multiple-attributes use a standard prefix for all related
    attributes. For example, if you are defining multiple attrs all
    related to the blob store, then prefix them all with zimbraBlob*.
```

Prashantsurana (changes requested):

common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
shubhamCS03: Corrected.

store/conf/attrs/zimbra-attrs.xml
shubhamCS03: Corrected . Thank you for Highlighting it!

store/src/java/com/zimbra/cs/service/admin/GenerateSecretKey.java
k-kato, line 1: add license block to new files
* store/src/java/com/zimbra/cs/service/admin/GenerateSecretKey.java
* store/src/java/com/zimbra/cs/service/util/SecretKey.java

k-kato (approved):

store/src/java/com/zimbra/cs/service/admin/GenerateSecretKey.java
shubhamCS03: Done.

shrutig0510 (approved):

zimsuchitgupta (approved):

swaatiTelus:
The mail recall feature is not planned in the upcoming release so this PR should not be merged.

store/src/java/com/zimbra/cs/service/util/SecretKey.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * returns the randomly generated String
+     *
+     * @return randomly generated String
+     * @throws ServiceException if an error occurred
+     */
+    public static String generateRandomString() throws ServiceException {
+        try {
+            SecureRandom random = SecureRandom.getInstance("SHA1PRNG");
```
ghost: Consider using:
SecureRandom random = new SecureRandom();

https://stackoverflow.com/questions/27622625/securerandom-with-nativeprng-vs-sha1prng

store/src/java/com/zimbra/cs/service/util/SecretKey.java
```diff
@@ -0,0 +1,96 @@
...
+            byte[] key = new byte[KEY_SIZE_BYTES];
+            random.nextBytes(key);
+            return new String(Hex.encodeHex(key));
```
ashishkataria86: Consider using Base64 for encoding to make the key more compact and URL-safe.

store/src/java/com/zimbra/cs/service/util/SecretKey.java
```diff
@@ -0,0 +1,96 @@
...
+     * Create a digest of the given input with the given algorithm.
+     *
+     * @param input
+     * @throws ServiceException
+     */
+    private static String getHashForMessageVerification(String input) throws ServiceException {
+        try {
+            MessageDigest md = MessageDigest.getInstance(MSGVRFY_ALGORITHM_NAME);
+            byte[] messageDigest = md.digest(input.getBytes(StandardCharsets.UTF_8));
+            BigInteger number = new BigInteger(1, messageDigest);
```
ashishkataria86: You convert the BigInteger to a string and then encode it to Base64. This could be simplified:
return Base64.getEncoder().encodeToString(messageDigest);

store/src/java/com/zimbra/cs/service/util/SecretKey.java
```diff
@@ -0,0 +1,96 @@
...
+     * provide the Hash for Message-Verification header field.
+     *
+     * @param id
+     * @param date
+     * @param from
+     * @throws ServiceException
+     */
+    public static String getMessageVerificationHeaderValue(String id, String date, String from) throws MessagingException, ServiceException {
+        String secretKey = Provisioning.getInstance().getConfig().getFeatureMailRecallSecretKey();
+        String guid = (id + date + from + secretKey);
```
log2akshat: Repeated use of the `+` operator can degrade performance as strings are immutable in Java. Each concatenation creates a new String object, leaving old strings for garbage collection, which can increase memory usage and slow down execution. Leverage Java 8 and above features.

store/src/java/com/zimbra/cs/service/util/SecretKey.java
```diff
@@ -0,0 +1,96 @@
...
+        String guidHash = getHashForMessageVerification(guid);
+        String hash = MSGVRFY_HEADER_PREFIX + guidHash;
+        return hash;
+    }
```
log2akshat, line 78: Maybe you can use `Optional.ofNullable(...).orElseThrow(...)` to ensure `secretKey` is not `null` and throw an exception if it is, to  avoid potential null pointer and `String.join(...)` for concatenation.

store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -22,17 +22,7 @@
...
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
log2akshat: Don't use implicit imports

store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -61,13 +52,7 @@
...
 import com.zimbra.common.util.SystemUtil;
 import com.zimbra.common.util.ZimbraLog;
-import com.zimbra.cs.account.AccessManager;
-import com.zimbra.cs.account.Account;
-import com.zimbra.cs.account.AuthToken;
-import com.zimbra.cs.account.DataSource;
-import com.zimbra.cs.account.Domain;
-import com.zimbra.cs.account.Identity;
-import com.zimbra.cs.account.Provisioning;
+import com.zimbra.cs.account.*;
```
log2akshat: Don't use implicit imports

store/src/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -1015,7 +1003,12 @@ void updateHeaders(MimeMessage mm, Account acct, Account authuser, OperationCont
         mm.setFrom(from);
         mm.setSender(sender);
 
-        mm.setSentDate(new Date());
+        Date date = new Date();
+        mm.setSentDate(date);
+
+        String value = SecretKey.getMessageVerificationHeaderValue(mm.getMessageID(), mailDateFormat.format(date), from.getAddress());
```
log2akshat: Make the code more expressive, and ensure it's null-safe by leveraging Java 8+ features such as `Optional`, method references etc.

store/src/java/com/zimbra/cs/account/callback/GenerateSecretKeyCallback.java
```diff
@@ -0,0 +1,53 @@
...
+            if(value.equals("TRUE")) {
+                secretKey = Provisioning.getInstance().getConfig().getFeatureMailRecallSecretKey();
+                if (Strings.isNullOrEmpty(secretKey)) {
+                    SoapProvisioning sp = SoapProvisioning.getAdminInstance();
+                    sp.invokeJaxb(new GenerateSecretKeyRequest());
+                }
+            }
+        } catch (ServiceException e) {
+            throw ServiceException.FAILURE("Unable to initialize SecureRandom for mail recall", e);
+        }
```
log2akshat, line 61: Please optimize the code to improve the readability and efficiency of the code. Consider using Java 8 and above features such as:
- `Optional` to wrap the `value` and `filter` it to only proceed if it equals `TRUE`,
- `method references`,
- `try-with-resources`
- `lambda expression` to encapsulate the logic for handling the secret key and its invocation, to reduce nested conditions, improve resource management.

Also, consider catching specific exceptions.

store/src/java/com/zimbra/cs/service/admin/GenerateSecretKey.java
```diff
@@ -0,0 +1,56 @@
...
+    private void flushCache() {
+        try {
+            SoapProvisioning sp = new SoapProvisioning();
+            sp.soapSetURI(SoapProvisioning.getLocalConfigURI());
+            sp.soapZimbraAdminAuthenticate();
+            sp.flushCache(CacheEntryType.config.toString(), null, true);
+        } catch (ServiceException e) {
+            ZimbraLog.misc.warn("Encountered exception during FlushCache after creating Secret Key", e);
+        }
+    }
```
log2akshat, line 53: Consider using `Try-with-Resources`

store/src/java/com/zimbra/cs/service/util/SecretKey.java
```diff
@@ -0,0 +1,96 @@
...
+        try {
+            MessageDigest md = MessageDigest.getInstance(MSGVRFY_ALGORITHM_NAME);
+            byte[] messageDigest = md.digest(input.getBytes(StandardCharsets.UTF_8));
+            BigInteger number = new BigInteger(1, messageDigest);
+            StringBuilder hexString = new StringBuilder(number.toString(16));
+
+            while (hexString.length() < 64) {
+                hexString.insert(0, '0');
+            }
+
```
log2akshat: You can use `String.format("%064x", ...)` to format the BigInteger directly into a zero-padded hexadecimal string of the desired length and remove `StringBuilder`.

common/src/java/com/zimbra/common/service/ServiceException.java
```diff
@@ -303,6 +303,10 @@ public static ServiceException LICENSE_ERROR(String message, Throwable cause) {
         return new ServiceException("license error: "+message, LICENSE_ERROR, RECEIVERS_FAULT, cause);
     }
 
+    public static ServiceException ERROR_MESSAGE(String str, Throwable cause){
+        return new ServiceException("error: " + str, null, SENDERS_FAULT, cause);
```
log2akshat: Avoid using `+` operator for String concatenation, consider using `String.format` or `formatted` if upgraded to Java 15 or above.

log2akshat (changes requested):

store/src/java/com/zimbra/cs/mailbox/MailSender.java
shubhamCS03: done

store/src/java/com/zimbra/cs/mailbox/MailSender.java
shubhamCS03: done

store/src/java/com/zimbra/cs/service/util/SecretKey.java
shubhamCS03: done

store/src/java/com/zimbra/cs/service/util/SecretKey.java
shubhamCS03: done

common/src/java/com/zimbra/common/service/ServiceException.java
shubhamCS03: done

store/src/java/com/zimbra/cs/service/admin/GenerateSecretKey.java
shubhamCS03: We can’t use `Try-with-Resources` as SoapProvisioning does not implement the AutoCloseable interface, which is a requirement for objects to be used in Try-with-Resources.

store/src/java/com/zimbra/cs/service/util/SecretKey.java
shubhamCS03: done

store/src/java/com/zimbra/cs/service/util/SecretKey.java
shubhamCS03: Done

store/src/java/com/zimbra/cs/service/util/SecretKey.java
shubhamCS03: this is not required now

store/src/java/com/zimbra/cs/service/util/SecretKey.java
shubhamCS03: done

store/src/java/com/zimbra/cs/mailbox/MailSender.java
shubhamCS03: done

store/src/java/com/zimbra/cs/account/callback/GenerateSecretKeyCallback.java
shubhamCS03: done

soap/src/java/com/zimbra/soap/mail/message/MailRecallResponse.java
```diff
@@ -0,0 +1,65 @@
...
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ * ***** END LICENSE BLOCK *****
+ */
+
+package com.zimbra.soap.mail.message;
+
+
```
shrutig0510: remove extra space

store/src/java/com/zimbra/cs/mailbox/MailSender.java
shrutig0510: Use java 8 practices
` Optional.ofNullable(SecretKey.getMessageVerificationHeaderValue(
        mm.getMessageID(),
        mailDateFormat.format(date),
        from.getAddress()
)).ifPresent(value -> mm.addHeader(X_MESSAGE_VERIFICATION, value));
`

soap/src/java/com/zimbra/soap/mail/message/MailRecallResponse.java
shubhamCS03: done

store/src/java/com/zimbra/cs/mailbox/MailSender.java
shubhamCS03: I tried that but encountered an implementation issue because the lambda expression used in Optional.ofNullable(...).ifPresent(...) cannot handle checked exceptions like MessagingException. This is a limitation of Java's lambda expressions, which don't natively support checked exceptions.

shubhamCS03: done

shrutig0510 (approved):

soap/src/java/com/zimbra/soap/JaxbUtil.java
```diff
@@ -1681,7 +1687,7 @@ public static ModifyItemNotification getModifiedItemSOAP(BaseItemInfo mod, int r
         ImapMessageInfo messageInfo = new ImapMessageInfo(mod.getIdInMailbox(), mod.getImapUid(), mod.getMailItemType().toString(), mod.getFlagBitmask(), tags);
         return new ModifyNotification.ModifyItemNotification(messageInfo, reason);
     }
-
```
chitranshu198, line 1686: revert this change

soap/src/java/com/zimbra/soap/mail/message/MailRecallRequest.java
```diff
@@ -0,0 +1,39 @@
...
+import com.zimbra.common.soap.MailConstants;
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlAccessorType(XmlAccessType.NONE)
+@XmlRootElement(name = MailConstants.E_MAIL_RECALL_REQUEST)
+public class MailRecallRequest {
+    @XmlAttribute(name = MailConstants.ITEM_ID /* more */, required = true)
```
chitranshu198:
```suggestion
    @XmlAttribute(name = MailConstants.ITEM_ID, required = true)
```

store/src/java/com/zimbra/cs/account/callback/GenerateSecretKeyCallback.java
```diff
@@ -0,0 +1,67 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2023 Synacor, Inc.
```
chitranshu198:
```suggestion
 * Copyright (C) 2025 Synacor, Inc.
```

chitranshu198: make this change in all new files

store/src/java/com/zimbra/cs/account/ZAttrConfig.java
```diff
@@ -18823,6 +18823,78 @@ public Map<String,Object> unsetFeatureMailRecallEnabled(Map<String,Object> attrs
...
+     *
+     * @param zimbraFeatureMailRecallSecretKey new value
+     * @param attrs existing map to populate, or null to create a new map
+     * @return populated map to pass into Provisioning.modifyAttrs
+     *
+     * @since ZCS 10.1.5
+     */
+    @ZAttr(id=4136)
+    public Map<String,Object> setFeatureMailRecallSecretKey(String zimbraFeatureMailRecallSecretKey, Map<String,Object> attrs) {
+        if (attrs == null) attrs = new HashMap<String,Object>();
```
mamtadahiwal, line 18864: if (attrs == null) {
    attrs = new HashMap<String, Object>();
}
```suggestion
        if (attrs == null) attrs = new HashMap<String,Object>();
```

mamtadahiwal, line 18864:
```suggestion
      if (attrs == null) {
    attrs = new HashMap<String, Object>();
}
```

store/src/java/com/zimbra/cs/account/ZAttrConfig.java
```diff
@@ -18823,6 +18823,78 @@ public Map<String,Object> unsetFeatureMailRecallEnabled(Map<String,Object> attrs
...
+     * Secret key used in Mail Recall to make it more secure from spoof.
+     *
+     * @param zimbraFeatureMailRecallSecretKey new value
+     * @throws com.zimbra.common.service.ServiceException if error during update
+     *
+     * @since ZCS 10.1.5
+     */
+    @ZAttr(id=4136)
+    public void setFeatureMailRecallSecretKey(String zimbraFeatureMailRecallSecretKey) throws com.zimbra.common.service.ServiceException {
+        HashMap<String,Object> attrs = new HashMap<String,Object>();
```
mamtadahiwal, line 18848: variable name should be proper , while reading code we should understand it is map like -->  attributeMap

store/src/java/com/zimbra/cs/account/ZAttrConfig.java
```diff
@@ -18823,6 +18823,78 @@ public Map<String,Object> unsetFeatureMailRecallEnabled(Map<String,Object> attrs
...
+    /**
+     * Secret key used in Mail Recall to make it more secure from spoof.
+     *
+     * @throws com.zimbra.common.service.ServiceException if error during update
+     *
+     * @since ZCS 10.1.5
+     */
+    @ZAttr(id=4136)
+    public void unsetFeatureMailRecallSecretKey() throws com.zimbra.common.service.ServiceException {
+        HashMap<String,Object> attrs = new HashMap<String,Object>();
```
mamtadahiwal, line 18878: change variable name

store/src/java/com/zimbra/cs/account/ZAttrConfig.java
```diff
@@ -18823,6 +18823,78 @@ public Map<String,Object> unsetFeatureMailRecallEnabled(Map<String,Object> attrs
...
+     * Secret key used in Mail Recall to make it more secure from spoof.
+     *
+     * @param attrs existing map to populate, or null to create a new map
+     * @return populated map to pass into Provisioning.modifyAttrs
+     *
+     * @since ZCS 10.1.5
+     */
+    @ZAttr(id=4136)
+    public Map<String,Object> unsetFeatureMailRecallSecretKey(Map<String,Object> attrs) {
+        if (attrs == null) attrs = new HashMap<String,Object>();
```
mamtadahiwal, line 18893: add { } after if

store/src/java/com/zimbra/cs/account/callback/GenerateSecretKeyCallback.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public void preModify(CallbackContext context, String attrName, Object value,
+            Map attrsToModify, Entry entry) throws ServiceException {
+        try {
+            Optional.ofNullable(value)
+                    .map(Object::toString)
+                    .filter("TRUE"::equals)
+                    .ifPresent(this::generateMailRecallSecretKey);
+        } catch (RuntimeException e) {
```
mamtadahiwal, line 43: can we catch particular exception instead of generic

store/src/java/com/zimbra/cs/account/callback/GenerateSecretKeyCallback.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    private void generateMailRecallSecretKey(String value) {
+        try {
+            String secretKey = Provisioning.getInstance()
+                    .getConfig()
+                    .getFeatureMailRecallSecretKey();
+
+            if (Strings.isNullOrEmpty(secretKey)) {
```
mamtadahiwal, line 54: what if secretKey = "  " , because nullOrEmpty checks only null and ""

store/src/java/com/zimbra/cs/service/admin/GenerateSecretKey.java
```diff
@@ -0,0 +1,56 @@
...
+import com.zimbra.cs.account.soap.SoapProvisioning;
+import com.zimbra.cs.service.util.SecretKey;
+import com.zimbra.soap.ZimbraSoapContext;
+import com.zimbra.soap.admin.type.CacheEntryType;
+
+import java.util.Map;
+
+public class GenerateSecretKey extends AdminDocumentHandler {
+    public Element handle(Element request, Map<String, Object> context) throws ServiceException {
+        ZimbraSoapContext zsc = getZimbraSoapContext(context);
```
mamtadahiwal, line 34: rename reference from zsc to zimbraSoapContext

store/src/java/com/zimbra/cs/service/admin/GenerateSecretKey.java
```diff
@@ -0,0 +1,56 @@
...
+        Provisioning.getInstance().getConfig().setFeatureMailRecallSecretKey(randomString);
+        flushCache();
+
+        Element response = zsc.createElement(AdminConstants.GENERATE_SECRET_KEY_RESPONSE);
+        return response;
+    }
+
+    private void flushCache() {
+        try {
+            SoapProvisioning sp = new SoapProvisioning();
```
mamtadahiwal: rename reference from sp to  soapProvisioning

store/src/java/com/zimbra/cs/service/util/SecretKey.java
```diff
@@ -0,0 +1,97 @@
...
+     */
+    public static String generateRandomString() throws ServiceException {
+        try {
+            SecureRandom random = new SecureRandom();
+            byte[] key = new byte[KEY_SIZE_BYTES];
+            random.nextBytes(key);
+            return Base64.getEncoder().encodeToString(key);
+        } catch (SecurityException e) {
+            throw ServiceException.FAILURE("Security exception occurred while initializing SecureRandom", e);
+        } catch (Exception e) {
```
mamtadahiwal: try to catch particular exception

store/src/java/com/zimbra/cs/service/util/SecretKey.java
```diff
@@ -0,0 +1,97 @@
...
+     * provide the Hash for Message-Verification header field.
+     *
+     * @param id
+     * @param date
+     * @param from
+     * @throws ServiceException
+     */
+    public static String getMessageVerificationHeaderValue(String id, String date, String from) throws MessagingException, ServiceException {
+        String secretKey = Provisioning.getInstance().getConfig().getFeatureMailRecallSecretKey();
+        StringBuilder sb = new StringBuilder();
```
mamtadahiwal: rename reference from sb to stringBuilder

store/src/java/com/zimbra/cs/service/util/SecretKey.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Create a digest of the given input with the given algorithm.
+     *
+     * @param input
+     * @throws ServiceException
+     */
+    private static String getHashForMessageVerification(String input) throws ServiceException {
+        try {
+            MessageDigest md = MessageDigest.getInstance(MSGVRFY_ALGORITHM_NAME);
```
mamtadahiwal: rename refernce from md to MessageDigest

soap/src/java/com/zimbra/soap/mail/message/MailRecallRequest.java
shubhamCS03: done

store/src/java/com/zimbra/cs/account/callback/GenerateSecretKeyCallback.java
shubhamCS03: done

store/src/java/com/zimbra/cs/account/ZAttrConfig.java
shubhamCS03: System generated code, we can’t modify it manually.

store/src/java/com/zimbra/cs/account/ZAttrConfig.java
shubhamCS03: System generated code, we can’t modify it manually.

store/src/java/com/zimbra/cs/account/ZAttrConfig.java
shubhamCS03: System generated code, we can’t modify it manually.

store/src/java/com/zimbra/cs/account/ZAttrConfig.java
shubhamCS03: System generated code, we can’t modify it manually.

store/src/java/com/zimbra/cs/service/admin/GenerateSecretKey.java
shubhamCS03: done

store/src/java/com/zimbra/cs/service/admin/GenerateSecretKey.java
shubhamCS03: I can’t rename as this naming pattern is used in our project.

store/src/java/com/zimbra/cs/service/util/SecretKey.java
shubhamCS03: done

store/src/java/com/zimbra/cs/service/util/SecretKey.java
shubhamCS03: we can’t as this name is already used in below line.

store/src/java/com/zimbra/cs/service/util/SecretKey.java
shubhamCS03: done

store/src/java/com/zimbra/cs/account/callback/GenerateSecretKeyCallback.java
shubhamCS03: done

store/src/java/com/zimbra/cs/service/util/SecretKey.java
mamtadahiwal: is this system generated code?

store/src/java/com/zimbra/cs/account/callback/GenerateSecretKeyCallback.java
shubhamCS03: No, we can’t as generateMailRecallSecretKey method is throwing RuntimeException because Optional.ifPresent cannot handle checked exceptions directly.

mamtadahiwal (approved):

chitranshu198 (approved):

shrutig0510 (approved):

PR #1475, opened 2023-04-11
ZCS-12912: LDAP global attributes latest value not available across node

swaatiTelus:
Problem ([ZCS-12912](https://jira.corp.synacor.com/browse/ZCS-12912)):
When we add or delete external S3 volume on one server, the LDAP global attribute `zimbraGlobalExternalStoreConfig` value is not updated across other mail servers. So on other servers, we have to execute `zmprov fc all`, to get the updated value of LDAP reflected.

Solution:
The cache should be flushed at _config level_ at the time of creating and deleting the S3 Volume. Cache should be  flushed for all the servers other than the server where changes were made, so that the updated LDAP value is reflected among all the servers right away.

Fix:
Executing a flushCache method after each S3 external volume addition and deletion.

Testing:
Done testing on multi-node server after adding and deleting multiple volumes.

Other PR: https://github.com/Zimbra/zm-hsm-store/pull/42

### DIFF
--- a/store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
+++ b/store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
@@ -6,6 +6,8 @@
  */
 package com.zimbra.util;
 
+import java.util.Map;
+
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
@@ -17,6 +19,11 @@ import com.zimbra.common.soap.GlobalExternalStoreConfigConstants;
 import com.zimbra.common.util.StringUtil;
 import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.cs.account.Provisioning;
+import com.zimbra.cs.service.admin.AdminDocumentHandler;
+import com.zimbra.cs.service.admin.FlushCache;
+import com.zimbra.soap.admin.message.FlushCacheRequest;
+import com.zimbra.soap.admin.type.CacheEntryType;
+import com.zimbra.soap.admin.type.CacheSelector;
 import com.zimbra.soap.admin.type.VolumeExternalInfo;
 import com.zimbra.soap.admin.type.VolumeExternalOpenIOInfo;
 import com.zimbra.soap.admin.type.VolumeInfo;
@@ -72,7 +79,7 @@ public class ExternalVolumeInfoHandler {
     /**
      * Delete server level external volume properties
      * @param volumeId
-     * @param server
+     * @param serverId
      * @throws JSONException, ServiceException
      */
     public void deleteServerProperties(int volumeId, String serverId) throws ServiceException, JSONException {
@@ -81,6 +88,7 @@ public class ExternalVolumeInfoHandler {
             String serverExternalStoreConfigJson = provisioning.getLocalServer().getServerExternalStoreConfig();
             JSONObject currentJsonObject = new JSONObject(serverExternalStoreConfigJson);
             JSONArray currentJsonArray = currentJsonObject.getJSONArray("server/stores");
+            boolean isUnifiedVolume = isUnifiedVolume(volumeId);
 
             // step 2: Create new/empty updated JSON state array
             JSONArray updatedJsonArray = new JSONArray();
@@ -102,8 +110,10 @@ public class ExternalVolumeInfoHandler {
             // step 6: Set ldap attribute
             provisioning.getLocalServer().setServerExternalStoreConfig(currentJsonObject.toString());
 
-            editGlobalConfigOnDeleteVolume(volumeId, serverId);
-
+            //if volume is unified, then only we should call the below method to modify global external store
+            if (isUnifiedVolume) {
+                editGlobalConfigOnDeleteVolume(volumeId, serverId);
+            }
         } catch (JSONException e) {
             throw e;
         }
@@ -111,8 +121,8 @@ public class ExternalVolumeInfoHandler {
 
     /**
      * Add server level external volume properties
-     * @param volumeId
-     * @param server
+     * @param volInfo
+     * @param serverId
      * @throws JSONException, ServiceException
      */
     public void addServerProperties(VolumeInfo volInfo, String serverId) throws ServiceException, JSONException {
@@ -158,7 +168,6 @@ public class ExternalVolumeInfoHandler {
                     && volInfo.getVolumeExternalInfo().isUnified()) {
                 editGlobalConfigOnAddVolume(volInfo, serverId);
             }
-
         } catch (JSONException e) {
             throw e;
         }
@@ -274,6 +283,7 @@ public class ExternalVolumeInfoHandler {
            }
         }
     }
+
     /**
      * Modify server level external volume properties
      * @param volumeId
@@ -359,5 +369,42 @@ public class ExternalVolumeInfoHandler {
             }
         }
         return false;
+    }
+
+    /**
+     * Performs flush cache at config level on all the server
+     * @param adminDocumentHandler, context
+     * @return nothing
+     */
+    public static void flushConfigLevelCacheOnAllServers(AdminDocumentHandler adminDocumentHandler, Map<String, Object> context) {
+        CacheSelector cacheSelector = new CacheSelector(true, CacheEntryType.config.toString());
+        try {
+            FlushCache.doFlushCache(adminDocumentHandler, context, new FlushCacheRequest(cacheSelector));
+        } catch (ServiceException se) {
+            ZimbraLog.misc.error("Encountered exception during FlushCache: ", se);
+        }
+    }
+
+    /**
+     * Checks if volume is unified or not
+     * @param volumeId
+     * @return true if volume is unified else false
+     */
+    public boolean isUnifiedVolume(int volumeId){
+        boolean isUnifiedVolume = false;
+        try {
+            JSONObject currentJsonObject = new JSONObject(provisioning.getLocalServer().getServerExternalStoreConfig());
+            JSONArray serverStoreJsonArray = currentJsonObject.getJSONArray("server/stores");
+            for (int i = 0; i < serverStoreJsonArray.length(); i++) {
+                JSONObject tempJsonObj = serverStoreJsonArray.getJSONObject(i);
+                if (volumeId == tempJsonObj.getInt(AdminConstants.A_VOLUME_ID)) {
+                    isUnifiedVolume = Boolean.valueOf(tempJsonObj.optString(AdminConstants.A_VOLUME_UNIFIED));
+                    break;
+                }
+            }
+        } catch (JSONException | ServiceException e) {
+            ZimbraLog.misc.error("Failed to find the unified volume to perform flush cache: ", e);
+        }
+        return isUnifiedVolume;
     }
 }
--- a/store/src/java/com/zimbra/cs/service/admin/CreateVolume.java
+++ b/store/src/java/com/zimbra/cs/service/admin/CreateVolume.java
@@ -35,6 +35,7 @@ import com.zimbra.soap.admin.message.CreateVolumeRequest;
 import com.zimbra.soap.admin.message.CreateVolumeResponse;
 import com.zimbra.soap.admin.type.StoreManagerRuntimeSwitchResult;
 import com.zimbra.soap.admin.type.VolumeInfo;
+import com.zimbra.util.ExternalVolumeInfoHandler;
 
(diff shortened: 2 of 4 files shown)


store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -157,13 +164,52 @@ public void addServerProperties(VolumeInfo volInfo, String serverId) throws Serv
...
+     */
+    private void flushCache(ZimbraSoapContext zsc) {
+        //In this option we have the dependency on zcs parameter.
+        try {
+          CacheSelector cacheSelector = new CacheSelector();
+          cacheSelector.setTypes(CacheEntryType.config.name());
+          FlushCache.flushConfigCacheOnAllServers(zsc, new FlushCacheRequest(cacheSelector));
+        } catch (ServiceException e) {
+          ZimbraLog.account.info("Encountered exception during FlushCache after creating unified volume" + e.getMessage());
+          e.printStackTrace();
```
RakeshAMore: Use logger instead

RakeshAMore:
As discussed please make changes accordingly.

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -271,9 +317,11 @@ public void editGlobalConfigOnDeleteVolume(int volumeId, String serverId) throws
                    globalExtStoreConfJson.put(GlobalExternalStoreConfigConstants.A_S3_UNIFIED_VOLUME, unifiedJsonObject);
                    Provisioning.getInstance().getConfig().setGlobalExternalStoreConfig(globalExtStoreConfJson.toString());
                 }
+                //flushCache(zsc);
```
abhishekdeshmukh2209: commented code

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -157,13 +164,52 @@ public void addServerProperties(VolumeInfo volInfo, String serverId) throws Serv
                     && AdminConstants.A_VOLUME_S3.equalsIgnoreCase(volInfo.getVolumeExternalInfo().getStorageType())
                     && volInfo.getVolumeExternalInfo().isUnified()) {
                 editGlobalConfigOnAddVolume(volInfo, serverId);
+                //flushCache(zsc);
```
abhishekdeshmukh2209: commented code

store/src/java/com/zimbra/cs/service/admin/FlushCache.java
```diff
@@ -328,4 +328,10 @@ private static String sanitizeImapCacheTypes(String cacheTypes) {
         }
         return imapTypes.isEmpty() ? null : Joiner.on(",").join(imapTypes);
     }
+
```
abhishekdeshmukh2209: extra lines

swaatiTelus: removed.

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
swaatiTelus: The approach was under discussion in the draft PR. Uncommented it. Thanks!

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
swaatiTelus: Uncommented it.

ronstra-synacor (approved):

store/src/java/com/zimbra/cs/service/admin/FlushCache.java
```diff
@@ -328,4 +328,9 @@ private static String sanitizeImapCacheTypes(String cacheTypes) {
...
+    public static void flushConfigCacheOnAllServers(ZimbraSoapContext zsc, FlushCacheRequest req) throws ServiceException {
```
RakeshAMore: Since it's named to flush the cache of the config only then how will it prevent another type of cache flush?

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -157,13 +163,28 @@ public void addServerProperties(VolumeInfo volInfo, String serverId) throws Serv
...
             throw e;
         }
     }
 
+    /**
+     * Flushes config level cache of all other server
+     * @param zsc
+     * @return nothing
+     */
+    public void flushCache(ZimbraSoapContext zsc) {
```
RakeshAMore: I would name it to specific attributes

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -157,13 +163,28 @@ public void addServerProperties(VolumeInfo volInfo, String serverId) throws Serv
...
+        try {
+          CacheSelector cacheSelector = new CacheSelector();
+          cacheSelector.setTypes(CacheEntryType.config.toString());
+          FlushCache.flushConfigCacheOnAllServers(zsc, new FlushCacheRequest(cacheSelector));
+        } catch (ServiceException e) {
+          ZimbraLog.account.info("Encountered exception during FlushCache after creating unified volume " + e.getMessage());
```
RakeshAMore: ZimbraLog.account.info("Encountered exception during FlushCache after creating unified volume ", e);

swaatiTelus: done

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
swaatiTelus: Renamed method name to flushConfigLevelCacheOnAllServers().

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
swaatiTelus: Done.

store/src/java/com/zimbra/cs/service/admin/FlushCache.java
swaatiTelus: Yes right, so now I have moved the logic of defining the cache type inside this method itself. Thanks!

store/src/java/com/zimbra/cs/service/admin/FlushCache.java
```diff
@@ -328,4 +328,16 @@ private static String sanitizeImapCacheTypes(String cacheTypes) {
...
         return imapTypes.isEmpty() ? null : Joiner.on(",").join(imapTypes);
     }
+
+    /**
+     * Flushes config level cache of all other server
+     * @param zimbraSoapContext
+     * @return nothing
+     * @throws ServiceException
+     */
+    public static void flushConfigCacheOnAllServers(ZimbraSoapContext zimbraSoapContext) throws ServiceException {
```
silentsakky: not sure if we should start adding different methods based on type of cache which needs to be flushed
any caller can do this to flush cache on all servers
```
                CacheSelector cacheSel = new CacheSelector(true /* allServers */, CacheEntryType.config.toString());
                FlushCacheRequest fcReq = new FlushCacheRequest(cacheSel);
                try {
                    FlushCache.doFlushCache(this, context, fcReq);
                } catch (ServiceException se) {
                    // Error log
                }
```

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -102,7 +107,10 @@ public void deleteServerProperties(int volumeId, String serverId) throws Service
             // step 6: Set ldap attribute
             provisioning.getLocalServer().setServerExternalStoreConfig(currentJsonObject.toString());
 
-            editGlobalConfigOnDeleteVolume(volumeId, serverId);
+            //if volume is unified, then only we should call the below method to modify global external store
+            if(isUnified) {
```
dasiyogesh:
```suggestion
            if (isUnified) {
```

store/src/java/com/zimbra/cs/service/admin/FlushCache.java
RakeshAMore: It is better to restrict it to config only because flushing another type of cache on all servers may have an impact on the performance so it's better to not add if this is not required :)(and if we have to add then impact should be analyzed before.)

silentsakky: if you check my code then it's restricting to config cache only :)

RakeshAMore: Got it. Swati will make changes accordingly. Thanks

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -102,15 +113,19 @@ public void deleteServerProperties(int volumeId, String serverId) throws Service
             // step 6: Set ldap attribute
             provisioning.getLocalServer().setServerExternalStoreConfig(currentJsonObject.toString());
 
-            editGlobalConfigOnDeleteVolume(volumeId, serverId);
-
+            //if volume is unified, then only we should call the below method to modify global external store
+            if (isUnifiedVolume) {
+                editGlobalConfigOnDeleteVolume(volumeId, serverId);
```
silentsakky, line 115: can't we call flushConfigLevelCacheOnAllServers from here itself? we are changing function signatures to return isUnifiedVolume flag and at the end based on that flag calling flushConfigLevelCacheOnAllServers, which can be done here also

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -360,4 +376,18 @@ public Boolean isVolumePresentInJson(int volumeId) throws ServiceException, JSON
...
+     * Performs flush cache at config level on all the server
+     * @param adminDocumentHandler, context
+     * @return nothing
+     */
+    public static void flushConfigLevelCacheOnAllServers(AdminDocumentHandler adminDocumentHandler, Map<String, Object> context) {
+        CacheSelector cacheSelector = new CacheSelector(true, CacheEntryType.config.toString());
+        try {
+            FlushCache.doFlushCache(adminDocumentHandler, context, new FlushCacheRequest(cacheSelector));
+        } catch (ServiceException se) {
+            ZimbraLog.account.error("Encountered exception during FlushCache: ", se);
```
silentsakky: not sure we should log to account category as this operation is not related to any account

swaatiTelus: Done. Changed it to `misc` as used in FlushCache utility.

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
swaatiTelus: Yes right, we could have done it by 2 ways: 
1. Carry the AdminDocumentHandler reference as a parameter from DeleteVolume API up to here to handler class. and
2. to send back the boolean value from handler.
I opted for option 2, to maintain the uniformity with other API, as we are calling `flushConfigLevelCacheOnAllServers` method from API's itself.

Please suggest if I should change it.?

silentsakky: ok I missed the handler part, option 1 is ok

silentsakky (approved):

silentsakky (approved):

store/src/java/com/zimbra/util/ExternalVolumeInfoHandler.java
```diff
@@ -360,4 +370,41 @@ public Boolean isVolumePresentInJson(int volumeId) throws ServiceException, JSON
...
+    /**
+     * Checks if volume is unified or not
+     * @param volumeId
+     * @return true if volume is unified else false
+     */
+    public boolean isUnifiedVolume(int volumeId){
+        boolean isUnifiedVolume = false;
+        try {
+            JSONObject currentJsonObject = new JSONObject(provisioning.getLocalServer().getServerExternalStoreConfig());
+            JSONArray serverStoreJsonArray = currentJsonObject.getJSONArray("server/stores");
```
vishnukantjangid, line 397: We should use constant for the key "server/stores"

zimsuchitgupta (approved):

vishnukantjangid (approved):

PR #1476, opened 2023-04-20
ZCS-10630: Replaced zimbraDataSourceHost www.googleapis.com with apid…

zimsuchitgupta:
**Problem**
Google calendar not syncing

**Reason** 
https://www.googleapis.com/.well-known/caldav is the endpoint which is currently used to fetch current-user-principal via PROPFIND. For this request endpoint we were seeing 404 error.

But when changing base URL to https://apidata.googleusercontent.com/ and when tried to fetch current-user-principal with PROPFIND endpoint(https://apidata.googleusercontent.com/caldav/v2) response was coming properly.
So seems like we have to change/update the base URL and respective endpoints to resolve this exception.

### DIFF
--- a/store/src/java/com/zimbra/cs/datasource/CalDavDataImport.java
+++ b/store/src/java/com/zimbra/cs/datasource/CalDavDataImport.java
@@ -154,7 +154,10 @@ public class CalDavDataImport extends MailItemImport {
             url.append("http://");
         break;
         }
-        url.append(ds.getHost()).append(":").append(ds.getPort());
+        url.append(ds.getHost());
+        if (ds.getPort() != null) {
+            url.append(":").append(ds.getPort());
+        }
         return url.toString();
     }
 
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/datasource/CalDavDataImport.java
```diff
@@ -154,7 +154,10 @@ protected String getTargetUrl() {
             url.append("http://");
         break;
         }
-        url.append(ds.getHost()).append(":").append(ds.getPort());
+        url.append(ds.getHost());
```
dasiyogesh, line 157: is there a port needed? may be this was earlier requirement for endpoint.

preshitaagrawal (approved):

store/src/java/com/zimbra/cs/datasource/CalDavDataImport.java
zimsuchitgupta: Yes, this port is required here. I think it may be used by other services.

zimsuchitgupta:
No need of these changes, issue has been fixed by doing changes in zm-oauth-social repo only.
Closing this PR.

PR #1480, opened 2023-04-28
ZCS-13340:Add an LDAP attribute to force users not to use username in…

shubhamCS03:
Requirement of ticket [[ZCS-13340](https://jira.corp.synacor.com/browse/ZCS-13340)]:

Add a new LDAP attribute `zimbraFeatureAllowUsernameInPassword` to restrict/allow the use of usernames in the password when users reset or change their passwords to achieve greater password security.

Solution:

One new attributes `zimbraFeatureAllowUsernameInPassword` is added and their setters/getters are generated.

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -10375,4 +10375,8 @@ TODO: delete them permanently from here
   <globalConfigValue>TRUE</globalConfigValue>
   <desc>Whether to enable/disable modern ui option on the login screen.</desc>
 </attr>
+<attr id="4098" name="zimbraFeatureAllowUsernameInPassword" type="boolean" cardinality="single" optionalIn="account,cos,domain" flags="accountInfo,accountCosDomainInherited" since="10.1.0">
+  <defaultCOSValue>TRUE</defaultCOSValue>
+  <desc>To restrict/allow the use of username in the password when user reset or change their password to achieve greater password security.</desc>
+</attr>
 </attrs>
--- a/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
+++ b/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
@@ -6706,6 +6706,15 @@ public class ZAttrProvisioning {
     public static final String A_zimbraFeatureAdvancedSearchEnabled = "zimbraFeatureAdvancedSearchEnabled";
 
     /**
+     * To restrict/allow the use of username in the password when user reset
+     * or change their password to achieve greater password security.
+     *
+     * @since ZCS 10.1.0
+     */
+    @ZAttr(id=4098)
+    public static final String A_zimbraFeatureAllowUsernameInPassword = "zimbraFeatureAllowUsernameInPassword";
+
+    /**
      * whether or not to enable rerouting spam messages to Junk folder in
      * ZCS, exposing Junk folder and actions in the web UI, and exposing Junk
      * folder to IMAP clients.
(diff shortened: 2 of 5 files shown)


store/conf/attrs/zimbra-attrs.xml
```diff
@@ -10357,4 +10357,9 @@ TODO: delete them permanently from here
   <globalConfigValue>TRUE</globalConfigValue>
   <desc>whether to enable CountObjectsRequest with userAccount/account type on Zimbra Admin Console</desc>
 </attr>
+
+<attr id="4098" name="zimbraFeatureAllowUsernameInPassword" type="boolean" cardinality="single" optionalIn="account,cos,domain" flags="accountInfo,accountCosDomainInherited" since="10.1.0">
```
zimsuchitgupta, line 10378: I can see last id for attribute is 4096 so next should be 4097. Any specific reason to not choose this?

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -10357,4 +10357,9 @@ TODO: delete them permanently from here
...
+  <defaultCOSValue>TRUE</defaultCOSValue>
+  <desc>To restrict/allow the use of usernames in the password when users reset or change their passwords to achieve greater password security.</desc>
```
zimsuchitgupta:
```suggestion
  <desc>To restrict/allow the use of username in the password when user reset or change their password to achieve greater password security.</desc>
```

zimsuchitgupta (changes requested):

ronstra-synacor (approved):

swaatiTelus (approved):

store/conf/attrs/zimbra-attrs.xml
shubhamCS03: Suchit In this PR[https://github.com/Zimbra/zm-mailbox/pull/1468/files] we are using the above Id. It will be merged soon.

zimsuchitgupta: Ideally if someone is adding a new attribute then next incremented id should be in use irrespective of how many developers are working on same work. Whoever is going to merge code first verify the attribute id and if it conflict update and push the code and then merge.

store/conf/attrs/zimbra-attrs.xml
zimsuchitgupta: Please run   ant generate-getters again so that description part can be updated for all files.

store/conf/attrs/zimbra-attrs.xml
shrutig0510, line 10378: We also have same type of LC attribute in LC.java 
public static final KnownKey allow_username_within_password = KnownKey.newKey(true);

Why is separate LDAP attribute needed again?

dasiyogesh: for Z8 & Z9 LDAP attrs cannot be updated in patch so released as LC and now this ldap attr can be released in major release for Z10. LC needs updating on all mailbox nodes but ldap attr can be updated globally.

shrutig0510 (approved):

store/conf/attrs/zimbra-attrs.xml
shubhamCS03: Done

zimsuchitgupta (approved):

store/conf/attrs/zimbra-attrs.xml
silentsakky, line 10378: a question related to requirements, should we remove LC value now and add some migration step to move LC value to ldap attribute so customers don't loose data in migration/upgrade

swaatiTelus: Saket, under this PR we have just introduced the placeholder for LDAP attribute so it can be shipped to next release. the implementation will be done as a part of ZCS-13218, where in we should take care of the above case you mentioned.

silentsakky: Thanks

PR #1489, opened 2023-05-18
ZBUG-3204: Webclient and AdminConsole both are not loading due to zim…

vishnukantjangid:
**Issue:** WebClient and AdminConsole both are not loading if we set the value of zimbraFileUploadMaxSize greater than 2147483647

**Solution:** Modify the value of file-size-threshold with 1048576 instead of zimbraFileUploadMaxSize.

- **fileSizeThreshold**: The file size in bytes after which the file will be temporarily stored on disk.

### DIFF
--- a/store/conf/web.xml.production
+++ b/store/conf/web.xml.production
@@ -222,7 +222,8 @@
       <location>/opt/zimbra/data/tmp</location>
       <max-file-size>%%zimbraFileUploadMaxSize%%</max-file-size>
       <max-request-size>%%zimbraMailContentMaxSize%%</max-request-size>
-      <file-size-threshold>%%zimbraFileUploadMaxSize%%</file-size-threshold>
+      <!-- The file size in bytes after which the file will be temporarily stored on disk -->
+      <file-size-threshold>1048576</file-size-threshold>
     </multipart-config>
   </servlet>
 


store/conf/web.xml.production
```diff
@@ -222,7 +222,8 @@
       <location>/opt/zimbra/data/tmp</location>
       <max-file-size>%%zimbraFileUploadMaxSize%%</max-file-size>
       <max-request-size>%%zimbraMailContentMaxSize%%</max-request-size>
-      <file-size-threshold>%%zimbraFileUploadMaxSize%%</file-size-threshold>
+      <!-- The file size in bytes after which the file will be temporarily stored on disk -->
```
silentsakky, line 225: same issue could happen with zimbraFileUploadMaxSize or zimbraMailContentMaxSize

vishnukantjangid: No, it will not happen we have long supported value for them.

silentsakky: ok should we make this value configurable? or is it not required?

vishnukantjangid: It is just a threshold value and by default is 0 but we are taking 1MB here to save some disk I/O and It will not require to change so I think it's not required. 
what is your opinion?

silentsakky: I am ok if we feel this should not be configurable, I was just curious.

silentsakky (approved):

preshitaagrawal (approved):

shrutig0510 (approved):

dasiyogesh (approved):

zimsuchitgupta (approved):

PR #1491, opened 2023-05-31
ZBUG-3376: Partially resolved ZBUG-2891

HimanshuSojitraSynacor:
**Issue**: For users with language set to German, the emails notifications for Calendar folders share/revoke displays the word "calendar" instead of "Kalender".
**Solution**: Add translation for the word Calendar.

### DIFF
--- a/store/src/java/com/zimbra/cs/util/AccountUtil.java
+++ b/store/src/java/com/zimbra/cs/util/AccountUtil.java
@@ -29,6 +29,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 
@@ -41,6 +42,7 @@ import javax.mail.internet.MimeBodyPart;
 import javax.mail.internet.MimeMessage;
 import javax.mail.internet.MimeMultipart;
 
+import com.zimbra.common.util.L10nUtil;
 import org.apache.commons.codec.binary.Hex;
 
 import com.sun.mail.smtp.SMTPMessage;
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/util/AccountUtil.java
```diff
@@ -1042,4 +1044,19 @@ public static void isAdminAccount(Account acct) throws ServiceException {
...
+     * Utils method to translate system folder name
+     * @param name
+     * @param locale
+     * @return
+     */
+    public static String getTranslatedFolderName(String name, Locale locale) {
+        if ("Calendar".equals(name)) {
+            return L10nUtil.getMessage(L10nUtil.MsgKey.calendar, locale);
+        } else if ("Trash".equals(name)) {
+            return L10nUtil.getMessage(L10nUtil.MsgKey.calendar, locale);
```
silentsakky, line 1058: trash is also using calendar?

store/src/java/com/zimbra/cs/util/AccountUtil.java
```diff
@@ -1042,4 +1044,19 @@ public static void isAdminAccount(Account acct) throws ServiceException {
...
     }
+    
+    /**
+     * Utils method to translate system folder name
+     * @param name
+     * @param locale
+     * @return
+     */
+    public static String getTranslatedFolderName(String name, Locale locale) {
+        if ("Calendar".equals(name)) {
```
silentsakky: can we check on folder id instead of name?

zimsuchitgupta: In 'name' variable we get 'Calendar' as value if operation for Calendar. So id can not used here.

store/src/java/com/zimbra/cs/util/AccountUtil.java
zimsuchitgupta: > trash is also using calendar?

It should be trash as if operation is for 'Trash' folder then variable 'name' will contain value 'Trash' and this should be translated.

store/src/java/com/zimbra/cs/util/AccountUtil.java
HimanshuSojitraSynacor: Can we pass sid when calling this utility, extract the id and check if it matches any of the system folder ids?

ronstra-synacor (approved):

silentsakky (approved):

rcyarrapothu (approved):

zimsuchitgupta (approved):

PR #1500, opened 2023-06-27
ZCS-13512 no need to check for patch while license activation.

shrutig0510:
**Problem:**
License re-activation needed for patch upgrade Ex: while upgrading from 10.0.0 to 10.0.1

**Fix:**
no patch related check from versioning is needed

**Test-cases**
After applying changes, zmlicense -c is ok and no error for license on admin console found

### DIFF
--- a/common/src/java-test/com/zimbra/common/util/VersionTest.java
+++ b/common/src/java-test/com/zimbra/common/util/VersionTest.java
@@ -151,4 +151,17 @@ public class VersionTest {
         }
     }
 
+    @Test
+    public void checkPatchRelease() throws Exception {
+        Version currentVersion = new Version("10.0.0_GA_4518");
+        Version lastVersion = new Version("10.1.0_GA_4549");
+        Assert.assertFalse(currentVersion.isSameMinorRelease(lastVersion));
+        currentVersion = new Version("10.0.0_GA_4518");
+        lastVersion = new Version("11.0.0_GA_4549");
+        Assert.assertFalse(currentVersion.isSameMinorRelease(lastVersion));
+        currentVersion = new Version("10.0.0_GA_4518");
+        lastVersion = new Version("10.0.1_GA_454");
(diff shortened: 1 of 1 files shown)


common/src/java/com/zimbra/common/util/Version.java
```diff
@@ -177,10 +177,6 @@ public int compareTo(Version version) {
         if (r != 0)
             return r;
 
-        r = mPatch - version.mPatch;
```
silentsakky: please update unit tests also zm-mailbox/common/src/java-test/com/zimbra/common/util/VersionTest.java

silentsakky: instead of changing a very generic code which changes behaviour of it, can't we change caller code to accomplish this?
https://github.com/Zimbra/zm-license-tools/blob/develop/src/java/com/zimbra/cs/license/License.java#L490 We can update it to use isSameMinorRelease method instead of this change

shrutig0510: Updated the VersionTest.java according to current scenario. Passing test scenarios snapshot added in the ticket.  I will add new test cases too

shrutig0510: The method changed is used at these places where it has been actively used in zimbra desktop and Zimbra/zm-touch-client repo· https://github.com/search?q=org%3AZimbra%20Version.compare(&type=code
The method will be unused in that case if these repos are not being actively used

silentsakky: touch client is only present in 8815, and this fix will be part of Zimbra 10.x only
about zimbra desktop you are referring to deprecated version of zimbra desktop, which we don't use or release anymore

shrutig0510: Done. Thanks

silentsakky (approved):

gautamdg (approved):

preshitaagrawal (approved):

PR #1510, opened 2023-08-03
ZCS-13306: changed to allow derefGroupMember in GetContactsRequest even when id is not set for print contacts

k-kato:
**Issue:**
When "Print Contacts" is clicked on Classic UI, the detail (name, email etc.) of contact group member(s) is not shown in the print page. Only contact id is shown.
When "Print selected contact(s) is clicked, the detail is shown.

**Root cause:**
When "Print Contacts" is clicked, the access URL is `https://HOST/h/printcontacts?sfi=[folder_id]`. In this case, a member of contact group is not derefed.
When "Print selected contact(s) is clicked the access URL is `https://HOST/h/printcontacts?id=[id1,id2,...]`. `derefGroupMember true` is set through taglib and ZMailbox class. Then a member is derefed and the detail is shown in Print page.

**Fix:**
* `store/src/java/com/zimbra/cs/service/mail/GetContacts.java`
   * change to pass `derefGroupMember` for searching contacts without `id` parameter as well.
   * Note that `derefGroupMember` was not allowed when `id` was not specified in GetContactsRequest. The logic was added when derefGroupMember parameter was added. According to the comment in the code, it was for performance reason.
      https://bugzilla.zimbra.com/show_bug.cgi?id=56772
    * `derefGroupMember` without specifying `id` can be blocked by setting `contact_allow_deref_group_member` to `false` in case the performance issue occurs. In that case, the detail of a contact group member is not shown in Print page.

* `common/src/java/com/zimbra/common/localconfig/LC.java`
    * added new localconfig key `contact_allow_deref_group_member`. Default value is `true`.

* `soap/src/java/com/zimbra/soap/mail/message/GetContactsRequest.java`
    * add a note about the performance 

* `client/src/java/com/zimbra/client/ZMailbox.java`
    * add `boolean forceDerefGroupMember` to `getContactsForFolder`.
    * add an overloading function for backward compatibility

**Related PRs:**
* https://github.com/Zimbra/zm-taglib/pull/49
* https://github.com/Zimbra/zm-web-client/pull/832

### DIFF
--- a/client/src/java/com/zimbra/client/ZMailbox.java
+++ b/client/src/java/com/zimbra/client/ZMailbox.java
@@ -2118,6 +2118,21 @@ public class ZMailbox implements ToZJSONObject, MailboxStore {
      * @throws ServiceException on error
      */
     public List<ZContact> getContactsForFolder(String folderid, String ids, ContactSortBy sortBy, boolean sync, List<String> attrs) throws ServiceException {
+        return getContactsForFolder(folderid, ids, sortBy, sync, attrs, false);
+    }
+
+    /**
+     * Fetch contacts for a given folder or contacts ids
+     * @param folderid folder id of the contact folder
+     * @param ids comma-separated list of contact ids
+     * @param attrs limit attrs returns to given list
+     * @param sortBy sort results (null for no sorting)
+     * @param sync if true, return modified date on contacts
+     * @param forceDerefGroupMember if true and if ids is empty or null, derefGroupMember is set
+     * @return list of contacts
+     * @throws ServiceException on error
+     */
+    public List<ZContact> getContactsForFolder(String folderid, String ids, ContactSortBy sortBy, boolean sync, List<String> attrs, boolean forceDerefGroupMember) throws ServiceException {
         List<ZContact> result = new ArrayList<ZContact>();
         int batchSize = LC.zimbra_activesync_remote_sync_batch_size.intValue();
         if (!StringUtil.isNullOrEmpty(ids)) {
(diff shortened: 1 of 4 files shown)


sitaram-mulik-synacor (approved):

common/src/java/com/zimbra/common/localconfig/LC.java
```diff
@@ -1094,6 +1094,7 @@ public final class LC {
     @Supported
     public static final KnownKey contact_ranking_enabled = KnownKey.newKey(true);
 
+    public static final KnownKey contact_allow_deref_group_member = KnownKey.newKey(true);
```
silentsakky, line 1097: should we by default disable it if it has performance issues?

k-kato: The detail of contact group member is not shown if `contact_allow_deref_group_member` is false.
It is trade-off between contact group print functionality and performance. Performance issue may happen when lots of contact groups exists or contact group(s) has lots of members.
Which value is better as default?

zimsuchitgupta (approved):

client/src/java/com/zimbra/client/ZMailbox.java
```diff
@@ -2129,8 +2144,7 @@ public List<ZContact> getContactsForFolder(String folderid, String ids, ContactS
                 String[] temp = Arrays.copyOfRange(arr, start, end);
                 String subIds = String.join(",", temp);
 
-                Element req = getGetContactsRequest(folderid, sortBy, sync, attrs);
-                req.addAttribute(MailConstants.A_DEREF_CONTACT_GROUP_MEMBER, true);
+                Element req = getGetContactsRequest(folderid, sortBy, sync, attrs, true);
```
dasiyogesh, line 2147: since we have forceDerefGroupMember param available here why are we defaulting this request with true here?

k-kato: `req.addAttribute(MailConstants.A_DEREF_CONTACT_GROUP_MEMBER, true);` is always required when id is specified.
This is just moving the code in previous line 2133 to the new line 2183.

common/src/java/com/zimbra/common/localconfig/LC.java
silentsakky: this LC is only used when we pass derefGroupMember = true in API, which we are doing in case of contact print only so I think it's ok to go ahead with this default value.

silentsakky (approved):

dasiyogesh (approved):

PR #1512, opened 2023-08-17
ZCS-13783 : Part 1- Update locator for external storage | In-place up…

rendurama:
Problem : AFter the zimbra 10 in place upgrade, When we try to read those mail stored on external storage we are getting missing blob because according to Zimbra 10 Storage Management locator values format should be volume_id@@mailbox_id/file_name for external volume but there is only external volume id present in the locator, according to zimbra9 standard.

Solution: 
We update locators in the mail_item table for external volumes only to read blobs from external storage according to Zimbra 10 Storage Management implementation.  
This utility will work with having the below argument. 
**Update locator based on account**

Link to another PR: [#44](https://github.com/Zimbra/zm-modules-porter/pull/44)

### DIFF
--- a/common/src/java/com/zimbra/common/soap/AdminConstants.java
+++ b/common/src/java/com/zimbra/common/soap/AdminConstants.java
@@ -1622,5 +1622,12 @@ public final class AdminConstants {
             "com_zextras_client", "com_zimbra_connect_classic", "com_zimbra_connect_modern", "com_zextras_docs",
             "com_zimbra_docs_modern", "com_zimbra_drive_modern", "com_zextras_drive", "com_zextras_drive_open",
             "com_zextras_chat_open", "com_zextras_talk", "zimbra-zimlet-briefcase-edit-lool");
-
+    public static final String E_ACCOUNT_NAME = "accounts";
+    public static final String E_MAIL_BOXES = "mboxNumbers";
+    public static final String E_IS_UPDATE_ALL = "isUpdateAllMailBoxes";
+    public static final String E_ZIMBRA10_LOCATOR_UPGRADE_REQUEST = "Zimbra10LocatorUpgradeRequest";
+    public static final String E_STATUS_NAME = "status";
+    public static final String E_ZIMBRA10_LOCATOR_UPGRADE_RESPONSE = "Zimbra10LocatorUpgradeResponse";
+    public static final QName LOCATOR_UPDATE_ZIMBRA10_REQUEST = QName.get(E_ZIMBRA10_LOCATOR_UPGRADE_REQUEST, NAMESPACE);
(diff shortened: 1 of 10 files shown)


common/src/java/com/zimbra/common/soap/AdminConstants.java
```diff
@@ -1628,4 +1628,21 @@ public final class AdminConstants {
             "com_zimbra_docs_modern", "com_zimbra_drive_modern", "com_zextras_drive", "com_zextras_drive_open",
             "com_zextras_chat_open", "com_zextras_talk", "zimbra-zimlet-briefcase-edit-lool");
 
+    public static final String E_ACCOUNT_NAME = "accounts";
+
```
swaatiTelus: Extra line spaces are not required here after every constant.

common/src/java/com/zimbra/common/util/Constants.java
```diff
@@ -50,4 +50,7 @@ public class Constants {
     public static final String BEARER= "Bearer";
     public static final String JWT_SALT_SEPARATOR = "|";
 
+
+    public static final Integer LOCATOR_UPDATE_THREAD_COUNT  = 10;
+
```
swaatiTelus, line 53: Remove extra line

store/src/java/com/zimbra/cs/db/DbMailItem.java
```diff
@@ -5483,4 +5483,64 @@ public static void setPreviousFolder(Mailbox mbox, int id, String prevFolders) t
             DbPool.closeStatement(stmt);
         }
     }
+
+    public static List<Integer> getAllRecordsForSentAndRecieved(Mailbox mbox, String senderEmail, boolean isUpdated)  throws ServiceException {
```
swaatiTelus: Please add method documentation

common/src/java/com/zimbra/common/soap/AdminConstants.java
rendurama: done

common/src/java/com/zimbra/common/util/Constants.java
rendurama: done

store/src/java/com/zimbra/cs/db/DbMailItem.java
rendurama: done

vishnukantjangid:
Move this code to zm-modules-porter repo and for accessing DB here is the reference https://github.com/Zimbra/zm-backup-store/blob/db7d1b1af9796fdaf328ab11f4efe02765a471df/src/java/com/zimbra/cs/db/DbBackup.java#L59

PR #1513, opened 2023-08-23
Zcs 13728 Part 2- Update locator for external storage | In-place upgrade

rendurama:
Problem : AFter the zimbra 10 in place upgrade, When we try to read those mail stored on external storage we are getting missing blob because according to Zimbra 10 Storage Management locator values format should be volume_id@@mailbox_id/file_name for external volume but there is only external volume id present in the locator, according to zimbra9 standard.

Solution:
We update locators in the mail_item table for external volumes only to read blobs from external storage according to Zimbra 10 Storage Management implementation.
This utility will work with having the below argument.
Update locator based on account, selected mailbox and all mailboxes

Link to another Part1 PR : https://github.com/Zimbra/zm-mailbox/pull/1512

### DIFF
--- a/store/src/java/com/zimbra/cs/db/DbMailItem.java
+++ b/store/src/java/com/zimbra/cs/db/DbMailItem.java
@@ -5496,17 +5496,17 @@ public class DbMailItem {
         ResultSet rs = null;
         DbConnection conn = null;
         try {
-             conn = mbox.getOperationConnection();
+            conn = mbox.getOperationConnection();
             List<Integer> recordIds = new ArrayList<>();
             String WHERE_CLAUSE_CHECK_UPDATED = "";
             if (isUpdated) {
                 WHERE_CLAUSE_CHECK_UPDATED = "AND LOCATOR NOT LIKE '%@@%'";
             }
-             preparedStatement = conn.prepareStatement(" SELECT id FROM " + getMailItemTableName(mbox,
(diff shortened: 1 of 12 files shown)


store/src/java/com/zimbra/cs/db/DbMailItem.java
```diff
@@ -5483,4 +5479,104 @@ public static void setPreviousFolder(Mailbox mbox, int id, String prevFolders) t
...
+            List<Integer> recordIds = new ArrayList<>();
+            String WHERE_CLAUSE_CHECK_UPDATED = "";
+            if (isUpdated) {
+                WHERE_CLAUSE_CHECK_UPDATED = "AND LOCATOR NOT LIKE '%@@%'";
+            }
+             preparedStatement = conn.prepareStatement(" SELECT id FROM " + getMailItemTableName(mbox,
+                    false) + " where ((sender like concat('%',?,'%')) or (recipients like concat('%',?,'%')))  and folder_id in (2,3,4,5,6,9) " + WHERE_CLAUSE_CHECK_UPDATED);
+            preparedStatement.setString(1, senderEmail);
+            preparedStatement.setString(2, senderEmail);
+             rs = preparedStatement.executeQuery();
```
swaatiTelus: Please re-check the formatting of this file

rendurama: fixed

PR #1516, opened 2023-08-28
ZCS-13783 : Update locator for external storage

rendurama:
Problem: After the Zimbra 10 in-place upgrade, When we try to read those mail stored on external storage we are getting missing blob because according to Zimbra 10 Storage Management locator values format should be volume_id@@mailbox_id/file_name for external volume but there is only external volume id present in the locator, according to zimbra9 standard.

Solution:
We update locators in the mail_item table for external volumes only to read blobs from external storage according to Zimbra 10 Storage Management implementation. The porter application calls the newly created soap api for this.

This LC attribute defines the number of threads the executor uses for running parallel tasks

Link to another PR: [zm-modules-porter](https://github.com/Zimbra/zm-modules-porter/pull/44)


swaatiTelus (approved):
LGTM

vishnukantjangid (approved):

PR #1518, opened 2023-10-06
ZCS-13901 method for unloading extension

RakeshAMore:
Added method for unloading individual extensions.

### DIFF
--- a/store/src/java/com/zimbra/cs/extension/ExtensionUtil.java
+++ b/store/src/java/com/zimbra/cs/extension/ExtensionUtil.java
@@ -16,6 +16,14 @@
  */
 package com.zimbra.cs.extension;
 
+import com.zimbra.common.localconfig.LC;
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.util.Log.Level;
+import com.zimbra.common.util.ZimbraLog;
+import com.zimbra.cs.ephemeral.EphemeralStore;
+import com.zimbra.cs.ephemeral.EphemeralStore.Factory;
+import com.zimbra.cs.redolog.op.RedoableOp;
+
 import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URI;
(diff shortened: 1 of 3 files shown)


dasiyogesh (approved):

vishnukantjangid (approved):

rcyarrapothu (approved):

store/src/java/com/zimbra/cs/extension/ExtensionUtil.java
```diff
@@ -236,6 +236,18 @@ public static synchronized void destroyAll() {
         sInitializedExtensions.clear();
     }
 
+    public static synchronized void destroy(String extName) {
+        ZimbraExtension ext = getExtension(extName);
+        try {
+            RedoableOp.deregisterClassLoader(ext.getClass().getClassLoader());
+            ext.destroy();
```
desouzas, line 243: If we're going to make use of this then the following method needs to be updated to keep track of which extension registered the specified path (no change to signature):
https://github.com/Zimbra/zm-mailbox/blob/develop/store/src/java/com/zimbra/cs/extension/ExtensionDispatcherServlet.java#L59

Then the following method needs to be updated to make sure all paths registered by the extension via the above register method are removed (no change to signature):
https://github.com/Zimbra/zm-mailbox/blob/develop/store/src/java/com/zimbra/cs/extension/ExtensionDispatcherServlet.java#L130

This needs to happen to make sure extensions that register multiple paths can unregister them all.

store/src/java/com/zimbra/cs/extension/ExtensionUtil.java
```diff
@@ -236,6 +236,18 @@ public static synchronized void destroyAll() {
         sInitializedExtensions.clear();
     }
 
+    public static synchronized void destroy(String extName) {
+        ZimbraExtension ext = getExtension(extName);
```
desouzas, line 240: What happens if this is null in the next line? (npe)

store/src/java/com/zimbra/cs/extension/ExtensionUtil.java
RakeshAMore: Ok let's connect to discuss this.

store/src/java/com/zimbra/cs/extension/ExtensionUtil.java
RakeshAMore: Yes.

store/src/java/com/zimbra/cs/extension/ExtensionUtil.java
RakeshAMore: Created a separate story for handling this.

PR #1521, opened 2023-10-10
ZBUG-3132: Changes to include missing external email warning in html …

chitranshu198:
Ticket - https://synacor.atlassian.net/browse/ZBUG-3132

ZBUG-3132 - External Email Warning for all messages NOT coming from gmail domain

Fix - Used base email warning from plain text of message and appended the warning in HTML form to response of those external email where the content was missing.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/ToXML.java
+++ b/store/src/java/com/zimbra/cs/service/mail/ToXML.java
@@ -22,7 +22,17 @@ import java.io.Reader;
 import java.io.StringWriter;
 import java.io.UnsupportedEncodingException;
 import java.io.Writer;
-import java.util.*;
+import java.util.Arrays;
+import java.util.Calendar;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 import javax.mail.MessagingException;
 import javax.mail.internet.InternetAddress;
@@ -162,7 +172,7 @@ import com.zimbra.soap.mail.type.XParam;
 import com.zimbra.soap.mail.type.XProp;
 import com.zimbra.soap.type.MsgContent;
 import com.zimbra.soap.type.WantRecipsSetting;
-import com.zimbra.cs.lmtpserver.ExternalEmailWarning;
+
 /**
  * Class containing static methods for encoding various MailItem-derived objects into XML.
  *
@@ -209,7 +219,7 @@ public final class ToXML {
     }
 
     public static Element encodeItem(Element parent, ItemIdFormatter ifmt, OperationContext octxt, MailItem item,
-            int fields) throws ServiceException {
+                                     int fields) throws ServiceException {
         if (item instanceof Folder) {
             return encodeFolder(parent, ifmt, octxt, (Folder) item, fields);
         } else if (item instanceof Tag) {
@@ -266,19 +276,19 @@ public final class ToXML {
     }
 
     public static Element encodeFolder(Element parent, ItemIdFormatter ifmt, OperationContext octxt, Folder folder)
-    throws ServiceException {
+            throws ServiceException {
         return encodeFolder(parent, ifmt, octxt, folder, NOTIFY_FIELDS);
     }
 
     public static Element encodeFolder(Element parent, ItemIdFormatter ifmt, OperationContext octxt,
-            Folder folder, int fields)
-    throws ServiceException {
+                                       Folder folder, int fields)
+            throws ServiceException {
         return encodeFolder(parent, ifmt, octxt, folder, fields, false);
     }
 
     public static Element encodeFolder(Element parent, ItemIdFormatter ifmt, OperationContext octxt,
-            Folder folder, int fields, boolean exposeAclAccessKey)
-    throws ServiceException {
+                                       Folder folder, int fields, boolean exposeAclAccessKey)
+            throws ServiceException {
         if (folder instanceof SearchFolder) {
             return encodeSearchFolder(parent, ifmt, (SearchFolder) folder, fields);
         } else if (folder instanceof Mountpoint) {
@@ -426,8 +436,8 @@ public final class ToXML {
 
             Element eGrant = eACL.addNonUniqueElement(MailConstants.E_GRANT);
             eGrant.addAttribute(MailConstants.A_ZIMBRA_ID, grant.getGranteeId())
-                .addAttribute(MailConstants.A_GRANT_TYPE, ACL.typeToString(granteeType))
-                .addAttribute(MailConstants.A_RIGHTS, ACL.rightsToString(grant.getGrantedRights()));
+                    .addAttribute(MailConstants.A_GRANT_TYPE, ACL.typeToString(granteeType))
+                    .addAttribute(MailConstants.A_RIGHTS, ACL.rightsToString(grant.getGrantedRights()));
             if (grant.getExpiry() != 0) {
                 eGrant.addAttribute(MailConstants.A_EXPIRY, grant.getExpiry());
             }
@@ -566,12 +576,12 @@ public final class ToXML {
     }
 
     public static Element encodeMountpoint(Element parent, ItemIdFormatter ifmt, OperationContext octx,
-            Mountpoint mpt) throws ServiceException {
+                                           Mountpoint mpt) throws ServiceException {
         return encodeMountpoint(parent, ifmt, octx, mpt, NOTIFY_FIELDS);
     }
 
     public static Element encodeMountpoint(Element parent, ItemIdFormatter ifmt, OperationContext octx,
-            Mountpoint mpt, int fields) throws ServiceException {
+                                           Mountpoint mpt, int fields) throws ServiceException {
 
         Element el = parent.addNonUniqueElement(MailConstants.E_MOUNT);
         encodeFolderCommon(el, ifmt, mpt, fields);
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/service/mail/ToXML.java
```diff
@@ -1631,7 +1621,36 @@ private static Element encodeMessageAsMPHelper(boolean bestEffort, Element paren
             List<MPartInfo> parts = Mime.getParts(mm, getDefaultCharset(msg));
             if (parts != null && !parts.isEmpty()) {
                 Set<MPartInfo> bodies = Mime.getBody(parts, wantHTML);
-                addParts(m, parts.get(0), bodies, part, maxSize, neuter, false, getDefaultCharset(msg), bestEffort, wantContent);
+                // added condition to check if email is external and if it contains warning in html content
```
silentsakky: I think we already have code at lmtp or smtp level to add warning message based on zimbra_external_email_warning_message value, so isn't that code working here? and why we are adding completely new code instead of fixing that?

store/src/java/com/zimbra/cs/service/mail/ToXML.java
```diff
@@ -1631,7 +1621,36 @@ private static Element encodeMessageAsMPHelper(boolean bestEffort, Element paren
...
+                            ZimbraLog.lmtp.info(" originator email : "+emails.get(0));   //originator's email
+                            ZimbraLog.lmtp.info(" receiver email : "+emails.get(1));   //receivers's email
+                            boolean isExternal = ExternalEmailWarning.getInstance().isExternal(emails.get(1),emails.get(0));
+                            if (isExternal) {
+                                String textContent = parts.get(1).getMimePart().getContent().toString();
+                                String htmlContent = parts.get(2).getMimePart().getContent().toString();
+                                String[] textContents = textContent.split("\n", 2);
+                                String baseEmailWarning = textContents[0];
+                                ZimbraLog.lmtp.info("Base Email Warning : "+baseEmailWarning);
+                                if (!htmlContent.contains(baseEmailWarning)) {
```
silentsakky: check below scenario and how it will behave with this code
1) admin sets warning for external mails, let's say as "Disclaimer 1"
2) user sends some mails
3) admin updates warning for external mails, let's say as "Disclaimer 2"
4) if user opens mails sent in step 2, what he will see Disclaimer 1 or 2?

store/src/java/com/zimbra/cs/service/mail/ToXML.java
```diff
@@ -2775,7 +2827,12 @@ private static Element addPart(VisitPhase phase, Element parent, Element root, M
             }
 
             try {
-                addContent(el, mpi, maxSize, neuter, defaultCharset, wantContent);
+                //checking ig content type is html
+                if (mpi.getContentType().equals("text/html") && warningMsg.length()>0 && warningMsg!=null) {
+                    addContentWithExternalWarning(el, mpi, maxSize, neuter, defaultCharset, wantContent, warningMsg);
+                }else{
+                    addContent(el, mpi, maxSize, neuter, defaultCharset, wantContent);
```
silentsakky: aren't we adding warning in plain text mail?

store/src/java/com/zimbra/cs/service/mail/ToXML.java
```diff
@@ -2967,6 +3024,143 @@ private static void addContent(Element elt, MPartInfo mpi, int maxSize, boolean
...
+            data = StringUtil.stripControlCharacters(data);
+            if (wasTruncated) {
+                elt.addAttribute(MailConstants.A_TRUNCATED_CONTENT, true);
+                if (data.length() > maxSize) {
+                    data = data.substring(0, maxSize);
+                }
+            }
+            if (!data.contains("<html>")) {
+                String bodyContent = data;
+                String htmBodyWithWarning = "<html><body><div><p>"+warningMsg+"</p><hr /></div><div style=\"font-size: 12pt; font-family: arial, helvetica, sans-serif; color: rgb(0, 0, 0);\"><div>&nbsp;</div><div>"+bodyContent+"</div>\n";
```
silentsakky: bodyContent may have `<html>` tag already, so are we duplicating it?

store/src/java/com/zimbra/cs/service/mail/ToXML.java
chitranshu198: yes, we have code to add external warning to incoming email. It is working as expected for plain text but failing for html content for Emails coming from gmail domain. 

Current code changes are on presentation layer wherein below conditions are checked -
1.  If Email is external
2. If html content contains base warning(extracted from plain text content)

if the warning is missing in html content we are transforming plain text warning to html warning and adding it to response.

store/src/java/com/zimbra/cs/service/mail/ToXML.java
chitranshu198: User will see "Disclaimer 1"
During the api call to fetch the message, message are extracted from storage. Since Disclaimer 1 one was the warning when email was saved, it will fetched and passed back in response.

store/src/java/com/zimbra/cs/service/mail/ToXML.java
chitranshu198: Plain text warning are added successfully while processing the mail at LMTP/SMTP. So we are only making changes for HTML content warning of external emails where the warning is missing(gmail domain).

store/src/java/com/zimbra/cs/service/mail/ToXML.java
chitranshu198: This is a check to see if the html content is already present. If the tag is not present, we are creating the html content(warning + body).

store/src/java/com/zimbra/cs/service/mail/ToXML.java
silentsakky: can't we fix code to add warning at LMTP/SMTP level for html mails also?

silentsakky: Consider scenario:
1) currently admin has not added a warning attribute
2) user gets some mails
3) admin adds new warning attribute
4) if user opens earlier mails it will add warning, I guess should not be done

store/src/java/com/zimbra/cs/service/mail/ToXML.java
```diff
@@ -2967,6 +3024,143 @@ private static void addContent(Element elt, MPartInfo mpi, int maxSize, boolean
         }
         // TODO: CDATA worth the effort?
     }
+    /*
+    * Adds Warning Content to html content of Response
+    * */
+    private static void addContentWithExternalWarning(Element elt, MPartInfo mpi, int maxSize, boolean neuter, String defaultCharset, MsgContent wantContent,String warningMsg)
```
vishnukantjangid: Why have we created a new duplicate method addContentWithExternalWarning() instead of adding some code block in the existing method addContent()? We can add the below code in addContent() to achieve the same.

`if (!data.contains("<html>")) {
                String bodyContent = data;
                String htmBodyWithWarning = "<html><body><div><p>"+warningMsg+"</p><hr /></div><div style=\"font-size: 12pt; font-family: arial, helvetica, sans-serif; color: rgb(0, 0, 0);\"><div>&nbsp;</div><div>"+bodyContent+"</div>\n";
data = htmBodyWithWarning;
}`

store/src/java/com/zimbra/cs/service/mail/ToXML.java
```diff
@@ -1687,6 +1706,39 @@ private static Element encodeMessageAsMPHelper(boolean bestEffort, Element paren
...
+            index = lowerCaseTextString.indexOf(lowerCaseWord, index);
+            if (index != -1) {
+                indexes.add(index);
+                index++;
+            }
+        }
+        return indexes;
+    }
+
+    private static List<String> getSenderReceiverEmailAddresses(Element parent) {
```
vishnukantjangid: We would use the existing methods to find the mail sender and recipients using the Message object.

chitranshu198: Refactored the code to get senders and recipients mail from message object

store/src/java/com/zimbra/cs/service/mail/ToXML.java
chitranshu198: To avoid redundancy, added one more parameter to addContent with a logic to add warning and removed addContentWithExternalWarning.
Made necessary changes to the methods calling addContent

store/src/java/com/zimbra/cs/service/mail/ToXML.java
chitranshu198: In such scenarios default warning will be display.

For eg.

1. Admin has not set the warning and receives external email (**M1**). In such scenario default warning will be used (default warning ~ This message originated outside of your organization.) and email will be stored.
2. Admin Edits the warning msg and receives another external email. In this scenario edited warning will be used and mail will be stored.
3. If the User tries to access M1 later, it will show the default warning as it was warning at the time of processing/storage.

Warning methods at LTMP/SMTP are invoked only when a new email is received, so if we make changes at LMTP/SMTP layers, only new emails will contain warning in both plain & html content and issue will still persist for existing email. To avoid this, changes are made at presentation layer where response is created.

silentsakky: I still feel, we should fix this at LMTP/SMTP layer only and not at presentation layer.

store/src/java/com/zimbra/cs/service/mail/ToXML.java
```diff
@@ -2959,6 +2979,13 @@ private static void addContent(Element elt, MPartInfo mpi, int maxSize, boolean
                     data = data.substring(0, maxSize);
                 }
             }
+            if(warningMsg.length()>0){
```
vishnukantjangid: You can club both conditions in a single if

chitranshu198: clubbed both the conditions now

PR #1524, opened 2023-10-19
ZCS-13071:  Drive module with shared files loss on sharee end if it is later moved to 10.0 and sharer has already been moved

shrutig0510:
**Problem:**
Drive module with shared files loss on sharee end if it is later moved to 10.0 and sharer has already been moved

**Fix:**
While exporting the sharee data, also export data of its remote shares from other users.
while importing sharee data, process these remote shares data on sharer end.

**Test-cases:**

1. sharer is moved first to 10.0 (all export and import done) and then sharee is moved to 10.0.
2. sharee is moved first to 10.0 (all export and import done) and then sharer is moved to 10.0.
3. sharer and sharee both moved to 10.0, but first sharer data is imported and then sharee data is imported.
4. sharer and sharee both moved to 10.0, but first sharee data is imported and then sharer data is imported.
In all the above cases, sharee remote files data is restored.

### DIFF
--- a/store/src/java/com/zimbra/cs/mailbox/Mailbox.java
+++ b/store/src/java/com/zimbra/cs/mailbox/Mailbox.java
@@ -147,10 +147,8 @@ import com.zimbra.cs.mailbox.ACL.Grant;
 import com.zimbra.cs.mailbox.CalendarItem.AlarmData;
 import com.zimbra.cs.mailbox.CalendarItem.Callback;
 import com.zimbra.cs.mailbox.CalendarItem.ReplyInfo;
-import com.zimbra.cs.mailbox.Flag.FlagInfo;
 import com.zimbra.cs.mailbox.FoldersTagsCache.FoldersTags;
 import com.zimbra.cs.mailbox.MailItem.CustomMetadata;
-import com.zimbra.cs.mailbox.MailItem.CustomMetadata.CustomMetadataList;
 import com.zimbra.cs.mailbox.MailItem.PendingDelete;
 import com.zimbra.cs.mailbox.MailItem.TargetConstraint;
 import com.zimbra.cs.mailbox.MailItem.Type;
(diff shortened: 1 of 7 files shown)


preshitaagrawal (approved):

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -9555,6 +9555,66 @@ public Document createDocument(OperationContext octxt, int folderId, ParsedDocum
         }
     }
 
+    public Document createDocument(OperationContext octxt, int folderId, ParsedDocument pd, MailItem.Type type,
```
vishnukantjangid, line 9502: Use method overloading instead of creating the same duplicate method.

shrutig0510: This is method overloading only, introduced one new parameter at end as nodeId which will mostly be used for migration. Not touched already available methods so as to restrict the changes for migration mostly.
>public Document createDocument(OperationContext octxt, int folderId, String filename, String mimeType,
            String author, String description, InputStream data)

>public Document createDocument(OperationContext octxt, int folderId, ParsedDocument pd, MailItem.Type type,
            int flags, MailItem parent, CustomMetadata custom, boolean indexing, String nodeId)

vishnukantjangid: Yes, you can achieve the same thing without having duplicate methods. like 

```
public Document createDocument(OperationContext octxt, int folderId, ParsedDocument pd, MailItem.Type type,
             int flags, MailItem parent, CustomMetadata custom, boolean indexing) {
             createDocument(OperationContext octxt, int folderId, ParsedDocument pd, MailItem.Type type,
             int flags, MailItem parent, CustomMetadata custom, boolean indexing, null);
             }`
```

and method signature will be for the existing method with the addition of a new parameter.
```
public Document createDocument(OperationContext octxt, int folderId, ParsedDocument pd, MailItem.Type type,
            int flags, MailItem parent, CustomMetadata custom, boolean indexing, String nodeId)
```

shrutig0510: Done

vishnukantjangid (approved):

zimsuchitgupta (approved):

PR #1525, opened 2023-10-19
Feature/zcs 14115 fix extension dispatcher unregistration issue

RakeshAMore:
**Description**
Extension’s HTTP handlers registered using method https://github.com/Zimbra/zm-mailbox/blob/develop/store/src/java/com/zimbra/cs/extension/ExtensionDispatcherServlet.java#L59 can not deregistered using the method https://github.com/Zimbra/zm-mailbox/blob/develop/store/src/java/com/zimbra/cs/extension/ExtensionDispatcherServlet.java#L130 since registration of handlers tracking is incorrect, while deregisterd code is not honoring correct key to remove handlers.

**Solution**
Define data structure that can keep tracking of handlers based on the extensions https://github.com/Zimbra/zm-mailbox/blob/develop/store/src/java/com/zimbra/cs/extension/ExtensionDispatcherServlet.java#L50 so while removing it will look for handlers with respect to the extensions


store/src/java/com/zimbra/cs/extension/ExtensionDispatcherServlet.java
```diff
@@ -129,19 +136,17 @@ public void service(HttpServletRequest req, HttpServletResponse resp) throws IOE
...
-                    ExtensionHttpHandler handler = (ExtensionHttpHandler) sHandlers.get(path);
-                    try {
-                        handler.destroy();
-                    } catch (Exception e) {
-                        ZimbraLog.extensions.warn("Error in destroy for handler " + handler.getClass(), e);
-                    } finally {
-                        it.remove();
-                    }
+            String extName = ext.getName();
+            for (ExtensionHttpHandler extHandler: EXTENSION_HANDLERS.get(extName)) {
```
dasiyogesh, line 141: why for loop, are there more extensions with same name?

store/src/java/com/zimbra/cs/extension/ExtensionDispatcherServlet.java
```diff
@@ -60,9 +62,14 @@ public static void register(ZimbraExtension ext, ExtensionHttpHandler handler) t
         handler.init(ext);
         String name = handler.getPath();
         synchronized (sHandlers) {
-            if (sHandlers.containsKey(name))
+            if (sHandlers.containsKey(name)) {
                 throw ServiceException.FAILURE("HTTP handler already registered: " + name, null);
+            }
             sHandlers.put(name, handler);
+            if (!EXTENSION_HANDLERS.containsKey(ext.getName())) {
+                EXTENSION_HANDLERS.put(ext.getName(), new ArrayList<ExtensionHttpHandler>());
```
dasiyogesh, line 71: are there multiple handlers for same extension?

RakeshAMore: Yes.

store/src/java/com/zimbra/cs/extension/ExtensionDispatcherServlet.java
RakeshAMore: Loop is for multiple handlers for extension.

store/src/java/com/zimbra/cs/extension/ExtensionDispatcherServlet.java
```diff
@@ -129,19 +136,17 @@ public void service(HttpServletRequest req, HttpServletResponse resp) throws IOE
...
-                        ZimbraLog.extensions.warn("Error in destroy for handler " + handler.getClass(), e);
-                    } finally {
-                        it.remove();
-                    }
+            String extName = ext.getName();
+            for (ExtensionHttpHandler extHandler: EXTENSION_HANDLERS.get(extName)) {
+                try {
+                    extHandler.destroy();
+                } catch (Exception e) {
+                    ZimbraLog.extensions.warn("Error in destroy for handler " + extHandler.getClass(), e);
```
log2akshat: `+` operator for string concatenation is an inefficient way to concatenate, it creates new string objects each time, leading to a lot of object creation and garbage collection, which we have seen a lot of problems of memory leaks in Zimbra in the past. And `+` operator can be slower than other methods due to the object creation and garbage collection overhead.

store/src/java/com/zimbra/cs/extension/ExtensionDispatcherServlet.java
```diff
@@ -129,19 +136,17 @@ public void service(HttpServletRequest req, HttpServletResponse resp) throws IOE
...
+                } finally {
```
log2akshat, line 146: Use `try-with-resources `here, as it automatically handles cleanup, improves code readability, enhances exception handling, reduces resource leak risks, and simplifies null checking compared to traditional `finally` blocks.

store/src/java/com/zimbra/cs/extension/ExtensionDispatcherServlet.java
```diff
@@ -48,6 +47,9 @@
  */
 public class ExtensionDispatcherServlet extends ZimbraServlet {
     private static Map sHandlers = Collections.synchronizedMap(new HashMap());
+
+    private static Map<String, ArrayList<ExtensionHttpHandler>> EXTENSION_HANDLERS = Collections.synchronizedMap(new HashMap());
```
log2akshat: For improved performance in high-concurrency scenarios, it is generally recommended to use `ConcurrentHashMap` over a `synchronizedMap`, as it employs fine-grained locking and provides efficient concurrent access to map data.

RakeshAMore: hmm. that's true, since we are not EXTENSION_HANDLERS as lock here we can go ahead with the ConcurrentHashMap here.

store/src/java/com/zimbra/cs/extension/ExtensionDispatcherServlet.java
RakeshAMore: That's good in other cases. but here it seems `try-catch-finally` is the best choice. since we are not opening try-on `sHandlers`

store/src/java/com/zimbra/cs/extension/ExtensionDispatcherServlet.java
RakeshAMore: Good catch

vishnukantjangid (approved):

zimsuchitgupta (approved):

PR #1528, opened 2023-11-02
ZBUG-3132 : Changes to append external email warning to gmail's mime …

chitranshu198:
Ticket - https://synacor.atlassian.net/browse/ZBUG-3132

ZBUG-3132 - External Email Warning for all messages NOT coming from gmail domain

Fix - Created a new regex to match gmail's html content.

### DIFF
--- a/store/src/java/com/zimbra/cs/lmtpserver/ExternalEmailWarning.java
+++ b/store/src/java/com/zimbra/cs/lmtpserver/ExternalEmailWarning.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.lang.StringEscapeUtils;
 
 import com.zimbra.common.localconfig.LC;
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/lmtpserver/ExternalEmailWarning.java
```diff
@@ -190,9 +193,59 @@ public String getUpdatedContent(String content) {
                     sb.append(getTextHtmlWarning());
                 sb.append(content.substring(end));
                 content = sb.toString();
+            } else if (isIncomingMimeFromGmail(content)) {
```
silentsakky: will this fix work if mails are coming from yahoo.com or hotmail.com or any other external domains?

chitranshu198: No, this will only work for gmail domain

silentsakky: so mails coming from the above mails will get the disclaimer or not?

chitranshu198: if their mime matches existing html regex (Content-Type: text/html.*?\r\n\r\n.*?<body.*?>) then it will get the disclaimer

vishnukantjangid: @chitranshu198 you can achieve the same by just making changes in the regex pattern. It will work for other domains too. The below change is not required in that case.
Use `Content-Type: text/html.*?\\r\\n\\r\\n.*?<div.*?>` regex for others.

PR #1534, opened 2023-11-27
ZBUG-3621 - Added condition to replace pkcs7-signature content type w…

chitranshu198:
Ticket - https://synacor.atlassian.net/browse/ZBUG-3621

Zbug 3621 - When compose a new message, and try to attach a .p7m file, it isn't attached and don't receive any errors in zcs10.

Fix - Added a condition at file upload servlet to replace pkcs7-signature content type with pcks7-mime.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/FileUploadServlet.java
+++ b/store/src/java/com/zimbra/cs/service/FileUploadServlet.java
@@ -161,11 +161,12 @@ public class FileUploadServlet extends ZimbraServlet {
             } else {
                 mimeType = getMimeType(file);
                 contentType = mimeType.toString();
-
+                extension = filename == null ? "" : FilenameUtils.getExtension(filename).trim();
+                if (MimeConstants.FILE_EXTENSION_P7M.equalsIgnoreCase(extension)
+                        && MimeConstants.CT_APPLICATION_SMIME_SIGNATURE.equalsIgnoreCase(contentType)) {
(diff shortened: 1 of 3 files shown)


rcyarrapothu (approved):

zimsuchitgupta (approved):

store/src/java/com/zimbra/cs/service/FileUploadServlet.java
```diff
@@ -161,11 +161,12 @@ public static final class Upload {
             } else {
                 mimeType = getMimeType(file);
                 contentType = mimeType.toString();
-
+                extension = filename == null ? "" : FilenameUtils.getExtension(filename).trim();
```
silentsakky: have we checked why tika is not able to detect p7m content type correctly? why are we putting hacks in our code when third party library should be able to handle this scenario
check https://github.com/apache/tika/blob/main/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml#L790

dasiyogesh: If for some reason its failing to detect the mime type from library, we can try and add mime type into https://github.com/Zimbra/zm-mailbox/blob/develop/store-conf/conf/custom-mimetypes.xml

chitranshu198: This seems to be an open issue with Tika. Please find below resource for more details - 
https://issues.apache.org/jira/browse/TIKA-1997

PR #1538, opened 2023-11-29
Zbug 3565 -  Need improvement in logging of auth requests

sumitkumar1110:
ZBUG-3565 - Need improvement in logging of auth requests in audit.log and mailbox.log 

Fix - added a LC attribute which when true logs each logging attempt to audit.log and mailbox.log  with ip.

### DIFF
--- a/common/src/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/java/com/zimbra/common/localconfig/LC.java
@@ -268,6 +268,9 @@ public final class LC {
     public static final KnownKey zimbra_rights_delegated_admin_supported = KnownKey.newKey(true);
 
     @Supported
+    public static final KnownKey zimbra_additional_logging = KnownKey.newKey(false);
+
+    @Supported
     public static final KnownKey zimbra_spam_report_queue_size = KnownKey.newKey(100);
 
     @Supported
--- a/store/src/java/com/zimbra/cs/service/account/Auth.java
+++ b/store/src/java/com/zimbra/cs/service/account/Auth.java
@@ -20,6 +20,7 @@
  */
 package com.zimbra.cs.service.account;
 
+import com.zimbra.common.localconfig.LC;
 import io.jsonwebtoken.Claims;
 
 import java.util.HashMap;
@@ -294,6 +295,10 @@ public class Auth extends AccountDocumentHandler {
         }
 
         if (acct == null) {
+            if (LC.zimbra_additional_logging.booleanValue()) {
+                ZimbraLog.security.info(String.format("Error occurred during authentication:" +
+                        " authentication failed for [%s]. Reason: account not found.", acctValue));
+            }
             AuthFailedServiceException e = AuthFailedServiceException.AUTH_FAILED(acctValue,
                 acctValuePassedIn, "account not found");
             AuthListener.invokeOnException(e);


store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
```diff
@@ -1133,6 +1133,10 @@ Account getAccountByName(String emailAddress, boolean loadFromMaster, boolean ch
             }
         }
 
+        if (LC.zimbra_additional_logging.booleanValue() && account == null) {
+            ZimbraLog.security.info(String.format("Error occurred during authentication: authentication failed for [%s]. Reason: account not found.", emailAddress));
```
gautamdg: Formatting, long string should be in next line.

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
```diff
@@ -1133,6 +1133,14 @@ Account getAccountByName(String emailAddress, boolean loadFromMaster, boolean ch
             }
         }
 
+        // using this clause because other clients like imap, pop get logged though nginx lookup
+        if (LC.zimbra_additional_logging.booleanValue() && account == null) {
+            String contextString = ZimbraLog.getContextString();
+            if (contextString != null && (contextString.contains("oip") || contextString.contains("ip"))) {
```
vishnukantjangid:
```suggestion
            if (contextString != null && contextString.contains("ip")) {
```

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
```diff
@@ -1133,6 +1133,14 @@ Account getAccountByName(String emailAddress, boolean loadFromMaster, boolean ch
             }
         }
 
+        // using this clause because other clients like imap, pop get logged though nginx lookup
```
shrutig0510: remove this comment from here and add it at line no 1139

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
```diff
@@ -1133,6 +1133,14 @@ Account getAccountByName(String emailAddress, boolean loadFromMaster, boolean ch
             }
         }
 
+        if (LC.zimbra_additional_logging.booleanValue() && account == null) {
+            String contextString = ZimbraLog.getContextString();
+            // using this clause because other clients like imap, pop get logged though nginx lookup
+            if (contextString != null && contextString.contains("ip")) {
+                ZimbraLog.security.info(String.format("Error occurred during authentication:" +
```
silentsakky: we are assuming that this will be auth call but function name is getAccountByName, so internally if we are calling this method then it could give wrong information in logs

sumitkumar1110: Thank you for the feedback. I've updated Auth.java to address the logging issue with 'getAccountByName'. Please review the changes and let me know if there are any more concerns

silentsakky (approved):

shrutig0510 (approved):

zimsuchitgupta (approved):

PR #1542, opened 2023-12-21
ZCS-14416:Date filter payload modification due to timezone issue

shubhamCS03:
Requirement of ticket: ZCS-14416

Requirements:
Now Zimbra UI is sending only _Date_ in timestamp so, in the backend we have to use `zimbraPrefTimeZoneId` to handle the **timezone issue** for parsing and formatting the _Date_.

Solution:
Modified the _DateParser_ class and _RuleManger_ class to handle timezone changes according to the `zimbraPrefTimeZoneId` value for the **formatter**.

Testing:
Tested on custom build with all the filters.

### DIFF
--- a/common/src/java/com/zimbra/common/util/DateParser.java
+++ b/common/src/java/com/zimbra/common/util/DateParser.java
@@ -19,6 +19,7 @@ package com.zimbra.common.util;
 import java.text.ParsePosition;
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.TimeZone;
 
 /**
  * Wrapper around {@link SimpleDateFormat} that allows it to be used in a
@@ -30,6 +31,7 @@ public class DateParser {
 
     private ThreadLocal<SimpleDateFormat> mFormatterHolder = new ThreadLocal<SimpleDateFormat>();
     private String mDatePattern;
+    private String timezone;
     
     public DateParser(String datePattern) {
         mDatePattern = datePattern;
@@ -38,17 +40,27 @@ public class DateParser {
     public Date parse(String s) {
         return getFormatter().parse(s, new ParsePosition(0));
     }
-    
+
     public String format(Date date) {
         return getFormatter().format(date);
     }
-    
+
     private SimpleDateFormat getFormatter() {
         SimpleDateFormat formatter = mFormatterHolder.get();
         if (formatter == null) {
             formatter = new SimpleDateFormat(mDatePattern);
             mFormatterHolder.set(formatter);
         }
+
+        formatter.setTimeZone(TimeZone.getTimeZone(timezone));
         return formatter;
+    }
+
+    public String getTimezone() {
+        return timezone;
+    }
+
+    public void setTimezone(String timezone) {
+        this.timezone = timezone;
     }
 }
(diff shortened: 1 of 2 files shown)


client/src/java/com/zimbra/client/ZFilterCondition.java
```diff
@@ -957,7 +956,12 @@ public ZDateCondition(DateOp op, String dateStr) throws ServiceException {
 
         ZDateCondition(FilterTest.DateTest test) throws ServiceException {
             this.op = DateOp.of(Sieve.DateComparison.fromString(test.getDateComparison()), test.isNegative());
-            this.date = new Date(test.getDate() * 1000);
+            //this.date = new Date(test.getDate() * 1000);
```
swathy-shaji: Do we need to keep this commented line

client/src/java/com/zimbra/client/ZFilterCondition.java
```diff
@@ -957,7 +956,12 @@ public ZDateCondition(DateOp op, String dateStr) throws ServiceException {
...
+            try {
+                this.date = new SimpleDateFormat("yyyyMMdd").parse(test.getDate());
```
swathy-shaji: can we use a constant to hold the pattern

client/src/java/com/zimbra/client/ZFilterCondition.java
shubhamCS03: Removed

client/src/java/com/zimbra/client/ZFilterCondition.java
shubhamCS03: not needed

k-kato (changes requested):
We should not change existing API. It will break communication with UI. If a Zimbra user has applied any customization around filter page, it will also be broken.
If we need to consider daylight saving time, it should be calculated on server side using the passed date in `long` type (as it is) and `zimbraPrefTimeZoneId` of the account. Basically, zimbraPrefTimeZoneId should be set to the same timezone of a user's computer.
If we have any concern about using `zimbraPrefTimeZoneId`, we should add `tz` (timezone) element in `ModifyFilterRulesRequest > filterRule > filterTests > dateTest` so that a client can pass its timezone.
If it is not passed, use `zimbraPrefTimeZoneId` of an account, cos or domain (`accountCosDomainInherited`) as a default value.  If it is empty in all levels, use server timezone.
In any way, maintain backward compatibility on API.

k-kato:
> we should add tz (timezone) element

Not an element but an attribute may be better.

k-kato (commented):
As we discussed, timezone does not need to be passed from client side. `zimbraPrefTimeZoneId` is used at filter creation.

common/src/java/com/zimbra/common/util/DateParser.java
```diff
@@ -42,13 +44,22 @@ public Date parse(String s) {
     public String format(Date date) {
         return getFormatter().format(date);
     }
-
```
k-kato, line 45: nit: please do not remove a blank line for readability.

k-kato (approved):

common/src/java/com/zimbra/common/util/DateParser.java
shubhamCS03: Done

swaatiTelus (approved):

zimsuchitgupta (approved):

PR #1543, opened 2023-12-22
ZCS-14412: change default value false for ldap attributes

vishnukantjangid:
**Implementation:** Change the default value to FALSE for the below LDAP attributes. It will help disable the feature for the new account.

1. zimbraFeatureSharingEnabled
2. zimbraFeatureBriefcasesEnabled
3. zimbraFeatureDocumentEditingEnabled
4. zimbraExternalSharingEnabled
5. zimbraPublicSharingEnabled

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -1822,7 +1822,7 @@ TODO - add support for multi-line values in globalConfigValue and defaultCOSValu
 </attr>
 
 <attr id="335" name="zimbraFeatureSharingEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInfo,accountInherited">
-  <defaultCOSValue>TRUE</defaultCOSValue>
+  <defaultCOSValue>FALSE</defaultCOSValue>
   <defaultExternalCOSValue>FALSE</defaultExternalCOSValue>
   <desc>enabled sharing</desc>
 </attr>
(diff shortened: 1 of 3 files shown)


swaatiTelus (approved):
We should make these changes in ZimbraOS repo as well.
The rest looks fine.

dasiyogesh:
@vishnukantjangid  how is this going to be handled for 10.0.x ?
Also when the entitlements are enabled from Zentitle, how are these LDAP attributes going to be enabled?

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -1822,7 +1822,7 @@ TODO - add support for multi-line values in globalConfigValue and defaultCOSValu
 </attr>
 
 <attr id="335" name="zimbraFeatureSharingEnabled" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInfo,accountInherited">
-  <defaultCOSValue>TRUE</defaultCOSValue>
+  <defaultCOSValue>FALSE</defaultCOSValue>
```
silentsakky, line 1825: we should check how this will behave in case of upgrade scenario
if customer is upgrading from 9.x.x to 10.1.x then existing default cos will have value TRUE only not FALSE, which is different then our expectation

RakeshAMore: If they have license entitlement they can enbled it back. Lets discuss this tomorrow

dasiyogesh: enabling for large user base will be problem

PR #1556, opened 2024-01-19
ZBUG-3249 Remove ZMG related code

ghost:
I have tested this change by compiling a FOSS build of Zimbra 10.0.6 but used _develop_ for zm-mailbox and zm-web-client. For zm-web-client I had to revert some (unrelated) recent changes in login.jsp and printcontacts for it to build. I also had to add a try..catch around ZmListView.prototype.addHeaderItemInTagGroup in 
/opt/zimbra/jetty_base/webapps/zimbra/js/zimbraMail/share/view/ZmListView.js

Then I was able to log in using: /?client=advanced&dev=1

I ran build.pl using `-DskipTests=true`.

For reviewers, please take an extra look at these two changes:

https://github.com/barrydegraaff/zm-mailbox/commit/91d8854c0eb215550660c0a2c5f5cfe279f5828a#diff-ec6294f6a1ccd7dfc9853dc54e1f74dc50cf2786f92e7f670449d17c6a2d9a23L380

https://github.com/barrydegraaff/zm-mailbox/commit/91d8854c0eb215550660c0a2c5f5cfe279f5828a#diff-76ff182fd704770358367a7c3f80b27e5ae12a54ea2c45a0e95b96c1c1fbcf53L100


common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
```diff
@@ -15212,14 +15074,6 @@ public static TwoFactorAuthSecretEncoding fromString(String s) throws ServiceExc
     @ZAttr(id=638)
     public static final String A_zimbraPrefZimletTreeOpen = "zimbraPrefZimletTreeOpen";
 
-    /**
-     * Option to turn on/off zimbra mobile gateway push notifications
-     *
-     * @since ZCS 8.7.0,9.0.0
-     */
-    @ZAttr(id=1952)
-    public static final String A_zimbraPrefZmgPushNotificationEnabled = "zimbraPrefZmgPushNotificationEnabled";
```
silentsakky, line 15221: if I understand correctly then these provisioning files are auto generated from zimbra-attrs.xml file, and we haven't removed ldap attributes from zimbra-attrs.xml file so any developer will regenerate these files and the code will be back

k-kato:
Regarding `ZAttrProvisioning` and `ZAttrAccount|Config|Cos`, we always update them using `ant generate-getters`.
https://github.com/Zimbra/zm-build/#adding-a-new-ldap-attribute

We must not remove any existing ldap attribute but add `deprecatedSince` flag in `zimbra-attrs.xml` (and run the `ant` command)
Removing an attribute from the definition, or from ldap schema will cause an issue at ZCS upgrade.

common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
silentsakky: also we usually just deprecate ldap attributes, we can't remove it as it can cause issues in upgrade paths.
I might be wrong here, backend devs will have more idea about this restriction.

ghost: This has been addressed via https://github.com/Zimbra/zm-mailbox/pull/1570

ghost:
> Regarding `ZAttrProvisioning` and `ZAttrAccount|Config|Cos`, we always update them using `ant generate-getters`. https://github.com/Zimbra/zm-build/#adding-a-new-ldap-attribute
> 
> We must not remove any existing ldap attribute but add `deprecatedSince` flag in `zimbra-attrs.xml` (and run the `ant` command) Removing an attribute from the definition, or from ldap schema will cause an issue at ZCS upgrade.

This has been addressed via https://github.com/Zimbra/zm-mailbox/pull/1570

PR #1568, opened 2024-02-15
ZBUG-2859 Bugzilla 80485 admin login honors zimbraAuthFallbackToLocal

ghost:
ZBUG-2859 Bugzilla 80485 admin login honors zimbraAuthFallbackToLocal


k-kato:
It was introduced in https://bugzilla.zimbra.com/show_bug.cgi?id=5106
Is there any case that an administrator wants to use a local password even when zimbraAuthFallbackToLocal is FALSE? Don't we need to add ldap attr or localconfig to switch fallback logic for admin account considering backward compatibility?

ghost:
> It was introduced in https://bugzilla.zimbra.com/show_bug.cgi?id=5106 Is there any case that an administrator wants to use a local password even when zimbraAuthFallbackToLocal is FALSE? Don't we need to add ldap attr or localconfig to switch fallback logic for admin account considering backward compatibility?

No I think the reasoning in 5106 is not sensible from a security perspective .

k-kato (approved):
I see. Then the change LGTM.

ghost:
> I see. Then the change LGTM.

But just in case someone runs into that issue again, I added an instruction here: https://wiki.zimbra.com/wiki/How_To_Create_an_Admin_Account

silentsakky:
Changes moved to https://github.com/Zimbra/zm-mailbox/pull/1597 for merging as this branch was having merge commits.

PR #1573, opened 2024-02-26
ZCS-14667: User is able to access email present in external storage w…

shubhamCS03:
Ticket : [ZCS-14667](https://synacor.atlassian.net/browse/ZCS-14667)
We are getting NumberFormatException, while reading the messages when primary is internal and secondary is external when license is expired. Also, we should block user reading external blobs from S3 volume, so to handle this we are providing a custom exception message for it.

Solution :
to handle this we are providing a custom exception message for it.

Testing:
Tested the changes remotely.

[ZCS-14667]: https://synacor.atlassian.net/browse/ZCS-14667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/store/src/java/com/zimbra/cs/store/file/FileBlobStore.java
+++ b/store/src/java/com/zimbra/cs/store/file/FileBlobStore.java
@@ -367,12 +367,17 @@ public final class FileBlobStore extends StoreManager {
 
     @Override
     public MailboxBlob getMailboxBlob(Mailbox mbox, int itemId, int revision, String locator, boolean validate) throws ServiceException {
-        short volumeId = Short.valueOf(locator);
-        File file = getMailboxBlobFile(mbox, itemId, revision, volumeId, validate);
-        if (file == null) {
-            return null;
+        try {
+            short volumeId = Short.valueOf(locator);
+            File file = getMailboxBlobFile(mbox, itemId, revision, volumeId, validate);
+            if (file == null) {
+                return null;
+            }
+            return new VolumeMailboxBlob(mbox, itemId, revision, locator, new VolumeBlob(file, volumeId));
+        } catch (NumberFormatException e) {
+            ZimbraLog.store.info("NumberFormatException:", e);
+            throw ServiceException.FAILURE("Operation can not be completed because store is not accessible", null);
         }
-        return new VolumeMailboxBlob(mbox, itemId, revision, locator, new VolumeBlob(file, volumeId));
     }
 
     @Override


store/src/java/com/zimbra/cs/store/file/FileBlobStore.java
```diff
@@ -367,12 +367,20 @@ private boolean deleteFile(File file) throws IOException {
...
+            short volumeId = Short.valueOf(locator);
+            File file = getMailboxBlobFile(mbox, itemId, revision, volumeId, validate);
+            if (file == null) {
+                return null;
+            }
+            return new VolumeMailboxBlob(mbox, itemId, revision, locator, new VolumeBlob(file, volumeId));
+        } catch (NumberFormatException e) {
+            // this exception occurs when primary is internal and secondary is external, while reading volume id we get custom locator
+            // Also, when license is expired, we should block user reading external blobs from S3 volume
+            ZimbraLog.store.debug("NumberFormatException:", e.getMessage());
```
RakeshAMore: `ZimbraLog.store.info("NumberFormatException:", e)`

store/src/java/com/zimbra/cs/store/file/FileBlobStore.java
```diff
@@ -367,12 +367,20 @@ private boolean deleteFile(File file) throws IOException {
...
+            throw ServiceException.FAILURE("Operation can not be completed because License is expired or not activated",
```
RakeshAMore: `"Operation can not be completed because store is not accessible"`

store/src/java/com/zimbra/cs/store/file/FileBlobStore.java
shubhamCS03: done

store/src/java/com/zimbra/cs/store/file/FileBlobStore.java
shubhamCS03: done

RakeshAMore (approved):

swaatiTelus (approved):

shubhamCS03:
squashed and rebased

store/src/java/com/zimbra/cs/store/file/FileBlobStore.java
```diff
@@ -367,12 +367,17 @@ private boolean deleteFile(File file) throws IOException {
 
     @Override
     public MailboxBlob getMailboxBlob(Mailbox mbox, int itemId, int revision, String locator, boolean validate) throws ServiceException {
-        short volumeId = Short.valueOf(locator);
-        File file = getMailboxBlobFile(mbox, itemId, revision, volumeId, validate);
-        if (file == null) {
-            return null;
+        try {
+            short volumeId = Short.valueOf(locator);
```
log2akshat, line 371: Check for `locator` is  a valid number or not and immediately throws an exception and stops further processing which will reduce the unnecessary execution of the remaining code. This will also avoids relying on exceptions for the control flow.

log2akshat (commented):
While raising the Pull Request please follow the proper Zimbra standards in the following format:

Problem Statement
Approach and Fix
Testing Done
Links to any other related pull requests or resources.

store/src/java/com/zimbra/cs/store/file/FileBlobStore.java
RakeshAMore: It is good to have but It should be always fail fast, and relying on exception for controll flow its varies from case to case.

swathy-shaji: For now we cannot do modifications in this PR as the ticket is already closed on May 17th. We need to do again the Pre-UAT even for a single change which we cannot accommodate right now. We will be considering it next time

PR #1578, opened 2024-03-12
ZCS-14701: Refactored code and returning 2FA attributes

zimsuchitgupta:
**Description**

Refactored code and added implementation to return 2FA attributes in AuthResponse

### DIFF
--- a/store/src/java/com/zimbra/cs/service/AuthProvider.java
+++ b/store/src/java/com/zimbra/cs/service/AuthProvider.java
@@ -817,7 +817,7 @@ public abstract class AuthProvider {
         if (null != authProviderExp) {
             throw authProviderExp;
         }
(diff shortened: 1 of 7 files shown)


preshitaagrawal (approved):

k-kato (commented):
Don't we update account/Auth.java? 
It still calls a private function `addTwoFactorAttributes` rather than `AccountUtil.addTwoFactorAttributes`.

https://github.com/Zimbra/zm-mailbox/blob/bd685991066662061c6c62c47ef1d802020310ee/store/src/java/com/zimbra/cs/service/account/Auth.java
Lines 492 and L614

vishnukantjangid (approved):

k-kato (approved):

PR #1583, opened 2024-03-18
ZCS-14787 Set zimbraMobileForceSamsungProtocol25 default to TRUE for activesync

shrutig0510:
**Problem:** 
While connecting on samsung mail app using activesync, by default zimbraMobileForceSamsungProtocol25 is FALSE and it gets connected with 16.0 version

**FIx:**
Setting zimbraMobileForceSamsungProtocol25 default to TRUE. It connects with 2.5 vesrion default and able to send emails from samsung mail app.

**Test-cases:**
Tested on various apps and verified the impact:
Outlook iOS
iPhone mail
Android gmail
Samsung mail app

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -7912,7 +7912,7 @@ TODO: delete them permanently from here
 </attr>
 
 <attr id="1572" name="zimbraMobileForceSamsungProtocol25" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInfo,accountInherited,domainAdminModifiable" since="8.0.7">
-  <defaultCOSValue>FALSE</defaultCOSValue>
+  <defaultCOSValue>TRUE</defaultCOSValue>
   <desc>Whether to force Samsung devices using Active Sync 2.5</desc>
 </attr>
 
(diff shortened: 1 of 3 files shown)


preshitaagrawal (approved):

zimsuchitgupta (approved):

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -7912,7 +7912,7 @@ TODO: delete them permanently from here
 </attr>
 
 <attr id="1572" name="zimbraMobileForceSamsungProtocol25" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInfo,accountInherited,domainAdminModifiable" since="8.0.7">
-  <defaultCOSValue>FALSE</defaultCOSValue>
+  <defaultCOSValue>TRUE</defaultCOSValue>
```
silentsakky, line 7915: can this be applied in a patch?

shrutig0510: The target version is 10.1.0

silentsakky: ok and will this be effective for already created cos in case of upgrade scenario or only for new installations?

shrutig0510: I think it will be effective also in case of upgrade.

dasiyogesh: Default COS value won't get changed for existing one and so the inheritance for the account within the COS. It will need updating the COS for existing ones, for new ones it will work.

PR #1592, opened 2024-04-05
ZBUG-3863 Gracefully handling OOM for reindex account

shrutig0510:
**Problem:** 
OOM on reindexing account having corrupted file.

**Fix:**
The file cannot be fixed if corrupted, so only handling it here to stop mailbox crash.

### DIFF
--- a/store/src/java/com/zimbra/cs/mime/MimeHandler.java
+++ b/store/src/java/com/zimbra/cs/mime/MimeHandler.java
@@ -177,11 +177,14 @@ public abstract class MimeHandler {
      */
     public final String getContent() throws MimeHandlerException {
         if (!DebugConfig.disableMimePartExtraction) {
-            String toRet = getContentImpl();
-            if (toRet == null)
-                return "";
-            else
-                return toRet;
+            String toRet = "";
+            try {
+                toRet = getContentImpl();
+            } catch (OutOfMemoryError e) {
+                // java.lang.OutOfMemoryError: Java heap space handled here
+                ZimbraLog.index.warn("File %s is corrupted causing OOM", filename);
+            }
+            return toRet;
         } else {
             if (dataSource != null && !mDrainedContent) {
                 InputStream is = null;
--- a/store/src/java/com/zimbra/cs/mime/MimeHandler.java
+++ b/store/src/java/com/zimbra/cs/mime/MimeHandler.java
@@ -27,6 +27,7 @@ import java.util.Locale;
 import javax.activation.DataSource;
 import javax.mail.internet.MimeUtility;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.lucene.document.Document;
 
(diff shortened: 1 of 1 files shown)


zimsuchitgupta (approved):

preshitaagrawal (approved):

store/src/java/com/zimbra/cs/mime/MimeHandler.java
```diff
@@ -177,11 +177,14 @@ public void setSize(long value) {
      */
     public final String getContent() throws MimeHandlerException {
         if (!DebugConfig.disableMimePartExtraction) {
-            String toRet = getContentImpl();
-            if (toRet == null)
-                return "";
-            else
-                return toRet;
+            String toRet = "";
```
PrasadpLad: can we try to use import org.apache.commons.lang3.StringUtils package instead of ""

store/src/java/com/zimbra/cs/mime/MimeHandler.java
```diff
@@ -177,11 +177,14 @@ public void setSize(long value) {
...
+            try {
+                toRet = getContentImpl();
+            } catch (OutOfMemoryError e) {
+                // java.lang.OutOfMemoryError: Java heap space handled here
+                ZimbraLog.index.warn("File %s is corrupted causing OOM", filename);
```
dasiyogesh, line 186: This handles the error but the usage of heap for the mime wil still continue right?

PrasadpLad (commented):
approved

store/src/java/com/zimbra/cs/mime/MimeHandler.java
shrutig0510: We have no control there. but i see 
now its exiting from reindexing that account.Earlier it was giving OOM and sometime crashing mailbox.

dasiyogesh: can we apply mime content size check with LDAP attrs we have along with the exception handling to avoid the exception

shrutig0510: Are you indicating about mime content size check "convertd_max_response_size = (long) 20 * 1024 * 1024)"
In that case we can make following change from where this method is called.
` if ((isMainBody && (!handler.runsExternally() || indexAttachments)) ||
                            (indexAttachments && !DebugConfig.disableIndexingAttachmentsTogether && handler.getSize() < DebugConfig.convertdMaxResponseSize)) {
                    toRet = handler.getContent();
                }`

If not then which LDAP attr you are talking about?

dasiyogesh: I guess you were checking `mailboxd_java_heap_size` for heap size, can we use this to check the attachment size before loading to stop the error

PR #1611, opened 2024-05-10
ZBUG-3468: Fetching name from Content-Type if it is invalid as per RFC-2231 in Content-Disposition

zimsuchitgupta:
**Issue**
Broken utf-8 encoding in the mail attachment filenames when sender uses Outlook for Mac

**Reason**
As per analysis, this issue was coming because the filename was on the same line as in  Content-Disposition statement. Due to this, it was considering the filename string from beginning to end the same as it is.

**Solution**
If the name is not valid as per RFC-2231 then ignoring it and falling back to Content-Type and fetching from here.
https://greenbytes.de/tech/webdav/rfc2231.html

### DIFF
--- a/common/src/java-test/com/zimbra/common/mime/MimeCompoundHeaderTest.java
+++ b/common/src/java-test/com/zimbra/common/mime/MimeCompoundHeaderTest.java
@@ -22,7 +22,6 @@ import org.junit.Test;
 public class MimeCompoundHeaderTest {
     private void test(boolean isContentType, String description, String raw, String value, String[] params) {
         MimeCompoundHeader mch = isContentType ? new ContentType(raw) : new ContentDisposition(raw);
-
         String primary = isContentType ? ((ContentType) mch).getContentType() : ((ContentDisposition) mch).getDisposition();
         Assert.assertEquals(description, value, primary);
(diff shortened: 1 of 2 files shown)


preshitaagrawal (approved):

store/src/java/com/zimbra/cs/mime/Mime.java
```diff
@@ -1122,7 +1122,7 @@ public static String getFilename(MimePart mp) {
         // first, check the Content-Disposition header for the "filename" parameter
         try {
             String cdisp = mp.getHeader("Content-Disposition", null);
-            if (cdisp != null) {
+            if (cdisp != null && MimeConstants.CD_ATTACHMENT.equals(cdisp.trim())) {
```
desouzas: Please see commentary on the analysis ticket before moving forward with this approach.

preshitaagrawal (approved):

shrutig0510 (approved):

shrutig0510 (approved):

preshitaagrawal (approved):

dasiyogesh (approved):

dasiyogesh (approved):

PR #1625, opened 2024-06-12
ZCS-15319: Exiting command execution after completion

zimsuchitgupta:
**Problem**
zmplayredo utility is not exiting after execution

**Solution**
Command was not getting terminated after successful execution so added `System.exit(0)`.

### DIFF
--- a/store/src/java/com/zimbra/cs/redolog/util/PlaybackUtil.java
+++ b/store/src/java/com/zimbra/cs/redolog/util/PlaybackUtil.java
@@ -375,6 +375,7 @@ public class PlaybackUtil {
             player.playback();
         } finally {
             teardown();
+            System.exit(0);
         }
     }
 


preshitaagrawal (approved):

store/src/java/com/zimbra/cs/redolog/util/PlaybackUtil.java
```diff
@@ -375,6 +375,7 @@ public static void main(String[] cmdlineargs) throws Throwable {
             player.playback();
         } finally {
             teardown();
+            System.exit(0);
```
dasiyogesh, line 378: why is this needed now?

ronstra-synacor: Yeah, does this actually stop it from hanging? It seemed like the hangup was something in the teardown(), that calls Zimbra.shutdown(), which just shuts down a lot of threads for the server, but if there was a problem, it's still not going to make it to the next System.exit(0) statement, right?

dasiyogesh: there seems to be something going in that shutdown, Suchit is checking on it.

log2akshat: The above solution is not very convincing. Perhaps a thread dump analysis can uncover the mystery.

Try filtering out the stack trace for the state of the threads during `Zimbra.shutdown()`, which, as @ronstra-synacor mentioned, is responsible for shutting down many server threads. Focus primarily on differentiating between `RUNNABLE` and `BLOCKED` threads.

dasiyogesh: Some new threads which are not exiting are causing the issue which will be refactored and fixed in the respective projects. In the mean time we can go with this change for the tool which has no functional impact.

dasiyogesh (approved):

zimsuchitgupta:
Closing this PR as this will be taken care of in another ticket.

PR #1628, opened 2024-06-19
ZCS-15266 : Updated the code to handle external email and Distribution List scenario for SaveDraft.

Gopal-Moolchandani:
Issue: Null Pointer Exception thrown for saveDraftRequest when composing Email.
Fix: Added Null pointer checks for delegator Account.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/SaveDraft.java
+++ b/store/src/java/com/zimbra/cs/service/mail/SaveDraft.java
@@ -100,14 +100,19 @@ public class SaveDraft extends MailDocumentHandler {
         // if mail is not delegated to be Send As, then the delegator account is the same as the account of the mailbox
         String mailAddress = null;
         List<Element> elements = msgElem.listElements(MailConstants.E_EMAIL);
-        for (Element element : elements) {
-            if (element.getAttribute(MailConstants.A_ITEM_TYPE, null).equals(MailConstants.A_FLAGS)) {
-                mailAddress = element.getAttribute(MailConstants.A_ADDRESS, null);
+        if (elements != null) {
+            for (Element element : elements) {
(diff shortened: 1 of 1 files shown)


zimsuchitgupta (approved):

store/src/java/com/zimbra/cs/service/mail/SaveDraft.java
```diff
@@ -177,18 +182,20 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
                 }
-                mm = ParseMimeMessage.parseMimeMsgSoap(zsc, oct, delegatorMbox, msgElem, null, mimeData);
-                SendMsg.doSendMessage(oct, delegatorMbox, mm, mimeData.uploads, null, "r", null, null, false,
-                        false, false, false);
+                if (oct != null && delegatorMbox != null) {
+                    mm = ParseMimeMessage.parseMimeMsgSoap(zsc, oct, delegatorMbox, msgElem, null, mimeData);
+                    SendMsg.doSendMessage(oct, delegatorMbox, mm, mimeData.uploads, null, "r", null, null, false,
+                            false, false, false);
+                }
                 Element response = zsc.createElement(MailConstants.SAVE_DRAFT_RESPONSE);
```
dasiyogesh, line 193: if above  SendMsg.doSendMessage doesn't executes whats the response here?

Gopal-Moolchandani: This particular method i.e. SendMsg.doSendMessage is executed when we proxy the request to delegator Mbox. But the save draft call is actually triggered from Delegatee Mbox. So the response will be specific to Delegatee Mbox not delegator Mbox. That's why we are returning the empty response here and it doesn't depend on execution of SendMsg.doSendMessage.

dasiyogesh (approved):

PR #1630, opened 2024-06-24
ZCS-15392: REGEXP_ESCAPED['+']=true

ronstra-synacor:
Mailboxes with a plus sign would not match any search filter. It seems we had not included '+' as a regexp character to be escaped when converting IMAP wildcards into regexp Patterns. This fixed that, so when you search for "BOB+ALICE" you'll find a literal "BOB+ALICE" but not "BOBBBBBBBBALICE" or nothing at all.

ImapHandlerTest intermittently fails, but it does that on develop, also. And it doesn't seem to fail on my new test, so I'm chalking it up to something wrong with the existing ones. I'll take a look, but this isn't the ideal place to fix them.

### DIFF
--- a/store/src/java/com/zimbra/cs/imap/ImapPath.java
+++ b/store/src/java/com/zimbra/cs/imap/ImapPath.java
@@ -58,8 +58,7 @@ public class ImapPath implements Comparable<ImapPath> {
             FOLDER_ENCODING_CHARSET = Charset.forName("imap-utf-7");
         } catch (Exception e) {
             ZimbraLog.imap.error(
-                "could not load imap-utf-7 charset (perhaps zimbra-charset.jar is not in the jetty endorsed directory)",
-                e);
+                "could not load imap-utf-7 charset (perhaps zimbra-charset.jar is not in the jetty endorsed directory)");
             FOLDER_ENCODING_CHARSET = Charset.forName("utf-8");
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/imap/ImapPath.java
```diff
@@ -58,8 +58,7 @@ enum Scope { UNPARSED, NAME, CONTENT, REFERENCE };
             FOLDER_ENCODING_CHARSET = Charset.forName("imap-utf-7");
         } catch (Exception e) {
             ZimbraLog.imap.error(
-                "could not load imap-utf-7 charset (perhaps zimbra-charset.jar is not in the jetty endorsed directory)",
-                e);
+                "could not load imap-utf-7 charset (perhaps zimbra-charset.jar is not in the jetty endorsed directory)");
```
dasiyogesh, line 61: why is the exception obj logging removed?

ronstra-synacor: It was injecting noise into the test output with a stack trace that adds nothing. I can put it back. It just doesn't seem necessary with a message that descriptive.

zimsuchitgupta (approved):

dasiyogesh (approved):

PR #1632, opened 2024-06-27
ZCS-14337: Fetching StoreManager classes from DB to handle blobcheck utility issue

zimsuchitgupta:
**Issue**
If a customer is using S3 external storage for secondary volume or primary volume and the customer wants to check the consistency of a mailbox by running zmblobchk, then zmblobchk reports the “blob not found“ for the messages which are on the external storage.

**Reason**
1 - This issue only comes when external volume is configured but local config is set to FileBlobStore as zimbra_class_store.
2 - Due to this, code skipped external case and consider blobs are on local volumes and see mentioned errors.

**Solution**
As we have separate implementation of internal and external so fetching volumes information from data base and provided implementation for both Storage classes ie FileBlobStore and ExternalStoreManager.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/admin/CheckBlobConsistency.java
+++ b/store/src/java/com/zimbra/cs/service/admin/CheckBlobConsistency.java
@@ -18,12 +18,14 @@
 package com.zimbra.cs.service.admin;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.soap.AdminConstants;
 import com.zimbra.common.soap.Element;
+import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.cs.account.accesscontrol.AdminRight;
 import com.zimbra.cs.mailbox.Mailbox;
 import com.zimbra.cs.mailbox.MailboxManager;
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/service/admin/CheckBlobConsistency.java
```diff
@@ -65,63 +67,82 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
-
-            for (int mboxId : mailboxIds) {
-                ExternalBlobConsistencyChecker checker = new ExternalBlobConsistencyChecker();
-                BlobConsistencyChecker.Results results = checker.check(null, mboxId, checkSize, reportUsedBlobs);
-                if (results.hasInconsistency() || reportUsedBlobs) { //or checking used blobs
-                    Element mboxEl = response.addElement(AdminConstants.E_MAILBOX).addAttribute(AdminConstants.A_ID, mboxId);
-                    results.toElement(mboxEl);
+        List<Element> volumeElementList = request.listElements(AdminConstants.E_VOLUME);
+        List<Short> volumeIds = new ArrayList<Short>();
+        if (!volumeElementList.isEmpty()) {
```
shrutig0510, line 72: can we have null and empty check here

store/src/java/com/zimbra/cs/service/admin/CheckBlobConsistency.java
```diff
@@ -65,63 +67,82 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
-                    switch (vol.getType()) {
-                        case Volume.TYPE_MESSAGE:
-                        case Volume.TYPE_MESSAGE_SECONDARY:
-                            volumeIds.add(vol.getId());
-                            break;
-                    }
+        }
+
+        List<Volume> vols = VolumeManager.getInstance().getAllVolumes();
+        HashMap<StoreType, StoreManager> storeManagerMap = new HashMap<>();
```
shrutig0510:
```suggestion
        Map<StoreType, StoreManager> storeManagerMap = new HashMap<>();
```

store/src/java/com/zimbra/cs/service/admin/CheckBlobConsistency.java
zimsuchitgupta: As per implementation of `request.listElements(AdminConstants.E_VOLUME)` returned list never be null.

store/src/java/com/zimbra/cs/service/admin/CheckBlobConsistency.java
```diff
@@ -65,63 +67,82 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
-            for (int mboxId : mailboxIds) {
-                BlobConsistencyChecker checker = new BlobConsistencyChecker();
-                BlobConsistencyChecker.Results results = checker.check(volumeIds, mboxId, checkSize, reportUsedBlobs);
-                if (results.hasInconsistency() || reportUsedBlobs) {
-                    Element mboxEl = response.addElement(AdminConstants.E_MAILBOX).addAttribute(AdminConstants.A_ID, mboxId);
-                    results.toElement(mboxEl);
+            } else if (sm instanceof FileBlobStore) {
+                for (int mboxId : mailboxIds) {
+                    BlobConsistencyChecker checker = new BlobConsistencyChecker();
+                    checker.setStoreManager(sm);
```
shrutig0510: The lines from 116 to 120 is same as 106 to 110. We can create one method for these lines of codes and use instanceOf for checker class.

store/src/java/com/zimbra/cs/store/external/ExternalBlobConsistencyChecker.java
```diff
@@ -22,6 +22,7 @@
 import java.util.List;
 
 import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.util.ZimbraLog;
```
shrutig0510: Remove unused import

shrutig0510 (approved):

preshitaagrawal (approved):

dasiyogesh (approved):

PR #1644, opened 2024-08-02
DESC-147 ZRFE-44 Support CSP in Modern UI

ghost:
# Enable CSP on Zimbra Modern

```
su - zimbra
zmlocalconfig -e zimbra_modern_csp="manifest-src 'self' blob:; connect-src 'self' wss: https://zimbra.org; default-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; object-src 'none'; frame-ancestors 'none'; form-action 'self';"
zmprov mcf +zimbraResponseHeader "X-Barry-Test: can-I-still-set-headers?"
zmmailboxdctl restart
```


ghost:
THIS PR IS FOR DEBUG/VALIDATION BUILD AND SHOULD NOT BE MERGED

PR #1655, opened 2024-08-30
ZCS-15708: eml file Import is not working with Zimbra10 server

Gopal-Moolchandani:
Issue: eml file Import is not working with Zimbra10 server
Fix: Modified the code so that server uses nativeFormatter instead of onlyOffice Formatter in case of POST Requests.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/UserServlet.java
+++ b/store/src/java/com/zimbra/cs/service/UserServlet.java
@@ -907,7 +907,7 @@ public class UserServlet extends ZimbraServlet {
         if (context.formatter == null) {
             context.formatter = FormatterFactory.mFormatters.get(context.format);
             // check if its native formatter, and if onlyoffice formatter is available
-            if (context.formatter.getType().equals(FormatType.HTML_CONVERTED)) {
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/service/UserServlet.java
```diff
@@ -907,7 +907,7 @@ private void resolveFormatter(UserServletContext context) throws UserServletExce
         if (context.formatter == null) {
             context.formatter = FormatterFactory.mFormatters.get(context.format);
             // check if its native formatter, and if onlyoffice formatter is available
-            if (context.formatter.getType().equals(FormatType.HTML_CONVERTED)) {
+            if (context.formatter.getType().equals(FormatType.HTML_CONVERTED) && (context.getFmt() == null || !context.getFmt().equals("eml"))){
```
dasiyogesh: why do we need to check specific format here? Should not have specific format in this servlet.

Gopal-Moolchandani: If we don't check the format, it will use OnlyOffice formatter which doesn't support eml files.

zimsuchitgupta (approved):

preshitaagrawal (approved):

dasiyogesh (approved):

PR #1656, opened 2024-09-05
ZCS-15769: Login issue with 10.1.1 OSE build

Gopal-Moolchandani:
Issue: When Modern UI wasn't installed we were getting 404 (Not Found) error.
Fix: Modified the response of `GetDomainInfo` class to send TRUE value for `zimbraModernWebClientDisabled` Attribute when Modern UI is not present.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/admin/GetDomainInfo.java
+++ b/store/src/java/com/zimbra/cs/service/admin/GetDomainInfo.java
@@ -16,6 +16,7 @@
  */
 package com.zimbra.cs.service.admin;
 
+import java.io.File;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
@@ -30,6 +31,7 @@ import com.zimbra.common.localconfig.LC;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.soap.AdminConstants;
 import com.zimbra.common.soap.Element;
+import com.zimbra.common.util.ZimbraLog;
 import com.zimbra.cs.account.AttributeFlag;
 import com.zimbra.cs.account.AttributeManager;
 import com.zimbra.cs.account.AuthToken;
@@ -41,6 +43,7 @@ import com.zimbra.soap.ZimbraSoapContext;
 
 public class GetDomainInfo extends AdminDocumentHandler {
 
+    private static final String zmlicensePath = "/opt/zimbra/bin/zmlicense";
     private static final Set<String> bootstrapInfoAttrs = ImmutableSet.of(
             ZAttrProvisioning.A_zimbraSkinLogoURL,
             ZAttrProvisioning.A_zimbraSkinLogoAppBanner,
@@ -127,6 +130,18 @@ public class GetDomainInfo extends AdminDocumentHandler {
         if (AttributeManager.getInstance().isMultiValued(key)) {
             for (String value : entry.getMultiAttr(key)) {
                 element.addNonUniqueElement(AdminConstants.E_A).addAttribute(AdminConstants.A_N, key).setText(value);
+            }
+        } else if (key.equals(ZAttrProvisioning.A_zimbraModernWebClientDisabled)) {
+            // checking whether it is a FOSS build or a NETWORK build
+            File zmLicenseFile = new File(zmlicensePath);
+            if (!zmLicenseFile.exists()) {
+                element.addNonUniqueElement(AdminConstants.E_A).addAttribute(AdminConstants.A_N, key).setText("TRUE");
+                ZimbraLog.account.info("Since it is a FOSS build, switching to Classic UI");
+            } else {
+                String value = entry.getAttr(key, null);
+                if (value != null) {
+                    element.addNonUniqueElement(AdminConstants.E_A).addAttribute(AdminConstants.A_N, key).setText(value);
+                }
             }
         } else {
             String value = entry.getAttr(key, null);


silentsakky (approved):

store/src/java/com/zimbra/cs/service/admin/GetDomainInfo.java
```diff
@@ -129,8 +134,13 @@ private void addAttrElementIfNotNull(Entry entry, Element element, String key) t
                 element.addNonUniqueElement(AdminConstants.E_A).addAttribute(AdminConstants.A_N, key).setText(value);
             }
         } else {
+            // Checking whether Modern UI is installed
```
dasiyogesh: comment should start with small case

store/src/java/com/zimbra/cs/service/admin/GetDomainInfo.java
```diff
@@ -129,8 +134,13 @@ private void addAttrElementIfNotNull(Entry entry, Element element, String key) t
...
+            File modernUIFile = new File(modernUIFolderPath);
             String value = entry.getAttr(key, null);
-            if (value != null) {
+            if (key.equals(ZAttrProvisioning.A_zimbraModernWebClientDisabled) && !(modernUIFile.exists() && modernUIFile.isDirectory())) {
```
dasiyogesh, line 140: shouldn't this be OR condition when zimbraModernWebClientDisabled is TRUE or modern UI is not installed in either of cases response should be TRUE. Also if zimbraModernWebClientDisabled is TRUE we won't need to check it right as that's explicitly disable even if the Modern UI exists

store/src/java/com/zimbra/cs/service/admin/GetDomainInfo.java
```diff
@@ -129,8 +133,13 @@ private void addAttrElementIfNotNull(Entry entry, Element element, String key) t
                 element.addNonUniqueElement(AdminConstants.E_A).addAttribute(AdminConstants.A_N, key).setText(value);
             }
         } else {
+            // checking whether Modern UI is installed
+            File modernUIFile = new File(modernUIFolderPath);
             String value = entry.getAttr(key, null);
-            if (value != null) {
+            if (key.equals(ZAttrProvisioning.A_zimbraModernWebClientDisabled) && !(modernUIFile.exists() && modernUIFile.isDirectory())) {
+                element.addNonUniqueElement(AdminConstants.E_A).addAttribute(AdminConstants.A_N, key).setText(TRUE);
```
dasiyogesh: don't see the need of var TRUE here, we can simply pass the string "TRUE" to response.

store/src/java/com/zimbra/cs/service/admin/GetDomainInfo.java
```diff
@@ -129,8 +133,13 @@ private void addAttrElementIfNotNull(Entry entry, Element element, String key) t
                 element.addNonUniqueElement(AdminConstants.E_A).addAttribute(AdminConstants.A_N, key).setText(value);
             }
         } else {
+            // checking whether Modern UI is installed
+            File modernUIFile = new File(modernUIFolderPath);
             String value = entry.getAttr(key, null);
-            if (value != null) {
+            if (key.equals(ZAttrProvisioning.A_zimbraModernWebClientDisabled) && !(modernUIFile.exists() && modernUIFile.isDirectory())) {
```
k-kato: Is the condition correct?
* If zimbraModernWebClientDisabled is TRUE, modern client is disabled. It is okay to return TRUE as it is.
* If zimbraModernWebClientDisabled is FALSE, modern client is enabled. However, if the modern directory does not exist, Modern UI is disabled. (i.e. zimbraModernWebClientDisabled should be set to TRUE)
* If zimbraModernWebClientDisabled is empty, process in the same way as FALSE.

dasiyogesh: `key.equals(ZAttrProvisioning.A_zimbraModernWebClientDisabled)` is just checking whether its the attribute on which we want to add a check for

k-kato: Oh, sorry it was my misunderstanding. I though it checked a value. Thanks @dasiyogesh

store/src/java/com/zimbra/cs/service/admin/GetDomainInfo.java
ronstra-synacor: If this was an existing constant for boolean values in XML attributes, we should probably keep using it, to properly transfer the canonical string representation, e.g. "TRUE" instead of "true" or "t" or "yes" or "1" or whatever. I recall using these TRUE/FALSE constants somewhere in Provisioning...

store/src/java/com/zimbra/cs/service/admin/GetDomainInfo.java
```diff
@@ -129,8 +132,13 @@ private void addAttrElementIfNotNull(Entry entry, Element element, String key) t
...
+                element.addNonUniqueElement(AdminConstants.E_A).addAttribute(AdminConstants.A_N, key).setText("TRUE");
+                ZimbraLog.account.info("Modern UI is not installed, switching to Classic UI");
+            } else if (value != null) {
                 element.addNonUniqueElement(AdminConstants.E_A).addAttribute(AdminConstants.A_N, key).setText(value);
```
ronstra-synacor: Maybe put the key equals check into this top-level `else/if` to avoid going into that block at all if we're not dealing with the relevant key. I doubt instantiating a new `File` object (almost a no-op) and fetching an `Entry` attribute aren't expensive, but we could avoid them nonetheless.

dasiyogesh (approved):

preshitaagrawal (approved):

ronstra-synacor (approved):

PR #1667, opened 2024-10-14
folder

mamtadahiwal:
Fix for Folder Structure Null check

### DIFF
--- a/store/src/java-test/com/zimbra/cs/service/admin/ModifyAccountTest.java
+++ b/store/src/java-test/com/zimbra/cs/service/admin/ModifyAccountTest.java
@@ -1,0 +1,76 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2024 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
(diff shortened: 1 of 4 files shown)


chitranshu198 (approved):

chitranshu198 (approved):

shrutig0510:
Please rename the title of PR with ticket no. Check any PR for reference

mamtadahiwal:
> Please rename the title of PR with ticket no. Check any PR for reference

@shrutig0510  - changed name , Please check

shrutig0510 (approved):

store/src/java/com/zimbra/cs/mailbox/Folder.java
```diff
@@ -806,7 +812,11 @@ boolean canContain(MailItem child) {
             // may not contain our parents or grandparents (c.f. Back to the Future)
             for (Folder folder = this; folder.getId() != Mailbox.ID_FOLDER_ROOT; folder = folder.parent) {
                 if (folder.getId() == child.getId()) {
-                    return false;
+                        return false;
+                }
+                if (folder.parent == null) {
+                    folder.setupParent();
```
shrutig0510: @mamtadahiwal It will always return false even if the parent is set in this method. You need to double check for null here

mamtadahiwal: if (folder.parent == null) {
                    folder.setupParent();
                    if (folder.parent == null) {
                        return false;
                    }
                } like this ur saying

PR #1673, opened 2024-10-29
ZBUG-4384 : "Send Later" scheduled messages disappearing on nonexistent email address

mamtadahiwal:
Problem:  "Send Later" scheduled messages disappear without sending when using a nonexistent email address for "zimbraAllowFromAddress"

Fix: code changes is depends on ZBUG:2956  , in ZBUG-2956 one scenario as remain to handle that if delegator account is null .

so in this ticket we have handle that scenario.

Test-cases:

### DIFF
--- a/store/src/java/com/zimbra/cs/mailbox/AutoSendDraftTask.java
+++ b/store/src/java/com/zimbra/cs/mailbox/AutoSendDraftTask.java
@@ -118,32 +118,36 @@ public class AutoSendDraftTask extends ScheduledTask<Object> {
         mailSender.setIdentity(StringUtil.isNullOrEmpty(msg.getDraftIdentityId()) ? null : mbox.getAccount().getIdentityById(msg.getDraftIdentityId()));
         Provisioning provisioning = Provisioning.getInstance();
         Account delegatorAccount = provisioning.getAccountById(getDelegatorAccountId());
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/mailbox/AutoSendDraftTask.java
```diff
@@ -118,32 +118,36 @@ public void setRequest(Element request) {
         mailSender.setIdentity(StringUtil.isNullOrEmpty(msg.getDraftIdentityId()) ? null : mbox.getAccount().getIdentityById(msg.getDraftIdentityId()));
         Provisioning provisioning = Provisioning.getInstance();
         Account delegatorAccount = provisioning.getAccountById(getDelegatorAccountId());
-        if (delegatorAccount != null && Provisioning.onLocalServer(delegatorAccount)) {
-            delegatorMbox = MailboxManager.getInstance().getMailboxByAccount(delegatorAccount);
+        if (delegatorAccount != null) {
```
chitranshu198:
```suggestion
        if (null != delegatorAccount) {
```

chitranshu198 (approved):

shrutig0510 (approved):

Gopal-Moolchandani (approved):

PR #1682, opened 2024-12-05
Bugfix/zbug 3254

mamtadahiwal:
issue is  - dav - Request: requested content-type: text/xml, actual content-type: text/plain

we are getting text/plain because of filename is not ending with any extension
so we have added custom detector for this spacial case .

image

### DIFF
--- a/store/src/java/com/zimbra/cs/dav/DavContext.java
+++ b/store/src/java/com/zimbra/cs/dav/DavContext.java
@@ -113,6 +113,7 @@ public class DavContext {
     private String mCollectionPath;
     private RequestProp mResponseProp;
     private String mDavCompliance;
+    public static boolean noFileExtension;
     /**
      * "actingAsDelegateFor" is used to form part of the scheduling inbox or scheduling outbox URL used
      * for scheduling on behalf of another user (the principal we are acting as a delegate for).
@@ -254,6 +255,11 @@ public class DavContext {
                 URI uri = HttpUtil.getUriFromFragments(fragments, req.getQueryString(), true, false);
                 mUri = uri.getPath();
             }
+            int lastDot = mUri.lastIndexOf(".");
+            noFileExtension = mUri.endsWith("/")
+                    || !(lastDot > 0
+                    && lastDot < mUri.length() - 1
+                    && mUri.substring(lastDot + 1).matches("[a-zA-Z0-9]+"));
 
             int index = mUri.indexOf('/', 1);
             if (index > 0) {
@@ -363,7 +369,6 @@ public class DavContext {
         }
         return null;
     }
-
     public String getActingAsDelegateFor() {
         return actingAsDelegateFor;
     }
--- a/store/src/java/com/zimbra/cs/dav/carddav/CustomXmlDetector.java
+++ b/store/src/java/com/zimbra/cs/dav/carddav/CustomXmlDetector.java
@@ -1,0 +1,66 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2004, 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019, 2020, 2021, 2022, 2023,2024 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ * ***** END LICENSE BLOCK *****
+ */
+package com.zimbra.cs.dav.carddav;
+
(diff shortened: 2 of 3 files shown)


store/src/java/com/zimbra/cs/dav/carddav/CustomXmlDetector.java
```diff
@@ -0,0 +1,49 @@
+package com.zimbra.cs.dav.carddav;
```
dasiyogesh, line 17: @mamtadahiwal have you tried verifying the content type from Tika mime type detection using https://github.com/Zimbra/zm-mailbox/blob/develop/store-conf/conf/custom-mimetypes.xml

mamtadahiwal: yes , it is not detecting using custom-mimetypes.xml , I have tried changing dependency version as well as i have try with adding some xml tag and changing in existing file also

store/src/java/com/zimbra/cs/dav/carddav/CustomXmlDetector.java
Gopal-Moolchandani, line 1: @mamtadahiwal 

If possible can you please attach any existing reference of the detection issue with apache tika detector. If its not available I would recommend testing the xml file with a simple tika detector program to be sure that we are going the right way.

mamtadahiwal: @Gopal-Moolchandani 
Please find this conversation → [Apache Tika unable to detect Content-Type using file content](https://stackoverflow.com/questions/69448197/apache-tika-unable-to-detect-content-type-using-file-content?newreg=27a5a196c2e04a258fe12ce076ff20e9) 


We are also passing the file with extension .tmp in fileitem reference

mamtadahiwal: image

Gopal-Moolchandani (approved):

chitranshu198 (approved):

store/src/java/com/zimbra/cs/dav/DavContext.java
```diff
@@ -113,6 +113,7 @@ public static void removeExtension(Extension ex) {
     private String mCollectionPath;
     private RequestProp mResponseProp;
     private String mDavCompliance;
+    public static boolean hasFileExtension;
```
dasiyogesh: @mamtadahiwal this is a request for the account property does not get share across

ashwinsahu19 (approved):

store/src/java/com/zimbra/cs/dav/carddav/CustomXmlDetector.java
shrutig0510, line 17: Add license block

store/src/java/com/zimbra/cs/dav/carddav/CustomXmlDetector.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Custom detector for XML content types .
+ * Extends Apache Tika's Detector interface to provide specialized XML detection.
+ */
+public class CustomXmlDetector implements Detector {
+
+    private final Detector defaultDetector;
+    private final FileItem file;
+    private String ctype;
```
shrutig0510:
```suggestion
    private String cType;
```

store/src/java/com/zimbra/cs/dav/carddav/CustomXmlDetector.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param input The input stream to analyze
+     * @param metadata Additional metadata about the content
+     * @return MediaType.APPLICATION_XML if content type is "text/xml",
+     *         otherwise delegates to default detector
+     * @throws IOException If an error occurs while reading the input stream
+     */
+    @Override
+    public MediaType detect(InputStream input, Metadata metadata) throws IOException {
+        ctype = file.getContentType();
+        if (ctype != null && ctype.equalsIgnoreCase("text/xml")) {
```
shrutig0510:
```suggestion
        if (ctype != null && ctype.equalsIgnoreCase(MimeConstants.CT_TEXT_XML)) {
```

store/src/java/com/zimbra/cs/dav/DavContext.java
```diff
@@ -113,6 +113,7 @@ public static void removeExtension(Extension ex) {
     private String mCollectionPath;
     private RequestProp mResponseProp;
     private String mDavCompliance;
+    public static boolean noFileExtension;
```
mamtadahiwal, line 116: @dasiyogesh this boolean flag we have specifically added to check at FileUpload class if the file has extension or not
FileUploadServlet is used extensively so adding a new flag to it would have been difficult.
This flag is false by default, it is turned true only in this class and once processed at fileUploadServlet it is turned back to false to make sure there is no impact on existing flow.

store/src/java/com/zimbra/cs/dav/DavContext.java
mamtadahiwal: @dasiyogesh this boolean flag we have specifically added to check at FileUpload class if the file has extension or not
FileUploadServlet is used extensively so adding a new flag to it would have been difficult.
This flag is false by default, it is turned true only in this class and once processed at fileUploadServlet it is turned back to false to make sure there is no impact on existing flow.

store/src/java/com/zimbra/cs/dav/DavContext.java
dasiyogesh: add `vcard` to https://github.com/Zimbra/zm-mailbox/blob/develop/store-conf/conf/custom-mimetypes.xml#L64 and check if this identifies the request CT properly

PR #1683, opened 2024-12-06
ZCS-16289 : SaveDraft request not failing on exceeding zimbraMtaMaxMessageSize

viditZim:
https://synacor.atlassian.net/browse/ZCS-16289

Problem:- On Modern/Classic, SaveDraftRequest does not fail on exceeding message size limit of zimbraMtaMaxMessageSize.

Solution:- Modified the condition such that when the mailAttachmentMaxSize is not set, it will consider mtaMaxMsgSize as the limit.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/ParseMimeMessage.java
+++ b/store/src/java/com/zimbra/cs/service/mail/ParseMimeMessage.java
@@ -284,7 +284,7 @@ public final class ParseMimeMessage {
                 long mailAttachmentMaxSize = account.getMailAttachmentMaxSize();
                 Config config = Provisioning.getInstance().getConfig();
                 long mtaMaxMsgSize = config.getLongAttr(Provisioning.A_zimbraMtaMaxMessageSize, -1);
-                if (mailAttachmentMaxSize > mtaMaxMsgSize) {
+                if (mailAttachmentMaxSize > mtaMaxMsgSize || mailAttachmentMaxSize == 0) {
                     maxSize = mtaMaxMsgSize;
                 } else {
                     maxSize = mailAttachmentMaxSize;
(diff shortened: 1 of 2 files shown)


preshitaagrawal (approved):

store/src/java/com/zimbra/cs/service/mail/ParseMimeMessage.java
```diff
@@ -284,7 +284,7 @@ private static class ParseMessageContext {
                 long mailAttachmentMaxSize = account.getMailAttachmentMaxSize();
                 Config config = Provisioning.getInstance().getConfig();
                 long mtaMaxMsgSize = config.getLongAttr(Provisioning.A_zimbraMtaMaxMessageSize, -1);
-                if (mailAttachmentMaxSize > mtaMaxMsgSize) {
+                if (mailAttachmentMaxSize > mtaMaxMsgSize || mailAttachmentMaxSize == 0) {
```
dasiyogesh, line 287: @viditZim could you check if the request fails if the mailAttachmentMaxSize & mtaMaxMsgSize is not set and maxsize is -1

dasiyogesh: its max value then

dasiyogesh (approved):

store/src/java/com/zimbra/cs/service/mail/ParseMimeMessage.java
viditZim: In that case, the limit will be unlimited because we're setting as Long.MAX_VALUE for negative.

https://github.com/Zimbra/zm-mailbox/blob/36f7286f28527de70341b10f614903ff936e406a/store/src/java/com/zimbra/cs/service/mail/ParseMimeMessage.java#L295

rcyarrapothu (approved):

Gopal-Moolchandani (approved):

ashwinsahu19 (approved):

PR #1698, opened 2025-01-30
ZBUG-4409: Missing Blobs on a restored account from backup with S3 storage

Gopal-Moolchandani:
Issue: With-Blob Backup was not supported for external storage.

Fix: Modified the Backup and Restore code to support with-blob backup. For external storage, blobs will be restored in the same bucket.


store/src/java/com/zimbra/cs/store/external/ExternalStoreManager.java
```diff
@@ -272,6 +273,31 @@ public StagedBlob stage(Blob blob, Mailbox mbox) throws IOException, ServiceExce
         }
     }
 
+    public StagedBlob stage(Blob blob, Mailbox mbox, Volume volume) throws IOException, ServiceException {
+        if (supports(StoreFeature.RESUMABLE_UPLOAD) && blob instanceof ExternalUploadedBlob) {
+            ZimbraLog.store.debug("blob already uploaded, just need to commit");
+            String locator = ((ExternalResumableUpload) this).finishUpload((ExternalUploadedBlob) blob);
+            if (locator != null) {
+                ZimbraLog.store.debug("wrote to locator %s",locator);
```
preshitaagrawal:
```suggestion
                ZimbraLog.store.debug("wrote to locator %s", locator);
```

store/src/java/com/zimbra/cs/store/external/ExternalStoreManager.java
```diff
@@ -303,7 +329,35 @@ public StagedBlob stage(InputStream in, long actualSize, Mailbox mbox) throws Se
...
+        try {
+            digest = MessageDigest.getInstance("SHA-256");
+        } catch (NoSuchAlgorithmException e) {
+            throw ServiceException.FAILURE("SHA-256 digest not found", e);
+        }
+        ByteUtil.PositionInputStream pin = new ByteUtil.PositionInputStream(new DigestInputStream(in, digest));
 
+        try {
+            String locator = writeStreamToStore(pin, actualSize, mbox, volume.getId());
+            if (locator != null) {
```
mamtadahiwal:
```suggestion
            if (null != locator) {
```

store/src/java/com/zimbra/cs/store/StoreManager.java
```diff
@@ -306,6 +321,21 @@ public StagedBlob stage(InputStream data, Mailbox mbox)
         return stage(data, -1, mbox);
```
mamtadahiwal: instead of hard code value can we declare variable for -1 , because its confusing

store/src/java/com/zimbra/cs/store/external/ExternalStoreManager.java
```diff
@@ -272,6 +273,31 @@ public StagedBlob stage(Blob blob, Mailbox mbox) throws IOException, ServiceExce
         }
     }
 
+    public StagedBlob stage(Blob blob, Mailbox mbox, Volume volume) throws IOException, ServiceException {
+        if (supports(StoreFeature.RESUMABLE_UPLOAD) && blob instanceof ExternalUploadedBlob) {
+            ZimbraLog.store.debug("blob already uploaded, just need to commit");
+            String locator = ((ExternalResumableUpload) this).finishUpload((ExternalUploadedBlob) blob);
+            if (locator != null) {
```
mamtadahiwal:
```suggestion
            if (null != locator) {
```

store/src/java/com/zimbra/cs/store/external/ExternalStoreManager.java
```diff
@@ -272,6 +273,31 @@ public StagedBlob stage(Blob blob, Mailbox mbox) throws IOException, ServiceExce
...
+                localCache.put(locator, getContent(blob));
+            } else {
+                ZimbraLog.store.warn("blob staging returned null locator");
+            }
+            return new ExternalStagedBlob(mbox, blob.getDigest(), blob.getRawSize(), locator);
+        } else {
+            InputStream is = getContent(blob);
+            try {
+                StagedBlob staged = stage(is, blob.getRawSize(), mbox, volume);
+                if (staged != null && staged.getLocator() != null) {
```
mamtadahiwal:
```suggestion
                if (null != staged &&  null != staged.getLocator()) {
```

store/src/java/com/zimbra/cs/store/external/ExternalBlobIO.java
```diff
@@ -38,6 +38,21 @@ public interface ExternalBlobIO {
...
+     * @param actualSize: size of data in stream, or -1 if size is unknown. To be used by implementor for optimization where possible
+     * @param mbox:       Mailbox which contains the blob. Can optionally be used by store for partitioning
+     * @param destVolId:  Destination Volume id on which we need to store the blob
+     * @return locator string for the stored blob, unique identifier created by storage protocol
+     * @throws IOException
+     * @throws ServiceException
+     */
+    default String writeStreamToStore(InputStream in, long actualSize, Mailbox mbox, short destVolId) throws IOException, ServiceException {
+        return null;
+    }
```
viditZim, line 54: Not able to understand why have we created the default method which is directly returning null?

dasiyogesh: @viditZim this being an interface is a way to fallback for methods to not require implementation for classes already implemented this interface to not break them.
@Gopal-Moolchandani pls check on the returning null.

RakeshAMore: You can throw exception such as `NotImplementedException` or `UnsupportedOperation`

mamtadahiwal (approved):

store/src/java/com/zimbra/cs/store/external/ExternalStoreManager.java
```diff
@@ -303,7 +329,35 @@ public StagedBlob stage(InputStream in, long actualSize, Mailbox mbox) throws Se
...
+            digest = MessageDigest.getInstance("SHA-256");
+        } catch (NoSuchAlgorithmException e) {
+            throw ServiceException.FAILURE("SHA-256 digest not found", e);
+        }
+        ByteUtil.PositionInputStream pin = new ByteUtil.PositionInputStream(new DigestInputStream(in, digest));
 
+        try {
+            String locator = writeStreamToStore(pin, actualSize, mbox, volume.getId());
+            if (null != locator) {
+                ZimbraLog.store.debug("wrote to locator %s",locator);
```
preshitaagrawal, line 352:
```suggestion
                ZimbraLog.store.debug("wrote to locator %s", locator);
```

ashwinsahu19 (approved):

preshitaagrawal (approved):

store/src/java-test/com/zimbra/cs/store/StoreManagerAfterReset.java
```diff
@@ -59,6 +60,11 @@ public StagedBlob stage(InputStream data, long actualSize, Mailbox mbox) throws
         return null;
     }
 
+    @Override
+    public StagedBlob stage(InputStream data, long actualSize, Mailbox mbox, Volume volume) throws IOException, ServiceException {
+        return null;
```
RakeshAMore: What about `NotImplementedException` or `UnsupportedOperation`

store/src/java/com/zimbra/cs/store/file/FileBlobStore.java
```diff
@@ -118,6 +118,12 @@ public VolumeStagedBlob stage(InputStream in, long actualSize, Mailbox mbox)
         return new VolumeStagedBlob(mbox, (VolumeBlob) blob).markStagedDirectly();
     }
 
+    @Override
+    public StagedBlob stage(InputStream data, long actualSize, Mailbox mbox, Volume volume) throws IOException, ServiceException {
+        // mailbox store is on the same volume as incoming directory, so no need to stage the blob
+        throw ServiceException.FAILURE("Operation can not be completed because ExternalStoreManager is not available", null);
```
RakeshAMore: It is not necessary this method will be always executed for external storage, this message should be generic

dasiyogesh (approved):

store/src/java/com/zimbra/cs/store/external/ExternalBlobIO.java
log2akshat: @dasiyogesh Returning `null` from a default interface method is a bad design choice, regardless of backward compatibility concerns. This silently introduces `NullPointerExceptions` and violates the principle of least surprise.

The correct and safe approach is to throw an `UnsupportedOperationException` or similar instead of silently returning a `null` that could fail unpredictably at runtime. This is like introducing bug purposely and then fixing it.

dasiyogesh: right, it was returning null earlier thats what was to be checked and handle it for which above change is done.

PR #1714, opened 2025-04-22
ZCS-16703 Logic for handling external storage in secondary volume for mailboxmove from 9.0 to 10.0

ashwinsahu19:
Mailbox move from 9.0 to 10.1.6 is not compatible because there are major updates made in 10 version release for handling external storage and backup.
When mbox move starts:

First an API is fired to source to check if external storage exists “GetMailboxCurrentVolumes”. This API is not available in 9.0 and I tried with other API’s but those are not returning data for secondary volumes(external storage). Added this API in 9.0 with default value 2 for storage_type , storage_type variable in unavailable in version 9(this external volume flag in response is required to run the importExternalBlobs() flow in 10 version)

For storing data in the same volume as of source to the destination, while taking the backup “volume id” is added to the path of the blob folder. While extracting the blobs in 10.0, the same “volume id” is extracted and the data get stored into the respective volume of destination server. This code to add volume id info into the blob path is missing in version 9.0 . So without volume id its not possible to map the data in the destination server to primary or secondary volumes. I have added code changes in version 9 for the same but with multiple workarounds.

9.0 misses few columns that were added in 10.0 to handle the external storage flow(also the zm-store-manager repo added in 10.0 for externalStorage management). Adding new columns to the 9 version TABLE would not be appropriate approach as this will be a one time change and we will revert back to the previous jars. As a workaround, I have added 2 LC attribute to handle the store_manager_class value that comes out of volume table in 10.0 for finding out the StorageManager instance. This Store Manager class is further used in the code for extracting the file from internal or external storage.

This code has the multithreading changes too for performance improvement while taking the backup in movemailbox flow. So LC attribute for thread count also needs to be configured(by default its 5)

These changes will only handle the MOVEMAILBOX flow. It will not be handle any other flow like restore.

### DIFF
--- a/common/src/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/java/com/zimbra/common/localconfig/LC.java
@@ -1438,6 +1438,8 @@ public final class LC {
 
     public static final KnownKey enable_delegated_admin_ldap_access = KnownKey.newKey(true);
 
+    public static final KnownKey primary_store_manager_class = KnownKey.newKey("");
+    public static final KnownKey secondary_store_manager_class = KnownKey.newKey("");
     // OAuth2 Social
     public static final KnownKey zm_oauth_classes_handlers_yahoo = KnownKey.newKey("com.zimbra.oauth.handlers.impl.YahooOAuth2Handler");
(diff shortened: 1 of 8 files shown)


Gopal-Moolchandani (commented):
We can add a fallback which could inform user that new local configs are unset.
LGTM

PR #1716, opened 2025-04-24
ZCS-15799:Ignite - Password expiry reminder

viditZim:
Problem: As an admin, I should be able to set password reminders for the users so that they are informed about the upcoming password expiry of their account and accordingly can take the required action

As a user, I should get a password reminder email when it is about to expire so that I can take appropriate actions.

Solution: Implemented the complete password expiry reminder feature.


shrutig0510 (approved):

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -10574,5 +10574,10 @@ TODO: delete them permanently from here
 <attr id="4140" name="zimbraPrefMailDeliveryStatusNotification" type="boolean" cardinality="single" optionalIn="account,cos" flags="accountInherited" since="10.1.8">
   <defaultCOSValue>FALSE</defaultCOSValue>
   <desc>whether web UI should always request delivery status notification for outgoing messages</desc>
+ </attr>
+
+<attr id="4141" name="zimbraFeaturePasswordExpiryReminderEnabled" type="boolean" cardinality="single" optionalIn="cos,account" flags="accountInherited,accountInfo" since="10.1.7">
```
umagmrit: Can we update the `since` field to 10.1.9?

store/src/java/com/zimbra/cs/service/mail/EmailTracker.java
```diff
@@ -0,0 +1,91 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2024 Synacor, Inc.
```
umagmrit:
```suggestion
 * Copyright (C) 2025 Synacor, Inc.
```

store/src/java/com/zimbra/cs/service/mail/EmailTracker.java
```diff
@@ -0,0 +1,91 @@
...
+package com.zimbra.cs.service.mail;
+
+import com.zimbra.common.util.ZimbraLog;
+
+import java.io.*;
+import java.util.HashSet;
+import java.util.Set;
+
+public class EmailTracker {
+    private static final String SENT_EMAILS_FILE = "/tmp/sent_emails.txt";
```
umagmrit: Since `sent_emails.txt` is stored in `/tmp/`, the file may be lost on server reboot, resulting in the loss of tracking for already sent emails.

viditZim: @shrutig0510 @dasiyogesh what do you suggest here?

shrutig0510: Can we store this file somewhere in /opt/zimbra?
There are already some files getting stored at /opt/zimbra/zmstat
Cc: @umagmrit

umagmrit: can we use localconfig `zimbra_tmp_directory` ?

dasiyogesh: I think we can use `zimbra_tmp_directory` to hold temp data

viditZim: done.

PR #1739, opened 2025-07-13
ZCS-17599: The user is receiving password expiry notifications even though the preference is disabled

viditZim:
Problem: If any use does not have a COS assigned apart from default and has password about to expire within the deadline was receiving the reminder even though the feature was not explicitly enabled at account or COS level.

Solution: Added a separate check at account level for the feature. Also, checking the Password Max age at account level instead of COS level.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/PasswordExpiryNotifier.java
+++ b/store/src/java/com/zimbra/cs/service/mail/PasswordExpiryNotifier.java
@@ -167,8 +167,7 @@ public class PasswordExpiryNotifier implements Runnable {
         ZLdapFilterFactory filterFactory = ZLdapFilterFactory.getInstance();
         // filtering Coses
         ZLdapFilter filterPasswordExpiryReminderEnabledCos = filterFactory.cosesWithLdapFeatureCheck(Provisioning.A_zimbraFeaturePasswordExpiryReminderEnabled, "TRUE");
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/service/mail/PasswordExpiryNotifier.java
```diff
@@ -228,16 +229,18 @@ private static void clearSentEmailsFile() {
...
-            Date lastChange = account.getGeneralizedTimeAttr(Provisioning.A_zimbraPasswordModifiedTime, null);
-            if (lastChange != null) {
-                long last = lastChange.getTime();
-                long curr = System.currentTimeMillis();
-                long expires = last + (Constants.MILLIS_PER_DAY * maxAge);
-                String expiresOn = formatter.format(new Date(expires));
-                long deadline = Math.round((float) (curr - expires) / -86400000);
-                if (deadline > 0 && deadline <= 10) {
-                    sendMail(account, deadline, expiresOn);
+            if (account.isFeaturePasswordExpiryReminderEnabled()) {
```
dasiyogesh, line 232: wouldn't this trigger LDAP query for the account?

ashwinsahu19 (approved):

store/src/java/com/zimbra/cs/service/mail/PasswordExpiryNotifier.java
viditZim: @dasiyogesh as discussed, this will first try to fetch the value from the cache and will not trigger LDAP query.
Also, tested the change with 100k accounts...did not observe any memory issues.

dasiyogesh (approved):

PR #1756, opened 2025-09-16
ZBUG-4620: Fix login issue by resolving domain on backend using serverName

sonali033:
Issue:
Login was failing when users entered only their username without a domain while the "password must change" policy was enforced.

Fix Implemented:
Updated the UI to call a backend method, passing the server name.
On the server side, the domain is retrieved using the provided serverName.
The backend constructs the complete username.

### DIFF
--- a/store/src/java/com/zimbra/cs/account/Provisioning.java
+++ b/store/src/java/com/zimbra/cs/account/Provisioning.java
@@ -1086,6 +1086,13 @@ public abstract class Provisioning extends ZAttrProvisioning {
         return get(acctSel.getBy().toKeyAccountBy(), acctSel.getKey());
     }
 
+    public Account get(AccountSelector acctSel, String serverName)
+            throws ServiceException {
+        String key = acctSel.getKey();
+        key = resolveDomainForUsername(key, serverName);
+        return get(acctSel.getBy().toKeyAccountBy(), key);
+    }
+
     public static interface EagerAutoProvisionScheduler {
         // returns whether a shutdown has been request to the scheduler
         public boolean isShutDownRequested();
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/account/Provisioning.java
```diff
@@ -1086,6 +1086,13 @@ public Account get(AccountSelector acctSel)
         return get(acctSel.getBy().toKeyAccountBy(), acctSel.getKey());
     }
 
+    public Account get(AccountSelector acctSel, String serverName)
```
shrutig0510, line 1089: Is this api only called on change password?

store/src/java/com/zimbra/cs/account/Provisioning.java
```diff
@@ -2798,4 +2805,21 @@ public String createAddressList(Domain domain, String name, String desc, Map<Str
     }
     
     public abstract String sendMdmEmail(String status, String timeInterval) throws ServiceException;
+
+    public String expandWithDomain(String username, String virtualHostname) throws ServiceException {
```
k-kato: Please check if there is any better name to indicate its function.

store/src/java/com/zimbra/cs/account/Provisioning.java
```diff
@@ -2798,4 +2805,21 @@ public String createAddressList(Domain domain, String name, String desc, Map<Str
...
+        if (username.contains("@")) {
+            return username;
+        }
+        if (virtualHostname == null) {
+            throw ServiceException.INVALID_REQUEST("Server name cannot be null for domain resolution", null);
+        }
+        Domain domain = get(Key.DomainBy.virtualHostname, virtualHostname.toLowerCase());
+        if (domain == null) {
+            throw ServiceException.INVALID_REQUEST("Unable to resolve domain for username: " + username, null);
+        }
```
k-kato, line 2822: What will occur when non-vitrualHostname like IP address is used in access URL? Will be the exception thrown? If so, frontend needs to do an appropriate error handling for a case that an exception is thrown.
Technically, any hostname (server name) can be set in an access url by adding a hostname to `hosts` file on user's local computer.

omkar-oak-01 (commented):
can we add one unit test case where valid and invalid host/domain names can be added which will validate the username

sonali033:
> can we add one unit test case where valid and invalid host/domain names can be added which will validate the username

This is discussed in separate call, no need of it.

k-kato (approved):

rcyarrapothu (approved):

PR #1758, opened 2025-09-26
ZCS-18009 : New Ldap attributes for Legacy BackupRestore 2.0

Gopal-Moolchandani:
Added following new ldap attributes for the new legacy backup 2.0 

Deduplication: zimbraBackupDeduplicationEnabled (default: TRUE).

Compression: Introduced new value ZSTD for zimbraBackupBlobsCompressType , which will be supported only when zimbraBackupDeduplicationEnabled is enabled.

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -10608,4 +10608,9 @@ TODO: delete them permanently from here
   <desc>Package Usage Reporting Scheduler last successful run date</desc>
 </attr>
 
+<attr id="4146" name="zimbraPURDomainDetailsCollectionEnabled" type="boolean" cardinality="single" optionalIn="globalConfig" requiresRestart="mailbox" since="10.1.12">
+  <globalConfigValue>FALSE</globalConfigValue>
+  <desc>Flag to enable or disable the collection of domain details</desc>
+</attr>
+
 </attrs>
--- a/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
+++ b/common/src/java/com/zimbra/common/account/ZAttrProvisioning.java
@@ -9517,15 +9517,6 @@ public class ZAttrProvisioning {
     public static final String A_zimbraLicenseNotificationEmail = "zimbraLicenseNotificationEmail";
 
     /**
-     * License Pre expiry notification mail enabled info. The default value
-     * is TRUE
-     *
-     * @since ZCS 10.1.11
-     */
-    @ZAttr(id=4143)
-    public static final String A_zimbraLicensePreExpiryMailAlertEnabled = "zimbraLicensePreExpiryMailAlertEnabled";
-
-    /**
      * Email address for receiving Zimbra license notifications for License
      * Pre Expiry
      *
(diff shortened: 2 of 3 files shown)


viditZim:
Please add changes in zm-mailbox-zok also.

viditZim (approved):

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -10604,4 +10605,9 @@ TODO: delete them permanently from here
   <desc>Email address for receiving Zimbra license notifications for License Pre Expiry</desc>
 </attr>
 
+<attr id="4145" name="zimbraBackupDeduplicationEnabled" type="boolean" cardinality="single" flags="serverInherited" optionalIn="globalConfig,server" since="10.2.0">
```
dasiyogesh: since value should be 10.1.13

store/conf/attrs/zimbra-attrs.xml
```diff
@@ -9964,12 +9964,13 @@ TODO: delete them permanently from here
   <desc>The number of days backups will be kept. Backup Scheduler deletes backups older than the set value.</desc>
 </attr>
 
-<attr id="4013" name="zimbraBackupBlobsCompressType" type="enum" value="zip,zipStore,noZip" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="10.0.0">
+<attr id="4013" name="zimbraBackupBlobsCompressType" type="enum" value="zip,zipStore,noZip,ZSTD" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" callback="BackupBlobsCompressCallback"  since="10.0.0">
```
dasiyogesh: ldap attr modification not supported for patch release and this change is targeted for 10.1.13

rcyarrapothu: we should test this for backward compatibility. If someone has already set this value and then upgrades to the newer version of Zimbra, we need to check whether changing the existing LDAP value could trigger any issues.

store/src/java/com/zimbra/cs/account/callback/BackupBlobsCompressCallback.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.Map;
+
+public class BackupBlobsCompressCallback extends AttributeCallback {
+    @Override
+    public void preModify(CallbackContext context, String attrName, Object attrValue, Map attrsToModify, Entry entry) throws ServiceException {
+
+        boolean backupDedupEnabled = entry.getBooleanAttr("zimbraBackupDeduplicationEnabled", true);
+
+        if(!backupDedupEnabled && attrValue.equals("ZSTD"))  {
+            throw ServiceException.INVALID_REQUEST("BackupBlobsCompress cannot be ZSTD if the zimbraBackupDeduplication is disabled", null);
```
rcyarrapothu:
```suggestion
            throw ServiceException.INVALID_REQUEST("BackupBlobsCompress cannot be set to ZSTD when zimbraBackupDeduplicationEnabled is disabled.", null);
```

PR #1774, opened 2025-12-08
ZBUG-3974: OLK Mobile is not recognizing daylight savings from appointments created within ZWC

viditZim:
Mapped standard name and daylight name with format compatible with windows and modified the WindowsTimeZoneInformation logic.

### DIFF
--- a/store/src/java/com/zimbra/cs/mailbox/calendar/WindowsTimeZoneConversion.java
+++ b/store/src/java/com/zimbra/cs/mailbox/calendar/WindowsTimeZoneConversion.java
@@ -1,0 +1,152 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2013, 2014, 2016, 2025 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software Foundation,
+ * version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/mailbox/calendar/WindowsTimeZoneConversion.java
```diff
@@ -0,0 +1,1057 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2013, 2014, 2016 Synacor, Inc.
```
preshitaagrawal: please correct copyright year

preshitaagrawal (approved):

store/src/java/com/zimbra/cs/mailbox/calendar/WindowsTimeZoneConversion.java
```diff
@@ -0,0 +1,1057 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public final class WindowsTimeZoneConversion {
+
+    private static final Map<String, String> ICAL_TO_WINDOWS_STANDARD_NAME = new HashMap<>();
+
+    private static final Map<String, String> ICAL_TO_WINDOWS_DAYLIGHT_NAME = new HashMap<>();
+
+    static {
```
dasiyogesh, line 46: @viditZim this timezone data mapping that is being set here with day light offset how does it map with timezone data we have in https://github.com/Zimbra/zm-timezones/blob/develop/conf/timezones.ics ?

viditZim: @dasiyogesh I used the timezones from the timezones.ics to add in the map.
For example in the timezones.ics, for TZID Europe/Brussels the standard name is CET and Daylight name is CEST...so in the map I created I added Tzid Brussels for which Standard Name would be Romance Standard Time instead of CET/CEST for active-sync.

viditZim: @dasiyogesh modified the logic to dynamically create Map based on zm-timezones files.

preshitaagrawal (approved):

shubhamCS03 (approved):

chitranshu198 (approved):

PR #1779, opened 2026-01-13
ZBUG-5011 Restrict exportDir from web-exposed jetty paths

aakashsingh2000:
**Problem:** The ExportAndDeleteItemsRequest SOAP API currently accepts arbitrary filesystem paths via the exportDir parameter, without adequate validation.Allow users to provide any exportDir path, but strictly restrict it from pointing to /opt/zimbra/jetty/webapps or any subdirectory that may be web-exposed.

**Solution:**

Reject if canonical path starts with:

/opt/zimbra/jetty/webapps/
/opt/zimbra/jetty_base/webapps/


aakashsingh2000:
PR not required

PR #1785, opened 2026-01-23
ZBUG-4613: Added soft quota for IMAP copy for Trash folder instead of unlimited quota

viditZim:
Two issues (move to trash):

Scenario 1 : Mailbox size almost near to quota → COPY to trash → mailbox size exceeds the quota → Copy fails. This issue is currently fixed by not including trash folder for quota check.

Scenario 2 : COPY to trash → NO EXPUNGE due to some error → Mailbox size keeps on increasing way beyond Quota. 

We need Scenario 1 to work (quota bypass for deletes) while preventing Scenario 2 (unbounded growth).

Solution: Have a configurable Soft Quota Limit for Trash Folder(100-120%)

Instead of enforcing hard quota (100%) on trash, use a soft limit as a safety buffer.

### DIFF
--- a/store/src/java/com/zimbra/cs/mailbox/Mailbox.java
+++ b/store/src/java/com/zimbra/cs/mailbox/Mailbox.java
@@ -1500,6 +1500,36 @@ public class Mailbox implements MailboxStore {
         }
     }
 
+    /**
+     * Validates whether moving data to Trash would exceed account or domain quota limits.
+     *
+     * <p>The method checks the effective account quota and applies the configured
+     * soft limit percentage. If the resulting allowed quota is exceeded by the
+     * provided size, a {@link MailServiceException#QUOTA_EXCEEDED(long)} is thrown.</p>
+     *
+     * <p>It also verifies the domain aggregate quota. If the domain is already
+     * over its aggregate limit and receiving data over the limit is not allowed,
+     * a {@link MailServiceException#DOMAIN_QUOTA_EXCEEDED(long)} is thrown.</p>
+     *
+     * @param newSize the expected mailbox size after the Trash size change
+     * @throws ServiceException if the account or domain quota restrictions are violated
+     */
+    public void checkSizeChangeForTrash(long newSize) throws ServiceException {
+        Account acct = getAccount();
+        long acctQuota = AccountUtil.getEffectiveQuota(acct);
+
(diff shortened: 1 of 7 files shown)


ghen2:
A default soft margin of 110% seems quite large already, for a 5 GB mailbox quota it's an additional 512 MB.
Something like 101% seems like a more conservative default, as the original ZBUG-1375 issue needs wiggle room for just a few mails at a time?

And 100% (no additional soft margin) should be an allowed minimum value, currently it has `min="110"`.

adriangibanelbtactic:
> A default soft margin of 110% seems quite large already, for a 5 GB mailbox quota it's an additional 512 MB.


The way I understand the code at: https://github.com/Zimbra/zm-mailbox/blob/36ef36c774bc7ea527393778826b9982c96fc6c0/store/src/java/com/zimbra/cs/mailbox/Mailbox.java#L1504-L1507 it will mean 5 GB for the mailbox **and** 5 GB + 512 MB for the Trash in the worst case scenario.

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
```diff
@@ -1500,6 +1500,22 @@ public void checkSizeChange(long newSize) throws ServiceException {
         }
     }
 
+    public void checkSizeChangeForTrash(long newSize) throws ServiceException {
```
rcyarrapothu, line 1517: add a comment explaining the intent of this method.

rcyarrapothu (approved):

store/src/java/com/zimbra/cs/mailbox/Mailbox.java
viditZim: done

preshitaagrawal (approved):

adriangibanelbtactic:
> > A default soft margin of 110% seems quite large already, for a 5 GB mailbox quota it's an additional 512 MB.
> 
> The way I understand the code at:
> 
> https://github.com/Zimbra/zm-mailbox/blob/36ef36c774bc7ea527393778826b9982c96fc6c0/store/src/java/com/zimbra/cs/mailbox/Mailbox.java#L1504-L1507
> it will mean 5 GB for the mailbox **and** 5 GB + 512 MB for the Trash in the worst case scenario.

After re-reading the code, I should correct myself. It just means 5 GB + 512 MB for all the account quota.

Also, right now the default has been updated from 110 to 100 as it was requested.

PR #1796, opened 2026-03-23
ZBUG-5378: IMAP copy loop issue with iPhone mail app

preshitaagrawal:
**Problem:** When the default Mail App - performs a mass delete/Move in Zimbra environment where primary volume is set to S3, it takes a considerable amount of time to process the emails and the client times out. The client send the copy for all the emails again which causes duplicate emails in destination folder. This loop continues endlessly growing the mailbox size.
The IMAP protocol processes the Move/Delete operation in 3 steps, 1st it copies the emails which are sent in request then Flags those emails and then Expunge those. These all commands are sent by the client sequentially. But when the timeout occurs, it does not send the Flag and Expunge commands, retaining the original email in source folder.

**Solution:** Implemented [rfc9585](https://datatracker.ietf.org/doc/html/rfc9585) to send INPROGRESS responses every 10 seconds so that the client waits for the copy to complete.


viditZim (approved):

dasiyogesh:
@preshitaagrawal check for junits

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -198,9 +203,15 @@ protected enum ImapExtension { CONDSTORE, QRESYNC }
     private static final byte RETURN_XLIST      = 0x04;
 
     private final int SUGGESTED_BATCH_SIZE = 100;
-    private final int SUGGESTED_COPY_BATCH_SIZE = 50;
+
+    private final int SUGGESTED_COPY_BATCH_SIZE = 5;
```
dasiyogesh: why is the batch size reduced? any impact on performance due to smaller batch size

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -320,6 +331,23 @@ public ImapConfig getConfig() {
...
+            Executors.newSingleThreadScheduledExecutor();
+
+    protected void sendInProgress() {
+        notifier = scheduler.scheduleAtFixedRate(() -> {
+            if (!running.get()) {
+                ZimbraLog.imap.trace("Stopping scheduler loop");
+                return;
+            }
+            try {
+                this.sendLine("* OK [INPROGRESS] Hang in there...", true);
```
Gopal-Moolchandani: Wouldn't it be better to use constant here?

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: There is no impact of reducing the batch size on performance as it gets processed sequentially and each I/O will require mailbox lock.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: done

viditZim (approved):

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -4587,12 +4614,6 @@ protected boolean doCOPY(String tag, String sequenceSet, ImapPath path, boolean
                 copyuid = "[COPYUID " + uvv + ' ' + ImapFolder.encodeSubsequence(srcUIDs) + ' ' +
                         ImapFolder.encodeSubsequence(copyUIDs) + "] ";
             }
-        } catch (IOException e) {
```
dasiyogesh, line 4590: how is copy failure handled as we removed this exception handling?

store/src/java/com/zimbra/cs/imap/ImapHandler.java
dasiyogesh: ok, why is batch size reduced?

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: We are sending INPROGRESS response to client every 10 seconds so that the client does not times out.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: If the batch size is large, the mailbox lock is held for a longer duration, which can cause other operations on the same mailbox to fail with a LockFailedException due to timeouts. This is particularly noticeable when the primary storage is configured on S3, as I/O operations take comparatively longer to complete.

store/src/java/com/zimbra/cs/imap/NioImapHandler.java
```diff
@@ -204,6 +204,7 @@ public void setLoggingContext() {
 
     @Override
     protected void sendLine(String line, boolean flush) throws IOException {
+        ZimbraLog.imap.info("S: %s", line);
```
RakeshAMore: Do you really want this to be info level?

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -320,6 +333,23 @@ public ImapConfig getConfig() {
 
     protected abstract String getRemoteIp();
 
+    private final ScheduledExecutorService scheduler =
```
RakeshAMore: Is this going to be scheduled per client connection?

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -320,6 +333,23 @@ public ImapConfig getConfig() {
...
+            Executors.newSingleThreadScheduledExecutor();
+
+    protected void sendInProgress() {
+        notifier = scheduler.scheduleAtFixedRate(() -> {
```
RakeshAMore: how this notifier will be stopped?

store/src/java/com/zimbra/cs/imap/NioImapHandler.java
preshitaagrawal: yes, as it prints the responses we send to the client

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: As soon as the copy operation completes, finally gets executed and notifier is cancelled.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: yes, per client when copy operation is triggered.

desouzas: Please do a thread dump after you've triggered these COPY commands with inprogress in parallel test across many accounts. Observe how many of these extra threads have spawned.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -198,9 +203,17 @@ protected enum ImapExtension { CONDSTORE, QRESYNC }
...
     private final int SUGGESTED_BATCH_SIZE = 100;
-    private final int SUGGESTED_COPY_BATCH_SIZE = 50;
+
+    private final int SUGGESTED_COPY_BATCH_SIZE = 5;
+
     private final int SUGGESTED_DELETE_BATCH_SIZE = 30;
 
+    private AtomicBoolean running = null;
+
+    private ScheduledFuture<?> notifier = null;
```
desouzas: Does this need to be a class property if it is created and closed only in doCOPY? Move to local and adjust sendInProgress to return a ref to it.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -198,9 +203,17 @@ protected enum ImapExtension { CONDSTORE, QRESYNC }
     private static final byte RETURN_XLIST      = 0x04;
 
     private final int SUGGESTED_BATCH_SIZE = 100;
-    private final int SUGGESTED_COPY_BATCH_SIZE = 50;
+
+    private final int SUGGESTED_COPY_BATCH_SIZE = 5;
+
     private final int SUGGESTED_DELETE_BATCH_SIZE = 30;
 
+    private AtomicBoolean running = null;
```
desouzas: Does this need to be a class property if it is set and used only in doCOPY? Move to local and adjust sendInProgress to accept as a param?

store/src/java/com/zimbra/cs/imap/ImapHandler.java
desouzas: Consider LC for batch size.

store/src/java/com/zimbra/cs/imap/NioImapHandler.java
desouzas: This is too much info to add to logs. (NioLoggingFilter also exists)

store/src/java/com/zimbra/cs/imap/TcpImapHandler.java
```diff
@@ -219,7 +219,7 @@ protected String getRemoteIp() {
 
         @Override
         protected void sendLine(String line, boolean flush) throws IOException {
-            ZimbraLog.imap.trace("S: %s", line);
+            ZimbraLog.imap.info("S: %s", line);
```
desouzas: Similarly, why is this being changed? Too much info.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: done

store/src/java/com/zimbra/cs/imap/NioImapHandler.java
preshitaagrawal: done

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: done

store/src/java/com/zimbra/cs/imap/TcpImapHandler.java
preshitaagrawal: done

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: we need to adjust it in doCopy and access it in thread so, kept it on class level.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -3611,7 +3620,7 @@ private boolean expungeMessages(String tag, ImapFolder i4folder, String sequence
                 // send a gratuitous untagged response to keep pissy clients from closing the socket from inactivity
                 long now = System.currentTimeMillis();
                 if (now - checkpoint > MAXIMUM_IDLE_PROCESSING_MILLIS) {
-                    sendIdleUntagged();
+                    sendOkInProgress();
```
rcyarrapothu, line 3632: @preshitaagrawal Regarding the external store processing, how does it behave in the following scenario:

- let’s say x number of items are moved during the COPY operation and we’re sending in-progress updates while processing them
- what happens if the process gets aborted midway? does the client get informed about the x number of items already processed, or does it start from the beginning?
- not sure if this is known already, but we may need to test tshis scenario before and after this change. May be checking after what would happens if the request are getting timedout after few records on S3 etc.

rcyarrapothu: guess the behavior wouldnt change in this case as well

`        } catch (ServiceException e) {
            // 6.4.7: "If the COPY command is unsuccessful for any reason, server implementations
            //         MUST restore the destination mailbox to its state before the COPY attempt."`
            
I guess the duplicates would still exist when there's an issue during the processing. But we could still add a test case to confirm that

common/src/java/com/zimbra/common/localconfig/LC.java
```diff
@@ -1499,6 +1499,8 @@ public enum PUBLIC_SHARE_VISIBILITY { samePrimaryDomain, all, none };
     // imap folder pagination enabled
     public static final KnownKey zimbra_imap_folder_pagination_enabled =  KnownKey.newKey(false);
 
+    public static final KnownKey SUGGESTED_COPY_BATCH_SIZE =  KnownKey.newKey(5);
```
rcyarrapothu: add @supported annotation, it will show up in `zmlocalconfig -i`

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -4611,6 +4618,12 @@ protected boolean doCOPY(String tag, String sequenceSet, ImapPath path, boolean
             }
             sendNO(tag, rcode + command + " failed");
```
desouzas, line 4632: Need to stop notifier before here on error

common/src/java/com/zimbra/common/localconfig/LC.java
preshitaagrawal: done

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: done

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: scheduler.shutdown(); will stop the notifier.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: done

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: done

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: @rcyarrapothu We currently do not have a rollback mechanism implemented. If a failure occurs during processing, the destination folder should ideally be restored to its previous state, but that is not handled at the moment.

In this PR, the focus is only on addressing the client-side timeout by sending in-progress updates during the operation.

So, these scenarios will fail and client will start copy from the beginning.
ex: if mailbox is restarted during copy, the client starts copy from the beginning.

@divyam37 please add these cases to the testing sheet

store/src/java/com/zimbra/cs/imap/OperationProgress.java
shubhamCS03, line 1: Please add LICENSE BLOCK for this class

preshitaagrawal: done

common/src/java/com/zimbra/common/localconfig/LC.java
```diff
@@ -1499,6 +1499,12 @@ public enum PUBLIC_SHARE_VISIBILITY { samePrimaryDomain, all, none };
     // imap folder pagination enabled
     public static final KnownKey zimbra_imap_folder_pagination_enabled =  KnownKey.newKey(false);
 
+    @Supported
+    public static final KnownKey suggested_batch_copy_size =  KnownKey.newKey(5);
```
prashantajari: @preshitaagrawal Please add documentation for newly introduced zmlocalconfig keys. This will help admins to look for configuration changes as per help information.
e.g.
```
zmlocalconfig -i mailboxd_java_options
mailboxd_java_options:
JVM options to use when launching Mailboxd.
* Changes are in effect after server restart.

zmlocalconfig -i suggested_batch_copy_size -> No information
zmlocalconfig -i imap_copy_thread_pool_size -> No information
```

suggested_batch_copy_size localconfig name -> can this start with imap prefix, similar to imap_copy_thread_pool_size to maintain consistency and suggest this is related to imap operation?

preshitaagrawal: done

```
$ zmlocalconfig -i imap_in_progress_response_thread_pool_size
imap_in_progress_response_thread_pool_size:
Max threads in pool for sending InProgress response to client
* Changes are in effect after server restart.

$ zmlocalconfig -i imap_suggested_batch_copy_size
imap_suggested_batch_copy_size:
Number of messages to process per batch during IMAP COPY operations
* Changes are in effect after server restart.
```

viditZim (approved):
Please address the check-style comments. Rest LGTM.

shubhamCS03 (approved):

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -4621,6 +4641,28 @@ protected boolean doCOPY(String tag, String sequenceSet, ImapPath path, boolean
         return true;
     }
 
+    private ScheduledFuture<?> sendInProgressResponses(OperationProgress copyStatus) {
+        ScheduledFuture<?> notifier =  IN_PROGRESS_SCHEDULAR.scheduleAtFixedRate(() -> {
+            if (!copyStatus.getIsRunning()) {
+                ZimbraLog.imap.info("Scheduled notifier is about to shutdown");
```
desouzas: debug

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -197,10 +203,23 @@ protected enum ImapExtension { CONDSTORE, QRESYNC }
     private static final byte RETURN_CHILDREN   = 0x02;
     private static final byte RETURN_XLIST      = 0x04;
 
+    protected static final ScheduledExecutorService IN_PROGRESS_SCHEDULAR = Executors.newScheduledThreadPool(0,
```
desouzas: INPROGRESS_SCHEDULER

store/src/java/com/zimbra/cs/imap/OperationProgress.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public String getTag() {
+        return tag;
+    }
+
+    public Boolean getIsRunning() {
+        return isRunning.get();
+    }
+
+    public Integer getPercentageComplete() {
```
desouzas, line 64: Calculate value and return, remove class property.

Zero divide safety.

store/src/java/com/zimbra/cs/imap/OperationProgress.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+package com.zimbra.cs.imap;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class OperationProgress {
+
+    private AtomicInteger completedCount;
```
desouzas, line 25: Make the rest of int-related properties except isRunning regular ints, this and that are the only ones being modified and read.

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -4621,6 +4641,28 @@ protected boolean doCOPY(String tag, String sequenceSet, ImapPath path, boolean
...
     }
 
+    private ScheduledFuture<?> sendInProgressResponses(OperationProgress copyStatus) {
+        ScheduledFuture<?> notifier =  IN_PROGRESS_SCHEDULAR.scheduleAtFixedRate(() -> {
+            if (!copyStatus.getIsRunning()) {
+                ZimbraLog.imap.info("Scheduled notifier is about to shutdown");
+                return;
+            }
+            try {
+                if (copyStatus.getCompletedCount() != null && copyStatus.getCompletedCount() != 0) {
```
desouzas: Update condition:
* not null
* greater than 0
* less than total - [per rfc MUST](https://datatracker.ietf.org/doc/html/rfc9585#:~:text=therefore,%20it%20MUST%20be%20greater%20than%20PROGRESS)

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -197,10 +203,23 @@ protected enum ImapExtension { CONDSTORE, QRESYNC }
...
 
+    protected static final ScheduledExecutorService IN_PROGRESS_SCHEDULAR = Executors.newScheduledThreadPool(0,
+            new ThreadFactoryBuilder().setNameFormat("imap-inprogress-notifier-%d").build()
+    );
+
+    static {
+        ScheduledThreadPoolExecutor executor = (ScheduledThreadPoolExecutor) IN_PROGRESS_SCHEDULAR;
+        executor.setMaximumPoolSize(LC.imap_in_progress_response_thread_pool_size.intValue());
+        executor.setKeepAliveTime(LC.imap_in_progress_response_thread_keep_alive.intValue(), TimeUnit.SECONDS);
+        executor.allowCoreThreadTimeOut(true);
```
rcyarrapothu: you may need to check the value `imap_in_progress_response_thread_keep_alive`, if it's greater than zero before setting this to true. see the docs on the method `allowCoreThreadTimeOut()`:



`Sets the policy governing whether core threads may time out and terminate if no tasks arrive within the keep-alive time, being replaced if needed when new tasks arrive. When false, core threads are never terminated due to lack of incoming tasks. When true, the same keep-alive policy applying to non-core threads applies also to core threads. To avoid continual thread replacement, the keep-alive time must be greater than zero when setting true. This method should in general be called before the pool is actively used.`

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -4587,12 +4613,6 @@ protected boolean doCOPY(String tag, String sequenceSet, ImapPath path, boolean
                 copyuid = "[COPYUID " + uvv + ' ' + ImapFolder.encodeSubsequence(srcUIDs) + ' ' +
                         ImapFolder.encodeSubsequence(copyUIDs) + "] ";
             }
-        } catch (IOException e) {
```
rcyarrapothu, line 4590: wouldn't remove this unless we for sure know that we don't run into this.

desouzas: It was thrown from sendLine

store/src/java/com/zimbra/cs/imap/ImapHandler.java
desouzas: > also when it's equal to 0, return 100 by default?

no need, the else returns a generic INPROGRESS response

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: done

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: done

store/src/java/com/zimbra/cs/imap/OperationProgress.java
preshitaagrawal: done

store/src/java/com/zimbra/cs/imap/OperationProgress.java
preshitaagrawal: done

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: done

store/src/java/com/zimbra/cs/imap/ImapHandler.java
preshitaagrawal: done

Gopal-Moolchandani (approved):

ashwinsahu19 (approved):

dasiyogesh (approved):
LGTM

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -4621,6 +4641,30 @@ protected boolean doCOPY(String tag, String sequenceSet, ImapPath path, boolean
...
+            }
+            try {
+                if (copyStatus.getCompletedCount() != null
+                        && copyStatus.getCompletedCount() > 0
+                        && copyStatus.getCompletedCount() < copyStatus.getTotalCount()) {
+                    this.sendLine("* OK [INPROGRESS (\"" + copyStatus.getTag() + "\" "
+                            + copyStatus.getCompletedCount() + " " + copyStatus.getTotalCount()
+                            + ")] Processed " + copyStatus.getPercentageComplete() + "% of the items", true);
+                } else {
+                    this.sendLine("*" + INPROGRESS_RESPONSE, true);
```
dasiyogesh: format for response here needs space between * & INPROGRESS_RESPONSE as per RFC
`this.sendLine("* " + INPROGRESS_RESPONSE, true);`

dasiyogesh (changes requested):

store/src/java/com/zimbra/cs/imap/ImapHandler.java
```diff
@@ -4621,6 +4641,30 @@ protected boolean doCOPY(String tag, String sequenceSet, ImapPath path, boolean
         return true;
     }
 
+    private ScheduledFuture<?> sendInProgressResponses(OperationProgress copyStatus) {
+        return INPROGRESS_SCHEDULER.scheduleAtFixedRate(() -> {
+            if (!copyStatus.getIsRunning()) {
```
Gopal-Moolchandani: setLoggingContext();

dasiyogesh (approved):

PR #1797, opened 2026-03-26
ZBUG-5063 :handling_zimbraFeatureMailForwardingInFiltersEnabled_flag_at_account_level

omkar-oak-01:
This PR is regarding utilizing zimbraFeatureMailForwardingInFiltersEnabled filter at account level to enable filters 

Jira: [ZBUG-5063](https://synacor.atlassian.net/browse/ZBUG-5063)

[ZBUG-5063]: https://synacor.atlassian.net/browse/ZBUG-5063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/store/src/java-test/com/zimbra/cs/filter/FilterUtilTest.java
+++ b/store/src/java-test/com/zimbra/cs/filter/FilterUtilTest.java
@@ -16,17 +16,16 @@
  */
 package com.zimbra.cs.filter;
 
+import static com.zimbra.cs.filter.JsieveConfigMapHandler.CAPABILITY_VARIABLES;
 import static org.junit.Assert.fail;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
-
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.util.StringUtil;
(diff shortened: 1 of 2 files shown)


adriangibanelbtactic:
Regarding your RuleManager.java change:

> boolean applyRules = account.isFeatureMailForwardingInFiltersEnabled();

Maybe I'm missing something because I have not checked the full file but it would seem to me that you are actually disabling all of the rules even if they are not mail forwarding rules which it's what the feature is supposed to be doing in the first place.

As a proof-of-concept to check if you are able to read properly the feature/flag it's fine... but soon or later you would have to implement an specific 'is a Mail forwarding rule' check somewhere.

yamayimok:
The issue where a particular filter action in non per-user Sieve rules remained active even when the mail forwarding feature was disabled in the Web client was supposed to be fixed.  But the current changes disable the entire filter evaluation process simply according to the Web-client feature enabling/disabling attribute.  That's not the fix that was intended.

dasiyogesh:
> The issue where a particular filter action in non per-user Sieve rules remained active even when the mail forwarding feature was disabled in the Web client was supposed to be fixed. But the current changes disable the entire filter evaluation process simply according to the Web-client feature enabling/disabling attribute. That's not the fix that was intended.

That's right, its WIP and not final fix.

PR #1798, opened 2026-03-26
ZCS-18799:Adding new target to upload artifacts

chetanlanje74:


build-common.xml
```diff
@@ -8,6 +8,13 @@
 
     <property name="dev.home"  value="${user.home}"/>
 
+   <!-- Nexus Credentials Configuration -->
+    <property environment="env"/>
+    <property name="nexus.base.url" value="${env.NEXUS_BASE_URL}"/>
+    <property name="nexus.repo" value="develop-snapshot"/>
```
umagmrit: `nexus.repo` is hardcoded as `develop-snapshot`. It would be better to make it configurable so it can be used for different environments.

build-common.xml
```diff
@@ -720,6 +727,20 @@
       <ivy:publish settingsRef="dev.settings" pubrevision="${dev.version}" resolver="local" overwrite="true" />
     </target>
 
+    <!-- publish-artifact target -->
+    <target name="publish-artifact" depends="publish-local"
+            description="Publish artifacts to Nexus Artifactory">
+
+       <property name="ivy.publish.dir" value="${dev.home}/.zcs-deps/${ivy.organisation}/${ivy.module}"/>
+       <property name="published.jar" value="${ivy.publish.dir}/${ivy.module}-${dev.version}.jar"/>
+       <exec executable="curl" failonerror="true">
```
umagmrit: Instead of using curl for uploading artifacts, can we use Ivy’s native publish mechanism ?

build-common.xml
```diff
@@ -720,6 +727,20 @@
       <ivy:publish settingsRef="dev.settings" pubrevision="${dev.version}" resolver="local" overwrite="true" />
     </target>
 
+    <!-- publish-artifact target -->
+    <target name="publish-artifact" depends="publish-local"
```
umagmrit: Can we have an option to skip publishing artifacts?

umagmrit: can we use target name `publish-remote` ?

umagmrit (approved):

PR #1817, opened 2026-04-27
ZBUG-5063: Allow redirection from zimbraAdminSieveScriptBefore when zimbraFeatureMailForwardingInFiltersEnabled is FALSE

omkar-oak-01:
This PR is related to redirection use case of zimbra when zimbraFeatureMailForwardingInFiltersEnabled flag is set as true or false for any user the filter script added using zimbraAdminSieveScriptBefore flag do not workj 

**Fix**: Even if the zimbraFeatureMailForwardingInFiltersEnabled is true or false filters or any redirection conditions applied in filter will work

Jira: [ZBUG-5063](https://synacor.atlassian.net/browse/ZBUG-5063)

[ZBUG-5063]: https://synacor.atlassian.net/browse/ZBUG-5063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/store/src/java-test/com/zimbra/cs/filter/FilterUtilTest.java
+++ b/store/src/java-test/com/zimbra/cs/filter/FilterUtilTest.java
@@ -14,9 +14,15 @@
  * If not, see <https://www.gnu.org/licenses/>.
  * ***** END LICENSE BLOCK *****
  */
+
 package com.zimbra.cs.filter;
 
-import static org.junit.Assert.fail;
+import com.zimbra.cs.mailbox.Mailbox;
+import com.zimbra.cs.mailbox.MailboxManager;
+import com.zimbra.cs.mailbox.MailboxTestUtil;
+import com.zimbra.cs.mailbox.Message;
+import com.zimbra.cs.service.util.ItemId;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -26,7 +32,6 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
-
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.util.StringUtil;
 import com.zimbra.cs.account.Account;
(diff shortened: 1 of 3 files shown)


store/src/java-test/com/zimbra/cs/filter/FilterUtilTest.java
```diff
@@ -14,33 +15,29 @@
  * If not, see <https://www.gnu.org/licenses/>.
  * ***** END LICENSE BLOCK *****
  */
+
 package com.zimbra.cs.filter;
 
-import static org.junit.Assert.fail;
+import com.zimbra.cs.mailbox.*;
```
chitranshu198: Use specific imports

store/src/java/com/zimbra/cs/filter/ZimbraMailAdapter.java
```diff
@@ -262,6 +264,9 @@ public void addAction(Action action) {
                 }
             }
         }
+        if (isUserScriptExecuting) {
```
dasiyogesh, line 267: in what scenario `isUserScriptExecuting` becomes true?

omkar-oak-01: Hi Yogesh, it becomes true when user script is executing and it adds all the user script related actions in userScriptActions set

yamayimok:
The code fix itself looks fine. However, I don't understand why the generated test cases only cover the scenario where the filter action was "fileinto";  even though this fix primarily targets "redirect" action.  Why don't you expand the test coverage to include combinations of `zimbraFeatureMailForwardingInFiltersEnabled: [TRUE|FALSE]` x `zimbraAdminSieveScriptBefore|After : ... redirect "test@zmc.com"` as well?     This would help ensure the functionality does not regress in future release and give us grater confidence in the long-term stability of the fix.

store/src/java-test/com/zimbra/cs/filter/FilterUtilTest.java
```diff
@@ -1,3 +1,4 @@
+
```
yamayimok: remove unnecessary brank line.

store/src/java-test/com/zimbra/cs/filter/FilterUtilTest.java
omkar-oak-01: done

store/src/java-test/com/zimbra/cs/filter/FilterUtilTest.java
omkar-oak-01: done

chitranshu198 (approved):

dasiyogesh (approved):

PR #1827, opened 2026-06-02
AENG-378: EAS POC

RakeshAMore:
AENG-378: EAS POC


adriangibanelbtactic:
I have just seen the work done in: https://github.com/Zimbra/zm-mailbox/commits/feat/eas-okta-mfa/ .

Relying on localconfig variables implies that this feature is not going to be multi-tenant friendly.

What happens if we run a SaaS and two different clients have different `zimbra_okta_org_url` values?

I'll assume for now that it's easier to draft PoC using xml for saving extra keys rather than messing around with ldap values... and that later on you will consider this to be a per-domain setting.

PR #1836, opened 2026-06-25
ZCS-19479: Add ROPC handler registry with Okta and KeyCloak handlers

sonali033:
Implements ROPC handler registry for EAS Multi-Factor Authentication.

- Added IRopcHandler interface with getName() and authenticate() methods
- Added OktaRopcHandler and KeyCloakRopcHandler handler implementations
- Added IROPCHandlerRegistry to register and retrieve handlers by provider name
- Okta registered as built-in provider via static initializer
- Added unit tests for registry and handler validation

### DIFF
--- a/store/src/java-test/com/zimbra/cs/account/auth/IROPCHandlerRegistryTest.java
+++ b/store/src/java-test/com/zimbra/cs/account/auth/IROPCHandlerRegistryTest.java
@@ -1,0 +1,58 @@
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+ * Zimbra Collaboration Suite Server
+ * Copyright (C) 2026 Synacor, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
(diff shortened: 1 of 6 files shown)


adriangibanelbtactic:
`package-info.java` files have: `Zimbra Collaboration Suite, Network Edition license.`

Could we maybe add pre-receive server-side hook Github equivalent so that pushes with that license are rejected when pushed to FOSS-only public facing repos ?

Thank you very much!

store/src/java-test/com/zimbra/cs/account/auth/IROPCHandlerRegistryTest.java
```diff
@@ -0,0 +1,52 @@
+package com.zimbra.cs.account.auth;
```
ashwinsahu19, line 17: Add Licensing block.

ashwinsahu19 (approved):

chitranshu198 (approved):
Please address check style comments, rest looks good to me

ashwinsahu19 (approved):

PR #1841, opened 2026-06-29
ZCS-19565 : Add new LDAP attributes for domain-level SAML configuration

aakashsingh2000:
**Problem**:
SAML configuration lacks support for per-domain customization, limiting multi-tenant deployments.

**Solution**:
Add new LDAP schema attributes to enable domain-level SAML configuration with global fallback. Purely additive — no existing attributes or behavior modified.

### DIFF
--- a/store/conf/attrs/zimbra-attrs.xml
+++ b/store/conf/attrs/zimbra-attrs.xml
@@ -10644,4 +10644,67 @@ TODO: delete them permanently from here
   <desc>Soft quota limit percentage applied over zimbraMailQuota</desc>
 </attr>
 
+  <attr id="4152" name="zimbraSamlSpEntityId" type="string" cardinality="single"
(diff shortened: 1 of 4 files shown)


store/conf/attrs/zimbra-attrs.xml
```diff
@@ -10644,4 +10644,67 @@ TODO: delete them permanently from here
   <desc>Soft quota limit percentage applied over zimbraMailQuota</desc>
 </attr>
 
+  <attr id="4152" name="zimbraSamlSpEntityId" type="string" cardinality="single"
+        optionalIn="globalConfig,domain"
+        flags="domainInherited,domainAdminModifiable"
+        since="10.1.18">
```
dasiyogesh, line 10650: update the since version to 10.1.20

PR #1851, opened 2026-07-13
ZCS-19490 : Implement OktaROPChandler with PUSH MFA support.

ashwinsahu19:
JIRA : https://synacor.atlassian.net/browse/ZCS-19490

Created OktaHandler with actual integration with OKTA.

It has ROPC flow for first step auth.
Second factor is capable of PUSH challenge.


chitranshu198 (approved):

store/src/java/com/zimbra/cs/account/auth/ropc/okta/OktaRopcHandler.java
```diff
@@ -0,0 +1,362 @@
...
+        String clientSecret = challenge.get(REQUEST_PARAM_CLIENT_SECRET);
+        String clientSecretAuth = challenge.get(REQUEST_PARAM_CLIENT_SECRET_AUTH_TYPE);
+        int httpConnectMs = NumberUtils.toInt(challenge.get(REQUEST_PARAM_CONNECTION_TIMEOUT),
+                CONNECTION_TIMEOUT_DEFAULT);
+        int httpSocketMs = NumberUtils.toInt(challenge.get(REQUEST_PARAM_SOCKET_TIMEOUT), SOCKET_TIMEOUT_DEFAULT);
+
+        headers.put(ACCEPT, APPLICATION_JSON);
+
+        if (CLIENT_SECRET_AUTH_TYPE_BASIC.equalsIgnoreCase(clientSecretAuth)) {
+            if (clientSecret != null && !clientSecret.trim().isEmpty()) {
```
dasiyogesh: this should be `if (clientSecret == null || clientSecret.trim().isEmpty()) {` right?

store/src/java/com/zimbra/cs/account/auth/ropc/okta/OktaRopcHandler.java
```diff
@@ -0,0 +1,362 @@
...
+    private MFAChallenge buildPushChallenge(IRopcAuthRequest req, OktaResponse initial)
+            throws ServiceException {
+        String oobCode = getSafeString(initial.getOobCode());
+        String mfaToken = getSafeString(initial.getMfaToken());
+        OktaResponse pushResponse = null;
+        if (oobCode.isEmpty()) {
+            pushResponse = triggerPushWithRetry(req, initial);
+        }
+
+        oobCode = pushResponse.getOobCode();
```
dasiyogesh, line 186: there should be null check here before using `pushResponse`

store/src/java/com/zimbra/cs/account/auth/IRopcCustomAuth.java
```diff
@@ -68,15 +69,42 @@ public void authenticate(Account acct, String password,
...
+            if (arg == null) {
+                continue;
+            }
+            String[] parts = arg.split("=", 2);
+
+            if (parts.length == 2) {
+                String key = parts[0].trim();
+                String value = parts[1].trim();
+                if (key.isEmpty()) {
+                    ZimbraLog.account.debug("Skipping configuration arguments with missing keys");
```
dasiyogesh, line 107: we should add `continue;` here after log instead of adding to map

store/src/java/com/zimbra/cs/account/auth/ropc/util/HttpUtilities.java
```diff
@@ -0,0 +1,85 @@
...
+
+        HttpPost post = new HttpPost(url);
+        post.setConfig(RequestConfig.custom()
+                .setConnectTimeout(connectTimeoutMs)
+                .setConnectionRequestTimeout(connectTimeoutMs)
+                .setSocketTimeout(socketTimeoutMs)
+                .build());
+        if (headers != null) {
+            for (Map.Entry<String, String> header : headers.entrySet()) {
+                if (header.getKey() != null & header.getValue() != null) {
```
dasiyogesh: should it be `&&` instead of `&`

store/src/java-test/com/zimbra/cs/account/auth/IRopcCustomAuthTest.java
```diff
@@ -112,46 +136,34 @@ public void testAuthenticateOutcomePolicyDenied() throws Exception {
...
+            ropcCustomAuth.authenticate(mockAccount, "password", mockContext, args);
             fail("Expected AuthFailedServiceException was not thrown");
         } catch (AccountServiceException.AuthFailedServiceException e) {
             assertTrue("Message should indicate timeout",
                     e.getMessage().contains("MFA request timed out. Please try again"));
         }
     }
 
-    @Test
-    public void testAuthenticateOutcomeDefaultError() throws Exception {
```
dasiyogesh, line 126: removal of this test do we have the failure coverage?

store/src/java/com/zimbra/cs/account/auth/ropc/util/HttpUtilities.java
```diff
@@ -0,0 +1,85 @@
+package com.zimbra.cs.account.auth.ropc.util;
```
dasiyogesh, line 18: add license block

store/src/java/com/zimbra/cs/account/auth/ropc/IRopcAuthEngine.java
```diff
@@ -17,84 +17,148 @@
...
+                return Outcome.ERROR;
+        }
+    }
 
-            Boolean result = handler.authenticate(request);
+    private static Outcome awaitPush(IRopcHandler handler, MFAChallenge challenge,
+                                     String user, String password) throws ServiceException {
+        MFAPollResult pollResult;
+        long startTime = System.currentTimeMillis();
+        // TODO : set the value of polling_timeout in MFAChallenge in buildPushChallenge
```
dasiyogesh, line 87: buildPushChallenge() puts INTERVAL into state but never puts POLLING_TIMEOUT. Here challenge.get(POLLING_TIMEOUT) will always be null, so the timeout always defaults to 30s, so thats intentional?

dasiyogesh (changes requested):

store/src/java/com/zimbra/cs/account/auth/ropc/okta/OktaRopcHandler.java
ashwinsahu19: OktaResponse will not be null , after parsing null check is applied 
https://github.com/Zimbra/zm-mailbox/blob/feature/ZCS-19490/store/src/java/com/zimbra/cs/account/auth/ropc/okta/OktaRopcHandler.java#L328

store/src/java-test/com/zimbra/cs/account/auth/IRopcCustomAuthTest.java
ashwinsahu19: IRopcCustomAuth.java code is changed.WIth the updated code ERROR and default will have separate logging and exceptions. Since the switch argument is enum, it should ideally never come to default value. But I have used the default value as a safe guard.

store/src/java/com/zimbra/cs/account/auth/ropc/IRopcAuthEngine.java
ashwinsahu19: Yes it is intentionally kept as 30 second default. In the MFAPolling service the code will get shifted. It could be configurable but not from the okta response as okta provides a duration of 5 minutes which will a very long duration to hold the thread. It could be configured from the zimbraAuthMech args. After confirmation on this I will update the value in MFAPollingServcie.

sonali033 (approved):

store/src/java/com/zimbra/cs/account/auth/ropc/util/HttpUtilities.java
```diff
@@ -0,0 +1,102 @@
...
+            ZimbraHttpConnectionManager.getExternalHttpConnMgr().getDefaultHttpClient().build();
+
+    private HttpUtilities() {
+    }
+
+    public static HttpResponseWrapper postForm(String url, Map<String, String> form, int connectTimeoutMs,
+                                               int socketTimeoutMs, Map<String, String> headers)
+            throws ServiceException {
+
+        if (url == null || url.isEmpty()) {
```
dasiyogesh: can use StrinUtil to check null or empty

dasiyogesh (approved):

PR #1859, opened 2026-07-21
ZBUG 5577 : Existing filters not shown in the Web UI in Corp

Gopal-Moolchandani:
Issue : Existing filters not shown in the Web UI in Corp

Fix : Added Null Pointer Check.

### DIFF
--- a/common/src/java/com/zimbra/common/util/DateParser.java
+++ b/common/src/java/com/zimbra/common/util/DateParser.java
@@ -52,7 +52,11 @@ public class DateParser {
             mFormatterHolder.set(formatter);
         }
 
-        formatter.setTimeZone(TimeZone.getTimeZone(timezone));
+        if (null != timezone) {
+            formatter.setTimeZone(TimeZone.getTimeZone(timezone));
+        } else {
+            ZimbraLog.mailbox.debug("timezone is null, proceeding with default timezone");
+        }
         return formatter;
     }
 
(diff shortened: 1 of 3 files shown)


common/src/java/com/zimbra/common/util/DateParser.java
```diff
@@ -52,7 +52,9 @@ private SimpleDateFormat getFormatter() {
             mFormatterHolder.set(formatter);
         }
 
-        formatter.setTimeZone(TimeZone.getTimeZone(timezone));
+        if (null != timezone) {
```
silentsakky, line 55: adding null check if ok, but if we don't have timezone data then will that affect execution of filters? and do we know from where timezone data is getting dropped?

Gopal-Moolchandani: Hi Saket

**_if we don't have timezone data then will that affect execution of filters?_**

No, because it was working fine without we explicitly set this timezone data i.e. before ZCS - 14416. It will proceed with default timezone.

_**do we know from where timezone data is getting dropped?**_

After mailbox restart, objects will get lost.

shrutig0510, line 55: Please add debug level log when timezone is null

shrutig0510 (approved):

chitranshu198 (approved):

satya73 (approved):

common/src/java/com/zimbra/common/util/DateParser.java
silentsakky: If we are dropping timezone data and using the default timezone, then doesn't it affect the functionality of filters requiring timezone data?

shrutig0510: @silentsakky The PR has been updated. Please review the changes.

silentsakky (approved):

PR #1869, opened 2026-08-07
ZBUG-5364 : Fix inputStream resource leak in ImapPartSpecifier on StartOutOfBoundsException

Gopal-Moolchandani:


shrutig0510 (approved):

store/src/java/com/zimbra/cs/imap/ImapPartSpecifier.java
```diff
@@ -257,8 +257,11 @@ private InputStreamWithSize getContentOctetRangeFromFullContents(InputStreamWith
                 }
                 is = ByteUtil.SegmentInputStream.create(is, start, start + realLength);
             } catch (StartOutOfBoundsException e) {
-                //return empty string {0} when start is out of range
-                ZimbraLog.imap.warn("IMAP part requested start out of range", e);
+                // closing the original stream to prevent resource leaks
```
dasiyogesh, line 260: Please check on adding JUnit for this change to cover closing IS.

dasiyogesh (approved):

store/src/java/com/zimbra/cs/imap/ImapPartSpecifier.java
Gopal-Moolchandani: Since it's a performance improvement change on the JVM side, we cannot validate it by adding any test case or a unit test since garbage collector might already collect that particular stream and return the same result after making the changes. This change is specifically meant for high load environments in performance improvement.

PR #1889, opened 2026-08-26
ZCS-20571: Add service context url pattern for CSRF filter

chitranshu198:
ZCS-20571: Add service context url pattern for CSRF filter


chitranshu198:
This change is no longer required, thus closing the PR.

PR #1892, opened 2026-08-28
ZBUG-4827: fix ContactAutoComplete slowness by restricting autocomplete folder scope to contact-view mountpoints only

satya73:
ticket: https://synacor.atlassian.net/browse/ZBUG-4827

**Fix Details:**
Fixed ContactAutoComplete slowness by restricting autocomplete folder scope to contact-view mountpoints only.

### DIFF
--- a/store/src/java/com/zimbra/cs/mailbox/ContactAutoComplete.java
+++ b/store/src/java/com/zimbra/cs/mailbox/ContactAutoComplete.java
@@ -830,7 +830,7 @@ public class ContactAutoComplete {
                 if (folder instanceof Mountpoint) {
                     Mountpoint mp = (Mountpoint) folder;
                     mountpoints.put(mp.getTarget(), mp);
-                    if (mIncludeSharedFolders) {
+                    if (mIncludeSharedFolders && mp.getDefaultView() == MailItem.Type.CONTACT) {
                         folders.add(folder);
                     }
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/mailbox/ContactAutoComplete.java
```diff
@@ -830,7 +830,7 @@ private Pair<List<Folder>, Map<ItemId, Mountpoint>> getLocalRemoteContactFolders
                 if (folder instanceof Mountpoint) {
                     Mountpoint mp = (Mountpoint) folder;
                     mountpoints.put(mp.getTarget(), mp);
-                    if (mIncludeSharedFolders) {
+                    if (mIncludeSharedFolders && mp.getDefaultView() == MailItem.Type.CONTACT) {
```
shrutig0510, line 833: Can  mp.getDefaultView() be null.
Check for NPE

satya73: Based on my understanding, `defaultView` will always be either an explicitly set value or defaults to Type.UNKNOWN. Even if mp.getDefaultView() is null by any chance, there will not be any NPE as enum comparison with the == operator is always null-safe.

shrutig0510 (approved):

Gopal-Moolchandani (approved):

dasiyogesh (approved):

